PR #9, opened 2015-03-02
Bug with renewing ot long running tasks fixed

ddobric:
Due the issue with SB time precision, this check-in contains a fix, which renews the lock in regular interval of 30 sec. It is recommended to make this interval configurable in dependence of the lock time defined on the entity.

### DIFF
--- a/Framework/TaskActivityDispatcher.cs
+++ b/Framework/TaskActivityDispatcher.cs
@@ -62,8 +62,7 @@ namespace DurableTask
             {
                 TraceHelper.TraceSession(TraceEventType.Information,
                     receivedMessage.SessionId,
-                    GetFormattedLog("New message to process: " + receivedMessage.MessageId + " [" +
-                                    receivedMessage.SequenceNumber + "]"));
+                    GetFormattedLog($"New message to process: {receivedMessage.MessageId} [{receivedMessage.SequenceNumber}], latency: {receivedMessage.DeliveryLatency()}ms"));
             }
 
             return receivedMessage;
--- a/Framework/BrokeredMessageExtensions.cs
+++ b/Framework/BrokeredMessageExtensions.cs
@@ -1,0 +1,40 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask
+{
+    using System;
+    using System.Diagnostics;
+    using Microsoft.ServiceBus.Messaging;
+    using Tracing;
+
+    /// <summary>
+    /// Extension methods for BrokeredMessage
+    /// </summary>
+    public static class BrokeredMessageExtensions
+    {
+        /// <summary>
+        /// Returns delivery latency of the message
+        /// </summary>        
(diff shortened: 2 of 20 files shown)


Framework/TaskActivityDispatcher.cs
```diff
@@ -152,45 +152,45 @@ public sealed class TaskActivityDispatcher : DispatcherBase<BrokeredMessage>
...
-                }
-
-                DateTime renewAt = message.LockedUntilUtc.Subtract(TimeSpan.FromSeconds(30));
-
-                // service bus clock sku can really mess us up so just always renew every 30 secs regardless of 
-                // what the message.LockedUntilUtc says. if the sku is negative then in the worst case we will be
-                // renewing every 5 secs
-                //
-                renewAt = AdjustRenewAt(renewAt);
+                int renewInterval = 30000;
```
affandar: We need to add the renewInterval as a configurable setting to TaskActivityDispatcherSettings. It should default to 30 secs.

Framework/TaskActivityDispatcher.cs
```diff
@@ -152,45 +152,45 @@ public sealed class TaskActivityDispatcher : DispatcherBase<BrokeredMessage>
...
 
                 while (!cancellationToken.IsCancellationRequested)
                 {
-                    await Task.Delay(TimeSpan.FromSeconds(5));
-
-                    if (DateTime.UtcNow >= renewAt)
+                    await Task.Delay(TimeSpan.FromSeconds(30));
```
affandar: This should actually use the renewInterval defined above.

Framework/TaskActivityDispatcher.cs
```diff
@@ -152,45 +152,45 @@ public sealed class TaskActivityDispatcher : DispatcherBase<BrokeredMessage>
...
-                            // might have been completed
-                            TraceHelper.TraceException(TraceEventType.Information, exception,
-                                "Failed to renew lock for message {0}", message.MessageId);
-                            break;
+                            try
+                            {
+                                TraceHelper.Trace(TraceEventType.Information, "Renewing lock for message id {0}",
+                                    message.MessageId);
+
+                                message.RenewLock();
```
affandar: Btw can we switch to await message.RenewLockAsync()? 

Not sure why we were not doing that already :)

Framework/TaskActivityDispatcher.cs
```diff
@@ -152,45 +152,45 @@ public sealed class TaskActivityDispatcher : DispatcherBase<BrokeredMessage>
...
+
+                                TraceHelper.Trace(TraceEventType.Information, "Next renew for message id '{0}' at '{1}'",
+                                    message.MessageId, DateTime.Now.AddMilliseconds(renewInterval));
```
affandar: We want to normalize to UTC time everywhere. Can this be changed to DateTime.UtcNow.AddMilliseconds() etc?

Framework/TaskActivityDispatcher.cs
```diff
@@ -152,45 +152,45 @@ public sealed class TaskActivityDispatcher : DispatcherBase<BrokeredMessage>
...
+        /// LockDuration, which is usually 1 minute by default.
+        /// It would be good to make default renew interval configurable.
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="cancellationToken"></param>
         async void RenewUntil(BrokeredMessage message, CancellationToken cancellationToken)
         {
             try
             {
-                if (message.LockedUntilUtc < DateTime.UtcNow)
```
affandar: Might be worth it to still keep this check around.

Framework/TaskActivityDispatcher.cs
```diff
@@ -202,6 +202,8 @@ public sealed class TaskActivityDispatcher : DispatcherBase<BrokeredMessage>
             }
         }
 
+
+
         DateTime AdjustRenewAt(DateTime renewAt)
```
affandar: I think this method can be deleted now that it is not used.

affandar:
Damir, thanks for sending this PR. It greatly simplifies the lock renewal code and I would love to accept the change after you go through the comments I've given above.

Also please make sure that all the unit tests are passing with this change.

ddobric:
What is the status of this?

simonporter:
Closing old pull requests.
Please do submit new requests if the changes are applicable.

PR #20, opened 2016-02-03
Missing Exception information related to which trace listener is throwing the exception

solankisamir:
We get weekly 3 alerts with this exception "Failed to log actual trace because one or more trace listeners threw an exception", but are not able to figure out the root cause.

Logging the stack trace can help us locate and fix the issue.


solankisamir:
This got fixed in TaskHub 1.0.0.9 version
[Fix](https://github.com/solankisamir/durabletask/blob/master/Framework/Tracing/TraceHelper.cs#L210)

PR #67, opened 2016-10-19
Changes to create nupkg files from Visual Studio build.

PradeepKadubandi:
This extension : "https://visualstudiogallery.msdn.microsoft.com/fbe9b9b8-34ae-47b5-a751-cb71a16f7e96" is a pre-req.


nuproj/DurableTask/DurableTask.nuproj
```diff
@@ -0,0 +1,47 @@
...
+      <Configuration>Debug</Configuration>
+      <Platform>AnyCPU</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|AnyCPU">
+      <Configuration>Release</Configuration>
+      <Platform>AnyCPU</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\DurableTask.Framework\DurableTask.Framework.csproj" />
```
PradeepKadubandi: @simonporter  : Note that since the above csproj uses the name 'DurableTaskFx' as the assembly name , the created nuget package also uses that name of the dll, I was thinking the project should be fixed and once that is done this should automatically generate the package in line with that name.

Also, I have seen that some package references are referring "net45" and some other "net451" from 'packages.config' of the above csproj - again something that needs fixing there and the .nuproj automatically generates the right nuspec file and nupkg. [Without fixing it, the generated package also uses the dependencies with the framework as specified in packages.config]

PradeepKadubandi:
Actually wait a bit for me to update the PR. I am not yet happy with the nuproj file...

simonporter:
Closing old pull requests.
Does not appear this one is still relevant.

PR #69, opened 2016-11-10
servicefabric provider (work in progress)

PradeepKadubandi:
An alternative provider implementation using service fabric reliable collections. Can refer to TestStatefulService class in the fabric test application for how to use the provider.

What is done:
------------------------
- Activities (ScheduleTask)
- Timers
- Sub orchestrations
- Only parts of instance store with a few quirks
  - Only the latest state for a given orchestration instance (instance_id + execution_id combination) is stored (for a period of 24 hour time window)
  - State can be queried only for orchestrations still running / pending OR orchestrations that are finished within an hour ago
  - State of orchestration that's completed a day ago will be cleaned up / reclaimed

What is not yet done:
--------------------------
- ContinueAsNew is not supported
- External Events is not tested / validated
- No history event tracking in instance store
- No support for querying orchestration state across executions (does not make sense without ContinueWith support)
- No Automatic lock expiry for activies / orchestrations that are fetched from store but neither completed nor abandoned
- No equivalent of dead letter support (any bad activity / session will keep running forever in the system)
- Fabric instance store using azure instance store for backup (so that state will be persisted for longer periods)
- Fabric orchestration service directly using azure/other instance store instead of fabric instance store

Design notes:
----------------
- Separate reliable dictionaries are used for storing sessions (orchestrations) , activities (tasks) and timers (scheduled messages)
- We use transaction support in reliable collections to perform operations across different dictionaries
- Activity provider is the simplest to understand and the other two are slight modifications over it
- This diagram shows the typical dispatcher (from framework) flow of operations and how the activities provider implements them using a reliable dictionary and an in-memory queue:
   - (Post Edit) Perhaps the first diagram should not show an arrow between Send and Receive , the dispatcher loop is really 'Receive / Complete / Abandon' and Send happens as separate step...
dtfx visio as image

- For sessions the difference is:
  - The value in reliable dictionary is of type PersistentSession (an immutable structure) which contains a list of messages for that session (with key being the sessionId)
  - Send is typically a new record with one message or an append to messages collection for an existing session
  - While receiving, we mark all messages in session as received 
  - When completing, we delete all messages that are marked received
  - Session will be dropped (deleted) from the reliable store when orchestration reaches a terminal state
- For times:
  - The inmemory structure is a ImmutableSortedSet sorted on the FireAt value.
  - We activate messages in a loop triggered based on the next timer and walk through all scheduled messages that are due and completing them in this store and adding them to sessions (within a transaction).
- An interesting detail of the code is that since there is transaction involved across multiple data stores, each store operation is a separate method and each in-memory operation is a separate
  method that gets invoked by the caller after successful transaction commit. (That adds a bit of complexity in terms of readability to the code)


Test/TestFabricApplication/TestStatefulService/TestStatefulService.cs
```diff
@@ -0,0 +1,162 @@
...
+        ReplicaRole currentRole;
+
+        public TestStatefulService(StatefulServiceContext context) : base(context)
+        {
+            var settings = new FabricOrchestrationProviderSettings();
+            settings.TaskOrchestrationDispatcherSettings.DispatcherCount = 5;
+            settings.TaskActivityDispatcherSettings.DispatcherCount = 5;
+
+            var fabricProviderFactory = new FabricOrchestrationProviderFactory(this.StateManager, settings);
+            this.worker = new TaskHubWorker(fabricProviderFactory.OrchestrationService);
```
PradeepKadubandi: This is how an application will start the TaskHubWorker and Client to use the SF provider.

NickDarvey:
Is there active work on this @PradeepKadubandi ? Any idea when it'll get merged into master?

Issue #68, opened 2016-11-10
Ability to set the size of the service bus queues on creation


matthawley:
We'd like to be able to change the size of the service bus queues when they are created as sometimes 1GB isn't sufficient.

PR #70, opened 2016-11-17
Adding RaiseEvent to client for scalability, Allow creating queue with specific size

matthawley:
A common pattern for long-running orchestrations is to create an orchestration then raise an event on it. This causes twice as many messages to service bus. This change will do this process in a batch send.

### DIFF
--- a/src/DurableTask.Framework/ServiceBusOrchestrationService.cs
+++ b/src/DurableTask.Framework/ServiceBusOrchestrationService.cs
@@ -969,29 +969,35 @@ namespace DurableTask
         }
 
         /// <summary>
-        ///    Send an orchestration message
+        ///    Sends a set of orchestration messages
         /// </summary>
-        /// <param name="message">The task message to be sent for the orchestration</param>
-        public async Task SendTaskOrchestrationMessageAsync(TaskMessage message)
+        /// <param name="messages">The task message to be sent for the orchestration</param>
+        public async Task SendTaskOrchestrationMessageAsync(params TaskMessage[] messages)
         {
-            BrokeredMessage brokeredMessage = await ServiceBusUtils.GetBrokeredMessageFromObjectAsync(
-                message,
-                Settings.MessageCompressionSettings,
-                Settings.MessageSettings,
-                message.OrchestrationInstance,
-                "SendTaskOrchestrationMessage",
-                this.BlobStore,
-                DateTime.MinValue);
-
-            // Use duplicate detection of ExecutionStartedEvent by addin messageId
-            var executionStartedEvent = message.Event as ExecutionStartedEvent;
-            if (executionStartedEvent != null)
+            var brokeredMessages = new BrokeredMessage[messages.Length];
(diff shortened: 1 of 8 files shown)


matthawley:
I also added ability to set the maximum service bus queue size (issue #68)

matthawley:
Also fixed issue #73 which was throwing a null ref exception when a work item was aborted.

src/DurableTask.Framework/ServiceBusOrchestrationService.cs
```diff
@@ -969,29 +972,40 @@ public Task AbandonTaskActivityWorkItemAsync(TaskActivityWorkItem workItem)
...
+            if (messages.Length == 0)
+            {
+                return;
+            }
 
-            // Use duplicate detection of ExecutionStartedEvent by addin messageId
-            var executionStartedEvent = message.Event as ExecutionStartedEvent;
-            if (executionStartedEvent != null)
+            var brokeredMessages = new BrokeredMessage[messages.Length];
+            for (int i = 0; i < messages.Length; i++)
```
simonporter: GetBrokeredMessageFromObjectAsync could end up going to blob storage or another slow external storage so running this as tasks would be better so the latency is not sequential.

src/DurableTask.Framework/Settings/ServiceBusOrchestrationServiceSettings.cs
```diff
@@ -61,6 +61,11 @@ public ServiceBusOrchestrationServiceSettings()
         public int MaxTrackingDeliveryCount { get; set; }
 
         /// <summary>
+        ///     Maximum queue size, in megabytes, for the service bus queues
+        /// </summary>
+        public long MaxQueueSizeInMegabytes { get; set; } = 1024;
```
simonporter: 1024L as default for consistency with the valid sizes list

src/DurableTask.Framework/TaskHubClient.cs
```diff
@@ -157,6 +157,69 @@ public Task<OrchestrationInstance> CreateOrchestrationInstanceAsync(string name,
...
         /// <summary>
+        ///     Creates an orchestration instance, and raises an event for it, which eventually causes the OnEvent() method in the
+        ///     orchestration to fire.
+        /// </summary>
+        /// <param name="orchestrationName">Name of the orchestration as specified by the ObjectCreator</param>
+        /// <param name="orchestrationVersion">Name of the orchestration as specified by the ObjectCreator</param>
+        /// <param name="instanceId">Instance id for the orchestration to be created, must be unique across the Task Hub</param>
+        /// <param name="eventName">Name of the event</param>
+        /// <param name="eventData">Data for the event</param>
+        public async Task RaiseEventAsync(
```
simonporter: I think the naming of RaiseEventAsync here is a bit confusing as the meaning for an already running instance is quite different.
 I think it would be better to have a CreateOrchestrationInstanceWithRaisedEventAsync (open to other naming suggestions) and have the various permutations for with/without tags etc.

PR #74, opened 2016-12-22
'Invalid format string' excepton in TraceExceptionCore

ailn:
GetFormattedString did string.Format. If format contains '{' or '}' then it fails with 'Invalid format string' exception. The situation was possible due to adding stacktrace to format string


simonporter:
This PR to master (https://github.com/Azure/durabletask/pull/71/files) has some additional validation to catch the issue. This PR improves on that but would be good to bring in the validation from that here as well before we merge.

ailn:
I'll add the validation

simonporter:
Closing old pull requests.
No longer needed based on change to EventSource

PR #86, opened 2017-04-26
Move tracing to ETW EventSource

alextolp:
Move all tracing in DTFx to EventSource


src/DurableTask.Framework/Tracing/DefaultEventSource.cs
```diff
@@ -0,0 +1,230 @@
...
+
+namespace DurableTask.Tracing
+{
+    using System;
+    using System.Diagnostics;
+    using System.Diagnostics.Tracing;
+    using System.Runtime.InteropServices;
+
+    [EventSource(
+        Name = "DurableTask-Default",
```
cgillum: Would it make sense to name this DurableTask-Core instead of DurableTask-Default? Not sure what "Default" means in this context. It's my understanding that the DTFx nuget package will be called "Core" so I assumed we'd want the event source naming convention to match.

alextolp: Good point! Changed it to core!

PR #94, opened 2017-05-26
Azure Storage Provider

cgillum:
This PR adds a new Azure Storage-based provider for DTFx. It supports most of the key scenarios, including ScheduleTask, CreateTimer, ContinueAsNew, etc.

Some useful things to call out:

## History Table
The history table is an Azure Storage table which contains the history events for all orchestration instances. As instances run, new rows are added to this table. The partition-key of this table is the instance ID of the orchestration, to ensure optimal distribution of internal partitions in Azure Storage. This table resembles the tracking table used by the Service Bus provider, making it useful for debugging.

## Work-Item Queue
There is exactly one work item queue per task hub. This is a basic queue and behaves very similarly to the work item queue in the Service Bus provider. When a task worker is scaled out to multiple nodes, these nodes will all compete to acquire work from the work-item queue.

## Control Queue
The control queue is more sophisticated than the simpler work-item queue and is used to drive the orchestrations, just like the *orchestrator* queue in Service Bus. Because Azure Storage queues do not support sessions, it's not possible to simply use a competing consumer model to distribute load across nodes. Instead, orchestrator control messages must be load-balanced across *multiple* control queues. **Scale out is not yet implemented**.

## Data Consistency
Azure Storage does not support transactions across entities like Service Bus does, so this provider instead relies on eventual consistency. This is done by ensuring all queue and table operations are done in a particular order such that any machine failure will not cause any orchestration to get stuck. The result of this design is that there could potentially be a lot more duplication of work if there is a failure, but the system should always be able to auto-recover. More testing is needed to ensure that this is working correctly in all possible failure cases.

## Tracing
This provider uses it's own ETW-based Event Source trace provider, separate from the core Event Source provider. All logging is structured and optimized for querying. Logging also makes extensive use of correlation IDs and trace transfers to make it easier to track the full flow of an orchestration just be looking at the traces. At some point we may need to look into whether this should somehow be unified with the core trace provider.

## Remaining Work
This provider is not yet production ready. There are a few things that need to be done first:
* Large message support (blob spillover)
* Multi-node scale-out
* Improved timer support (e.g. longer than 7 days)
* Support for child orchestrations
* General performance tuning
* History retention policies


src/DurableTask.AzureStorage/TableEntityConverter.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// Converts a data contract object into a <see cref="DynamicTableEntity"/>.
+        /// </summary>
+        public DynamicTableEntity ConvertToTableEntity(object obj)
+        {
+            if (obj == null)
+            {
+                throw new ArgumentNullException(nameof(obj));
+            }
+
+            Debug.Assert(obj.GetType().GetCustomAttribute<DataContractAttribute>() != null);
```
simonporter: Are you sure you want this only in Debug

src/DurableTask.AzureStorage/TableEntityConverter.cs
```diff
@@ -0,0 +1,317 @@
...
+                    if (typeof(string).IsAssignableFrom(memberValueType))
+                    {
+                        if (property != null)
+                        {
+                            getEntityPropertyFunc = o => EntityProperty.GeneratePropertyForString((string)property.GetValue(o));
+                            setObjectPropertyFunc = (o, e) => property.SetValue(o, e.StringValue);
+                        }
+                        else
+                        {
+                            getEntityPropertyFunc = o => EntityProperty.GeneratePropertyForString((string)field.GetValue(o));
```
simonporter: This duplication feels very boilerplate, can this be contacted into some ternary operators and a switch statement

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -0,0 +1,50 @@
...
+    static class Utils
+    {
+        static JsonSerializerSettings TaskMessageSerializerSettings = new JsonSerializerSettings
+        {
+            TypeNameHandling = TypeNameHandling.Objects
+        };
+
+        public static string SerializeMessageData(MessageData messageData)
+        {
+            string rawContent = JsonConvert.SerializeObject(messageData, TaskMessageSerializerSettings);
```
simonporter: Are you hardset on Newtonsoft for serialization, we have tried to make this optional (but I do believe there is still at least on place that is not).

DurableTask.AzureStorage.sln
```diff
@@ -0,0 +1,28 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
+VisualStudioVersion = 14.0.25123.0
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DurableTask.AzureStorage", "src\DurableTask.AzureStorage\DurableTask.AzureStorage.csproj", "{A53DDB55-A6B7-474D-B404-40E7C35005B6}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DurableTask.Framework", "src\DurableTask.Framework\DurableTask.Framework.csproj", "{6F5D2EAD-726D-4FE5-A575-AEB96D1CCE37}"
```
cgillum: I'm deleting this additional solution file. I already added my new project to the main solution.

Test/DurableTask.AzureStorage.Tests/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
```
cgillum: Great! Deleting this as well.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -0,0 +1,1145 @@
...
+            try
+            {
+                await this.workItemQueue.DeleteMessageAsync(
+                    context.MessageData.OriginalQueueMessage,
+                    this.settings.WorkItemQueueRequestOptions,
+                    context.StorageOperationContext);
+            }
+            catch (StorageException e) when (e.RequestInformation?.HttpStatusCode == 404)
+            {
+                // Message was already deleted
```
cgillum: Agreed. Adding a new trace event for this.

src/DurableTask.AzureStorage/ReceivedMessageContext.cs
```diff
@@ -0,0 +1,228 @@
...
+        {
+            get
+            {
+                if (this.messageData != null)
+                {
+                    return this.messageData.TaskMessage.OrchestrationInstance;
+                }
+                else
+                {
+                    return this.messageDataBatch[0].TaskMessage.OrchestrationInstance;
```
cgillum: No, but I suppose I could be more defensive about it.

src/DurableTask.AzureStorage/TableEntityConverter.cs
cgillum: Yes. It's a fixed set of types which should be caught in testing. I also didn't want this reflection overhead in production (most of the other reflection uses are cached).

src/DurableTask.AzureStorage/TableEntityConverter.cs
cgillum: I couldn't come up with a way to reduce it further. The main problem is that the EntityProperty methods/properties are type-specific. I'm open to suggestions if there is something I'm not seeing.

src/DurableTask.AzureStorage/Utils.cs
cgillum: I don't have any hard requirements, but I would probably treat that as a separate feature that could be added later.

PR #99, opened 2017-06-29
Fix the possible null ref when cancelling the orchestration

zheg:


simonporter:
PR won't merge due to refactoring, incorporated into vnext as part of: https://github.com/Azure/durabletask/pull/101

PR #105, opened 2017-07-12
Upgraded ServiceBuss package to 3.2.2. 

ddobric:
Current version of DTF does not run in .NET 461 host.
Error: "General: .Net 4.6.1+ compatibility fix. Fixing custom DNS IdentityVerifier so that we honor multiple DNS claims "
Bugfix was done by SB team (see See https://www.nuget.org/packages/WindowsAzure.ServiceBus/3.1.3).
I have upgraded DTF to the next higher version.


.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
simonporter: Add '.vs/' to .gitignore, should not include anything in that folder

ddobric:
Hi guys,

could you please upgrade nuget package with new reference to SB 3.2.2+.
Right now, we cannot use DTF projects from desktop applicationn, which reference net core libraries, which reference DTF.

PR #109, opened 2017-08-23
Azure Functions Consumption Plan integration

cgillum:
This change is the initial work to enable dynamic scale-out for the Durable Task Framework in Azure Functions (i.e. *Durable Functions*).  It consists of two parts:

1. Created a `DisconnectedPerformanceMonitor` class which is responsible for examining queue metrics and making scale decisions. I call it "disconnected" because it's designed to be run outside of the task hub worker process. The intent is to run it in the Azure Scale Controller component.
2. Created a `AzureStorageOrchestrationServiceStats` and used it to track how often we do various operations against Azure Storage. This is intended to work just like `ServiceBusOrchestrationServiceStats`, but is specific to the Azure Storage provider.

The intent is to plug this new functionality into Azure Functions so that scale-out can happen and so that it will be easier to measure performance-sensitive statistics (which is important when you start scaling things out).

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
@@ -19,6 +19,7 @@ namespace DurableTask.AzureStorage.Tests
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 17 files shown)


cgillum:
Updated PR with squashed commits.

PR #110, opened 2017-08-28
first step towards .net core 2.0

ElanHasson:
This PR is not yet meant for merging, it's add to the discussion at #96

PR #116, opened 2017-09-25
Add multi-targeting support for .NET standard 2.0.

cgillum:
This change enables multi-targeting of `DurableTask.Core` to both .NET 4.5.1 (the existing target) as well as .NET Standard 2.0 for .NET Core support. This is needed for the Azure Functions Durable Task extension, which needs to support .NET Core. The code changes are influenced by the [existing PR for .NET Core support](https://github.com/Azure/durabletask/pull/110), but this PR is only concerned with `DurableTask.Core` and it involves multi-targeting rather than a wholesale move to .NET Standard 2.0.

I arbitrarily bumped the version to 2.1.0.0-preview, but we can discuss this. I suspect it will need to remain as a preview release until the non-production NuGet dependencies can be removed or replaced with non-prerelease dependencies.

This change does not currently contain any support for signing the assemblies since that is not required for Azure Functions. This is another thing we can discuss if necessary.

There are no warnings or errors when built using Release configuration.

### DIFF
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -2,13 +2,20 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFrameworks>net451;netstandard2.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
   </PropertyGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
     <PackageReference Include="ImpromptuInterface" Version="6.2.2" />
-    <PackageReference Include="Newtonsoft.Json" version="7.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="7.0.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Dynamitey" Version="2.0.6-beta111" />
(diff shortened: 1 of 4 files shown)


src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -2,13 +2,20 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFrameworks>net451;netstandard2.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
simonporter: I'd prefer package creation is manual as the signing key need to be brought in

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -2,13 +2,20 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFrameworks>net451;netstandard2.0</TargetFrameworks>
```
simonporter: Can we do this on the emulator as well, it should have no weird dependencies

tools/DurableTask.props
```diff
@@ -23,9 +24,9 @@
   <!-- Nuget Package Settings -->
   <PropertyGroup>
     <PackageOutputPath>..\..\build_output\packages</PackageOutputPath>
-    <AssemblyVersion>2.0.0.2</AssemblyVersion>
-    <FileVersion>2.0.0.2</FileVersion>
-    <Version>2.0.0.2-preview</Version>
+    <AssemblyVersion>2.1.0.0</AssemblyVersion>
```
simonporter: I am thinking 2.0.0.3 we have not released 2.0 on nuget yet so its not a breaking change

PR #144, opened 2018-03-09
WIP: Orchestration caching (aka "sticky sessions")

cgillum:
# Overview
This PR is intended to dramatically improve the throughput of individual orchestrations. The primary motivation is to improve the performance of Durable Functions, especially large fan-out/fan-in scenarios like https://github.com/Azure/azure-functions-durable-extension/issues/81.

The meat of the change is in DurableTask.Core. There are two parts:

1. Caching and reusing of task executors to avoid history replay whenever possible.
2. Separating work-item fetching from runtime state fetching, since we can usually just get the runtime state from the cache.

# Performance Implications
The effectiveness of the caching depends on how the storage provider is implemented. For **DurableTask.AzureStorage**, instances are pinned to workers (in steady state) so it can benefit quite a lot from caching. **DurableTask.ServiceBus** on the other hand does not pin orchestration instances to workers, so additional changes would be required in order for `DurableTask.ServiceBus` to get the most out of caching.

Whenever there is a cache hit, we will skip loading the orchestration state from the task hub (session state for Service Bus and table storage for Azure Storage), and will skip replaying the orchestration history. This means that we can simply resume the most recent open task and keep the local memory in-tact. If large messages are involved, then we would not need to fetch those from blob storage either.

# Design
A new interface, **IOrchestrationCaching**, was created and can be optionally implemented by the provider. This interface has methods for getting the cache configuration as well as loading orchestration state separate from the orchestration work items. When not implemented, caching will be disabled. This design allows us to start with an implementation in `DurableTask.AzureStorage` and allows us to leave `DurableTask.ServiceBus` untouched until we're ready.

Users can decide how many instances they want to cache per worker (for example, 10 instances). I decided to make this configurable because caching introduces additional memory pressure, which the customer has to pay for when running in the Azure Functions consumption plan. Also, some orchestrations may be heavier than others and the VM hosting the worker may have memory constraints (thus some users may opt to configure smaller numbers).

The first execution of an orchestration instance will always go down the original path. If the orchestration hits an `await`, we will cache the instance state, context, and executor in an in-memory MRU cache. Subsequent executions will then try to use the cached instance data. If we find the instance data in the cache, we simply reuse the existing `TaskOrchestrationExecutor` and execute **only the new events which are part of the work item**.

# Performance Comparison
TODO

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -29,6 +29,7 @@ namespace DurableTask.AzureStorage
     using DurableTask.AzureStorage.Partitioning;
     using DurableTask.Core;
     using DurableTask.Core.History;
+    using DurableTask.Core.Settings;
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Queue;
     using Microsoft.WindowsAzure.Storage.Table;
@@ -40,6 +41,7 @@ namespace DurableTask.AzureStorage
     public class AzureStorageOrchestrationService :
         IOrchestrationService,
         IOrchestrationServiceClient,
+        IOrchestrationCaching,
         IPartitionObserver<BlobLease>
     {
         static readonly HistoryEvent[] EmptyHistoryEventList = new HistoryEvent[0];
@@ -256,6 +258,12 @@ namespace DurableTask.AzureStorage
         public int MaxConcurrentTaskActivityWorkItems
         {
             get { return this.settings.MaxConcurrentTaskActivityWorkItems; }
+        }
+
+        /// <inheritdoc />
+        public OrchestrationInstanceCacheSettings OrchestrationCacheSettings
+        {
+            get { return this.settings.OrchestrationCacheSettings; }
         }
 
         // We always leave the dispatcher counts at one unless we can find a customer workload that requires more.
@@ -698,6 +706,15 @@ namespace DurableTask.AzureStorage
 
                 return null;
             }
+        }
+
+        /// <summary>
+        /// Fetches the runtime state for a particular orchestration instance.
+        /// </summary>
+        public Task<OrchestrationRuntimeState> FetchOrchestrationRuntimeStateAsync(TaskOrchestrationWorkItem workItem)
+        {
+            // TODO, cgillum: Implement this in such a way that we can get the information we need from the work item.
+            throw new NotImplementedException("TODO");
         }
 
         Task<OrchestrationRuntimeState> GetOrchestrationRuntimeStateAsync(
--- a/src/DurableTask.Core/Common/MruCache.cs
+++ b/src/DurableTask.Core/Common/MruCache.cs
@@ -1,0 +1,229 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.Core.Common
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Simple MRU cache implementation. TODO, cgillum: Implement locking for thread safety!
+    /// </summary>
+    class MruCache<TKey, TValue> where TValue : class
+    {
+        readonly LinkedList<TKey> mruList;
+        readonly Dictionary<TKey, CacheEntry> items;
+
+        CacheEntry mruEntry;
+
+        public MruCache(int watermark)
+            : this(watermark * 4 / 5, watermark)
+        {
+        }
+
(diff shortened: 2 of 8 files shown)


src/DurableTask.Core/Common/MruCache.cs
```diff
@@ -0,0 +1,229 @@
...
+
+namespace DurableTask.Core.Common
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Simple MRU cache implementation. TODO, cgillum: Implement locking for thread safety!
+    /// </summary>
+    class MruCache<TKey, TValue> where TValue : class
```
tohling: >class [](start = 4, length = 5)

For consistency, add "public"?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -700,12 +690,38 @@ PendingMessageBatch StashMessagesAndGetNextBatch(IEnumerable<MessageData> queueM
...
+                context.StorageOperationContext,
+                CancellationToken.None);
+
+            if (runtimeState == null ||
+                (runtimeState.ExecutionStartedEvent != null &&
+                runtimeState.OrchestrationStatus != OrchestrationStatus.Running &&
+                runtimeState.OrchestrationStatus != OrchestrationStatus.Pending))
+            {
+                // The instance has already completed.
+                return null;
```
tohling: >null [](start = 23, length = 4)

Should we log here?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -847,82 +863,20 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
             bool addedWorkItemMessages = false;
 
-            CloudQueue currentControlQueue = await this.GetControlQueueAsync(instanceId);
+            CloudQueue currentControlQueue = await this.GetControlQueueAsync(workItem.InstanceId);
             int totalMessageCount = 0;
 
             // Second persistence step is to commit outgoing messages to their respective queues. If there is
             // any failures here, then the messages may get written again later.
-            var enqueueTasks = new List<Task>(newEvents.Count);
+            var enqueueTasks = new List<Task>(runtimeState?.NewEvents.Count ?? 16);
```
tohling: >16 [](start = 79, length = 2)

Sorry I have no context, but why 16?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1013,6 +967,76 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+
+            for (int i = 0; i < newEvents.Count; i++)
+            {
+                HistoryEvent historyEvent = newEvents[i];
+                DynamicTableEntity entity = this.tableEntityConverter.ConvertToTableEntity(historyEvent);
+
+                newEventList.Append(historyEvent.EventType.ToString()).Append(',');
+
+                // The row key is the sequence number, which represents the chronological ordinal of the event.
+                long sequenceNumber = i + (allEvents.Count - newEvents.Count);
```
tohling: >allEvents.Count - newEvents.Count [](start = 43, length = 33)

Will this return a negative value?

src/DurableTask.Core/Common/MruCache.cs
```diff
@@ -0,0 +1,229 @@
...
+        public bool TryGetValue(TKey key, out TValue value)
+        {
+            // remove the MRU item if it is invalid
+            if (this.mruEntry.Node != null && !IsValid(this.mruEntry))
+            {
+                this.Remove(this.mruEntry.Node.Value);
+            }
+
+            // first check our MRU item
+            if (this.mruEntry.Node != null && key != null && key.Equals(this.mruEntry.Node.Value))
```
tohling: >Value [](start = 91, length = 5)

Do we need to create a comparator for the CachedOrchestrationInstance object?

tohling: >Equals(this.mruEntry.Node.Value) [](start = 65, length = 32)

Do we need to create a comparator for the CachedOrchestrationInstance object?

tohling: >Value [](start = 91, length = 5)

Do we need to create a comparator for the CachedOrchestrationInstance object?

tohling:
>                         if (result.IsCompleted)

I think this is supposed to be if(IsCompleted) ?

---
Refers to: src/DurableTask.Core/TaskOrchestrationExecutor.cs:82 in b566b10. [](commit_id = b566b109ca8d5579c0c738607c115b6a22a62706, deletion_comment = False)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -36,7 +37,9 @@ public class TaskOrchestrationDispatcher
         readonly IOrchestrationService orchestrationService;
         readonly WorkItemDispatcher<TaskOrchestrationWorkItem> dispatcher;
         readonly DispatchMiddlewarePipeline dispatchPipeline;
+        readonly MruCache<string, CachedOrchestrationInstance> instanceCache;
```
tohling: >instanceCache [](start = 63, length = 13)

Some very good and significant changes here with the introduction caching instances.  Not sure if it's possible to add some tests around this. Would be nice if that's possible.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -38,7 +38,20 @@ internal class TaskOrchestrationExecutor
             this.taskOrchestration = taskOrchestration;
         }
 
+        public bool IsCompleted => this.result != null && (this.result.IsCompleted || this.result.IsFaulted);
```
tohling: >IsCompleted [](start = 20, length = 11)

Was this mean for Line 82?

PR #145, opened 2018-03-14
Azure Storage Provider: Adding Instances table

gled4er:
Hello @cgillum ,

Let's use this pull request for tracking the changes I am making for this issue - https://github.com/Azure/azure-functions-durable-extension/issues/183

The main changes are yet to come but I decided to open the pull request as soon as possible. 

Thank you!


gled4er:
Hello @cgillum ,

In the last commit I added the first attempt for implementing logic for adding rows to the new Instances table.

May I ask you to quickly check it when you have time just to know if I am on the right track?

Thank you!

Test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.playlist
```diff
@@ -0,0 +1 @@
...rage.Tests.AzureStorageScenarioTests.UnhandledActivityException" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.TimerExpiration" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.ValidateEventSource" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.ActorOrchestration" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScaleTests.CreateTaskHub" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.TimerCancellation" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScaleTests.DeleteTaskHub" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScaleTests.ScaleDecisionForIncreasingWorkItemCount" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScaleTests.TestInstanceAndMessageDistribution" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.OrchestrationConcurrency" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.UnhandledOrchestrationException" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.SequentialOrchestration" /><Add Test="DurableTask.AzureStorage.Tests.AzureStorageScenarioTests.HandledActivityException" /></Playlist>
```
cgillum: I'm not sure what this file is. Should we leave it out?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1455,6 +1466,40 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
...
+            }
+
+            DynamicTableEntity entity = new DynamicTableEntity(message.OrchestrationInstance.InstanceId, "")
+            {
+                Properties =
+                {
+                    ["Input"] = new EntityProperty(executionStartedEvent.Input),
+                    ["Output"] = new EntityProperty(string.Empty),
+                    ["CreatedTime"] = new EntityProperty(executionStartedEvent.Timestamp),
+                    ["LastUpdatedTime"] = new EntityProperty(string.Empty),
```
cgillum: Instead of `string.Empty`, let's remove **Output** and **LastUpdatedTime** so that they are null.

gled4er:
Hello @cgillum, 

Thank you very much for the great feedback.

I work on fixing the issues and working on the next items.

Thank you!

Test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.playlist
gled4er: This is a unit test playlist including only the Azure Storage ones. I am using it to validate the changes  I am making but I can leave it if you think that it is not required to be included in the project.

Thank you!

gled4er:
Hello @cgillum ,

I added the logic for updating the Instances table to the `CompleteTaskOrchestrationWorkItemAsync` method. 

May I ask you to review so we can be sure that the direction I am taking is correct?

Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -925,6 +959,48 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                    orchestrationInstanceUpdate.Properties.Add("Output", new EntityProperty(runtimeState.Output));
+                }
+
+                orchestrationInstanceUpdate.Properties.Add("LastUpdatedTime", new EntityProperty(orchestratorLastUpdatedTime));
+                orchestrationInstanceUpdate.Properties.Add("RuntimeStatus", new EntityProperty(orchestratorStatus.ToString()));
+
+                Stopwatch orchestrationInstanceUpdateStopwatch = Stopwatch.StartNew();
+                await this.instancesTable.ExecuteAsync(TableOperation.InsertOrReplace(orchestrationInstanceUpdate));
+
+                AnalyticsEventSource.Log.AppendedInstanceState(
```
cgillum: Can you create a new event for this case in `AnalyticsEventSource`? I think it should be tracked separately from history updates.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -925,6 +959,48 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                if (!string.IsNullOrEmpty(runtimeState.Output) && runtimeState.Output != "\"\"")
+                {
+                    orchestrationInstanceUpdate.Properties.Add("Output", new EntityProperty(runtimeState.Output));
+                }
+
+                orchestrationInstanceUpdate.Properties.Add("LastUpdatedTime", new EntityProperty(orchestratorLastUpdatedTime));
+                orchestrationInstanceUpdate.Properties.Add("RuntimeStatus", new EntityProperty(orchestratorStatus.ToString()));
+
+                Stopwatch orchestrationInstanceUpdateStopwatch = Stopwatch.StartNew();
+                await this.instancesTable.ExecuteAsync(TableOperation.InsertOrReplace(orchestrationInstanceUpdate));
```
cgillum: For doing performance analysis, it's important that we keep track of our storage account metrics. Please add the following lines after the table update:
```csharp
this.stats.TableEntitiesWritten.Increment();
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1006,7 +1082,7 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
             }
 
             await Task.WhenAll(enqueueTasks);
-            this.stats.StorageRequests.Increment(totalMessageCount);
+            this.stats.StorageRequests.Increment(totalMessageCount + requestsForUpdatingInstancesTable);
```
cgillum: Rather than creating a `requestsForUpdatingInstancesTable`, I think it is cleaner to just have a separate increment call at the point immediately after we insert into the table (see my other comment). In a separate PR, I have put updating table storage into a separate method, and it will be easier to combine my changes this way.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -869,6 +869,11 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
             var tableBatchBatches = new List<TableBatchOperation>();
             tableBatchBatches.Add(batchOperation);
 
+            bool orchestratorEventExisits = false;
```
cgillum: The **Instances** table will need to be updated every time we call this method. This is because we need to keep the **LastUpdatedTime** and **CustomStatus** columns up-to-date. Therefore I think we can remove this local variable.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -925,6 +959,48 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                {
+                    Properties =
+                    {
+                        ["CreatedTime"] = new EntityProperty(runtimeState.CreatedTime),
+                    }
+                };
+
+                if (!string.IsNullOrEmpty(runtimeState.Input) && runtimeState.Input != "\"\"")
+                {
+                    orchestrationInstanceUpdate.Properties.Add("Input", new EntityProperty(runtimeState.Input));
```
cgillum: If I understand correctly, the **Input** and **CreatedTime** properties will have already been set at creation time. Yet it seems you're re-saving them every time we update the **Instances** table. Why is this? To minimize I/O to Table Storage, I would expect we only update these columns at creation time, or when `historyEvent.EventType == EventType.ContinueAsNew`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -925,6 +959,48 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                }
+
+                if (!string.IsNullOrEmpty(runtimeState.Output) && runtimeState.Output != "\"\"")
+                {
+                    orchestrationInstanceUpdate.Properties.Add("Output", new EntityProperty(runtimeState.Output));
```
cgillum: It seems to me that we should only update the **Output** column when `historyEvent.EventType` is `EventType.ExecutionCompleted`, `EventType.ExecutionFailed`, or `EventType.ExecutionTerminated`. Also, we should not filter out empty string because that could be a valid output for the orchestration. Instead, just filter out `null`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -869,6 +869,11 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
             var tableBatchBatches = new List<TableBatchOperation>();
             tableBatchBatches.Add(batchOperation);
 
+            bool orchestratorEventExisits = false;
+            EventType orchestratorEventType = EventType.ExecutionStarted;
+            DateTime orchestratorLastUpdatedTime = DateTime.MinValue;
+            OrchestrationStatus orchestratorStatus = OrchestrationStatus.Pending;
+
```
cgillum: Instead of creating these temporary variables, I think it would make the code simpler if you created the `DynamicTableEntity` here and initialized it with all the values that will be required for every update (e.g. **LastUpdatedTime** and **CustomStatus**). You can then go through your `switch` statement and add columns appropriately depending on `historyEvent.EventType`, similar to what you're doing now. This design will help ensure that you only update the columns that need to be updated (see my comments further down), which is important from a performance perspective.

BTW, the **CustomStatus** column should be populated with the `runtimeState.Status` value. This is a new requirement for the feature work I am doing in another PR.

gled4er:
Hello @cgillum ,

Thank you very much for the great feedback.

I will work on the issues and let you know for the next round of review.

Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you very much! I will change it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Great idea! I will add it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you for pointing that. I will make the changes. I think I need better understanding here. When we have one of the Task events such as `TaskScheduled` , `TaskCompleted` or `TaskFailed` without **CustomStatus** provided do we need to update the Instances table? This was my main idea for introducing the filtering here.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you very much! I will fix that.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you! I will fix that.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: This is very elegant approach! I love it! Thank you for clarifying it! This answers the question I had in the comment above. Thank you!

gled4er:
Hello @cgillum ,

I addressed your comments in the last review iteration.

When you have time may I ask you to check my latest changes?

Thank you!

gled4er:
Hello @cgillum ,

I added conceptual implementation for the changed `GetOrchestrationStateAsync` method. I wasn't able to test it so far but I wanted first to check with you if this is the correct approach. I am worried that I am not able to provide the same information as the original implementation because we don't store everything in the **Instance** table.

How do you usually test the public methods in the service that does not have test in the project? Are you integrating the NuGet artifacts in the Durable Functions project?

Thank you!

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -302,5 +302,12 @@ public void LeaseRemovalFailed(string Account, string TaskHub, string WorkerName
             EnsureLogicalTraceActivityId();
             this.WriteEvent(134, Account, TaskHub, WorkerName, PartitionId, Token);
         }
+
+        [Event(135, Level = EventLevel.Informational, Message = "Host '{2}' successfully updated Instances table with '{5}' event")]
+        public void InstancesUpdate(string Account, string TaskHub, string InstanceId, string ExecutionId, int EventCount, string EventType ,long LatencyMs)
```
cgillum: `EventCount` will always be `1`, so let's not include this parameter.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1100,6 +1091,18 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
             await this.DeleteMessageBatchAsync(context, currentControlQueue);
         }
 
+        private void InsertOrUpdateDynamicTableEntityProperty(DynamicTableEntity dynamicTableEntity,  string propertyName, EntityProperty entityProperty)
+        {
+            if (dynamicTableEntity.Properties.ContainsKey(propertyName))
+            {
+                dynamicTableEntity.Properties[propertyName] = entityProperty;
```
cgillum: For normal dictionaries, you can always safely use `dict["property"] = value` for either insert or update. Is that not the case with `DynamicTableEntity`? Ideally we don't have to the `ContainsKey` check and we don't need to have the `InsertOrUpdateDynamicTableEntityProperty` method.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -869,10 +869,16 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
-            DateTime orchestratorLastUpdatedTime = DateTime.MinValue;
-            OrchestrationStatus orchestratorStatus = OrchestrationStatus.Pending;
+            EventType? orchestratorEventType = null;
+
+            DynamicTableEntity orchestrationInstanceUpdate = new DynamicTableEntity(instanceId, "")
+            {
+                Properties =
+                {
+                    ["CustomStatus"] = new EntityProperty(runtimeState.Status),
+                    ["LastUpdatedTime"] = new EntityProperty(DateTime.UtcNow),
```
cgillum: I think LastUpdatedTime needs to always be set to the time of the last history event.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -897,32 +903,43 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                        this.InsertOrUpdateDynamicTableEntityProperty(orchestrationInstanceUpdate, "RuntimeStatus", new EntityProperty(EventType.ExecutionTerminated.ToString()));
+                        break;
+                    case EventType.ContinueAsNew:
+                        orchestratorEventType = historyEvent.EventType;
+                        if (!string.IsNullOrEmpty(runtimeState.Input))
+                        {
+                            orchestrationInstanceUpdate.Properties.Add("Input", new EntityProperty(runtimeState.Input));
+                        }
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
+                        orchestrationInstanceUpdate.Properties.Add("CreatedTime", new EntityProperty(runtimeState.CreatedTime));
```
cgillum: For `ContinueAsNew`, I don't think we need to set Input or CreatedTime. I think both of those will be set once the new instance generation starts. If that's the case, then just setting "LastUpdatedTime" and "RuntimeStatus" should be enough.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -897,32 +903,43 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                 // Replacement can happen if the orchestration episode gets replayed due to a commit failure in one of the steps below.
                 batchOperation.InsertOrReplace(entity);
 
+
+
```
cgillum: nit: please remove these extra blank lines.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -897,32 +903,43 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
                 // Monitor for orchestration instance events 
                 switch (historyEvent.EventType)
                 {
                     case EventType.ExecutionStarted:
                         orchestratorEventType = historyEvent.EventType;
-                        orchestratorLastUpdatedTime = historyEvent.Timestamp;
-                        orchestratorStatus = OrchestrationStatus.Running;
-                        orchestratorEventExisits = true;
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
+                        orchestrationInstanceUpdate.Properties.Add("RuntimeStatus", new EntityProperty(EventType.ExecutionStarted.ToString()));
```
cgillum: For `EventType.Executionstarted`, you will need to cast `historyEvent` to `ExecutionStartedEvent` and save the `Name`, `Version`, and `Input` properties.

cgillum: For the other event cases, you may want to consider casting `historyEvent` to the appropriate type and using those properties as well instead of relying on the properties of `runtimeState`. I think `runtimeState` should only be used for saving "CustomStatus", like you are above.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -309,5 +309,12 @@ public void InstancesUpdate(string Account, string TaskHub, string InstanceId, s
             EnsureLogicalTraceActivityId();
             this.WriteEvent(135, Account, TaskHub, InstanceId, ExecutionId, EventCount, EventType, LatencyMs);
         }
+
+        [Event(136, Level = EventLevel.Informational)]
+        public void FetchedInstanceStatus(string Account, string TaskHub, string InstanceId, string ExecutionId, int EventCount, int RequestCount, long LatencyMs)
```
cgillum: Let's remove EventCount and RequestCount since they will always be `1`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1620,33 +1620,40 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
...
-                Name = runtimeState.Name,
-                Version = runtimeState.Version,
-                Status = runtimeState.Status,
-                Tags = runtimeState.Tags,
-                OrchestrationStatus = runtimeState.OrchestrationStatus,
-                CreatedTime = runtimeState.CreatedTime,
-                CompletedTime = runtimeState.CompletedTime,
-                LastUpdatedTime = runtimeState.Events.Last().Timestamp,
-                Size = runtimeState.Size,
-                CompressedSize = runtimeState.CompressedSize,
```
cgillum: Don't worry about setting `Size` and `CompressedSize`. Those are set by the **DurableTask.ServiceBus** provider but aren't really relevant to the **DurableTask.AzureStorage** provider.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1620,33 +1620,40 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
...
-                Input = runtimeState.Input,
-                Output = runtimeState.Output
-            };
+            OrchestrationInstanceStatus orchestrationInstanceStatus = (OrchestrationInstanceStatus)orchestration.Result;
+            if (orchestrationInstanceStatus != null)
+            {
+                orchestrationState.OrchestrationInstance = new OrchestrationInstance
+                {
+                    InstanceId = instanceId,
+                    ExecutionId = executionId
```
cgillum: It's very likely that the `executionId` parameter will be `null`. For that reason, we should always rely on the ExecutionId value that comes back from table storage.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1785,5 +1792,16 @@ public void Reset()
...
+
+        class OrchestrationInstanceStatus : TableEntity
+        {
+            public string Input { get; set; }
+            public string Output { get; set; }
+            public string CustomStatus { get; set; }
+            public DateTime CreatedTime { get; set; }
+            public DateTime LastUpdatedTime { get; set; }
+            public OrchestrationStatus RuntimeStatus { get; set; }
+
```
cgillum: nit: please remove this extra newline.

gled4er:
Hello @cgillum ,

Thank you very much for the great feedback!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you for catching this.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
gled4er: Yes, you are right. I will write 1 directly:

`this.WriteEvent(135, Account, TaskHub, InstanceId, ExecutionId, 1, EventType, LatencyMs);`

Is that OK with you or your prefer we skip completely 1 as a value we pass.

Thank you!

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
gled4er: Thank you!

I will change it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you! I will change it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Great! I will change it. Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Sure. I will fix it. Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: This clarifies a lot. Thank you very much!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Got it. Thank you very much for the clarification.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you for the clarification. I will add `ExecutionId` column to the Instance table and I will use it here. Let me know if this approach should be modified. Thank you!

gled4er:
Hello @cgillum ,

I addressed all your comments from the last review iteration.

May I ask you to review my changes when you have time?

Thank you!

Kanio

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -880,6 +903,42 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                        ExecutionCompletedEvent executionCompleted = (ExecutionCompletedEvent)historyEvent;
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(executionCompleted.Timestamp);
+                        orchestrationInstanceUpdate.Properties["Output"] = new EntityProperty(executionCompleted.Result);
+                        orchestrationInstanceUpdate.Properties["RuntimeStatus"] = new EntityProperty(executionCompleted.OrchestrationStatus.ToString());
+                        break;
+                    case EventType.ExecutionFailed:
+                        orchestratorEventType = historyEvent.EventType;
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
+                        orchestrationInstanceUpdate.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Failed.ToString());
+                        break;
```
cgillum: Don't we need to set the failure details here as **Output**?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -880,6 +903,42 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                    case EventType.ExecutionTerminated:
+                        orchestratorEventType = historyEvent.EventType;
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
+                        orchestrationInstanceUpdate.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Terminated.ToString());
+                        break;
```
cgillum: **Output** needs to be set to [ExecutionTerminatedEvent.Input](https://github.com/Azure/durabletask/blob/07bb1d34a7da626faf2b2f3cf7dee87df925d782/src/DurableTask.Core/History/ExecutionTerminatedEvent.cs#L47)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -880,6 +903,42 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+                    case EventType.ContinueAsNew:
+                        orchestratorEventType = historyEvent.EventType;
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
+                        orchestrationInstanceUpdate.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Running.ToString());
+                        break;
```
cgillum: Need to set **Output** to the value of `CompletedAsNewEvent.Result`.

gled4er:
Hello @cgillum ,

Thank you very much for the great feedback.

I will work on fixing them.

Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -903,43 +904,39 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
                     case EventType.ContinueAsNew:
                         orchestratorEventType = historyEvent.EventType;
-                        if (!string.IsNullOrEmpty(runtimeState.Input))
-                        {
-                            orchestrationInstanceUpdate.Properties.Add("Input", new EntityProperty(runtimeState.Input));
-                        }
                         orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
-                        orchestrationInstanceUpdate.Properties.Add("CreatedTime", new EntityProperty(runtimeState.CreatedTime));
-                        this.InsertOrUpdateDynamicTableEntityProperty(orchestrationInstanceUpdate, "RuntimeStatus", new EntityProperty(EventType.ContinueAsNew.ToString()));
+                        orchestrationInstanceUpdate.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Running.ToString());
```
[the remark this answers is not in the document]
gled4er: Thank you for the comment. I did the same in the beginning but we have one test - `ActorOrchestration` that is checking if the status is **Running** when we are waiting for an external event:

```csharp
 [TestMethod]
        public async Task ActorOrchestration()
        {
            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost())
            {
                await host.StartAsync();

                int initialValue = 0;
                var client = await host.StartOrchestrationAsync(typeof(Orchestrations.Counter), initialValue);

                // Need to wait for the instance to start before sending events to it.
                // TODO: This requirement may not be ideal and should be revisited.
                await client.WaitForStartupAsync(TimeSpan.FromSeconds(10));

                // Perform some operations
                await client.RaiseEventAsync("operation", "incr");

                // TODO: Sleeping to avoid a race condition where multiple ContinueAsNew messages
                //       are processed by the same instance at the same time, resulting in a corrupt
                //       storage failure in DTFx.
                await Task.Delay(2000);
                await client.RaiseEventAsync("operation", "incr");
                await Task.Delay(2000);
                await client.RaiseEventAsync("operation", "incr");
                await Task.Delay(2000);
                await client.RaiseEventAsync("operation", "decr");
                await Task.Delay(2000);
                await client.RaiseEventAsync("operation", "incr");
                await Task.Delay(2000);

                // Make sure it's still running and didn't complete early (or fail).
                var status = await client.GetStatusAsync();
                Assert.AreEqual(OrchestrationStatus.Running, status?.OrchestrationStatus);

                // The end message will cause the actor to complete itself.
                await client.RaiseEventAsync("operation", "end");

                status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(10));

                Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);
                Assert.AreEqual(3, JToken.Parse(status?.Output));

                // When using ContinueAsNew, the original input is discarded and replaced with the most recent state.
                Assert.AreNotEqual(initialValue, JToken.Parse(status?.Input));

                await host.StopAsync();
            }
        }
```csharp

I will make the change but maybe I am missing something.

Thank you!

File: src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs [hunk 24]
```diff
@@ -903,43 +904,39 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
                 {
                     case EventType.ExecutionStarted:
                         orchestratorEventType = historyEvent.EventType;
-                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(historyEvent.Timestamp);
-                        orchestrationInstanceUpdate.Properties.Add("RuntimeStatus", new EntityProperty(EventType.ExecutionStarted.ToString()));
+                        ExecutionStartedEvent executionStartedEvent = (ExecutionStartedEvent)historyEvent;
+                        orchestrationInstanceUpdate.Properties["Name"] = new EntityProperty(executionStartedEvent.Name);
+                        orchestrationInstanceUpdate.Properties["Version"] = new EntityProperty(executionStartedEvent.Version);
+                        orchestrationInstanceUpdate.Properties["Input"] = new EntityProperty(executionStartedEvent.Input);
+                        orchestrationInstanceUpdate.Properties["LastUpdatedTime"] = new EntityProperty(executionStartedEvent.Timestamp);
```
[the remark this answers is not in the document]
gled4er: Thank you very much for catching that! I will fix it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you very much! I will fix it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you very much! I will change it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Yes, you are right. I am not sure to what type to cast since I don't see `ExecutionFailedEvent` in the project. I think the output will be added by the `ExecutionCompletedEvent`. Do you think this is enough or we need to write the output here as well. The only way I can think of is to use the `OrchestrationRuntimeState` but we decided not to leverage it in the switch block. Please let me know what will be the best approach here. Thank you!

gled4er:
Hello @cgillum ,

I added the requested the changes. May I ask you for additional guidance how to add the details for `EventType.ExecutionFailed` as output?

Please review when you have time.

Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: I'm not sure what the best approach is. I suggest testing/debugging to see what is available. I think it's fine to use `OrchestrationRuntimeState` if that's the only way to get the output.

gled4er:
Hello @cgillum ,

In the last commit I relaxed a bit the condition for the `ActorOrchestration` test to check one of two options for **RuntimeStatus** - Running or **ContinuedAsNew**. Let me know if this needs to be changed. 

Thank you!

cgillum:
Thanks @gled4er, I'm happy with the test change. Regarding `EventType.ExecutionFailed`, I noticed the same thing. It may be possible that it's not actually used. I suggest debugging to see what the history event contains in the failure condition. It may be `EventType.ExecutionCompleted` as you suggested, in which case we just need to make sure that `case` statement does the right thing. Hopefully the automated tests are already configured to help validate this.

gled4er:
Hello @cgillum ,

Thank you for the feedback.

I will research further the behavior around `EventType.ExecutionFailed` and add my changes as soon as possible.

Thank you!

gled4er:
Hello @cgillum ,

I decided to remove the code related to `EventType.ExectuionFailed` because after further debugging I see the Failed status and additional info coming from the `EventType.ExectutionCompleted` event. It already contains everything we need so I think this should be enough.

Please let me know if this is OK and what should be the next issues I need to fix.

Thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -914,6 +969,21 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+
+            this.stats.StorageRequests.Increment();
+            this.stats.TableEntitiesWritten.Increment();
+
+            AnalyticsEventSource.Log.InstancesUpdate(
+                   this.storageAccountName,
+                   this.settings.TaskHubName,
+                   instanceId,
+                   executionId,
+                   orchestratorEventType?.ToString(),
```
cgillum: This API is not allowed to accept nulls. Can you instead specify this?
```csharp
orchestratorEventType?.ToString() ?? string.Empty
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1455,6 +1525,41 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
...
+                }
+            };
+
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            await this.instancesTable.ExecuteAsync(
+                TableOperation.Insert(entity));
+            this.stats.StorageRequests.Increment();
+            this.stats.TableEntitiesWritten.Increment(1);
+
+            AnalyticsEventSource.Log.AppendedInstanceState(
```
cgillum: Should this be `AnalyticsEventSource.Log.InstancesUpdate`?

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -302,5 +302,19 @@ public void LeaseRemovalFailed(string Account, string TaskHub, string WorkerName
             EnsureLogicalTraceActivityId();
             this.WriteEvent(134, Account, TaskHub, WorkerName, PartitionId, Token);
         }
+
+        [Event(135, Level = EventLevel.Informational, Message = "Host '{2}' successfully updated Instances table with '{5}' event")]
+        public void InstancesUpdate(string Account, string TaskHub, string InstanceId, string ExecutionId, string EventType ,long LatencyMs)
```
cgillum: For naming consistency with `FetchedInstanceStatus`, can we rename this to `InstanceStatusUpdate`?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Thank you very much! I will change it!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Great catch! Thank you very much! I will change it!

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
gled4er: Sure. Thank you very much!

gled4er:
Hello @cgillum ,

Thank you for the last round of comments.

I added the requested changes.

Please let me know what should be fixed next.

Thank you!

gled4er:
Hello @cgillum ,

Thank you very much for the great feedback and the support while working on this pull request!

Thank you!

PR #150, opened 2018-03-31
Add the ability to add a custom IOrchestrationInstanceStore for the …

adarsh1:


cgillum:
To be honest I'm not super comfortable with this change. It introduces a lot of forks in the logic. I think a cleaner approach would be to create a sub-class of `AzureStorageOrchestrationService` and override my logic with yours. That might reduce the likelihood of us regressing each other with changes. You could even be able to ship it as a separate nuget package that way if you wanted.

Thoughts?

adarsh1:
Extracted out the table writing functionality to an ITracking Store Interface. Backed by either the history and instance tables or an IOrchestrationInstanceStore.

cgillum:
I like this approach better. But can it wait until after we add [large message support](https://github.com/Azure/durabletask/pull/153)? We need to get that PR done ASAP and I would rather not have to merge it with these fairly substantial refactoring changes.

adarsh1:
What is our timeline for those? I looked through the large message PR and its changes don't seem to intersect all that much with mine.
Another question I had was those changes seem to be merging with the vnext_azurestorage branch, is the latest branch not the azure-functions one? Hoping you can clarify this for me.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1715,12 +1433,13 @@ public Task ForceTerminateTaskOrchestrationAsync(string instanceId, string reaso
         }
 
         /// <summary>
-        /// This method is not supported.
+        /// This method is supported.
```
cgillum: Let's add a description of what the method does in the summary now that it is supported. :)

Test/DurableTask.AzureStorage.Tests/TestHelpers.cs
```diff
@@ -15,11 +15,11 @@ public static TestOrchestrationHost GetTestOrchestrationHost()
...
             var service = new AzureStorageOrchestrationService(
                 new AzureStorageOrchestrationServiceSettings
-                {
+               {
                     StorageConnectionString = storageConnectionString,
                     TaskHubName = ConfigurationManager.AppSettings.Get("TaskHubName"),
-                });
-
+               });
+            //, new AzureTableInstanceStore(ConfigurationManager.AppSettings.Get("TaskHubName"), storageConnectionString)
```
cgillum: Should this commented-out line be deleted?

src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
```diff
@@ -1,10 +1,24 @@
...
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
 
 namespace DurableTask.AzureStorage
 {
+    using System;
+    using Microsoft.WindowsAzure.Storage.Table;
+
+    /// <summary>
+    /// Table Entity Representation of an organization Instance Status
```
cgillum: Did you mean to say "Table entity representation of an **orchestration** instance status"?

PR #153, opened 2018-04-04
Adding support for large message type in Durable Functions

tohling:
1. If message size >60KB, message will be compressed and uploaded to blob storage.
2. Added tests for large text and binary messages.


Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -323,6 +324,130 @@ public class AzureStorageScenarioTests
...
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost())
+            {
+                await host.StartAsync();
+
+                // Generate a medium random string payload
+                const int TargetPayloadSize = 128 * 1024; // 128 KB
+                const string Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 {}/<>.-";
+                var sb = new StringBuilder();
+                var random = new Random();
+                while (Encoding.UTF8.GetByteCount(sb.ToString()) < TargetPayloadSize)
```
cgillum: This should be `Encoding.Unicode`, not `Encoding.UTF8`, right?

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -323,6 +324,130 @@ public class AzureStorageScenarioTests
...
+        /// </summary>
+        [TestMethod]
+        public async Task LargeBinaryByteMessagePayloads()
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost())
+            {
+                await host.StartAsync();
+
+                // Construct byte array from large binary file 
+                string originalFileName = "jpn.pdf";
```
cgillum: How big is this file? Can you mention that in a comment?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -67,6 +68,10 @@ public class AzureStorageOrchestrationService :
 
         readonly object hubCreationLock;
 
+        readonly string InputProperty = "Input";
+        readonly string ResultProperty = "Result";
+        readonly string BlobNameProperty = "BlobName";
```
cgillum: Shouldn't these all be `const`?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -774,6 +781,15 @@ PendingMessageBatch StashMessagesAndGetNextBatch(IEnumerable<MessageData> queueM
...
 
+                // Find any table entitiy properties that have been compressed
+                if (mostRecentGeneration
+                    .Where(tableEntity => this.HasCompressedTableEntityProperty(tableEntity))
+                    .Select(e => e).ToList().Any())
+                {
+                    // Decompress the compressed table entity properties
+                    this.DecompressTableEntitiesProperties(mostRecentGeneration);
+                }
+
```
cgillum: There is a lot of redundant looping going on here. I don't think there is any reason we should be looping through these entities more than once. I suggest removing the LINQ calls and doing a single `foreach (DynamicTableEntity e in mostRecentGeneration)` and do all the checks/conversions inline. The output of this `foreach` loop should be the final `List<HistoryEvent>`.

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -60,5 +91,112 @@ public static MessageData DeserializeQueueMessage(CloudQueueMessage queueMessage
...
+                            SharedBufferManager.Instance.ReturnBuffer(buffer);
+                        }
+
+                    }
+                    return memory.ToArray();
+                }
+            }
+        }
+
+        internal static string DownloadAndDecompressAsBytes(string storageConnectionString, string blobName)
```
cgillum: This needs to be async.

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -60,5 +91,112 @@ public static MessageData DeserializeQueueMessage(CloudQueueMessage queueMessage
...
+        {
+            if (!string.IsNullOrEmpty(data) && Encoding.UTF8.GetByteCount(data) > MaxStorageQueuePayloadSizeInBytes)
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        internal static void CompressAndUploadAsBytes(string storageConnectionString, byte[] payloadBuffer, string activationId)
```
cgillum: I would make this a private method of AzureStorageOrchestrationService rather than a public static method of Utils. I think that may simplify a lot of things that are going on here.

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -15,36 +15,67 @@ namespace DurableTask.AzureStorage
...
         {
             TypeNameHandling = TypeNameHandling.Objects
         };
 
-        public static string SerializeMessageData(MessageData messageData)
+        public static string SerializeMessageData(string storageConnectionString, MessageData messageData)
         {
             string rawContent = JsonConvert.SerializeObject(messageData, TaskMessageSerializerSettings);
-            return rawContent;
+            messageData.TotalMessageSizeBytes = Encoding.UTF8.GetByteCount(rawContent);
```
cgillum: I see you're using `Encoding.UTF8` in a lot of places. Shouldn't these all be `Encoding.Unicode`. Azure Storage always stores string data as UTF-16, which corresponds to `Encoding.Unicode`.

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -15,36 +15,67 @@ namespace DurableTask.AzureStorage
...
+    using System.IO.Compression;
     using System.Text;
     using System.Threading.Tasks;
     using Microsoft.WindowsAzure.Storage.Queue;
     using Newtonsoft.Json;
 
     static class Utils
     {
+        const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
+        const int DefaultBufferSize = 4096;
```
cgillum: Given that your change only involves payloads that are 60KB or greater, it would seem more appropriate and performant to use a larger buffer size, like 64 KB.

src/DurableTask.AzureStorage/StorageUtil.cs
```diff
@@ -0,0 +1,126 @@
...
+        private static CloudBlobClient cloudBlobClient;
+        private static bool isInitialized;
+        private const string DefaultContainerName = "durable-compressedmessage-container";
+
+        private StorageUtil() { }
+
+        /// <summary>
+        /// The storage client instance.
+        /// </summary>
+        public static StorageUtil Instance
```
cgillum: Why are you using a singleton pattern here? Instances of your class don't have any state, so I don't see the point of doing this versus just having all static methods.

src/DurableTask.AzureStorage/StorageUtil.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        public bool IsInitialized()
+        {
+            return isInitialized;
+        }
+
+        /// <summary>
+        /// Initializes the storage client.
+        /// </summary>
+        public void Initialize(string connectionString)
```
cgillum: Rather than initializing, why not just pass in a `CloudBlobClient` as a parameter to your methods?  Alternatively, you could make this class a regular class which takes a `CloudBlobClient` in the constructor and save it as a member variable. I think either of those options would be simpler and cleaner.

tohling:
Sorry, it was based off an obsolete branch. New pull request is at https://github.com/Azure/durabletask/pull/154

PR #154, opened 2018-04-05
Adding support for large message type in DurableTask.AzureStorage (updated)

tohling:
1. If message size >60KB, message will be compressed and uploaded to blob storage.
2. Added tests for large text and binary messages.


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -791,6 +802,21 @@ PendingMessageBatch StashMessagesAndGetNextBatch(IEnumerable<MessageData> queueM
                         break;
                     }
 
+                    if (this.HasCompressedTableEntityByPropertyKey(entity, InputProperty))
```
cgillum: >HasCompressedTableEntityByPropertyKey [](start = 29, length = 37)

Why separate "Has" and "Set" methods? I haven't studied the details yet, but it feels like that could be less efficient since you're reading the same property details multiple times.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -971,6 +1017,8 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                 }
             }
 
+            await this.CompressLargeMessageAsync(orchestrationInstanceUpdate);
```
cgillum: >CompressLargeMessageAsync [](start = 23, length = 25)

Is this compression/upload necessary? I still have a lot to review, but my intuition was that this data would only need to be compressed and uploaded once, and both the History and Instances table could reference the same uploaded blob.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1006,7 +1054,7 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                     CloudQueue targetControlQueue = await this.GetControlQueueAsync(targetInstanceId);
 
                     enqueueTasks.Add(targetControlQueue.AddMessageAsync(
-                        context.CreateOutboundQueueMessage(taskMessage, targetControlQueue.Name),
+                        context.CreateOutboundQueueMessageAsync(this.messageManager, taskMessage, targetControlQueue.Name).Result,
```
cgillum: >.Result [](start = 122, length = 7)

Why `.Result` and not `await`?

To make this more efficient, we'll probably need to refactor the body of this `foreach` loop into a separate `async` method to make sure all of these enqueue operations can be run in parallel. The way it's written now we'll only enqueue one message at a time because of the I/O caused by `CreateOutboundQueueMessageAsync`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1030,7 +1078,7 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                     }
 
                     enqueueTasks.Add(currentControlQueue.AddMessageAsync(
-                        context.CreateOutboundQueueMessage(taskMessage, currentControlQueue.Name),
+                        context.CreateOutboundQueueMessageAsync(this.messageManager, taskMessage, currentControlQueue.Name).Result,
```
cgillum: >Result [](start = 124, length = 6)

Same concern about using `.Result`. This is usually very bad practice because it causes threads to get blocked which can lead to a variety of performance and scalability issues.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1045,7 +1093,7 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                 foreach (TaskMessage taskMessage in outboundMessages)
                 {
                     enqueueTasks.Add(this.workItemQueue.AddMessageAsync(
-                        context.CreateOutboundQueueMessage(taskMessage, this.workItemQueue.Name),
+                        context.CreateOutboundQueueMessageAsync(this.messageManager, taskMessage, this.workItemQueue.Name).Result,
```
cgillum: >Result [](start = 123, length = 6)

Same concern about using `.Result`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1059,7 +1107,7 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                 addedControlMessages = true;
 
                 enqueueTasks.Add(currentControlQueue.AddMessageAsync(
-                    context.CreateOutboundQueueMessage(continuedAsNewMessage, currentControlQueue.Name),
+                    context.CreateOutboundQueueMessageAsync(this.messageManager, continuedAsNewMessage, currentControlQueue.Name).Result,
```
cgillum: >.Result [](start = 129, length = 7)

Same concern about using `.Result`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1121,6 +1185,74 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
             }
         }
 
+        void SetProperyMessageToEmptyString(DynamicTableEntity entity)
```
cgillum: >SetProperyMessageToEmptyString [](start = 13, length = 30)

Typo. Should be `SetPropertyMessageToEmptyString`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1121,6 +1185,74 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+            }
+        }
+
+        byte[] GetPropertyMessageAsBytes(DynamicTableEntity entity)
+        {
+            byte[] messageBytes = new byte[0];
+
+            if (entity.Properties.ContainsKey(InputProperty) && entity.Properties[InputProperty] != null)
+            {
+                messageBytes = Encoding.Unicode.GetBytes(entity.Properties[InputProperty].StringValue);
```
cgillum: >Unicode [](start = 40, length = 7)

UTF8 is a better choice here because it will result in fewer bytes and therefore smaller blobs.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1121,6 +1185,74 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
+            }
+
+            return null;
+        }
+
+        string GetLargeTableEntityInternal(DynamicTableEntity entity, string propertyKey)
+        {
+            if (entity.Properties.ContainsKey(propertyKey)
+                && entity.Properties[propertyKey] != null
+                && this.ExceedsMaxTableEntitySize(entity.Properties[propertyKey].StringValue))
```
cgillum: This code is doing 3 dictionary lookups in every call. :(  Please refactor to only do a single lookup per call. For example:

```csharp
if (entity.Properties.TryGetValue(propertyKey, out EntityProperty value)
    && this.ExceedsMaxTableEntitySize(value?.StringValue))
{
    return propertyKey;
}

return null;
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1308,7 +1441,7 @@ public Task ReleaseTaskOrchestrationWorkItemAsync(TaskOrchestrationWorkItem work
             try
             {
                 await controlQueue.AddMessageAsync(
-                    context.CreateOutboundQueueMessage(responseTaskMessage, controlQueue.Name),
+                    context.CreateOutboundQueueMessageAsync(this.messageManager, responseTaskMessage, controlQueue.Name).Result,
```
cgillum: >.Result [](start = 120, length = 7)

Same comment about using `.Result`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1638,13 +1774,33 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
                 orchestrationState.Status = orchestrationInstanceStatus.CustomStatus;
                 orchestrationState.CreatedTime = orchestrationInstanceStatus.CreatedTime;
                 orchestrationState.LastUpdatedTime = orchestrationInstanceStatus.LastUpdatedTime;
-                orchestrationState.Input = orchestrationInstanceStatus.Input;
-                orchestrationState.Output = orchestrationInstanceStatus.Output;
+                orchestrationState.Input = await this.GetOrchestrationInputAsync(orchestrationInstanceStatus);
+                orchestrationState.Output = await this.GetOrchestrationOutputAsync(orchestrationInstanceStatus);
```
cgillum: Consider doing these two async operations in parallel.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -20,6 +20,10 @@
     <PackageReference Include="WindowsAzure.Storage" version="8.5.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.ServiceModel.Primitives" Version="4.4.1" />
```
cgillum: >System.ServiceModel.Primitives [](start = 31, length = 30)

Is this nuget reference required when targeting net451? I haven't researched it, but I assumed System.ServiceModel.dll already had everything that we needed for net451 and that a package like this could be included only when targeting netstandard2.0.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// The message manager for messages from MessageData, and DynamicTableEntities
+    /// </summary>
+    public class MessageManager
+    {
+        readonly CloudBlobContainer cloudBlobContainer;
+        readonly JsonSerializerSettings taskMessageSerializerSettings;
+        const string DefaultContainerName = "durable-compressedmessage-container";
```
cgillum: >durable-compressedmessage-container [](start = 45, length = 35)

The convention we have right now is that all storage entities are prefixed with the task hub name. That way we can easily delete all resources associated with a task hub. It also provides greater isolation if two function apps are using the same storage account.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+            string rawContent = JsonConvert.SerializeObject(messageData, this.taskMessageSerializerSettings);
+            messageData.TotalMessageSizeBytes = Encoding.Unicode.GetByteCount(rawContent);
+            MessageFormatFlags messageFormat = this.GetMessageFormatFlags(messageData);
+
+            if (messageFormat != MessageFormatFlags.InlineJson)
+            {
+                byte[] messageBytes = Encoding.Unicode.GetBytes(rawContent);
+
+                // Get Compressed bytes
+                string blobName = messageData.ActivityId.ToString();
```
cgillum: >ActivityId [](start = 46, length = 10)

Is using the ActivityId for the blob name safe? How do we ensure there are never duplicates?

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+                    return memory.ToArray();
+                }
+            }
+        }
+
+        internal async Task<string> DownloadAndDecompressAsBytesAsync(string blobName)
+        {
+            byte[] downloadBlobAsAsBytes = await this.DownloadBlobAsync(blobName);
+            byte[] decompressedBytes = this.Decompress(downloadBlobAsAsBytes);
+            return Encoding.Unicode.GetString(decompressedBytes);
```
cgillum: >Unicode [](start = 28, length = 7)

We should always be using UTF8 for serialization.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        /// <summary>
+        /// Downloads MessageData as bytes[] 
+        /// </summary>
+        internal async Task<byte[]> DownloadBlobAsync(string blobName)
+        {
+            CloudBlockBlob cloudBlockBlob = this.cloudBlobContainer.GetBlockBlobReference(blobName);
+
+            using (MemoryStream ms = new MemoryStream())
```
cgillum: >new MemoryStream() [](start = 37, length = 18)

We know that these blobs are going to be large, so to avoid unnecessary byte[] allocations, we should at least allocate a minimum of MaxStorageQueuePayloadSizeInBytes for each `MemoryStream` object we create:

```
new MemoryStream(MaxStorageQueuePayloadSizeInBytes)
```

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+        internal async Task<string> DownloadAndDecompressAsBytesAsync(string blobName)
+        {
+            byte[] downloadBlobAsAsBytes = await this.DownloadBlobAsync(blobName);
+            byte[] decompressedBytes = this.Decompress(downloadBlobAsAsBytes);
+            return Encoding.Unicode.GetString(decompressedBytes);
+        }
+
+        internal byte[] Decompress(byte[] gzip)
+        {
+            using (GZipStream stream = new GZipStream(new MemoryStream(gzip), CompressionMode.Decompress))
```
cgillum: >new MemoryStream(gzip) [](start = 54, length = 22)

If we're going to wrap these bytes in a MemoryStream anyways, why not just pass in a MemoryStream as a parameter rather than byte[]? That could lead to some additional performance improvements.

cgillum: In fact, I would take this one step further and say it would be much more CPU and memory efficient if we passed in the blob stream directly. Then we don't have to copy so many bytes around or create so many large objects in the CLR's managed heap.

---
In reply to: [180174633](https://github.com/Azure/durabletask/pull/154#discussion_r180174633) [](ancestors = 180174633)

cgillum: >stream [](start = 30, length = 6)

Please name this `gzipStream` to help remove ambiguity between the `MemoryStream` and the `GZipStream`.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                using (MemoryStream memory = new MemoryStream())
```
cgillum: >new MemoryStream() [](start = 45, length = 18)

We know that these blobs are going to be large, so to avoid unnecessary byte[] allocations, we should at least allocate a **minimum** of MaxStorageQueuePayloadSizeInBytes for each `MemoryStream` object we create:

```
new MemoryStream(MaxStorageQueuePayloadSizeInBytes)
```

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+        /// </summary>
+        public async Task<string> SerializeMessageDataAsync(MessageData messageData)
+        {
+            string rawContent = JsonConvert.SerializeObject(messageData, this.taskMessageSerializerSettings);
+            messageData.TotalMessageSizeBytes = Encoding.Unicode.GetByteCount(rawContent);
+            MessageFormatFlags messageFormat = this.GetMessageFormatFlags(messageData);
+
+            if (messageFormat != MessageFormatFlags.InlineJson)
+            {
+                byte[] messageBytes = Encoding.Unicode.GetBytes(rawContent);
```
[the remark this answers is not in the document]
tohling: I used UTF8 in my previous PR and changed all encoding to Unicode per your comment at [https://github.com/Azure/durabletask/pull/153#discussion_r179272718](https://github.com/Azure/durabletask/pull/153#discussion_r179272718) 

Kindly re-cconfirm that you want to use UTF8 instead of Unicode.

---
In reply to: [180168798](https://github.com/Azure/durabletask/pull/154#discussion_r180168798) [](ancestors = 180168798)

src/DurableTask.AzureStorage/MessageManager.cs
cgillum: After further review, I think it would actually be much better to rewrite this method to return the stream directly. Something like this:

return cloudBlockBlob.OpenReadAsync();

---
In reply to: [180174017](https://github.com/Azure/durabletask/pull/154#discussion_r180174017) [](ancestors = 180174017)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Sorry, I was refactoring from previous PR and left this in-flight change in by accident.  Will fix for the rest of similar issues.

---
In reply to: [180161664](https://github.com/Azure/durabletask/pull/154#discussion_r180161664) [](ancestors = 180161664)

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+
+                            gzip.Flush();
+                        }
+                        finally
+                        {
+                            SharedBufferManager.Instance.ReturnBuffer(buffer);
+                        }
+
+                    }
+                    return memory.ToArray();
```
cgillum: >memory.ToArray(); [](start = 27, length = 17)

I mentioned this elsewhere, but it's more efficient to return an ArraySegment<byte>:
``
return new ArraySegment<byte>(memory.GetBuffer(), 0, bytesWritten);
```

File: src/DurableTask.AzureStorage/MessageManager.cs [hunk 21]
```diff
@@ -0,0 +1,202 @@
...
+                envelope.MessageFormat = MessageFormatFlags.StorageBlob;
+            }
+
+            envelope.OriginalQueueMessage = queueMessage;
+            envelope.TotalMessageSizeBytes = Encoding.Unicode.GetByteCount(queueMessage.AsString);
+            envelope.QueueName = queueName;
+            return envelope;
+        }
+
+        internal async Task<string> CompressAndUploadAsBytesAsync(byte[] payloadBuffer, string activationId)
```
cgillum: >activationId [](start = 95, length = 12)

`activationId` isn't really meaningful in this context. I would name this parameter `blobName`.

src/DurableTask.AzureStorage/ReceivedMessageContext.cs
```diff
@@ -180,12 +182,13 @@ public DateTime GetNextMessageExpirationTimeUtc()
             }
         }
 
-        public CloudQueueMessage CreateOutboundQueueMessage(TaskMessage taskMessage, string queueName)
+        public async Task<CloudQueueMessage> CreateOutboundQueueMessageAsync(MessageManager messageManager, TaskMessage taskMessage, string queueName)
         {
-            return CreateOutboundQueueMessageInternal(this.storageAccountName, this.taskHub, queueName, taskMessage);
+            return await CreateOutboundQueueMessageInternalAsync(messageManager, this.storageAccountName, this.taskHub, queueName, taskMessage);
```
cgillum: >await [](start = 19, length = 5)

Please remove the `async` keyword from this method and return the task directly. Using `await` has a non-zero runtime cost associated with it and makes debugging more painful.

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// Returns a buffer to the buffer pool.
+        /// </summary>
+        public void ReturnBuffer(byte[] buffer)
+        {
+            // Return buffer to pool, and set the memory to null.
+            // Reference: https://www.wintellect.com/pooling-buffers-for-better-memory-management/
+            lock (sharedlock)
+            {
+                innerBufferManager.ReturnBuffer(buffer);
+                buffer = null;
```
cgillum: >buffer = null; [](start = 16, length = 14)

Why are you assigning this local variable to null?

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage
+{
+    using System;
+    using System.ServiceModel.Channels;
+    using System.Timers;
```
cgillum: >using System.Timers; [](start = 4, length = 20)

Use of System.Timers.Timer is generally not recommended. System.Threading.Timer is preferred and more likely to remain compatible with .NET Core.

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+            DateTime scheduledTime = DateTime.UtcNow;
+            if (nowTime >= scheduledTime)
+            {
+                scheduledTime = scheduledTime.AddMinutes(defaultClearBufferIntervalInMinutes);
+            }
+
+            double tickTime = (double)(scheduledTime - DateTime.Now).TotalMilliseconds;
+            timer = new Timer(tickTime);
+            timer.Elapsed += new ElapsedEventHandler(TimerElapsed);
+            timer.Start();
```
cgillum: This method is largely nonsensical to me, from start to finish. I assume it was copied from a blog post, which is a dangerous practice if the code is not fully understood.  I think this entire method can be removed.

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        /// <summary>
+        /// Initiate clearing buffer pool when timer fires
+        /// </summary>
+        internal static void TimerElapsed(object sender, ElapsedEventArgs e)
+        {
+            timer.Stop();
+            Clear();
+            ScheduleTimer();
```
cgillum: There is no need to stop and restart the timer since Clear() is an instantaneous operation.

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+        internal static void Initialize(long maxBufferPoolSize, int maxBufferSize)
+        {
+            ValidateParams(maxBufferPoolSize, maxBufferSize);
+            lock (sharedlock)
+            {
+                innerBufferManager = BufferManager.CreateBufferManager(maxBufferPoolSize, maxBufferSize);
+                isInitialized = true;
+            }
+            ScheduleTimer();
+        }
```
cgillum: Why do you need a separate Initialize method? Can't all this be done directly in the constructor? This really feels over-engineered and unnecessarily complex.

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+        static SharedBufferManager instance = null;
+        static readonly object padlock = new object();
+
+        static Timer timer;
+        static readonly object sharedlock = new object();
+        static BufferManager innerBufferManager;
+        static bool isInitialized;
+        static readonly long defaultMaxBufferPoolSize = 0; // Avoid pooling for large messages, reference: http://andriybuday.com/2011/08/wcf-configuration-caused-memory-leaks.html
+        static readonly int defaultBufferSize = 256 * 1024; // 256 KB
+        static readonly int defaultClearBufferIntervalInMinutes = 10;
```
cgillum: >static readonly [](start = 8, length = 15)

Shouldn't these all be `const`?

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Checks if buffer is initialized
+        /// </summary>
+        public bool IsInitialized()
+        {
+            return isInitialized;
+        }
```
cgillum: When will this ever return `false`?

src/DurableTask.AzureStorage/SharedBufferManager.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            timer.Stop();
+            Clear();
+            ScheduleTimer();
+        }
+
+        /// <summary>
+        /// Validates the maximum buffer pool and maximum buffer size.
+        /// </summary>
+        internal static void ValidateParams(long maxBufferPoolSize, int maxBufferSize)
```
cgillum: >internal [](start = 8, length = 8)

Why internal and not private? Can we remove this method and put argument validation in the constructor?

src/DurableTask.AzureStorage/MessageManager.cs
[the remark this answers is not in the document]
cgillum: There are places in this PR where we need Unicode/UTF16 and places were we should use UTF8. When checking to see whether a string is too large to put into a queue or table, we need to use Unicode/UTF16 since that is the string encoding used by queue and table storage. That was the context of my comment in the previous PR. For blobs we are uploading binary data, not strings, and want to use the more efficient UTF8 encoding.

---
In reply to: [180180787](https://github.com/Azure/durabletask/pull/154#discussion_r180180787) [](ancestors = 180180787,180168798)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

I have refactored this to read the blobNameKey property only once in the HasCompressedTableEntityByPropertyKey method.

Details:
Sorry if this looks confusing. Let me give more context.

For the method HasCompressedTableEntityByPropertyKey:
What I am trying to do is check if the table entity has a blobName property for the given property key. e.g. If there is a compressed blob for the "Input" property, then the table entity should have the following property structures

1. Input = "" //empty string
2. InputBlobName = <some Guid representing the blobName of the compressed large message>

For the method SetDecompressedTableEntityAsync:
Here I want to replace the "Input" property with the original large message (decompressed), such that

1. Input = <original large message>
2. InputBlobName = <some Guid representing the blobName of the compressed large message>




---
In reply to: [180158101](https://github.com/Azure/durabletask/pull/154#discussion_r180158101) [](ancestors = 180158101)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed. 

I have removed this and added the code to re-use the already compressed message in the construction of the 'orchestrationInstanceUpdate' in the SWITCH statement.

---
In reply to: [180159767](https://github.com/Azure/durabletask/pull/154#discussion_r180159767) [](ancestors = 180159767)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180182936](https://github.com/Azure/durabletask/pull/154#discussion_r180182936) [](ancestors = 180182936,180161664)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180162055](https://github.com/Azure/durabletask/pull/154#discussion_r180162055) [](ancestors = 180162055)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180162117](https://github.com/Azure/durabletask/pull/154#discussion_r180162117) [](ancestors = 180162117)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180162191](https://github.com/Azure/durabletask/pull/154#discussion_r180162191) [](ancestors = 180162191)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180165251](https://github.com/Azure/durabletask/pull/154#discussion_r180165251) [](ancestors = 180165251)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180165637](https://github.com/Azure/durabletask/pull/154#discussion_r180165637) [](ancestors = 180165637)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180165945](https://github.com/Azure/durabletask/pull/154#discussion_r180165945) [](ancestors = 180165945)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180162997](https://github.com/Azure/durabletask/pull/154#discussion_r180162997) [](ancestors = 180162997)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180162461](https://github.com/Azure/durabletask/pull/154#discussion_r180162461) [](ancestors = 180162461)

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
tohling: Fixed.

---
In reply to: [180167159](https://github.com/Azure/durabletask/pull/154#discussion_r180167159) [](ancestors = 180167159)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed. I have removed this.

---
In reply to: [180167869](https://github.com/Azure/durabletask/pull/154#discussion_r180167869) [](ancestors = 180167869)

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+        const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
+        const int DefaultBufferSize = 64 * 2014; // 64KB
+
+        /// <summary>
+        /// The message manager.
+        /// </summary>
+        public MessageManager(CloudBlobClient cloudBlobClient)
+        {
+            this.cloudBlobContainer = cloudBlobClient.GetContainerReference(DefaultContainerName);
+            this.cloudBlobContainer.CreateIfNotExistsAsync();
```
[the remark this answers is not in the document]
tohling: Fixed.

---
In reply to: [180168273](https://github.com/Azure/durabletask/pull/154#discussion_r180168273) [](ancestors = 180168273)

src/DurableTask.AzureStorage/MessageManager.cs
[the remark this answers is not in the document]
tohling: Fixed.

---
In reply to: [180200886](https://github.com/Azure/durabletask/pull/154#discussion_r180200886) [](ancestors = 180200886,180180787,180168798)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

---
In reply to: [180169836](https://github.com/Azure/durabletask/pull/154#discussion_r180169836) [](ancestors = 180169836)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

---
In reply to: [180187482](https://github.com/Azure/durabletask/pull/154#discussion_r180187482) [](ancestors = 180187482)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed

---
In reply to: [180187066](https://github.com/Azure/durabletask/pull/154#discussion_r180187066) [](ancestors = 180187066)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

---
In reply to: [180170410](https://github.com/Azure/durabletask/pull/154#discussion_r180170410) [](ancestors = 180170410)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

---
In reply to: [180176727](https://github.com/Azure/durabletask/pull/154#discussion_r180176727) [](ancestors = 180176727)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

---
In reply to: [180177580](https://github.com/Azure/durabletask/pull/154#discussion_r180177580) [](ancestors = 180177580)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

Thanks, I really like this idea! :-)

---
In reply to: [180175654](https://github.com/Azure/durabletask/pull/154#discussion_r180175654) [](ancestors = 180175654,180174633)

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,202 @@
...
+                            }
+                        }
+                        while (count > 0);
+
+                    }
+                    finally
+                    {
+                        SharedBufferManager.Instance.ReturnBuffer(buffer);
+                    }
+                    return memory.ToArray();
```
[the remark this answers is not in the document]
tohling: Fixed.

---
In reply to: [180185752](https://github.com/Azure/durabletask/pull/154#discussion_r180185752) [](ancestors = 180185752)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

Used OpenReadAsync per your recommendation at Line 128 and removed this method as it is not needed anymore.

---
In reply to: [180180830](https://github.com/Azure/durabletask/pull/154#discussion_r180180830) [](ancestors = 180180830,180174017)

src/DurableTask.AzureStorage/ReceivedMessageContext.cs
tohling: Fixed.

---
In reply to: [180189529](https://github.com/Azure/durabletask/pull/154#discussion_r180189529) [](ancestors = 180189529)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed. Removed as it is not needed anymore.

---
In reply to: [180192308](https://github.com/Azure/durabletask/pull/154#discussion_r180192308) [](ancestors = 180192308)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed.

---
In reply to: [180197079](https://github.com/Azure/durabletask/pull/154#discussion_r180197079) [](ancestors = 180197079)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed. Removed.

---
In reply to: [180197378](https://github.com/Azure/durabletask/pull/154#discussion_r180197378) [](ancestors = 180197378)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed.

---
In reply to: [180190099](https://github.com/Azure/durabletask/pull/154#discussion_r180190099) [](ancestors = 180190099)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed.

---
In reply to: [180196869](https://github.com/Azure/durabletask/pull/154#discussion_r180196869) [](ancestors = 180196869)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed.

---
In reply to: [180195597](https://github.com/Azure/durabletask/pull/154#discussion_r180195597) [](ancestors = 180195597)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed.

---
In reply to: [180197546](https://github.com/Azure/durabletask/pull/154#discussion_r180197546) [](ancestors = 180197546)

src/DurableTask.AzureStorage/SharedBufferManager.cs
tohling: Fixed.

---
In reply to: [180196200](https://github.com/Azure/durabletask/pull/154#discussion_r180196200) [](ancestors = 180196200)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -111,10 +120,16 @@ public AzureStorageOrchestrationService(AzureStorageOrchestrationServiceSettings
             string instancesTableName = $"{settings.TaskHubName}Instances";
             NameValidator.ValidateTableName(instancesTableName);
 
+            string compressedMessageBlobContainerName = $"{settings.TaskHubName.ToLowerInvariant()}-compressedmessages";
```
cgillum: >compressedmessages [](start = 100, length = 18)

Let's call this "largeMessages" instead of "compressedMessages" since that's the primary purpose of this container.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1085,6 +1117,55 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
...
 
+        async Task EnqueueMessageAsync(ReceivedMessageContext context, CloudQueue queue, TaskMessage taskMessage, TimeSpan? initialVisibilityDelay)
+        {
+            CloudQueueMessage message = await context.CreateOutboundQueueMessageAsync(this.messageManager, taskMessage, queue.Name);
+
+            await queue.AddMessageAsync(
+                message,
+                null /* timeToLive */,
+                initialVisibilityDelay,
+                this.settings.ControlQueueRequestOptions,
```
cgillum: >this.settings.ControlQueueRequestOptions [](start = 16, length = 40)

Looks like this needs to be parameterized as well. There is one case where we need `this.settings.WorkItemQueueRequestOptions`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -791,6 +807,23 @@ PendingMessageBatch StashMessagesAndGetNextBatch(IEnumerable<MessageData> queueM
                         break;
                     }
 
+                    string blobName;
```
cgillum: >blobName [](start = 27, length = 8)

If I'm reading this correctly, shouldn't this be `blobNameKey` rather than `blobName`?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1789,11 +1967,23 @@ class OrchestrationInstanceStatus : TableEntity
...
+            public string OutputBlobName { get; set; }
             public string CustomStatus { get; set; }
             public DateTime CreatedTime { get; set; }
             public DateTime LastUpdatedTime { get; set; }
             public string RuntimeStatus { get; set; }
         }
+
+        internal bool ExceedsMaxTableEntitySize(string data)
+        {
+            if (!string.IsNullOrEmpty(data) && Encoding.UTF8.GetByteCount(data) > MaxStorageQueuePayloadSizeInBytes)
```
cgillum: >UTF8 [](start = 56, length = 4)

This needs to be Unicode. We always use Unicode when comparing to the max message size. UTF-8 is only for the serialization encoding.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            envelope.OriginalQueueMessage = queueMessage;
+            envelope.TotalMessageSizeBytes = Encoding.Unicode.GetByteCount(queueMessage.AsString);
+            envelope.QueueName = queueName;
+            return envelope;
+        }
+
+        internal async Task CompressAndUploadAsBytesAsync(byte[] payloadBuffer, string blobName)
+        {
+            byte[] compressedBytes = this.Compress(payloadBuffer);
+            await this.UploadToBlobAsync(compressedBytes, compressedBytes.Length, blobName);
```
cgillum: >await [](start = 12, length = 5)

Please return the task directly instead of awaiting it. Consider this a best practice if nothing needs to happen after the await.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -356,6 +357,130 @@ public void ValidateEventSource()
...
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost())
+            {
+                await host.StartAsync();
+
+                // Generate a small random string payload
+                const int TargetPayloadSize = 1 * 1024; // 1 KB
+                const string Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 {}/<>.-";
+                var sb = new StringBuilder();
+                var random = new Random();
+                while (Encoding.UTF8.GetByteCount(sb.ToString()) < TargetPayloadSize)
```
cgillum: >UTF8 [](start = 32, length = 4)

You still need to use Unicode here and below since you're testing against the storage payload size limits. UTF8 is for serialization.

tohling: Fixed.

---
In reply to: [180295411](https://github.com/Azure/durabletask/pull/154#discussion_r180295411) [](ancestors = 180295411)

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                        {
+                            memory.Write(buffer, 0, count);
+                        }
+                    }
+                    finally
+                    {
+                        this.sharedBufferManager.ReturnBuffer(buffer);
+                    }
+
+                    return new ArraySegment<byte>(memory.GetBuffer(), 0, (int)memory.Length).Array;
```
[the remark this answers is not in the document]
tohling: Fixed.

---
In reply to: [180293664](https://github.com/Azure/durabletask/pull/154#discussion_r180293664) [](ancestors = 180293664)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180292943](https://github.com/Azure/durabletask/pull/154#discussion_r180292943) [](ancestors = 180292943)

src/DurableTask.AzureStorage/MessageManager.cs
tohling: Fixed.

---
In reply to: [180293155](https://github.com/Azure/durabletask/pull/154#discussion_r180293155) [](ancestors = 180293155)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.


---
In reply to: [180292449](https://github.com/Azure/durabletask/pull/154#discussion_r180292449) [](ancestors = 180292449)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Ah, good catch! Fixed.

---
In reply to: [180291751](https://github.com/Azure/durabletask/pull/154#discussion_r180291751) [](ancestors = 180291751)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
tohling: Fixed.

---
In reply to: [180287332](https://github.com/Azure/durabletask/pull/154#discussion_r180287332) [](ancestors = 180287332)

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            return this.UploadToBlobAsync(compressedSegment.Array, compressedSegment.Count, blobName);
+        }
+
+        internal ArraySegment<byte> Compress(byte[] payloadBuffer)
+        {
+            using (var originStream = new MemoryStream(payloadBuffer, 0, payloadBuffer.Length))
+            {
+                using (MemoryStream memory = new MemoryStream())
+                {
+                    using (GZipStream gZipStream = new GZipStream(memory, CompressionLevel.Optimal, true))
```
cgillum: >true [](start = 100, length = 4)

Any particular reason you changed this to leave the stream open?

tohling: Yes, it gets closed by default. I needed to open it so that I can read memory.Length

cgillum: Just curious: why did you decide to leave the stream open instead of capturing the length after the call to `gZipStream.Flush()`?

---
In reply to: [180477920](https://github.com/Azure/durabletask/pull/154#discussion_r180477920) [](ancestors = 180477920)

tohling: Per our conversation offline, this was necessary since the memory.Length obtained after gZipStream.Flush() was not accurate.

cgillum:
This PR was re-submitted via https://github.com/Azure/durabletask/pull/159.

PR #160, opened 2018-04-11
Adding ETW traces for uploading and downloading large messages to blob.

tohling:
Added ETW traces when large messages are uploaded and downloaded from blob storage.

Sample snapshot from PerfView for large message **Upload**:

etw-upload

Sample snapshot from PerfView for large message **Download**:

etw-download

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -128,7 +128,7 @@ namespace DurableTask.AzureStorage
 
             this.instancesTable = tableClient.GetTableReference(instancesTableName);
 
-            this.messageManager = new MessageManager(this.blobClient, compressedMessageBlobContainerName);
+            this.messageManager = new MessageManager(this.blobClient, compressedMessageBlobContainerName, this.storageAccountName, this.settings.TaskHubName);
 
             this.pendingOrchestrationMessageBatches = new LinkedList<PendingMessageBatch>();
             this.activeOrchestrationInstances = new ConcurrentDictionary<string, object>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -937,7 +944,11 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
                 HistoryEvent historyEvent = newEvents[i];
                 DynamicTableEntity entity = this.tableEntityConverter.ConvertToTableEntity(historyEvent);
 
-                await this.CompressLargeMessageAsync(entity);
+                await this.CompressLargeMessageAsync(
+                    entity, 
+                    historyEvent.EventType.ToString() ?? string.Empty,
```
cgillum: What do you expect to be potentially null in this expression?

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -122,11 +146,22 @@ internal ArraySegment<byte> Compress(byte[] payloadBuffer)
...
+
+            AnalyticsEventSource.Log.DownloadLargeMessage(
+                this.storageAccountName,
+                this.taskHubName,
+                eventType,
+                instanceId,
+                executionId,
+                this.cloudBlobContainer.Name,
+                blobName,
+                decompressedSegment.Count);
```
cgillum: This shows the decompressed byte count where as the upload trace shows the compressed byte count. I wonder if we should always show the compressed byte count? Or even both? I think you should be able to get the compressed byte count from `downloadBlobAsStream.Length`.

src/DurableTask.AzureStorage/ReceivedMessageContext.cs
```diff
@@ -133,7 +133,7 @@ public static ReceivedMessageContext CreateFromReceivedMessageBatch(string stora
             CloudQueueMessage queueMessage,
             string queueName)
         {
-            MessageData data = await messageManager.DeserializeQueueMessageAsync(queueMessage, queueName);
+            MessageData data = await messageManager.DeserializeQueueMessageAsync(queueMessage, queueName, null, null, null);
```
cgillum: It's worrisome that we have to pass nulls in so many places. It makes this trace a lot less useful if we can't correlate it to anything. I wonder if we need to investigate an alternate approach. Maybe instead of two new ETW events, we should expand our existing ones: `SendingMessage`, `ReceivedMessage`, `FetchedInstanceState`, `AppendedInstanceState`, `InstanceStatusUpdate`, and `FetchedInstanceStatus`. Those are all places where your new large message logic might take place, right?  I'm wondering if we can simply add **BlobName**, **BlobContainer**, and **SizeInBytes** to those events. Would that be possible?

cgillum:
Closing this PR since it is now stale.

PR #166, opened 2018-04-17
Adding first logic for avoiding split brain.

gled4er:
Hello @cgillum,

May I ask you when you have time to take a look on the changes I started to make for this issue - [#38](https://github.com/Azure/azure-functions-durable-extension/issues/38)?

I want to hear your opinion about the exception handling and if I am addressing all the cases you had in mind.

Thank you!

### DIFF
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -148,17 +148,17 @@ namespace DurableTask.AzureStorage
         }
 
         [Event(110, Level = EventLevel.Informational)]
-        public void FetchedInstanceState(string Account, string TaskHub, string InstanceId, string ExecutionId, int EventCount, int RequestCount, long LatencyMs)
+        public void FetchedInstanceState(string Account, string TaskHub, string InstanceId, string ExecutionId, int EventCount, int RequestCount, long LatencyMs, string ETag)
         {
             EnsureLogicalTraceActivityId();
-            this.WriteEvent(110, Account, TaskHub, InstanceId, ExecutionId, EventCount, RequestCount, LatencyMs);
+            this.WriteEvent(110, Account, TaskHub, InstanceId, ExecutionId, EventCount, RequestCount, LatencyMs, ETag);
         }
 
         [Event(111, Level = EventLevel.Informational)]
-        public void AppendedInstanceState(string Account, string TaskHub, string InstanceId, string ExecutionId, int NewEventCount, int TotalEventCount, string NewEvents, long LatencyMs)
+        public void AppendedInstanceState(string Account, string TaskHub, string InstanceId, string ExecutionId, int NewEventCount, int TotalEventCount, string NewEvents, long LatencyMs, string ETag)
         {
             EnsureLogicalTraceActivityId();
-            this.WriteEvent(111, Account, TaskHub, InstanceId, ExecutionId, NewEventCount, TotalEventCount, NewEvents, LatencyMs);
+            this.WriteEvent(111, Account, TaskHub, InstanceId, ExecutionId, NewEventCount, TotalEventCount, NewEvents, LatencyMs, ETag);
         }
 
         [Event(112, Level = EventLevel.Informational)]
@@ -322,6 +322,13 @@ namespace DurableTask.AzureStorage
         {
             EnsureLogicalTraceActivityId();
             this.WriteEvent(137, Account, TaskHub, Details);
+        }
+
+        [Event(138, Level = EventLevel.Error)]
+        public void SplitBrainDetected(string Account, string TaskHub, string InstanceId, string ExecutionId, int NewEventCount, int TotalEventCount, string NewEvents, long LatencyMs, string ETag)
+        {
+            EnsureLogicalTraceActivityId();
+            this.WriteEvent(138, Account, TaskHub, InstanceId, ExecutionId, NewEventCount, TotalEventCount, NewEvents, LatencyMs, ETag);
         }
     }
 }
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -324,7 +324,7 @@ namespace DurableTask.AzureStorage
             this.WriteEvent(137, Account, TaskHub, Details);
         }
 
-        [Event(138, Level = EventLevel.Error)]
+        [Event(138, Level = EventLevel.Warning)]
         public void SplitBrainDetected(string Account, string TaskHub, string InstanceId, string ExecutionId, int NewEventCount, int TotalEventCount, string NewEvents, long LatencyMs, string ETag)
         {
             EnsureLogicalTraceActivityId();
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -766,17 +766,9 @@ namespace DurableTask.AzureStorage
             string instanceId = workItem.InstanceId;
             string executionId = runtimeState.OrchestrationInstance.ExecutionId;
 
-            try
+            if (!await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId))
             {
-                await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId);
-            }
-            catch (StorageException ex)
-            {
-                if (ex.RequestInformation.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
-                {
-                    await this.AbandonTaskOrchestrationWorkItemAsync(workItem);
-                    return;
-                }
+                await this.AbandonTaskOrchestrationWorkItemAsync(workItem);
             }
 
             bool addedControlMessages = false;
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -766,7 +766,18 @@ namespace DurableTask.AzureStorage
             string instanceId = workItem.InstanceId;
             string executionId = runtimeState.OrchestrationInstance.ExecutionId;
 
-            await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId);
+            try
+            {
+                await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId);
+            }
+            catch (StorageException ex)
+            {
+                if (ex.RequestInformation.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
+                {
+                    await this.AbandonTaskOrchestrationWorkItemAsync(workItem);
+                    return;
+                }
+            }
 
             bool addedControlMessages = false;
             bool addedWorkItemMessages = false;
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -766,10 +766,7 @@ namespace DurableTask.AzureStorage
             string instanceId = workItem.InstanceId;
             string executionId = runtimeState.OrchestrationInstance.ExecutionId;
 
-            if (!await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId))
-            {
-                await this.AbandonTaskOrchestrationWorkItemAsync(workItem);
-            }
+            await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId);
 
             bool addedControlMessages = false;
             bool addedWorkItemMessages = false;
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -146,7 +146,7 @@ namespace DurableTask.AzureStorage.Tracking
             if (expectedExecutionId != null)
             {
                 // Filter down to a specific generation.
-                // e.g. "PartitionKey eq 'pid' and ExecutionId eq 'eid'"
+                // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81' and ExecutionId eq '85f05ce1494c4a29989f64d3fe0f9089'"
                 filterCondition.Append(" and ExecutionId eq ").Append(Quote).Append(expectedExecutionId).Append(Quote);
             }
 
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.AzureStorage.Tracking
 {
     using System;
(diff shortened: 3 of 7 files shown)


src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -323,5 +323,12 @@ public void MonitorWarning(string Account, string TaskHub, string Details)
             EnsureLogicalTraceActivityId();
             this.WriteEvent(137, Account, TaskHub, Details);
         }
+
+        [Event(138, Level = EventLevel.Error)]
```
cgillum: This should be a warning.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -766,7 +766,18 @@ PendingMessageBatch StashMessagesAndGetNextBatch(IEnumerable<MessageData> queueM
...
             string executionId = runtimeState.OrchestrationInstance.ExecutionId;
 
-            await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId);
+            try
+            {
+                await this.trackingStore.UpdateStateAsync(runtimeState, instanceId, executionId);
+            }
+            catch (StorageException ex)
+            {
+                if (ex.RequestInformation.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
```
cgillum: We already have logic elsewhere which catches the `StorageException`. I think it would be better to simply change `UpdateStateAsync` to return `true` on success or `false` on a precondition failure. Here, we simply abandon the work item if `UpdateStateAsync` returns `false`.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -141,7 +144,8 @@ public override Task DeleteAsync()
             {
                 // Filter down to a specific generation.
                 // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81' and ExecutionId eq '85f05ce1494c4a29989f64d3fe0f9089'"
-                filterCondition.Append(" and ExecutionId eq ").Append(Quote).Append(expectedExecutionId).Append(Quote);
+                filterCondition.Append(" or ExecutionId eq ").Append(Quote).Append(expectedExecutionId).Append(Quote);
+                filterCondition.Append(" or RowKey eq ").Append(Quote).Append(SentinelRowKey).Append(Quote);
```
cgillum: Can you explain the change here and also reflect it in the comment above? I don't understand why we would change the `and ExecutionId eq` into `or ExecutionId eq`.

It seems to me that the new format would be something like:
*"PartitionKey eq 'pid' and (ExecutionId eq 'eid' or RowId eq 'sentinel')"*

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -190,6 +194,12 @@ public override Task DeleteAsync()
 
                 foreach (DynamicTableEntity entity in historyEventEntities)
                 {
+                    if (entity.RowKey == SentinelRowKey)
+                    {
+                        this.eTagValue = entity.ETag;
```
cgillum: This class has multiple threads accessing it concurrently. I don't think we can safely save the ETag in an instance variable.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -578,13 +612,38 @@ void SetTablePropertyForMessage(DynamicTableEntity entity, DynamicTableEntity or
...
+                        stopwatch.ElapsedMilliseconds,
+                        this.eTagValue);
+                    throw;
+                }
+            }
+
             this.stats.StorageRequests.Increment();
             this.stats.TableEntitiesWritten.Increment(historyEventBatch.Count);
 
+            TableResult sentinelItem = tableResultList?.FirstOrDefault(x => ((DynamicTableEntity)x.Result).RowKey == SentinelRowKey);
```
cgillum: In your current implementation, isn't the sentinel item always at the end of the list? If so, it would be much more efficient to search for it from the back.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
gled4er: Great! This will simplify the logic! Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
gled4er: Yes, you are right. I will improve this query. Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
gled4er: Thank you for this feedback. I will change it to `ConcurrentDictionary` with **InstanceId** as a key. Let me know if this is not a good fit. Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
gled4er: Great optimization! Thank you!

gled4er:
Hello @cgillum,

I worked on the issues you highlighted in the first round of feedback above.

May I ask you when you have time to check the latest changes and let me know what I should fix next?

Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -359,71 +373,118 @@ public override Task StartAsync()
...
             }
-
-            // First persistence step is to commit history to the history table. Messages must come after.
-            if (historyEventBatch.Count > 0)
+            catch (StorageException ex)
             {
-                await this.UploadHistoryBatch(instanceId, executionId, historyEventBatch, newEventListBuffer, newEvents.Count);
+                if (ex.RequestInformation.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
+                {
+                    return false;
```
cgillum: I thought about this more and I think we should not return `false` here. Instead, let's write a trace message and rethrow the exception. I believe this is better because it doesn't require us to change the interface. Also, I studied the DTFx code more closely and I now know that the batch will get aborted automatically when an unhandled exception is raised from this code path. This should also simplify the overall PR.

gled4er:
Hello @cgillum,

May I ask you when you have time to check the latest changes related to simplifying the pull request and reverting changes related to based classes and interfaces?

Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -423,9 +455,27 @@ public override Task StartAsync()
             // First persistence step is to commit history to the history table. Messages must come after.
             if (historyEventBatch.Count > 0)
             {
+                // Adding / updating sentinel entity
```
cgillum: Looks like this logic is redundant with the logic above. Can we move it into a helper method? Or even put it inside `UploadHistoryBatch`?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -578,13 +629,47 @@ void SetTablePropertyForMessage(DynamicTableEntity entity, DynamicTableEntity or
...
+            }
+
             this.stats.StorageRequests.Increment();
             this.stats.TableEntitiesWritten.Increment(historyEventBatch.Count);
 
+            if (tableResultList != null)
+            {
+                for (int i = tableResultList.Count - 1; i > 0; i--)
+                {
+                    if(((DynamicTableEntity)tableResultList[i].Result).RowKey == SentinelRowKey)
```
cgillum: Should the loop condition be `i >= 0`?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -593,7 +678,8 @@ void SetTablePropertyForMessage(DynamicTableEntity entity, DynamicTableEntity or
                 historyEventBatch.Count,
                 numberOfTotalEvents,
                 historyEventNamesBuffer.ToString(0, historyEventNamesBuffer.Length - 1), // remove trailing comma
-                stopwatch.ElapsedMilliseconds);
+                stopwatch.ElapsedMilliseconds,
+                this.GetETagValue(instanceId));
```
cgillum: We already fetched the new etag value in the for-loop above. Let's save it in a local variable and reference that here instead of doing a dictionary lookup.

gled4er:
Hello @cgillum,

Thank you for the review comments.

I will work on fixing the issues.

Thank you!

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
```diff
@@ -64,7 +64,7 @@ interface ITrackingStore
         /// <param name="runtimeState">The New RuntimeState</param>
         /// <param name="instanceId">InstanceId for the Orchestration Update</param>
         /// <param name="executionId">ExecutionId for the Orchestration Update</param>
-        /// <returns></returns>
+        /// <returns>Boolean marker for the success of the operation</returns>
```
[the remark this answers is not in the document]
gled4er: Great catch! I will remove it.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -378,8 +392,26 @@ public override Task StartAsync()
...
+                    DynamicTableEntity sentinelEntity = new DynamicTableEntity(instanceId, SentinelRowKey)
+                    {
+                        Properties =
+                        {
+                            ["ExecutionId"] = new EntityProperty(executionId),
+                        }
+                    };
+                    if (!string.IsNullOrEmpty(this.GetETagValue(instanceId)))
+                    {
+                        sentinelEntity.ETag = this.eTagValues[instanceId];
```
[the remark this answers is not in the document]
gled4er: Thank you! I will modify it.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -140,8 +146,9 @@ public override Task DeleteAsync()
             if (expectedExecutionId != null)
             {
                 // Filter down to a specific generation.
-                // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81' and ExecutionId eq '85f05ce1494c4a29989f64d3fe0f9089'"
-                filterCondition.Append(" and ExecutionId eq ").Append(Quote).Append(expectedExecutionId).Append(Quote);
+                // e.g. "PartitionKey eq 'pid' and (ExecutionId eq 'eid' or RowId eq 'sentinel')"
+                filterCondition.Append(" and (ExecutionId eq ").Append(Quote).Append(expectedExecutionId).Append(Quote);
+                filterCondition.Append(" or RowKey eq ").Append(Quote).Append(SentinelRowKey).Append(Quote).Append(")");
```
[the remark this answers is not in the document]
gled4er: Yes, great catch. I will modify this line.

gled4er:
Hello @cgillum,

May I ask you when you have time to check the latest changes for this pull request?

Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -575,16 +591,70 @@ void SetTablePropertyForMessage(DynamicTableEntity entity, DynamicTableEntity or
...
+                        this.storageAccountName,
+                        this.taskHubName,
+                        instanceId,
+                        executionId,
+                        historyEventBatch.Count,
+                        numberOfTotalEvents,
+                        historyEventNamesBuffer.ToString(0, historyEventNamesBuffer.Length - 1), // remove trailing comma
+                        stopwatch.ElapsedMilliseconds,
+                        eTagValue);
+                    throw;
```
cgillum: This `throw` needs to be moved outside the `if` block.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -440,6 +455,7 @@ public override Task StartAsync()
                 orchestratorEventType?.ToString() ?? string.Empty,
                 orchestrationInstanceUpdateStopwatch.ElapsedMilliseconds);
         }
+
```
cgillum: nit: please remove the extra newlines

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -575,16 +591,70 @@ void SetTablePropertyForMessage(DynamicTableEntity entity, DynamicTableEntity or
...
+            {
+                historyEventBatch.InsertOrReplace(sentinelEntity);
+            }
+
             Stopwatch stopwatch = Stopwatch.StartNew();
-            await this.historyTable.ExecuteBatchAsync(
-                historyEventBatch,
-                this.StorageTableRequestOptions,
-                null);
+            IList<TableResult> tableResultList = null;
```
cgillum: Remove the initialization to `null` here. If you did not initialize to `null`, then the compiler would have complained about the use of an uninitialized `tableResultList`. That would have been a signal to you that there is a bug in your try/catch logic (which I pointed out further down) because `tableResultList` should always be getting assigned.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
gled4er: Thank you for catching this!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
gled4er: Thank you! I will remove it.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
gled4er: Thank you very much for providing additional explanation about this issue. I will fix it. Thank you!

gled4er:
Hello @cgillum,

May I ask you when you have time to review once more the latest changes addressing your comments?

Thank you!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -140,7 +146,7 @@ public override Task DeleteAsync()
             if (expectedExecutionId != null)
             {
                 // Filter down to a specific generation.
-                // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81' and ExecutionId eq '85f05ce1494c4a29989f64d3fe0f9089'"
+                // e.g. "PartitionKey eq 'pid' and ExecutionId eq 'eid'"
```
cgillum: Can you revert the change to this comment?

gled4er:
Hello @cgillum,

May I ask you to confirm the latest changes to this pull request and let me know what should be fixed next?

Thank you!

PR #170, opened 2018-04-20
Sticky sessions support in DurableTask.Core

cgillum:
This PR enables "sticky-sessions", which is a feature to minimize the number of orchestration replays when multiple messages arrive in quick succession.  The previous flow for orchestrations was as follows:

1. Provider's `LockNextTaskOrchestrationWorkItemAsync` locks a session and returns a work item + new events
2. DTFx reconciles the new work item events into state
3. DTFx executions the orchestration from scratch
4. Provider's `CompleteTaskOrchestrationWorkItemAsync` checkpoints the new state
5. Provider's `ReleaseTaskOrchestrationWorkItemAsync` releases the session

This PR changes the flow as follows:

1. Provider's `LockNextTaskOrchestrationWorkItemAsync` locks a session and returns a work item with an attached `IOrchestrationSession`.
2. DTFx pulls new events from `IOrchestrationSession.FetchNewOrchestrationMessagesAsync` and reconciles them into state
3. DTFx executions the orchestration from where it left off
4. Provider's `CompleteTaskOrchestrationWorkItemAsync` checkpoints the new state
5. DTFx repeats (2) above, stopping only when `IOrchestrationSession.FetchNewOrchestrationMessagesAsync` returns `null`.
6. Provider's `ReleaseTaskOrchestrationWorkItemAsync` releases the session

I tested this with a refactored version of **DurableTask.AzureStorage** and verified that I can step through a sequential Fibonacci orchestration using the debugger and it behaves just as if it were normal C# method calls (i.e. no replays). I also verified that an unmodified **DurableTask.AzureStorage** continues to work as it did before, even with sticky-sessions enabled (demonstrating backwards compatibility).

This change is backwards compatible with existing provider code. This way Azure Functions (which is on a tight shipping schedule) can leverage it first in **DurableTask.AzureStorage** and then we can later add support to **DurableTask.ServiceBus**.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -30,13 +30,16 @@ namespace DurableTask.Core
     /// <summary>
     /// Dispatcher for orchestrations to handle processing and renewing, completetion of orchestration events
     /// </summary>
-    public class TaskOrchestrationDispatcher 
+    public class TaskOrchestrationDispatcher
     {
+        static readonly DataConverter DataConverter = new JsonDataConverter();
+        static readonly Task CompletedTask = Task.FromResult(0);
+
         readonly INameVersionObjectManager<TaskOrchestration> objectManager;
         readonly IOrchestrationService orchestrationService;
         readonly WorkItemDispatcher<TaskOrchestrationWorkItem> dispatcher;
         readonly DispatchMiddlewarePipeline dispatchPipeline;
-        static readonly DataConverter DataConverter = new JsonDataConverter();
+        readonly NonBlockingCountdownLock concurrentSessionLock;
 
         internal TaskOrchestrationDispatcher(
             IOrchestrationService orchestrationService,
@@ -60,6 +63,11 @@ namespace DurableTask.Core
                 DispatcherCount = orchestrationService.TaskOrchestrationDispatcherCount,
                 MaxConcurrentWorkItems = orchestrationService.MaxConcurrentTaskOrchestrationWorkItems
             };
+
+            // To avoid starvation, we only allow half of all concurrently execution orchestrations to
+            // leverage extended sessions.
+            int maxConcurrentSessions = (int)Math.Ceiling(this.dispatcher.MaxConcurrentWorkItems / 2.0);
+            this.concurrentSessionLock = new NonBlockingCountdownLock(maxConcurrentSessions);
         }
 
         /// <summary>
@@ -90,14 +98,19 @@ namespace DurableTask.Core
         public bool IncludeParameters { get; set; }
 
         /// <summary>
+        /// Gets or sets a flag indicating whether to enable extended sessions.
+        /// </summary>
+        public bool ExtendedSessionsEnabled { get; set; } = true;
+
+        /// <summary>
         /// Method to get the next work item to process within supplied timeout
         /// </summary>
         /// <param name="receiveTimeout">The max timeout to wait</param>
         /// <returns>A new TaskOrchestrationWorkItem</returns>
-        protected async Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveTimeout)
+        protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveTimeout)
         {
             // AFFANDAR : TODO : wire-up cancellation tokens
-            return await this.orchestrationService.LockNextTaskOrchestrationWorkItemAsync(receiveTimeout, CancellationToken.None);
+            return this.orchestrationService.LockNextTaskOrchestrationWorkItemAsync(receiveTimeout, CancellationToken.None);
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


cgillum:
Please see https://github.com/Azure/durabletask/pull/171, which contains these changes plus the changes to **DurableTask.AzureStorage**.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
-                // TODO : mark an orchestration as faulted if there is data corruption
-                TraceHelper.TraceSession(
-                    TraceEventType.Error, 
-                    "TaskOrchestrationDispatcher-DeletedOrchestration", 
-                    runtimeState.OrchestrationInstance?.InstanceId,
-                    "Received result for a deleted orchestration");
-                isCompleted = true;
-            }
-            else
+            while (!isCompleted)
```
affandar: taking the simple case when this.ExtendedSessionsEnabled = false, how does this loop break out when the orchestration does not get enough messages to actually get to the completed state?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
+                    {
+                        keepFetching = this.concurrentSessionLock.Acquire();
+                        isExtendedSession |= keepFetching;
+                    }
 
-                IList<OrchestratorAction> decisions = (await ExecuteOrchestrationAsync(runtimeState)).ToList();
+                    if (keepFetching)
+                    {
+                        newMessages = await workItem.Session.FetchNewOrchestrationMessagesAsync(workItem);
+                        workItem.NewMessages = null;
```
affandar: is setting workItem.NewMessages required? I don't see this being used after this..

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
+                var subOrchestrationMessages = new List<TaskMessage>();
+                bool continuedAsNew = false;
+
+                ExecutionStartedEvent continueAsNewExecutionStarted = null;
+                TaskMessage continuedAsNewMessage = null;
 
-                foreach (OrchestratorAction decision in decisions)
+                runtimeState.AddEvent(new OrchestratorStartedEvent(-1));
+
+                if (!ReconcileMessagesWithState(runtimeState, newMessages))
```
affandar: newMessages can be null if concurrent lock acquire failed above right? reconcile will throw a null ref

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
-                    "TaskOrchestrationDispatcher-ExecuteUserOrchestration-Begin",
-                    runtimeState.OrchestrationInstance,
-                    "Executing user orchestration: {0}",
-                    DataConverter.Serialize(runtimeState.GetOrchestrationRuntimeStateDump(), true));
+                if (this.ExtendedSessionsEnabled && newMessages == null && workItem.Session != null)
+                {
+                    bool keepFetching = true;
+                    if (!isExtendedSession && fetchCount >= 1)
+                    {
+                        keepFetching = this.concurrentSessionLock.Acquire();
```
affandar: if we failed to acquire lock then why not exit the loop and pack up this iteration? 

also it looks like we will get into high CPU if this lock can't be acquired, i.e I don't see a sleep or a timeout for this acquisition

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
+                        isExtendedSession |= keepFetching;
+                    }
 
-                IList<OrchestratorAction> decisions = (await ExecuteOrchestrationAsync(runtimeState)).ToList();
+                    if (keepFetching)
+                    {
+                        newMessages = await workItem.Session.FetchNewOrchestrationMessagesAsync(workItem);
```
affandar: trying to figure out why not pull the fetch logic into onfetchworkitem call above. the way I see this the onprocessworkitemasync is purely for processing a work item. the fetch logic should be tucked away into the on fetch call above. but there might be a reason.. will go through the whole CR to understand.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
-                                "TaskOrchestrationDispatcher-UnsupportedDecisionType",
-                                runtimeState.OrchestrationInstance,
-                                new NotSupportedException("decision type not supported"));
+                    if (cursor == null)
+                    {
+                        cursor = await ExecuteOrchestrationAsync(runtimeState);
                     }
+                    else
+                    {
+                        await ResumeOrchestrationAsync(cursor);
```
affandar: I think resume could be rolled into Execute if execute took a cursor

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +119,234 @@ public class TaskOrchestrationDispatcher
...
+                    foreach (OrchestratorAction decision in decisions)
                     {
                         TraceHelper.TraceInstance(
-                            TraceEventType.Information, 
-                            "TaskOrchestrationDispatcher-MaxMessageCountReached", 
+                            TraceEventType.Information,
+                            "TaskOrchestrationDispatcher-ProcessOrchestratorAction",
                             runtimeState.OrchestrationInstance,
-                            "MaxMessageCount reached.  Adding timer to process remaining events in next attempt.");
+                            "Processing orchestrator action of type {0}",
```
affandar: btw for most of the diffs showing up here I'm assuming there weren't really changes. just moving around the code?

pls flag if you changed some core logic in these places :)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
affandar: actually thinking more about it, pulling this logic into fetch will ultimately be better because it would force refactoring of sessions between taskactivitydispatcher and taskorchestrationdispatcher which is a good thing as it will enable activity cancellations and messaging between orchestration and activities while the activity is running (this is something that is quite useful)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: I believe I may have found and fixed this null-ref issue in #171.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Sure, let me look into this. I didn't understand the comments about how this could benefit activity execution, but it does feel like a cleaner model.

cgillum: I looked at this more and I'm not sure how this can be made to work. I see two issues:

1. I need a while loop inside of OnProcess because letting the method exit will result in a release of the work item/session.
2. OnFetch runs in a separate loop from OnProcess. Even if I move the actual fetching code back into Onfetch, I still need to loop and block to wait for the new events to arrive to ensure I'm not processing multiple event batches for the same orchestration concurrently.

Can you clarify your thoughts on this?

affandar: Hmm, yes it is more surgery than we'd be comfortable with for now.

Re you points:
1. the safe release is implemented by the provider and would look at the work item to see if it needs to release or not (e..g last fetch yielded zero messages so release otherwise don't release)
2. the fetchworkitem then also takes an ISession from the existing work item and tries to call fetch next on it (ISession.FetchNextMessagesAsync e.g). If this returns zero messages then it calls safereleaseworkitem and goes back into IOrchestrationService.fetch next
3. the way it is written now, OnProcess runs in parallel with the loop because the dispatcher assumes it is serialized due to the underlying provider having a lock on the specific session until  on process completes and you are right in that this assumption will be broken so this is more surgery.

At this point, I think your current approach is fine but can we pls keep the taskorchestrationdispatcher.onprocessworkitem() almost exactly the same as before and add the loop in a wrapper function (e.g, onprocessworkitembatches() or something. this would help keep the intent clearer.

cgillum: OK, thanks - I've refactored it based on the suggestion and simplified it a bit to reduce the number of changes. Hopefully that will make it easier to review. Updates are in #171.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: I've rearranged it a bit to reduce the overall number of diffs.

PR #171, opened 2018-04-20
"Sticky sessions" support in DurableTask.Core and DurableTask.AzureStorage

cgillum:
This PR contains the same changes as https://github.com/Azure/durabletask/pull/170 but also includes the changes to **DurableTask.AzureStorage** to support it.


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -106,192 +114,239 @@ public class TaskOrchestrationDispatcher
...
-                workItem,
-                newOrchestrationRuntimeState,
-                continuedAsNew ? null : messagesToSend,
-                subOrchestrationMessages,
-                continuedAsNew ? null : timerMessages,
-                continuedAsNewMessage,
-                instanceState);
+            if (isExtendedSession)
+            {
+                this.concurrentSessionLock.Release();
```
adarsh1: Should this be in a finally block?

cgillum: Yep, this is fixed in the latest push (which was a force update to make it easier for me to revert some of the unnecessary changes).

cgillum:
I've re-pushed the changes to this PR. The overall diff should be reduced now, making it a simpler review for the parts in DurableTask.Core. Please have a look.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -282,16 +338,18 @@ public class TaskOrchestrationDispatcher
                 continuedAsNew ? null : timerMessages,
                 continuedAsNewMessage,
                 instanceState);
+
```
affandar: the this.orchestrationService.CompleteTaskOrchestrationWorkItemAsync call above will release the session lock for session based providers. So we need to add a flag that specifies 'update only but not release'. In fact in addition to this flag, we'd need to move this to the wrapping onprocessworkitemsession method so that guy can periodically complete tasks (without releasing lock le). Otherwise the uncommitted 'transaction log' is going to continue growing which has two problems:

i) much much heavier replay on a reload
ii) runs into transaction limits (e.g. service bus can't save more than 100 message sends per transaction commit)

affandar: ..iii) without this the orchestration may stop moving forward, because completetaskorchestrationworkitemasync() is where activities are actually triggered (and respond)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -209,16 +265,16 @@ public class TaskOrchestrationDispatcher
                     if (this.orchestrationService.IsMaxMessageCountExceeded(totalMessages, runtimeState))
```
affandar: @cgillum have you thought though how this max message count limit mechanic play out with the sticky sessions?

when this limit hits, we need to completetaskorchestrationworkitem (without releasing the lock.. see comment below) so the messages getting queued up to be sent are actually sent.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: My understanding is that each call to Complete() will commit all the changes in SB (sending messages, etc.) but that the session doesn't get released until Release() is called. Is that not the case?  If that is the case, then I think the current implementation will work just fine for Service Bus.

affandar: yea nvm, I had forgotten my own code :)

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -24,6 +24,13 @@ namespace DurableTask.Core
     /// </summary>
     public abstract class OrchestrationContext
     {
+        /// <summary>
+        /// Thread-static variable used to signal whether the calling thread is the orchestrator thread.
+        /// The primary use case is for detecting illegal async usage in orchestration code.
+        /// </summary>
+        [ThreadStatic]
+        public static bool IsOrchestratorThread;
```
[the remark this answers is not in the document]
affandar: ..you should add this explanation to the comment in the code as well. 

Also you also point me to example usage in durable functions code?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: We are already calling `CompleteTaskOrchestrationWorkItem` for every batch (every loop iteration), so there is no real behavior change here. The only real change is that we're doing multiple Complete calls before we release the work item.

cgillum: Oh! I see what you're getting at - sorry I completely misunderstood the concern. **DurableTask.AzureStorage** is not impacted by this since we don't have a max message count, so I don't think this necessarily needs to be a blocker for this PR - i.e. it could be worked on once you guys on-board **DurableTask.ServiceBus**.

If I understand correctly, what you're doing is you're check-pointing the progress that was made thus far (scheduling 100 messages and updating the history accordingly) and then scheduling a no-op timer to run through the history a second time to make sure any non-processed tasks are re-executed. The next iteration will know which tasks were scheduled and which ones weren't because it will be reflected in the updated history.

With my current implementation, we won't replay the history, thus won't re-run those "aborted" tasks. I can see this would be a problem. In that case, I think it's best for us to break out of the sticky-session loop and re-execute from scratch on the next work item. I'll add that change.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -613,13 +616,38 @@ async delegate (CloudQueue controlQueue)
 
             this.controlQueueBackoff.Reset();
 
-            ReceivedMessageContext messageContext =
-                ReceivedMessageContext.CreateFromReceivedMessageBatch(
+            OrchestrationInstance instance = nextBatch.Messages[0].TaskMessage.OrchestrationInstance;
```
simonporter: >Messages[0] [](start = 55, length = 11)

Is this gauranteed to exist?

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -159,7 +185,7 @@ public class AzureStorageScenarioTests
                 // The end message will cause the actor to complete itself.
                 await client.RaiseEventAsync("operation", "end");
 
-                status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(10));
+                status = await client.WaitForCompletionAsync(Debugger.IsAttached ? TimeSpan.FromMinutes(5) : TimeSpan.FromSeconds(10));
```
simonporter: >Debugger.IsAttached ? TimeSpan.FromMinutes(5) : TimeSpan.FromSeconds(10) [](start = 61, length = 72)

nit: more to helper, duped 5 times in here already

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: Yes, a batch is only ever created if we have a message to put into it.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -282,16 +343,18 @@ public class TaskOrchestrationDispatcher
                 continuedAsNew ? null : timerMessages,
                 continuedAsNewMessage,
                 instanceState);
+
+            return isCompleted || isInterrupted;
```
cgillum: >return isCompleted || isInterrupted; [](start = 12, length = 36)

More testing revealed that we also need to return `true` here if `continuedAsNew` is true.

src/DurableTask.Core/OrchestrationContext.cs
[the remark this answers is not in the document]
cgillum: I don't have a PR for it yet since this sticky-sessions PR has yet to be approved. However, you can see where I'm adding it here: [https://github.com/Azure/azure-functions-durable-extension/blob/36e1388c02b1c34d986a3876eef080f79d0c00c1/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs#L453](https://github.com/Azure/azure-functions-durable-extension/blob/36e1388c02b1c34d986a3876eef080f79d0c00c1/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs#L453)

---
In reply to: [183552220](https://github.com/Azure/durabletask/pull/171#discussion_r183552220) [](ancestors = 183552220)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -60,6 +63,11 @@ public class TaskOrchestrationDispatcher
                 DispatcherCount = orchestrationService.TaskOrchestrationDispatcherCount,
                 MaxConcurrentWorkItems = orchestrationService.MaxConcurrentTaskOrchestrationWorkItems
             };
+
+            // To avoid starvation, we only allow half of all concurrently execution orchestrations to
+            // leverage extended sessions.
+            int maxConcurrentSessions = (int)Math.Ceiling(this.dispatcher.MaxConcurrentWorkItems / 2.0);
```
adarsh1: Should this be configurable?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -301,27 +364,41 @@ public class TaskOrchestrationDispatcher
...
+        Task ResumeOrchestrationAsync(OrchestrationExecutionCursor cursor)
+        {
+            var dispatchContext = new DispatchMiddlewareContext();
+            dispatchContext.SetProperty(cursor.RuntimeState.OrchestrationInstance);
+            dispatchContext.SetProperty(cursor.TaskOrchestration);
+            dispatchContext.SetProperty(cursor.RuntimeState);
+
+            return this.dispatchPipeline.RunAsync(dispatchContext, _ =>
+            {
+                cursor.LatestDecisions = cursor.OrchestrationExecutor.ExecuteNewEvents();
```
adarsh1: nit:  should we be exposing the OrchestrationExecutor from the cursor here? shouldn't the cursor update itself rather than its latest decisions being set by the dispatcher?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: We can make it configurable later. For now I think half is a good default.

---
In reply to: [183831739](https://github.com/Azure/durabletask/pull/171#discussion_r183831739) [](ancestors = 183831739)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Turns out the `continueAsNew` case completely derailed my session implementation in Azure Storage due to race conditions. I've had to refactor it.

---
In reply to: [183619482](https://github.com/Azure/durabletask/pull/171#discussion_r183619482) [](ancestors = 183619482)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -120,22 +120,38 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
+                        if (!isExtendedSession)
+                        {
+                            break;
+                        }
+                    }
+
+                    // Wait for new messages to arrive for the session. This call is expected to block (asynchronously)
+                    // until either new messages are available or until a provider-specific timeout has expired.
+                    workItem.NewMessages = await workItem.Session.FetchNewOrchestrationMessagesAsync(workItem);
+                    if (workItem.NewMessages == null)
```
affandar: non-blocking nit: please add tracing to this method. when things go wrong, there should be a log we can look at to see how long each fetch took and how many messages did it actually fetch. similar to the logging we have for the master fetch etc.

src/DurableTask.AzureStorage/AsyncAutoResetEvent.cs
```diff
@@ -99,5 +99,10 @@ public void Set()
                 }
             }
         }
+
+        public override string ToString()
+        {
+            return $"Signaled: {this.isSignaled}, Waiters: {this.waiters.Count}";
```
simonporter: >Count [](start = 73, length = 5)

nit: use tostring() to avoid boxing

src/DurableTask.Core/IOrchestrationSession.cs
```diff
@@ -0,0 +1,25 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.Core
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    /// <summary>TODO</summary>
```
simonporter: >TODO [](start = 17, length = 4)

Would be nice if we can populate this

cgillum: Sorry, I totally spaced on that. I'll fill these in.

---
In reply to: [184815743](https://github.com/Azure/durabletask/pull/171#discussion_r184815743) [](ancestors = 184815743)

PR #177, opened 2018-05-04
Fix issue with ETag check when fanning out to 100+ parallel activities

cgillum:
FYI @gled4er. Let me know if you think this fix will work. I haven't had a chance to test it yet, but it looks like we need a new test case which fans out to 100+ activity functions.


gled4er:
Hello @cgillum,

The change is very good.

I agree we need a test for it.

Thank you for catching this bug!

PR #179, opened 2018-05-11
Changed PackageUpgradeSerializationBinder to inherit from the Default…

adarsh1:
… SerializationBinder removing the Dependency on the NewtonsoftJsonSerializationBinder


adarsh1:
this should resolve #179

PR #185, opened 2018-05-17
Update AssemblyInfo.cs

sazeesha:


simonporter:
Thanks for your contribution. We cannot however take contributions that add arbitrary external dependencies to the internals.
It you have a dependency that needs something exposed as protected please file an issue or submit a pull request to expose that.

PR #187, opened 2018-05-21
DurableTask.AzureStorage API to enumerate instances

TsuyoshiUshio:
Hi @cgillum and @gled4er,
 
I open this pull request to start discussion the details. 

I have two topic to discuss. 

1. How many attribute do we need? 

In the OrchestrationState has a lot of members. Which attribute do we need for this purpose?

2. Style of Query

We'll full scan search for the instance table. Is it OK? or do we have any other clever way to fetch the on the fly instances? Currenty, I just return all record of the isntance table. 

This pull request is correspond to this issue. 
https://github.com/Azure/azure-functions-durable-extension/issues/316

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1544,6 +1544,16 @@ namespace DurableTask.AzureStorage
         }
 
         /// <summary>
+        /// Get states of the all orchestration instances
+        /// </summary>
+        /// <returns>List of <see cref="OrchestrationState"/></returns>
+        public async Task<IList<OrchestrationState>> GetStateAsync()
+        {
+            await this.EnsureTaskHubAsync();
+            return await this.trackingStore.GetStateAsync();
+        }
+
+        /// <summary>
         /// Force terminates an orchestration by sending a execution terminated event
         /// </summary>
         /// <param name="instanceId">Instance ID of the orchestration to terminate.</param>
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1547,10 +1547,10 @@ namespace DurableTask.AzureStorage
         /// Get states of the all orchestration instances
         /// </summary>
         /// <returns>List of <see cref="OrchestrationState"/></returns>
-        public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync()
+        public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
             await this.EnsureTaskHubAsync();
-            return await this.trackingStore.GetStateAsync();
+            return await this.trackingStore.GetStateAsync(cancellationToken);
         }
 
         /// <summary>
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1547,7 +1547,7 @@ namespace DurableTask.AzureStorage
         /// Get states of the all orchestration instances
         /// </summary>
         /// <returns>List of <see cref="OrchestrationState"/></returns>
-        public async Task<IList<OrchestrationState>> GetStateAsync()
+        public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync()
         {
             await this.EnsureTaskHubAsync();
             return await this.trackingStore.GetStateAsync();
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -330,7 +330,6 @@ namespace DurableTask.AzureStorage.Tracking
                 int previousCount = instancestatuses.Count;
                 instancestatuses.AddRange(segment.AsEnumerable<OrchestrationInstanceStatus>());
 
-                // TODO do we need these?
                 this.stats.StorageRequests.Increment();
                 this.stats.TableEntitiesRead.Increment(instancestatuses.Count - previousCount);
 
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -315,7 +315,7 @@ namespace DurableTask.AzureStorage.Tracking
             var query = new TableQuery<OrchestrationInstanceStatus>();
             TableContinuationToken token = null;
 
-            var instanceStatuses = new List<OrchestrationInstanceStatus>(100);
+            var orchestrationStates = new List<OrchestrationState>(100);
             var stopwatch = new Stopwatch();
             var requestCount = 0;
             bool finishedEarly = false;
@@ -327,11 +327,13 @@ namespace DurableTask.AzureStorage.Tracking
                 var segment = await this.instancesTable.ExecuteQuerySegmentedAsync(query, token); // TODO make sure if it has enough parameters
                 stopwatch.Stop();
 
-                int previousCount = instanceStatuses.Count;
-                instanceStatuses.AddRange(segment.AsEnumerable<OrchestrationInstanceStatus>());
+                int previousCount = orchestrationStates.Count;
+                var tasks = segment.AsEnumerable<OrchestrationInstanceStatus>().Select(async x => await ConvertFromAsync(x, x.PartitionKey));
+                OrchestrationState[] result = await Task.WhenAll(tasks);
+                orchestrationStates.AddRange(result);
 
                 this.stats.StorageRequests.Increment();
-                this.stats.TableEntitiesRead.Increment(instanceStatuses.Count - previousCount);
+                this.stats.TableEntitiesRead.Increment(orchestrationStates.Count - previousCount);
 
                 token = segment.ContinuationToken;
                 if (finishedEarly || token == null || cancellationToken.IsCancellationRequested)
@@ -340,18 +342,6 @@ namespace DurableTask.AzureStorage.Tracking
                 }      
             }
 
-            IList<OrchestrationState> orchestrationStates;
-
-            orchestrationStates = new List<OrchestrationState>(instanceStatuses.Count);
-
-            if (instanceStatuses.Count > 0)
-            {
-                foreach(OrchestrationInstanceStatus entity in instanceStatuses)
-                {
-                    var state = await ConvertFromAsync(entity, entity.PartitionKey);
-                    orchestrationStates.Add(state);
-                }
-            }
             return orchestrationStates;
         }
 
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -306,6 +306,86 @@ namespace DurableTask.AzureStorage.Tracking
         }
 
         /// <inheritdoc />
+        public override async Task<IList<OrchestrationState>> GetStateAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var query = new TableQuery();
+            TableContinuationToken token = null;
+            var instanceTableEntities = new List<DynamicTableEntity>(100);
+            var stopwatch = new Stopwatch();
+            var requestCount = 0;
+            bool finishedEarly = false;
+
+            while (true)
+            {
+                requestCount++;
+                stopwatch.Start();
+                var segment = await this.instancesTable.ExecuteQuerySegmentedAsync(query, token); // TODO make sure if it has enough parameters
+                stopwatch.Stop();
+
+                int previousCount = instanceTableEntities.Count;
+                instanceTableEntities.AddRange(segment);
+
+                // TODO do we need these?
+                this.stats.StorageRequests.Increment();
+                this.stats.TableEntitiesRead.Increment(instanceTableEntities.Count - previousCount);
+
+                token = segment.ContinuationToken;
+                if (finishedEarly || token == null || cancellationToken.IsCancellationRequested)
+                {
+                    break;
+                }      
+            }
+
+            IList<OrchestrationState> orchestrationStates;
+
+            orchestrationStates = new List<OrchestrationState>(instanceTableEntities.Count);
+
+            if (instanceTableEntities.Count > 0)
+            {
+                foreach(DynamicTableEntity entity in instanceTableEntities)
+                {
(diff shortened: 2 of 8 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -305,6 +305,86 @@ public override Task DeleteAsync()
...
+            IList<OrchestrationState> orchestrationStates;
+
+            orchestrationStates = new List<OrchestrationState>(instanceTableEntities.Count);
+
+            if (instanceTableEntities.Count > 0)
+            {
+                foreach(DynamicTableEntity entity in instanceTableEntities)
+                {
+                    var instance = new OrchestrationInstance();
+                    instance.InstanceId = entity.Properties["PartitionKey"].ToString();
```
cgillum: Take a look at the previous `GetStateAsync` method above. Since it already has logic to convert table storage rows into `OrchestrationState` objects, I think it would be best to reuse that code as much as possible. That way if we need to change the logic for reading from table storage, we can do it in one place instead of two places.

TsuyoshiUshio: Thanks! I’ll fix that!

gled4er: Hello @cgillum ,

Will it make sense to add additional partition in the Instance table with partition key "OnTheFly" (and row key the instance id) that will contain all the instances currently on the fly. Each of the instance will have also dedicated row with its instance id as partition key as well. We can have a setting that will enable the duplicate update of the instance in the "OnTheFly" partition and its own partition. When the instance completes we will delete the row in the "OnTheFly" partition. This will be one more call but we will know both the partition and the row key. 

Then the query for retrieving the on the fly instances will be straightforward. I am worried that the Instnaces table will grow over time and a query that will make table scan each time will have bad performance impact.

Thank you!

TsuyoshiUshio: Good point! lets discuss that. I also worry about the performance since storage table only have a key for partition key and the instance table might be growing. :)

TsuyoshiUshio: Hi @gled4er , One question. OrchestrationInstanceStatus might be a domain class of instance table right? Then which attribute is the instanceId?  I guess PartitionKey. Is it right?

cgillum:
I'm a little concerned about creating a new partition for "on the fly" for several reasons. I'm not sure if I know how to communicate it clearly. The high-level issue is that it pollutes the data in the instances table. and creates complexity in the data model. In general, data in the partition column should be as balanced as possible, and "on-the-fly" partitions will create a lot of imbalance and will eventually result in different kinds of performance issues. The design is also specific to a particular use case and may not add value to other use cases.

My thought is that we should keep this feature simple and not change the structure of the data. Yes there will be performance issues with large numbers of rows - however, i think this is unavoidable when using Azure Table storage, so it's better if we just document it. For many users, the simple implementation will still be very valuable. If there is a need for huge numbers of instances, then we should recommend the Event Grid approach to put the data into a different, optimized data store based on the customer's specific needs (maybe they will use Cosmos DB, or something else).

gled4er:
Hello @cgillum,

Thank you very much for the explanation.

I agree.

Thank you!

TsuyoshiUshio:
@cgillum , @gled4er I have one suggestion. How about adding some configuration to remove the old instances. e.g. 1day or 7 days or something on the `host.json`.  Then we can remove the old record which is already done. Then the number of the record might be limited. Also you don't need to add new column. Just asynchronously remove the record which is not used.

TsuyoshiUshio:
Hi @cgillum  One more thing. Could you enable CI for this pull request?  I refactor the code to keep DRY. I just did extract method. however, it fails the test. When I see the test code, it seems related timeout. Which seems flakiness. I'd like to see what happens.

TsuyoshiUshio:
@cgillum Make sense! Sure.

TsuyoshiUshio:
Hi @cgillum and @gled4er , 

I have a question for testing.  I add the query method for AzureTableTrackingStore.cs and AzureStorageOrchestrationService, and TaskHubClient and IOrchestrationServiceClient. Except for the first one, all of others are just delegation. 

I might need to write test on  the AzureStorageScenarioTests.cs with TestOrchestraionClient. 
Since this method fetch all instances from instance table, the test will be like this. 

1. Clear the instance table 
2. Insert data ( or start Orchestrator) 
3. Fetch the all data and compare the collect number and contents. 

However the tests are executed parallely, If I clear the instance table, it might cause a problem. In this case how do you test the method? Do you have similar example on your test?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -305,6 +309,53 @@ public override Task DeleteAsync()
...
+                stopwatch.Start();
+                var segment = await this.instancesTable.ExecuteQuerySegmentedAsync(query, token); // TODO make sure if it has enough parameters
+                stopwatch.Stop();
+
+                int previousCount = instancestatuses.Count;
+                instancestatuses.AddRange(segment.AsEnumerable<OrchestrationInstanceStatus>());
+
+                // TODO do we need these?
+                this.stats.StorageRequests.Increment();
+                this.stats.TableEntitiesRead.Increment(instancestatuses.Count - previousCount);
```
cgillum: Regarding the `TODO`: Yes, we need these. We track these statistics to know how much load we're putting on the Azure Storage account.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -305,6 +309,53 @@ public override Task DeleteAsync()
             return orchestrationState;
         }
 
+        /// <inheritdoc />
+        public override async Task<IList<OrchestrationState>> GetStateAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var query = new TableQuery<OrchestrationInstanceStatus>();
+            TableContinuationToken token = null;
+
+            var instancestatuses = new List<OrchestrationInstanceStatus>(100);
```
cgillum: nit: capitalize the "S" - `instanceStatuses`.

src/DurableTask.Core/IOrchestrationServiceClient.cs
```diff
@@ -80,6 +80,12 @@ public interface IOrchestrationServiceClient
         /// <returns>The OrchestrationState of the specified instanceId or null if not found</returns>
         Task<OrchestrationState> GetOrchestrationStateAsync(string instanceId, string executionId);
 
+        /// <summary>
+        /// Get states of the all orchestration instances
+        /// </summary>
+        /// <returns>List of <see cref="OrchestrationState"/></returns>
+        Task<IList<OrchestrationState>> GetStateAsync();
```
cgillum: We have to be very careful about making changes to DurableTask.Core. If we need any changes in this area, then we need to core DTFx team to approve them - and that may not be easy since this design might not work well for other implementations, like DurableTask.ServiceBus. Should we try to implement this first only in DurableTask.AzureStorage to avoid any potential conflict?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -305,6 +309,53 @@ public override Task DeleteAsync()
...
+            var stopwatch = new Stopwatch();
+            var requestCount = 0;
+            bool finishedEarly = false;
```
cgillum: It looks like `stopwatch`, `requestCount`, and `finishedEarly` aren't really used for anything. Can they be removed?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1543,6 +1543,16 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
             return await this.trackingStore.GetStateAsync(instanceId, executionId);
         }
 
+        /// <summary>
+        /// Get states of the all orchestration instances
+        /// </summary>
+        /// <returns>List of <see cref="OrchestrationState"/></returns>
+        public async Task<IList<OrchestrationState>> GetStateAsync()
```
cgillum: The other methods similar to this one are named `GetOrchestrationStateAsync`. I think we should name this one the same.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -305,6 +309,53 @@ public override Task DeleteAsync()
...
+
+            IList<OrchestrationState> orchestrationStates;
+
+            orchestrationStates = new List<OrchestrationState>(instancestatuses.Count);
+
+            if (instancestatuses.Count > 0)
+            {
+                foreach(OrchestrationInstanceStatus entity in instancestatuses)
+                {
+                    var state = await ConvertFromAsync(entity, entity.PartitionKey);
```
cgillum: Instead of calling `ConvertFromAsync` inside a second `foreach` loop, would it be more efficient to do it inside the previous `foreach` loop (so that we only loop through the list of instances once instead of twice)?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done. However, it eventually removed. lol

src/DurableTask.Core/IOrchestrationServiceClient.cs
TsuyoshiUshio: One concern is, I use Linq for the refactoring. I'm not sure if you prefer the Linq. I do like this. for make it single loop.

```
                var tasks = segment.AsEnumerable<OrchestrationInstanceStatus>().Select(async x => await ConvertFromAsync(x, x.PartitionKey));
                OrchestrationState[] result = await Task.WhenAll(tasks);
                orchestrationStates.AddRange(result);
```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done!

One concern is, I use Linq for the refactoring. I'm not sure if you prefer the Linq. I do like this. for make it single loop.

```
                var tasks = segment.AsEnumerable<OrchestrationInstanceStatus>().Select(async x => await ConvertFromAsync(x, x.PartitionKey));
                OrchestrationState[] result = await Task.WhenAll(tasks);
                orchestrationStates.AddRange(result);
```

TsuyoshiUshio:
Hi @cgillum 

For enable the testing, I add one method to the TestOrchestrationClient.cs. Since you said that to add Core we should be careful, however, the TestOrchestrationClient is depend on TaskHub and 
I needed to add this method. It is not cool way, however, I can't find any other good solution. 

```
internal AzureStorageOrchestrationService GetServiceClient()
{
       return (AzureStorageOrchestrationService)this.client.serviceClient;
}
```

https://github.com/Azure/durabletask/pull/187/commits/ba1afae5ca8e955d272d0dae8e839eed78e1e34f

Any feedback is welcome.

TsuyoshiUshio:
@cgillum Then what is the next step?  I have two options. 

1. Implement OrchestratorClient on the DurableTaskFramework extension repo
2. Implement TaskHub and other Core methods. 

Maybe 1. should be fine. Then How I can test it? Create a nuget package of this branch then put it on local, then implement OrchestratorClient and test it using AzureStorageOrchestrationService directly then move on 2?

TsuyoshiUshio:
One of the concern for the next step is, how can we handle the branching.  This branch comes from azure-storage branch.  You might want to release this one with azure-storage branch. After the branch is merged, we can refer it from Durable Framework extension repo.  Do you have a strategy how to manage two repos?

cgillum:
I still need to figure out the exact branching and release strategy, but I have some ideas. My goal is that DurableTask.AzureStorage can be updated and released independently as much as possible.

For now, I think the next step is for you is to build your DurableTask.AzureStorage nuget package locally and reference it locally from Microsoft.Azure.WebJobs.Extensions.DurableTask. This is the normal workflow I follow and this will unblock the next phase of development.

Once we are happy with the PR for Microsoft.Azure.WebJobs.Extensions.DurableTask, then we can merge your changes in DurableTask.AzureStorage into **azure-storage**. From there, I can publish to the myget feed, and that will unblock the azure-functions-durable-extension CI. Once the CI looks good, we can decide whether to go ahead and merge **azure-storage** into **master** and publish the updated DurableTask.AzureStorage to nuget.org.

Will that work for you? Let me know if you need help with instructions for deploying the DurableTask.AzureStorage nuget package locally. Kanio might know how to do this as well.

I think your suggestion to isolate the changes in DurableTask.AzureStorage is good and it's exactly what I would have done. I'll try to review it in more detail tomorrow. For now I think you should be unblocked though.

TsuyoshiUshio:
Thank you, @cgillum . 

I'll test and start implement the logic to the Durable Functions repo. 
Then, Can I folk a branch from dev branch as usual?

cgillum:
When you are ready, please increment the package version in `DurableTask.AzureStorage.csproj`. I will then merge your change, create a nuget package, and publish it to the staging location. If the CI in azure-functions-durable-extension passes, then we can publish this package to nuget.org.

TsuyoshiUshio:
Hi @cgillum  One question. 

When we use this feature from Http API, it might be like this. `code=` is host key of Azure Functions. 

```"http://localhost:7071/runtime/webhooks/DurableTaskExtension/instances/?taskHub=DurableFunctionsHub&connection=Storage&code=es4Bb3qqtascBKLtKo5z41IP7Ne3QuooyWIoVXSpaqcZ2E47bFlSZg==",
```

I was wondering how customer get TaskHub/Connection/Code. Simplest solution is use  CreateCheckStatusRespons API. However, it requires "instanceId". It might be not a big deal. However, I'd like to share with you in advance.

cgillum:
TaskHub and Connection should be known, since they are configured statically.  The code can be fetched from the Azure Functions host admin APIs. It's a good point that this may not be easy to acquire. We will need to think about how difficult this might be for users. Hopefully we can use tools to help make it easier. We could also consider creating a new API for getting your new instance query URL.

TsuyoshiUshio:
Yeah. That it will be the next pull request. :)

PR #190, opened 2018-06-11
Implementation of rewind functionality

cmkerner20:
- added RewindInstanceAsync in TaskHubClient
- finds last known good state and rewinds data store (updates history and instance tables)
- does not kick off new orchestration

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -296,23 +296,19 @@ namespace DurableTask.AzureStorage.Tests
         /// </summary>
         [DataTestMethod]
         //[DataRow(true)]
-        // TODO: figure out what to do about extended sessions
         [DataRow(false)]
         public async Task RewindOrchestrationsFail(bool enableExtendedSessions)
         {
             using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
             {
-                Orchestrations.FactorialFail.ShouldFail = true;
+                Orchestrations.FactorialOrchestratorFail.ShouldFail = true;
                 await host.StartAsync();
 
                 string singletonInstanceId1 = $"1_Test_{Guid.NewGuid():N}";
                 string singletonInstanceId2 = $"2_Test_{Guid.NewGuid():N}";
 
-                Activities.MultiplyFail.ShouldFail1 = false;
-                Activities.MultiplyFail.ShouldFail2 = false;
-
                 var client1 = await host.StartOrchestrationAsync(
-                    typeof(Orchestrations.FactorialFail),
+                    typeof(Orchestrations.FactorialOrchestratorFail),
                     input: 3, 
                     instanceId: singletonInstanceId1);
 
@@ -320,14 +316,14 @@ namespace DurableTask.AzureStorage.Tests
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Orchestrations.FactorialFail.ShouldFail = false;
+                Orchestrations.FactorialOrchestratorFail.ShouldFail = false;
 
                 var client2 = await host.StartOrchestrationAsync(
                 typeof(Orchestrations.SayHelloWithActivity),
                 input: "Catherine",
                 instanceId: singletonInstanceId2);
 
-                await client1.RewindAsync("rewind!"); 
+                await client1.RewindAsync("Rewind failed orchestration only"); 
 
                 var statusRewind = await client1.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
@@ -348,7 +344,7 @@ namespace DurableTask.AzureStorage.Tests
         {
             using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
             {
-                Activities.HelloFail.ShouldFail1 = false;
+                Activities.HelloFailFanOut.ShouldFail1 = false;
                 await host.StartAsync();
 
                 string singletonInstanceId = $"Test_{Guid.NewGuid():N}";
@@ -362,9 +358,9 @@ namespace DurableTask.AzureStorage.Tests
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Activities.HelloFail.ShouldFail2 = false;
+                Activities.HelloFailFanOut.ShouldFail2 = false;
 
-                await client.RewindAsync("rewind!");
+                await client.RewindAsync("Rewind orchestrator with failed parallel activity.");
 
                 var statusRewind = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
@@ -391,8 +387,6 @@ namespace DurableTask.AzureStorage.Tests
 
                 string singletonInstanceId = $"{Guid.NewGuid():N}";
 
-                Activities.HelloFail.ShouldFail1 = false;
-
                 var client = await host.StartOrchestrationAsync(
                     typeof(Orchestrations.SayHelloWithActivityFail),
                     input: "World",
@@ -402,9 +396,9 @@ namespace DurableTask.AzureStorage.Tests
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Activities.HelloFail.ShouldFail2 = false;
+                Activities.HelloFailActivity.ShouldFail = false;
 
-                await client.RewindAsync("REWIND");
+                await client.RewindAsync("Activity failure test.");
 
                 var statusRewind = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
@@ -426,10 +420,8 @@ namespace DurableTask.AzureStorage.Tests
 
                 string singletonInstanceId = $"Test_{Guid.NewGuid():N}";
 
-                Orchestrations.FactorialFail.ShouldFail = false;
-
                 var client = await host.StartOrchestrationAsync(
-                    typeof(Orchestrations.FactorialFail),
+                    typeof(Orchestrations.FactorialMultipleActivityFail),
                     input: 4,
                     instanceId: singletonInstanceId);
 
@@ -437,17 +429,17 @@ namespace DurableTask.AzureStorage.Tests
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Activities.MultiplyFail.ShouldFail1 = false;
+                Activities.MultiplyMultipleActivityFail.ShouldFail1 = false;
 
-                await client.RewindAsync("rewind!");
+                await client.RewindAsync("Rewind for activity failure 1.");
 
                 statusFail = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Activities.MultiplyFail.ShouldFail2 = false;
+                Activities.MultiplyMultipleActivityFail.ShouldFail2 = false;
 
-                await client.RewindAsync("rewind!");
+                await client.RewindAsync("Rewind for activity failure 2.");
 
                 var statusRewind = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
@@ -471,11 +463,8 @@ namespace DurableTask.AzureStorage.Tests
                 string ParentInstanceId = $"Parent_{Guid.NewGuid():N}";
                 string ChildInstanceId = $"Child_{Guid.NewGuid():N}";
 
-                Activities.HelloFail.ShouldFail1 = false;
-                Activities.HelloFail.ShouldFail2 = false;
-
                 var client_parent = await host.StartOrchestrationAsync(
-                    typeof(Orchestrations.ParentWorkflow),
+                    typeof(Orchestrations.ParentWorkflowSubOrchestrationFail),
                     input: true,
                     instanceId: ParentInstanceId);
 
@@ -484,22 +473,21 @@ namespace DurableTask.AzureStorage.Tests
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Orchestrations.ChildWorkflow.ShouldFail1 = false;
+                Orchestrations.ChildWorkflowSubOrchestrationFail.ShouldFail1 = false;
 
-                await client_parent.RewindAsync("rewind!");
+                await client_parent.RewindAsync("Rewind first suborchestration failure.");
 
                 statusFail = await client_parent.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
                 Assert.AreEqual(OrchestrationStatus.Failed, statusFail?.OrchestrationStatus);
 
-                Orchestrations.ChildWorkflow.ShouldFail2 = false;
+                Orchestrations.ChildWorkflowSubOrchestrationFail.ShouldFail2 = false;
 
-                await client_parent.RewindAsync("rewind!");
+                await client_parent.RewindAsync("Rewind second suborchestration failure.");
 
                 var statusRewind = await client_parent.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
 
                 Assert.AreEqual(OrchestrationStatus.Completed, statusRewind?.OrchestrationStatus);
-                //Assert.AreEqual("\"Hello, Catherine!\"", statusRewind?.Output);
 
                 await host.StopAsync();
             }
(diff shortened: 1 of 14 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -245,6 +245,146 @@ public override Task DeleteAsync()
...
+                    break;
+                }
+            }
+            string executionId;
+            string failRow;
+
+            if (failedEntities.Count == 1)
+            {
+                // The most recent generation will always be in the first history event.
+                executionId = failedEntities[0].Properties["ExecutionId"].StringValue;
```
cgillum: Another way to get the execution ID is to query the instances table. Querying the instances table also has the benefit of giving us a proper ETag if we want to use it for our subsequent update to the Instances table.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -245,6 +245,146 @@ public override Task DeleteAsync()
...
+                failRow = failedEntities[0].RowKey;
```
cgillum: Rather than getting a single row, we probably want to collect *all* the rows that we intend to "rewind". That way we don't need to issue another query to this table.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -245,6 +245,146 @@ public override Task DeleteAsync()
...
+            var filterCondition = new StringBuilder(200);
+
+            // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81'"
+            filterCondition.Append("PartitionKey eq ").Append(Quote).Append(instanceId).Append(Quote); // = TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, instanceId);"
+            filterCondition.Append(" and ExecutionId eq ").Append(Quote).Append(executionId).Append(Quote);
+            filterCondition.Append(" and RowKey ge ").Append(Quote).Append(failRow).Append(Quote);
+
+
+
+            TableQuery query = new TableQuery().Where(filterCondition.ToString());
```
cgillum: I don't think it's necessary for us to issue another query. Instead, why don't we just save the rows that we collected in the first query and use those to do the merge? That would save a significant performance cost.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -245,6 +245,146 @@ public override Task DeleteAsync()
...
+            // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81'"
+            filterCondition.Append("PartitionKey eq ").Append(Quote).Append(instanceId).Append(Quote); // = TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, instanceId);
+            filterCondition.Append(" and OrchestrationStatus eq ").Append(Quote).Append("Failed").Append(Quote); //IS THIS THE RIGHT EVENT TYPE???
+
+            TableQuery query = new TableQuery().Where(filterCondition.ToString());
+
+            // TODO: Write-through caching should ensure that we rarely need to make this call?
+            var failedEntities = new List<DynamicTableEntity>(100);
+
+            var stopwatch = new Stopwatch();
```
cgillum: I see we're creating a `Stopwatch` but not doing anything with it. I think we need to create a new method in the AnalyticsEventSource class where we trace the result of the rewind operation, including the overall time it took to do the rewind.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -378,6 +518,38 @@ public override Task PurgeHistoryAsync(DateTime thresholdDateTimeUtc, Orchestrat
...
+            };
+            //TODO: MERGE HERE
+
+            await this.CompressLargeMessageAsync(entity);
+
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            await this.instancesTable.ExecuteAsync(
+                TableOperation.Merge(entity));
+            this.stats.StorageRequests.Increment();
+            //this.stats.TableEntitiesWritten.Increment(1);
```
cgillum: Let's un-comment this.

src/DurableTask.AzureStorage/Tracking/TrackingStoreBase.cs
```diff
@@ -36,6 +36,12 @@ abstract class TrackingStoreBase : ITrackingStore
         /// <inheritdoc />
         public abstract Task<IList<HistoryEvent>> GetHistoryEventsAsync(string instanceId, string expectedExecutionId, CancellationToken cancellationToken = default(CancellationToken));
 
+        /// <inheritdoc />
+        public abstract Task RewindFindFailureEventsAsync(string instanceId,CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <inheritdoc />
+        public abstract Task RewindClearExecutionIds(string instanceId, string executionId, string failRow, CancellationToken cancellationToken = default(CancellationToken));
```
cgillum: Let's remove this - I don't think it's necessary to be in the base class.

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -508,6 +520,20 @@ public Task ReleaseTaskOrchestrationWorkItemAsync(TaskOrchestrationWorkItem work
             await SendTaskOrchestrationMessageAsync(taskMessage);
         }
 
+        /// <inheritdoc />
+        /// //IDK WHAT TO DO HERE
```
cgillum: Does this comment still apply? :)

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -508,6 +520,20 @@ public Task ReleaseTaskOrchestrationWorkItemAsync(TaskOrchestrationWorkItem work
...
+        public async Task RewindTaskOrchestrationAsync(string instanceId, string reason)
+        {
+            var taskMessage = new TaskMessage
+            {
+                OrchestrationInstance = new OrchestrationInstance { InstanceId = instanceId },
+                Event = new GenericEvent(-1, reason) //MAKE REWIND EVENT
```
cgillum: I agree it would be a good idea to create a rewind event.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -245,6 +245,146 @@ public override Task DeleteAsync()
...
 
+        public override async Task RewindFindFailureEventsAsync(string instanceId, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var filterCondition = new StringBuilder(200);
+
+            const char Quote = '\'';
+
+            // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81'"
+            filterCondition.Append("PartitionKey eq ").Append(Quote).Append(instanceId).Append(Quote); // = TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, instanceId);
+            filterCondition.Append(" and OrchestrationStatus eq ").Append(Quote).Append("Failed").Append(Quote); //IS THIS THE RIGHT EVENT TYPE???
```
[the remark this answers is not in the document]
cmkerner20: Yes this started to break down on other orchestrators...unfortunately Table Storage doesn't have support for OrderBy as far as I can tell. I'll probably have to query for all OrchestratorStarted events within the PartitionKey and then order them on my side of things. Please let me know if you can think of a better way!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -245,6 +245,146 @@ public override Task DeleteAsync()
...
+            {
+                throw new ArgumentException("No orchestration failure found here! :)");
+            }
+            else
+            {
+                throw new ArgumentException("Ambiguous orchestration failure! << really shouldn't happen :/");
+            }
+        }
+
+        public override async Task RewindClearExecutionIds(string instanceId, string executionId, string failRow, CancellationToken cancellationToken = default(CancellationToken))
```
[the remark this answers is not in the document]
cmkerner20: Yup, that makes sense.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cmkerner20: Having the ETag would be nice, but if it isn't necessary (I don't think it would ever be ambiguous but I could be wrong) doesn't the historyTable makes more sense here because we are looking for the most recent orchestratorStarted event?

cgillum:
Adding a link to https://github.com/Azure/azure-functions-durable-extension/issues/301, since that is what this PR is primarily addressing.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1187,17 +1173,40 @@ public override Task<string> RunTask(OrchestrationContext context, string input)
                 }
             }
 
-            [KnownType(typeof(Activities.HelloFail))]
-            internal class SayHelloWithActivityFail : TaskOrchestration<string, string>
+            //[KnownType(typeof(Activities.HelloFail))]
```
cgillum: It looks like there are still a few places where we have code commented out. Can these be deleted?

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1586,7 +1707,90 @@ public override async Task<int> RunTask(OrchestrationContext context, int iterat
 
         static class Activities
         {
-            internal class HelloFail : TaskActivity<string, string>
+            internal class HelloFailActivity : TaskActivity<string, string>
+            {
+                // HACK: selectively fails activity function/orchestration so rewind can revive it
```
cgillum: I see we use the comment prefix `HACK: ` in a few places. I suggest we not do that. There are various internal Microsoft code scanners which may flag this as problematic because of how the term "hack" can be very sensitive, especially with open source code. I think `NOTE: ` or no prefix at all would be better.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -391,6 +551,38 @@ public override async Task SetNewExecutionAsync(ExecutionStartedEvent executionS
                 Utils.ExtensionVersion);
         }
 
+
+        public override async Task UpdateExistingExecutionAsync(string instanceId, DateTime newTimestamp)
```
cgillum: >DateTime newTimestamp [](start = 83, length = 21)

A few things:
- We should probably remove the `newTimestamp` parameter from this class and from the interface definition in `ITrackingStore.cs`.
- Let's rename UpdateExistingExecutionAsync to `UpdateStatusForRewindAsync`

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -256,6 +256,161 @@ public override Task DeleteAsync()
...
+                    var TSFilterCondition = new StringBuilder(200);
+
+                    TSFilterCondition.Append("PartitionKey eq ").Append(Quote).Append(instanceId).Append(Quote);
+                    TSFilterCondition.Append(" and ExecutionId eq ").Append(Quote).Append(executionId).Append(Quote);
+                    TSFilterCondition.Append(" and EventId eq ").Append(taskScheduledId);
+                    TSFilterCondition.Append(" and EventType eq ").Append(Quote).Append("TaskScheduled").Append(Quote);
+
+                    var taskScheduledEntities = await QueryHistory(TSFilterCondition.ToString(), instanceId, cancellationToken);
+
+                    taskScheduledEntities[0].Properties["EventType"] = new EntityProperty("GenericEvent");
```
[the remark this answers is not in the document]
cgillum: Reminder on GenericEvent --> RewindEvent.

---
In reply to: [206624118](https://github.com/Azure/durabletask/pull/190#discussion_r206624118) [](ancestors = 206624118)

src/DurableTask.Core/History/EventType.cs
```diff
@@ -103,6 +103,11 @@ public enum EventType
         /// </summary>
         GenericEvent,
 
+        /// <summary>
+        /// Event for tracking instance rewinds
+        /// </summary>
+        RewindEvent,
```
simonporter: >RewindEvent [](start = 8, length = 11)

Are we serializing this as an int, if so we altering HistoryState, and this is breaking, we should append only to the enum

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -218,6 +218,12 @@ public Task SendTaskOrchestrationMessageBatchAsync(params TaskMessage[] messages
             return Task.FromResult<object>(null);
         }
 
+        /// <inheritdoc />
+        public Task RewindTaskOrchestrationAsync(string instanceId, string reason)
+        {
+            throw new NotImplementedException();
```
simonporter: >NotImplementedException [](start = 22, length = 23)

I would like to see this implemented in the LocaOrchestrationService

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -1049,6 +1049,7 @@ public Task SendTaskOrchestrationMessageAsync(TaskMessage message)
             return SendTaskOrchestrationMessageBatchAsync(message);
         }
 
+
         /// <summary>
```
simonporter: nit: remove added line

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -292,11 +292,295 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
         }
 
         /// <summary>
-        /// End-to-end test which validates the cancellation of durable timers.
+        /// End-to-end test which validates the Rewind functionality on more than one orchestration.
         /// </summary>
+        [DataTestMethod]
+        //[DataRow(true)]
```
simonporter: >//[DataRow(true)] [](start = 8, length = 17)

nit: remove commented code

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -860,9 +1153,60 @@ public override async Task<long> RunTask(OrchestrationContext context, int n)
...
+                    long result = 1;
+                    for (int i = 1; i <= n; i++)
+                    {
+                        result = await (context.ScheduleTask<long>(typeof(Activities.Multiply), new[] { result, i }));
                     }
+                    if (ShouldFail)
+                    {
+                        throw new Exception("Simulating a transient, unhandled exception");
+                    }
+                    return result;
```
simonporter: nit : blank line

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -256,6 +256,175 @@ public override async Task<IList<HistoryEvent>> GetHistoryEventsAsync(string ins
...
+                entities.Count,
+                requestCount,
+                stopwatch.ElapsedMilliseconds,
+                this.GetETagValue(instanceId),
+                Utils.ExtensionVersion);
+
+            return entities; 
+        }
+
+        public override async Task<IList<string>> RewindHistoryAsync(string instanceId, IList<string> failedLeaves, CancellationToken cancellationToken)
```
simonporter: >RewindHistoryAsync [](start = 50, length = 18)

This method seems a bit unwieldy, can it be broken into logical components

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -256,6 +256,175 @@ public override async Task<IList<HistoryEvent>> GetHistoryEventsAsync(string ins
...
+        {
+            //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+            // REWIND ALGORITHM:
+            // 1. Finds failed execution of specified orchestraion instance to rewind
```
simonporter: >orchestraion [](start = 54, length = 12)

nit: typo

Manuelrmz:
Hi Chris.

That new activity does it take in count when the activity has a reply configuration applied? or will it works only when the activity failed one time?

Regards.

cgillum:
@simonporter Thanks for the review feedback. Per our discussion over email, I created a new PR which removes the public surface area changes from DurableTask.Core: https://github.com/Azure/durabletask/pull/209. We can look into adding it back later if/when we can agree on the scenario and general design (until then, it's limited to just DurableTask.AzureStorage). I also addressed all your CR feedback.

@Manuelrmz Good question, I don't think we tested for that. We'll look into it.

PR #192, opened 2018-06-14
DurableTask.AzureStorage ETW tracing improvements

cgillum:
This addresses much of https://github.com/Azure/azure-functions-durable-extension/issues/326.

Summary of changes:
* The version of the DurableTask.AzureStorage nuget package is traced everywhere.
* Added several new traces to capture errors. Previously we didn't trace unhandled exceptions when calling storage, and this made diagnosing issues very painful. The exceptions do get traced to the DurableTask.Core provider, but those traces are very difficult to correlate back to specific customer incidents.
* Refactored message processing trace into two events (was previously just one) so we can see both when a message was dequeued and when it was scheduled for processing. In production we've seen a case where messages appear to be sent but never received, and this update will help us identify what's happening
* Added a trace for when a message has its visibility updated
* Added "target instance" information to the `SendMessage` trace to make it easier to correlate sub-orchestrations back to their parents
* Other misc. improvements, including code formatting

FYI @tohling and @kashimiz to help review

### DIFF
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -72,7 +72,7 @@ namespace DurableTask.AzureStorage
 #endif
         }
 
-        [Event(101, Level = EventLevel.Informational, Opcode = EventOpcode.Send)]
+        [Event(101, Level = EventLevel.Informational, Opcode = EventOpcode.Send, Task = Tasks.Enqueue, Version = 2)]
         public void SendingMessage(
             Guid relatedActivityId,
             string Account,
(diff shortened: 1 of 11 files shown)


src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -81,13 +81,28 @@ private static void EnsureLogicalTraceActivityId()
...
+                Account,
+                TaskHub,
+                EventType,
+                InstanceId ?? string.Empty,
+                ExecutionId ?? string.Empty,
+                SizeInBytes,
+                PartitionId,
+                TargetInstanceId,
+                TargetExecutionId ?? string.Empty,
+                ExtensionVersion);
```
tohling: >ExtensionVersion [](start = 16, length = 16)

Seems like the concept of ExtensionVersion is scoped to DurableTask.AzureStorage. Instead of having it plumbed for every single log method, we could also just call Utils.ExtensionVersion here.

cgillum: I considered that, but the EventSource implementation requires the method parameters to match the WriteEvent calls, so I can't just plum it in internally.

PR #193, opened 2018-06-15
Updated Newtonsoft.Json to v11.0.2, WindowsAzure.Storage to v8.6.0.

kashimiz:
Updating to fix: 

Azure/azure-functions-durable-extension/issues/342 (Caused by breaking change from between Newtonsoft.Json 10.0.3 and 11.0.2)
Azure/azure-functions-durable-extension/issues/343 (Caused by mismatched WindowsAzure.Storage versions between Azure Functions host and DurableTask)

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -19,8 +19,8 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
-    <PackageReference Include="WindowsAzure.Storage" version="8.5.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="WindowsAzure.Storage" version="8.6.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -13,7 +13,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="ImpromptuInterface" Version="7.0.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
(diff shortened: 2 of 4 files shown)


src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -13,7 +13,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="ImpromptuInterface" Version="7.0.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
```
cgillum: Ouch, this means we'll also need an updated deployment of DurableTask.Core. In that case, we'll also need to increment the version of this package to 2.0.0.6 (see [here](https://github.com/Azure/durabletask/blob/0a93eba6c9692f52776dac0f3741eccad3293ef6/tools/DurableTask.props#L35-L37)).

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -19,8 +19,8 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
-    <PackageReference Include="WindowsAzure.Storage" version="8.5.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="WindowsAzure.Storage" version="8.6.0" />
```
cgillum: Just to confirm, no code changes are necessary?

kashimiz: Confirmed. 

The [breaking change](https://www.fuget.org/packages/Newtonsoft.Json/11.0.2/lib/netstandard2.0/diff/10.0.3/) between Newtonsoft.Json v10 and v11 is a switch from DefaultSerializationBinder inheriting from Newtonsoft.Json.Json.SerializationBinder (v10) to System.Runtime.Serialization.SerializationBinder (v11). It doesn't change the API surface of the DefaultSerializationBinder that DurableTask.Core.Serializing.PackageUpgradeSerializationBinder inherits from. There were no breaking changes [between WindowsAzure.Storage 8.5.0 and 8.6.0](https://www.fuget.org/packages/WindowsAzure.Storage/8.6.0/lib/netstandard1.3/diff/8.5.0/).

PR #194, opened 2018-06-15
DurableTask.AzureStorage: Alternate fix for the 4 MB max entity size which covers more scenarios.

cgillum:
This is a revised version of a previous pull request: https://github.com/Azure/durabletask/pull/191.  After merging, I did some testing and found a few issues. This new version addresses those issues.

FYI @tohling, @gled4er

This addresses https://github.com/Azure/azure-functions-durable-extension/issues/339.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -35,6 +35,7 @@ namespace DurableTask.AzureStorage.Tracking
     /// </summary>
     class AzureTableTrackingStore : TrackingStoreBase
     {
+        const string NameProperty = "Name";
         const string InputProperty = "Input";
         const string ResultProperty = "Result";
         const string OutputProperty = "Output";
@@ -42,6 +43,14 @@ namespace DurableTask.AzureStorage.Tracking
         const string SentinelRowKey = "sentinel";
         const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
         const int GuidByteSize = 72;
+
+        static readonly string[] VariableSizeEntityProperties = new[]
+        {
+            NameProperty,
+            InputProperty,
+            ResultProperty,
+            OutputProperty,
+        };
 
         readonly string storageAccountName;
         readonly string taskHubName;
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -513,10 +535,22 @@ public override Task StartAsync()
...
+            // Assume at least 1 KB of data per entity to account for static-length properties
+            int estimatedByteCount = 1024;
+
+            // Count the bytes for variable-length properties, which are assumed to always be strings
+            foreach (string propertyName in VariableSizeEntityProperties)
+            {
+                EntityProperty property;
+                if (entity.Properties.TryGetValue(propertyName, out property) && !string.IsNullOrEmpty(property.StringValue))
+                {
+                    estimatedByteCount += Encoding.Unicode.GetByteCount(property.StringValue);
```
kashimiz: Curious; why UTF-16? Is this the format Table Storage uses?

cgillum: Yep. This basically means 2 bytes for every character in a string.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -513,10 +535,22 @@ public override Task StartAsync()
...
 
-        void UpdateEstimatedBytes(ref int estimatedBytes, string payload)
+        static int GetEstimatedByteCount(DynamicTableEntity entity)
         {
-            int payloadBytes = Encoding.Unicode.GetByteCount(payload);
-            estimatedBytes += payloadBytes > MaxStorageQueuePayloadSizeInBytes ? GuidByteSize : payloadBytes;
+            // Assume at least 1 KB of data per entity to account for static-length properties
+            int estimatedByteCount = 1024;
+
+            // Count the bytes for variable-length properties, which are assumed to always be strings
```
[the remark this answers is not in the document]
cgillum: Not all, but the ones in the `VariableSizeEntityProperties` collection do. With the exception of `Name` (and sometimes `Result`), these are all serialized JSON payloads. If we wanted, we could store these as binary data as well, but it's much easier to debug things when they are strings.

PR #196, opened 2018-07-03
Better concurrency throttling

cgillum:
This PR is intended to fix https://github.com/Azure/azure-functions-durable-extension/issues/370, which represents a significant performance bottleneck when running large number of orchestrations or activities.

### Problem
The WorkItemDispatcher sleeps for 5 seconds whenever it notices that we've reached the maximum configured concurrency. This looping 5 second sleep effectively blocks all orchestration processing and can have a significant impact on throughput, especially when orchestrations or activities normally execute quickly.

### Change
The change is to replace the looping 5 second sleep with a `SemaphoreSlim` named `concurrencyLock` to block additional fetches until we're under our throttle limitations. With this solution, there is no sleeping, and the dispatcher will **immediately** start a new fetch as soon as a previous work-item is completed, rather than waiting for an arbitrary number of seconds.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -593,59 +593,74 @@ namespace DurableTask.AzureStorage
 
             await this.EnsureTaskHubAsync();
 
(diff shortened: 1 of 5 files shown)


cgillum:
Verified all tests are passing and the release build has no warnings.

PR #200, opened 2018-07-31
Enabling conditional query

TsuyoshiUshio:
Hi @cgillum 

This PR is related this issue. I open this PR to start conversation. 

https://github.com/Azure/azure-functions-durable-extension/issues/358

I'd like to discuss some architecture decision.  Please review the initial idea and feel free to give me comments. 

#  Query parameter object 

I'd like to create a query object which represent the query parameter. I create OrchestrationInstanceStatusQueryBuilder for this purpose. I have several ideas however, I should talk with you before I implement whole.  If you prefer other implementation, please let me know.

## 1. Builder vs Query object

Now I'm implementing as a builder. However, I also can implement as Query object which has property of parameter with GetQuery() method to return TableQuery<OrchestrationInstanceStatus>. Both is OK for me. However, you seems hide OrchestrationInstanceStatus from public, then, we need to create a QueryParameter class in DurableTask.Core namespace which you want to avoid to change. 
I also can implement Parse method to the builder to enable to parse the query parameter object, however, since I'm not sure if I can  create the parameter object in DurableTask.Core, I postpone the decision. 

Or do you have any other cool idea for querying? 

IMO, creating query object  or build could gain the testability and feel safe to handle complex condition generation by a unit test. 

## 2. OrchestrationInstanceStatusQueryBuilder:  public TableQuery<T> Build<T>()  method

This method is only used for TableQuery<OrchestrationInstanceStatus> as T. However, If I make it `public TableQuery<OrchestrationInstanceStatus> Build()` method, it requires make OrchestrationInstanceStatus to public. However, I want to keep testability of this class. Do you have any other good idea?

## 3. scope of the class

I need to public a class for query parameter object. It should be in the DurableTask.Core however, maybe, we might start without changing DurableTask.Core. In this case, where should the query parameter class(currently, OrchestrationStatusQueryBuilder) sits? I mean namespace and directory. 
Looks good?


TsuyoshiUshio:
I found an issue for Builder solution. I'll make it Condition object solution. :)

cgillum:
It's an interesting idea, but how would you expose this to the public APIs? For example, what would a REST API look like?

TsuyoshiUshio:
RestAPI will be something like this.  

```
GET /runtime/webhooks/DurableTaskExtension/instances/?taskHub={taskHub}&connection={connection}&code={systemKey}&createdTimeFrom={createdTimeFrom}&createdTimeTo={createdTimeTo}&{lastUpdatedTimeFrom={lastUpdatedTimeFrom}&lastUpdatedTimeTo={lastUpdatedTimeTo}&runtimeStatus={runtimeStatus}
```
All additional parameters are optional. 

Then, we can change the HttpApiHander: HandleGetStatusRequestAsync into something like this. 

```
            DurableOrchestrationClientBase client = this.GetClient(request);
            IList<DurableOrchestrationStatus> statusForAllInstances = await client.GetStatusAsync(OrchestrationInstanceStatusQueryCondition.Parse(request.GetQueryNameValuePairs()));
```

https://github.com/Azure/azure-functions-durable-extension/blob/master/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs#L161

cgillum:
If the REST API just takes a series of optional query string parameters, why not have the same list of parameters for the `GetStateAsync(...)` method?

TsuyoshiUshio:
@cgillum 

In this case, we have 4 DateTime parameters CreatedTimeFrom, CreatedTimeTo,LastUpdatedTimeFrom, LastUpdatedTimeTo and RuntimeStatus. Then, if we want to query only CreatedTimeFrom, or only LastUpdatedTimeFrom, then the API might be messy. (we need to fill with null or default(DateTime)?))  Also, Condition can be complex in the future. In this case, if we create condition object, we can unit test it.

TsuyoshiUshio:
You mean, it is difficult to expose Condition object to the WebJobs.Extensions.DurableTask repo?

cgillum:
Yes, generally we don't want to expose types defined in DurableTask.* packages to Durable Functions customers. That allows us to more easily swap out dependencies in the future (for example, if we wanted to replace DurableTask.AzureStorage with a hypothetical DurableTask.CosmosDB. We also have to think about how we can translate this into other languages like JavaScript.

To avoid over-engineering and to avoid too many parameters, which should focus on the most important scenarios and add parameters for those. In the issue you referenced, Mark suggests two filters: status and creation date. I don't think there is a strong case for last updated date. So I would reduce it to three parameters (createdFrom, createdTo, and runtimeStatus).  If customers push us with stronger requirements, then we can revisit and consider adding other options.

TsuyoshiUshio:
Make sense. Thanks! I’ll do that!

TsuyoshiUshio:
Hi @cgillum,

I change the method signatures. If you like this, I'll create a nuget package and implement Durable Extension side. :) Please review this.

TsuyoshiUshio:
Maybe you might love GetStatusAsync() instead of QueryStatusAsync().

cgillum:
I like this much better.

I do wonder if we should use `GetStateAsync()` instead of `QueryStatusAsync`. The reason is mostly because we already have other methods named `GetStateAsync`, and it's more natural to developers if we have overloads. for different ways to fetch the state

TsuyoshiUshio:
Thanks. I'll change into the `GetStatusAsync` Then start developing the main repo.

TsuyoshiUshio:
@cgillum  Done. After testing to implement Durable Extensions Repo, then I'll tell you. If you need additional fix for this PR, please let me know.

TsuyoshiUshio:
When I tried without parameters, new logic caused an error.  The reason is, if there is no parameter on Where(...) method, it cause this error. This is fixed. 

```
[05/08/2018 07:43:19] Executed 'Query' (Failed, Id=31af8789-cd6b-4734-aa69-93d8a88dd4f3)
[05/08/2018 07:43:19] System.Private.CoreLib: Exception while executing function: Query. System.Linq: Sequence contains no elements.
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1760,6 +1760,20 @@ public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(Cancella
...
         }
 
+        /// <summary>
+        /// Get states which match the condition
+        /// </summary>
+        /// <param name="createdTimeFrom">fetch status greater than this createdTime</param>
+        /// <param name="createdTimeTo">fetch status less than this createdTime createdTime</param>
+        /// <param name="runtimeStatus">fetch status matches this runtimeStatus</param>
+        /// <param name="cancellationToken">Cancellation Token</param>
+        /// <returns>List of <see cref="OrchestrationState"/></returns>
```
cgillum: Looks like there are some formatting and consistency issues with your documentation comments. Can you please update them to be more consistent with the `GetOrchestrationStateAsync` overload above?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1760,6 +1760,20 @@ public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(Cancella
...
+        public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(DateTime createdTimeFrom, DateTime createdTimeTo, string runtimeStatus, CancellationToken cancellationToken = default(CancellationToken))
```
cgillum: Should `createdTimeTo` be optional?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -325,13 +325,26 @@ private async Task<OrchestrationState> ConvertFromAsync(OrchestrationInstanceSta
         public override async Task<IList<OrchestrationState>> GetStateAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
             var query = new TableQuery<OrchestrationInstanceStatus>();
-            TableContinuationToken token = null;
+            return await QueryStateAsync(query, cancellationToken);
```
cgillum: nit: remove `async` and `await` from this method. It is not necessary to have the extra overhead here. You can just do this:

```csharp
return this.QueryStateAsync(query, cancellationToken);
```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -325,13 +325,26 @@ private async Task<OrchestrationState> ConvertFromAsync(OrchestrationInstanceSta
...
+        }
 
-            var orchestrationStates = new List<OrchestrationState>(100);
+        public override async Task<IList<OrchestrationState>> GetStateAsync(DateTime createdTimeFrom, DateTime createdTimeTo, String runtimeStatus, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return await QueryStateAsync(new OrchestrationInstanceStatusQueryCondition
```
cgillum: Same comment about removing `async` and `await`.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Microsoft.WindowsAzure.Storage.Table;
```
cgillum: Please add the standard file header and put the `using` statements under the `namespace` declaration like the rest of the files in this project.

Test/DurableTask.AzureStorage.Tests/OrchestrationInstanceStatusQueryConditionTest.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
```
cgillum: Please add the standard file header and put the `using` statements under the `namespace` declaration like the rest of the files in this project.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: I update the expression.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done

Test/DurableTask.AzureStorage.Tests/OrchestrationInstanceStatusQueryConditionTest.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
```diff
@@ -89,7 +89,7 @@ interface ITrackingStore
         /// <param name="RuntimeStatus">RuntimeStatus</param>
         /// <param name="cancellationToken">cancellation token</param>
         /// <returns></returns>
-        Task<IList<OrchestrationState>> GetStateAsync(DateTime createdTimeFrom, DateTime createdTimeTo, String RuntimeStatus, CancellationToken cancellationToken = default(CancellationToken));
+        Task<IList<OrchestrationState>> GetStateAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<string> RuntimeStatus, CancellationToken cancellationToken = default(CancellationToken));
```
cgillum: * `RuntimeStatus` should be camel-case, like `runtimeStatus` (small r) since it is a C# parameter.
* Also, this parameter needs to be `IEnumerable<OrchestrationStatus>` instead of `IEnumerable<string>` to make it strongly-typed.

Test/DurableTask.AzureStorage.Tests/OrchestrationInstanceStatusQueryConditionTest.cs
```diff
@@ -0,0 +1,130 @@
...
+            };
+            Assert.AreEqual("((CreatedTime ge datetime'2018-01-10T01:10:10.0000000Z') and (CreatedTime le datetime'2018-01-10T01:10:50.0000000Z')) and (RuntimeStatus eq 'Running')", condition.ToTableQuery<OrchestrationInstanceStatus>().FilterString);
+
+        }
+        [TestMethod]
+        public void OrchestrationInstanceQuery_NoParameter()
+        {
+            var condition = new OrchestrationInstanceStatusQueryCondition();
+            var query = condition.ToTableQuery<OrchestrationInstanceStatus>();
+            Assert.IsNull(query.Expression);
```
cgillum: Do we want `null` or should this be `string.Empty`? Returning `null` can be dangerous if it can cause `NullReferenceException`.

Test/DurableTask.AzureStorage.Tests/OrchestrationInstanceStatusQueryConditionTest.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        [TestMethod]
+        public void OrchestrationInstanceQuery_MultipleRuntimeStatus()
+        {
+            var runtimeStatus = new List<string>();
+            runtimeStatus.Add("Running");
+            runtimeStatus.Add("Completed");
```
cgillum: Tip: you can use this as a shortcut:
```csharp
var runtimeStatus = new[] {"Running", "Completed" };
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1751,7 +1751,7 @@ public async Task<OrchestrationState> GetOrchestrationStateAsync(string instance
         }
 
         /// <summary>
-        /// Get states of the all orchestration instances
+        /// Get states all orchestration instances
```
cgillum: "Gets the state of all orchestration instances."

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1760,6 +1760,20 @@ public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(Cancella
             return await this.trackingStore.GetStateAsync(cancellationToken);
         }
 
+        /// <summary>
+        /// Get states which match the specified parameters.
```
cgillum: "Gets the state of all orchestration instances that match the specified parameters."

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -0,0 +1,109 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using System.Xml.Linq;
+
+    /// <summary>
+    /// OrchestrationInstanceStatusQueryBuilder is a builder to create a StorageTable Query
+    /// </summary>
+    public class OrchestrationInstanceStatusQueryCondition
```
cgillum: Can we make this type `internal`? That way you don't have to document all the members. Unit tests can create `internal` types.

cgillum:
Also, we just merged Catherine's changes today, so you will probably need to resolve any conflicts this created in your PR.

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
TsuyoshiUshio: Done

Test/DurableTask.AzureStorage.Tests/OrchestrationInstanceStatusQueryConditionTest.cs
TsuyoshiUshio: Hi @cgillum ,

This is no problem. ToTableQuery method returns TableQuery object.  it is not null.

We use this like this. We never touch the query.Expression. 

```
                var segment = await this.instancesTable.ExecuteQuerySegmentedAsync(query, token);
```

The Expression can be return null. (Not mentioned on the document, maybe) 

https://docs.microsoft.com/en-us/dotnet/api/microsoft.windowsazure.storage.table.tablequery-1.expression?view=azure-dotnet#Microsoft_WindowsAzure_Storage_Table_TableQuery_1_Expression

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
TsuyoshiUshio: I didn't know that. Thanks. 

```
#if !SIGN_ASSEMBLY
[assembly: InternalsVisibleTo("DurableTask.AzureStorage.Tests")]
#endif
```

Test/DurableTask.AzureStorage.Tests/OrchestrationInstanceStatusQueryConditionTest.cs
TsuyoshiUshio: Thanks! Done

TsuyoshiUshio:
Hi @cgillum , 
I've done all, and resolve the conflict. 

When I tested locally, I noticed that some test fails. It might be freaky ones. Both related time related assertion.

* AzureStorageScaleTests.TestInstanceAndMessageDistribution
* AzureStorageScenarioTests.OrchestrationConcurrency

Is it Ok?

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -69,7 +59,7 @@ private string GetConditions()
                 conditions.Add(TableQuery.GenerateFilterConditionForDate("CreatedTime", QueryComparisons.LessThanOrEqual, new DateTimeOffset(this.CreatedTimeTo)));
             }
 
-            if (this.RuntimeStatus != null)
+            if (this.RuntimeStatus != null && this.RuntimeStatus.Count() != 0)
```
cgillum: When dealing with `IEnumerable<T>`, it's better to say `this.RuntimeStatus.Any()` instead of `this.RuntimeStatus.Count() != 0`. This is because `.Count()` may actually walk through the entire enumerable, which is much more expensive than `.Any()` which only needs to look and see if a first element exists.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -23,25 +24,14 @@ namespace DurableTask.AzureStorage.Tracking
     /// <summary>
     /// OrchestrationInstanceStatusQueryBuilder is a builder to create a StorageTable Query
     /// </summary>
-    public class OrchestrationInstanceStatusQueryCondition
+    internal class OrchestrationInstanceStatusQueryCondition
     {
-        /// <summary>
-        /// RuntimeStatus
-        /// </summary>
         public IEnumerable<string> RuntimeStatus { get; set; }
```
cgillum: This should also be `IEnumerable<OrchestrationStatus>`.

TsuyoshiUshio: Thanks, Done

TsuyoshiUshio:
Yes. I run these test for the every time. :) It is safety guard for me. I'm a TDD guy, usually. :)

PR #203, opened 2018-08-11
Updated DurableTask.ServiceBus to be .NET Standard Compatible

Preecington:
I took a stab at making `DurableTask.ServiceBus` work in .NET standard - it turned out to be a far larger undertaking then I anticipated!

With the following PR, all tests pass in isolation and the stress test runs to completion with standard settings. There's a few caveats/nits here that may not make it suitable for immediate merge to master that are probably best looked at by someone more familiar with the history of the project:

- Because the `Microsoft.Azure.ServiceBus.Maintenance`  namespace has only [recently been added](https://github.com/Azure/azure-service-bus-dotnet/issues/65) to the .NET Standard 2.0 library, this PR is based on [ServiceBus 3.1.0-preview](https://www.nuget.org/packages/Microsoft.Azure.ServiceBus/3.1.0-preview). This may not be acceptable for production environments.
 - I was more focused on getting .NET Standard 2.0 / .NET Framework 4.6.1 to work in this PR, so I removed most of the 4.5.1 build settings. I'm not sure if this is still a requirement for this project so it might need more work to include 4.5.1 back in.
- The unit tests perform... *strangely*... when working together and I can't puzzle out why. In isolation they will all pass (except for one troublesome test in AzureStorage needing a few re-runs that I see is also temperamental on `master`) - but when you run them as a group some orchestrations will start hitting timeouts. I'm not sure if this is an issue with my machine, with my setup, with the .NET Core MSTest runner, or to do with the fact I'm using ReSharper for running tests.

All in all, I hope someone finds the below work useful and can use it as a base to get Durable Tasks working completely in .NET Core. 🎉


alextolp:
>   </ItemGroup>

Either we keep net451 target or we can remove this

---
Refers to: src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj:19 in 6917b19. [](commit_id = 6917b1976df91d810b34ea2c00fc1822ff5895f7, deletion_comment = False)

alextolp:
>   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">

and this

---
Refers to: src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj:21 in 6917b19. [](commit_id = 6917b1976df91d810b34ea2c00fc1822ff5895f7, deletion_comment = False)

alextolp:
>   <ItemGroup Condition="'$(TargetFramework)' == 'net451'">

here too

---
Refers to: src/DurableTask.Core/DurableTask.Core.csproj:9 in 6917b19. [](commit_id = 6917b1976df91d810b34ea2c00fc1822ff5895f7, deletion_comment = False)

src/DurableTask.Emulator/DurableTask.Emulator.csproj
```diff
@@ -2,7 +2,7 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net451</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
```
alextolp: here as well

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -345,27 +329,27 @@ public async Task<bool> HubExistsAsync()
         ///     Get the count of pending orchestrationsb
         /// </summary>
         /// <returns>Count of pending orchestrations</returns>
-        public long GetPendingOrchestrationsCount()
+        public async Task<long> GetPendingOrchestrationsCount()
         {
-            return GetQueueCount(orchestratorEntityName);
+            return await GetQueueCount(orchestratorEntityName);
```
alextolp: >await [](start = 19, length = 5)

nit: remove async and the await. you can simply return the Task

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -345,27 +329,27 @@ public async Task<bool> HubExistsAsync()
...
 
         /// <summary>
         ///     Get the count of pending work items (activities)
         /// </summary>
         /// <returns>Count of pending activities</returns>
-        public long GetPendingWorkItemsCount()
+        public async Task<long> GetPendingWorkItemsCount()
         {
-            return GetQueueCount(workerEntityName);
+            return await GetQueueCount(workerEntityName);
```
alextolp: >await [](start = 19, length = 5)

same here

cgillum:
FYI @simonporter. I see that this PR removes support for .NET 4.5.1 and targets only .NET Standard 2.0.  I'm okay with this from a Durable Functions perspective, but is this going to be acceptable for the other teams who depend on DTFx (especially since net451 is not compatible with .NET Standard 2.0)?

simonporter:
No this would be a problem for multiple users, we need to maintain net451 or we have to fork if that;s not possible, preferably the former

---
In reply to: [414771279](https://github.com/Azure/durabletask/pull/203#issuecomment-414771279) [](ancestors = 414771279)

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -345,27 +329,27 @@ public async Task<bool> HubExistsAsync()
         ///     Get the count of pending orchestrationsb
         /// </summary>
         /// <returns>Count of pending orchestrations</returns>
-        public long GetPendingOrchestrationsCount()
+        public async Task<long> GetPendingOrchestrationsCount()
```
simonporter: >GetPendingOrchestrationsCount [](start = 32, length = 29)

If you making methods async, please add Async to the name, here and other places

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -950,11 +934,10 @@ public async Task CompleteTaskActivityWorkItemAsync(TaskActivityWorkItem workIte
...
                     () => $@"Created new TaskActivity Transaction - txnid: {
                         Transaction.Current.TransactionInformation.LocalIdentifier
-                        } - msg seq and locktoken: [SEQ: {originalMessage.SequenceNumber} LT: {originalMessage.LockToken}]");
+                        } - msg seq and locktoken: [SEQ: {originalMessage.SystemProperties.SequenceNumber} LT: {originalMessage.SystemProperties.LockToken}]");
 
-                await Task.WhenAll(
-                    workerQueueClient.CompleteAsync(originalMessage.LockToken),
-                    orchestratorSender.SendAsync(brokeredResponseMessage));
+                await workerReceiver.CompleteAsync(originalMessage.SystemProperties.LockToken);
+                await orchestratorSender.SendAsync(brokeredResponseMessage);
```
simonporter: The semantics have been changed here which can increase latency, any reason we need this or can we maintain current behavior.

src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs
```diff
@@ -55,7 +56,15 @@ public static Task<BrokeredMessage> GetBrokeredMessageFromObjectAsync(object ser
 
             if (compressionSettings.Style == CompressionStyle.Legacy)
             {
-                return new BrokeredMessage(serializableObject) { SessionId = instance?.InstanceId };
+                using (var ms = new MemoryStream())
+                {
+                    var serialiser = (XmlObjectSerializer)typeof(DataContractBinarySerializer<>)
```
simonporter: >serialiser [](start = 24, length = 10)

serializer

src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs
```diff
@@ -55,7 +56,15 @@ public static Task<BrokeredMessage> GetBrokeredMessageFromObjectAsync(object ser
...
+                        .MakeGenericType(serializableObject.GetType())
+                        .GetField("Instance")
+                        ?.GetValue(null);
+                    serialiser?.WriteObject(ms,serializableObject);
```
simonporter: why is this needed?
Is this binary compatible with current implementation or is this a contract breaking change?

src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs
```diff
@@ -166,14 +179,14 @@ static BrokeredMessage GenerateBrokeredMessageWithCompressionTypeProperty(Stream
                 () => $"Saving the message stream in blob storage using key {blobKey}.");
             await orchestrationServiceBlobStore.SaveStreamAsync(blobKey, stream);
 
-            BrokeredMessage brokeredMessage = new BrokeredMessage();
-            brokeredMessage.Properties[ServiceBusConstants.MessageBlobKey] = blobKey;
-            brokeredMessage.Properties[FrameworkConstants.CompressionTypePropertyName] = compressionType;
+            Message brokeredMessage = new Message();
+            brokeredMessage.UserProperties[ServiceBusConstants.MessageBlobKey] = blobKey;
```
simonporter: >UserProperties [](start = 28, length = 14)

Are these compatible across SDKs, is a Property set with the old version === a UserProperty in this version

src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs
```diff
@@ -185,15 +198,16 @@ public static async Task<T> GetObjectFromBrokeredMessageAsync<T>(BrokeredMessage
...
             {
                 compressionType = (string)compressionTypeObj;
             }
 
             if (string.IsNullOrEmpty(compressionType))
             {
                 // no compression, legacy style
-                deserializedObject = message.GetBody<T>();
+                using (var ms = new MemoryStream(message.Body))
+                    deserializedObject = (T)DataContractBinarySerializer<T>.Instance.ReadObject(ms);
```
simonporter: >deserializedObject = (T)DataContractBinarySerializer<T>.Instance.ReadObject(ms); [](start = 20, length = 80)

put in brackets for readability

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -344,12 +344,13 @@ string GetSecondaryFilterExpression(OrchestrationStateQueryFilter filter)
             return filterExpression;
         }
 
+        private static readonly DateTimeOffset MinDateTime = new DateTimeOffset(1601, 1, 1, 0, 0, 0, TimeSpan.Zero);
```
simonporter: >private static readonly DateTimeOffset MinDateTime = new DateTimeOffset(1601, 1, 1, 0, 0, 0, TimeSpan.Zero); [](start = 8, length = 108)

Why was this extracted from TableContants

src/DurableTask.ServiceBus/Tracking/BlobStorageClient.cs
```diff
@@ -109,10 +109,20 @@ async Task<ICloudBlob> GetCloudBlockBlobReferenceAsync(string containerNameSuffi
...
+            do
+            {
+                var response = await this.blobClient.ListContainersSegmentedAsync(this.containerNamePrefix,continuationToken);
+                continuationToken = response.ContinuationToken;
+                results.AddRange(response.Results);
+            }
+            while (continuationToken != null);
+            return results;
         }
+
```
simonporter: nit: remove extra line

src/DurableTask.ServiceBus/Tracking/BlobStorageClient.cs
```diff
@@ -109,10 +109,20 @@ async Task<ICloudBlob> GetCloudBlockBlobReferenceAsync(string containerNameSuffi
         /// List all containers of the blob storage, whose prefix is containerNamePrefix, i.e., {hubName}-dtfx.
         /// </summary>
         /// <returns>A list of Azure blob containers</returns>
-        public IEnumerable<CloudBlobContainer> ListContainers()
+        public async Task<IEnumerable<CloudBlobContainer>> ListContainers()
```
simonporter: >ListContainers [](start = 59, length = 14)

name Async

test/DurableTask.Stress.Tests/App.config
```diff
@@ -15,7 +15,7 @@
   </appSettings>
   <connectionStrings>
     <!-- Service Bus connection string -->
-    <add name="Microsoft.ServiceBus.ConnectionString" connectionString="[Enter you connection string here]" />
```
simonporter: >[Enter you connection string here] [](start = 72, length = 34)

nit: keep this

simonporter:
>     <PackageReference Include="WindowsAzure.Storage" version="8.6.0" />

Tests are using a conflicting version @ , Chris is this ok? I thought there was a hard dependency on version.

---
Refers to: src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj:23 in 6917b19. [](commit_id = 6917b1976df91d810b34ea2c00fc1822ff5895f7, deletion_comment = False)

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -2,7 +2,7 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net451</TargetFrameworks>
```
simonporter: >net451 [](start = 37, length = 6)

Per comment in discussion. There are multiple users that have a dependency on this, is it possible to maintain compatibility?

Preecington: @simonporter Thanks for taking the time to look at my PR! I wasn’t sure if this was the case, and took the (easier) path of just supporting standard. I’m happy to take the time to rework 4.51 in this, but it might take a few days.

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
Preecington: @simonporter This is because the constant is now internal, for some strange reason: https://github.com/Azure/azure-storage-net/issues/627

Do you have any thoughts on better alternatives?

src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs
Preecington: @simonporter Yes, that’s correct.

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
Preecington: @simonporter This is due to what I think is a implementation change in the Service Bus library. Under unit tests, and other scenarios I tested, you were no longer allowed to access two different queues simultaneously in parallel in the same transaction - which is why I’ve added the change. 

I may have misunderstood the transaction change though and I’m happy to have someone verify my interpretation.

src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs
Preecington: @simonporter I was attempting to keep the implementation the same; strictly I don’t think this is necessary (the new DataContractSerializer is nothing fancy and I can reimplement it easily enough in DTFx) but I wanted to err on the side of maintaining compatibility just in case. Thoughts?

Preecington:
@simonporter @cgillum I've been giving some thought to re-adding support for .NET 4.5.2 into this PR and there's a few things we can try. I'd like to get some opinions before I go off and make the change. I'm really keen to try and get this merged so the rest of the community can use it as well. 👍 

Since the old Framework library and the new Standard library for Service Bus differ significantly, the options I see available are:

- Add an enormous amount of conditionals to the Orchestrator implementation and friends to switch between `BrokeredMessage` and `Message` implementations (This means we get a lot of messy duplicate code)
- Fork the csproj in this solution and have a "Framework" and "Standard" NuGet packages (we'll also have to duplicate the tests in most cases, and this also makes for a maintenance nightmare)
- Refactor the current csproj radically to 'funnel' the service bus dependency into a factory or similar pattern to build an abstracted unit of work that can work across both library implementations (This is a mix of the above: We reduce the conditionals to one place but end up changing the code a lot

I'm leaning towards the third options, but I'd like some direction (or even another idea entirely) from the team before I make some decisions.

test/DurableTask.Stress.Tests/DurableTask.Stress.Tests.csproj
```diff
@@ -2,23 +2,29 @@
...
 	<PropertyGroup>
-		<TargetFramework>net451</TargetFramework>
+	  <TargetFramework>netcoreapp2.1</TargetFramework>
+	  <StartupObject>DurableTask.Stress.Tests.Program</StartupObject>
+	  <ApplicationIcon />
+	  <OutputType>Exe</OutputType>
 	</PropertyGroup>
+	<ItemGroup>
+	  <None Remove="eventFlowConfig.json" />
+	</ItemGroup>
```
alextolp: nit: why is this needed?

simonporter:
I also think the 3rd option would be the cleanest. Multiple packages will be hard to maintain and test.


---
In reply to: [416005299](https://github.com/Azure/durabletask/pull/203#issuecomment-416005299) [](ancestors = 416005299)

simonporter:
@gfrancomsft is added to the review. #Closed

gfrancomsft:
3rd option seems the cleanest however to make it even easier to obsolete 4.5.1 in future releases can you put the Service Bus APIs calls behind an interface.
The final goal is to obsolete the current implementation that supports 4.5.1 with few changes as possible.
This could be achieved if we 
1 - Create an interface/s that maps 1-1 to the new Service Bus APIs calls.
2 - Create a class that implements these APIs, every API calls contains have a #define
- #define for .NET Standard maps 1 to 1 to APIs, basically every call will contain a single line with the call to the Service Bus new APIs
- #define for .NET 4.5.1 emulates the behavior of the new calls using the calls from 4.5.1
If we can do that, the provider code will be clean from versioning while the abstraction will contains the two implementations, once we decide to obsolete 4.5.1 in the future we just need to drop the file with the two implementations and just replace the abstracted class names in the provider with the new versions.


---
In reply to: [427882163](https://github.com/Azure/durabletask/pull/203#issuecomment-427882163) [](ancestors = 427882163,416005299)

Tadimsky:
Are there any plans to resume this work? I'd love to adopt Durable Tasks but our services are all .NET Core and so would be fantastic if this was done at some point soon :).

zumwald:
@Preecington is this a mythical-man-month situation, or could a second pair of hands help?

That is, offering my help if it would do anything more than slow you down.

Preecington:
Appreciate the offer @zumwald ! Right now it's all just stuck in my head and I'm working through it as I get to steal moments, so I think it's best I stick to it alone until I get to a point I'm able to explain to others and not just step on toes.

zumwald:
Fair enough, let me know when things change and I’ll roll up my sleeves!

adarsh1:
@Preecington This is really great stuff, do you have a timeline in mind? We would really love to have this in as soon as we can, would be happy to join the effort, if it might help make progress :)

malafreniere:
ETA ?

adarsh1:
@Preecington we would really like this in ASAP, would appreciate some sort of a timeline here.

Preecington:
Sorry to go communication-dark on this one @adarsh1 - I've been prototyping different abstraction solutions and I think I've finally found one that ticks all the boxes. At this point I'm hoping to use the period over Easter to nail down the final class (the orchestration) to work multi platform. Then I'll be looking for anyone who wants to help like yourself or @zumwald to validate the decisions I've made. It was a much harder problem then I originally envisioned which isn't helped by the sheer number of sealed classes inside the Fx Service Bus libraries. 

If I can't solve the abstraction piece over Easter I'll be looking to hand this one off to someone else, as I won't get another chunk of time to work on it in the near future and I think it's too important to keep on hold forever.

Preecington:
@adarsh1 @gfrancomsft @simonporter @cgillum It's done! I am sorry this took as long as it did.

I've created a set of abstraction classes that tries to wrap both API points as close as possible whilst sticking to the new .NET Stardard Service Bus Library signatures. It leads to some strange looking code (such as class names with interface naming standards) but overall I think it makes the orchestration code much neater without too many `#IFDEF`s around.

The unit tests and stress tests pass for both .NET Standard and .NET 4.5 implementations. I'm looking forward to your feedback so we can get this one finally merged in! 🎉 (@zumwald if you'd like to take a look as well, please be my guest. Open to any and all suggestions.)

scabana:
Hi, 

We would really like to see this PR merged! We have services that we would like to migrate to .net core + linux, and this lib is blocking this effort. 

I'm writing here because I've been following this PR for a while now and there seems to be about 2 months with no updates after code has been completed. Is there a will to merge this and get it released? Or should we migrate our codebase away from DurableTask?

Thanks a lot!

Preecington:
Hi @scabana ,

I'd love to get this in too! I'm ready and waiting to action any feedback requests. Since it's been since April I'll do a rebase later today just to make the reviewer's lives easier. I understand @adarsh1 was keen to get this in so I'm curious to as to what the hold up is.

src/DurableTask.ServiceBus/Common/Abstraction/ServiceBusAbstraction.cs
```diff
@@ -0,0 +1,959 @@
+﻿#pragma warning disable 1591
+namespace DurableTask.ServiceBus.Common.Abstraction
```
simonporter: >namespace [](start = 0, length = 9)

Missing copyright

test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj
```diff
@@ -2,15 +2,23 @@
...
-		<PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
-		<PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
-		<PackageReference Include="Newtonsoft.Json" version="7.0.1" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
+    <PackageReference Include="Newtonsoft.Json" version="7.0.1" />
+  </ItemGroup>
+
+	<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
+	  <PackageReference Include="EnterpriseLibrary.SemanticLogging.EventSourceAnalyzer.NetCore" Version="2.0.1406.4" />
```
simonporter: >NetCore [](start = 84, length = 7)

nit: this is the only package difference requiring the condition

src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
```diff
@@ -17,14 +17,15 @@
     <PackageReference Include="System.Spatial" version="5.6.4" />
     <PackageReference Include="WindowsAzure.ServiceBus" version="4.1.3" />
     <PackageReference Include="WindowsAzure.Storage" version="7.0.0" />
+    <Reference Include="System.Transactions" />
   </ItemGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\DurableTask.Core\DurableTask.Core.csproj" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="3.3.0" />
```
simonporter: >3.3 [](start = 68, length = 3)

we should probably bump to 3.4.0 before packaging the release

ltdavids:
I notice this has been merged yet I don't see that NuGet has been updated to support standard.  Am I missing something>

scabana:
@simonporter Hello, any idea on a possible release date? Also, I quickly looked, and there seems to be no beta feed of the durable task nuget packages. I'd be willing to live with a beta version for sure.

cgillum:
This is available now on nuget.org: [Microsoft.Azure.DurableTask.ServiceBus v2.1.3](https://www.nuget.org/packages/Microsoft.Azure.DurableTask.ServiceBus/2.1.3)

PR #209, opened 2018-08-23
Remove rewind functionality from core public API, other fixes

cgillum:
The main purpose of this PR is to remove Rewind* public surface area from DurableTask.Core. The rewind functionality has instead been isolated to DurableTask.Azure. This will allow us to have further discussions about how rewind should work in general across DTFx without locking us to a particular design.

I also noticed that some of the new unit tests were broken due to timezone-related issues (they only passed in Japan :)). FYI @tsuyoshiushio.

Lastly, I did a pass and cleaned up some minor code readability issues that I encountered along the way.


Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -294,12 +291,10 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
         /// </summary>
-        [DataTestMethod]
-        //[DataRow(true)]
-        [DataRow(false)]
-        public async Task RewindOrchestrationsFail(bool enableExtendedSessions)
+        [TestMethod]
+        public async Task RewindOrchestrationsFail()
         {
-            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
```
kashimiz: Do we also need to test this method and those below with enableExtendedSessions == false?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -349,7 +349,6 @@ public override async Task<IList<string>> RewindHistoryAsync(string instanceId,
 
                 if (entity.RowKey == SentinelRowKey)
                 {
-                    this.eTagValues[instanceId] = entity.ETag;
```
kashimiz: Why did we stop doing this?

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
cgillum: Most of these tests don't work with replays because the method Catherine uses to induce failures (static variables) makes the execution non-deterministic. For that reason, it's necessary to run these tests using extended sessions, which effectively removes all replay.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: It's not necessary to do anything with eTags as part of rewind since there is no potential for race conditions (these are manual operations). We still do eTag management in other, non-rewind code paths.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
kashimiz: Just an added note for anyone who might come across this note in the future: this PR only turns on extended sessions for the ease of testing; rewind will work for "live" functions that don't have extended sessions enabled.

PR #225, opened 2018-10-23
DurableTask.Core: Fix process crash caused by uncaught exception

### LINKED ISSUE #222: TaskActivityDispatcher>RenewUntil cancellation is not handled
In 2.0.9 cancellation token was added to 
`await Task.Delay(TimeSpan.FromSeconds(5), cancellationToken);`
https://github.com/Azure/durabletask/blob/master/src/DurableTask.Core/TaskActivityDispatcher.cs#L191

This causes unhanded OperationCancelledException to be thrown and there is no handling for it.

Rolling back to 2.0.8 works.

cgillum:
Resolves #222

### DIFF
--- a/src/DurableTask.Core/TaskActivityDispatcher.cs
+++ b/src/DurableTask.Core/TaskActivityDispatcher.cs
@@ -211,6 +211,10 @@ namespace DurableTask.Core
                     }
                 }
             }
+            catch (OperationCanceledException)
+            {
+                // cancellation token was triggered
+            }
             catch (ObjectDisposedException)
             {
                 // brokered message is already disposed probably through 


cgillum:
Got offline approval from Simon.

cgillum:
BTW @gfrancomsft it looks like there are a few more places in the code where these types of process crashes are possible. For example, [here](https://github.com/Azure/durabletask/blob/80ddcda54e8960779352640c88878df8656733ba/src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs#L1565). I suggest doing a code sweep of places where we use `Task.Delay` to see if we can find others.

PR #226, opened 2018-11-02
Add option to avoid retrieving orchestration instance inputs

glennamanns:
This is in response to issue 476 on the Durable Functions Extension: https://github.com/Azure/azure-functions-durable-extension/issues/476

My corresponding DF Extensions PR: https://github.com/Azure/azure-functions-durable-extension/pull/490

**Summary:**
User(s) with large orchestration inputs would like to increase perf by not retrieving the orchestration input when retrieving the status. 

**Some things to note:**
- For the Durable Functions Extension, the HTTP query parameter name is `showInput`. 
- In DTFx, the name is `fetchInput`, rather than `showInput`, because the name `showInput` implies that the input is being retrieved but not shown. I opted to not retrieve the input at all because there's no point in wasting bandwidth on it. 
- `AzureStorageOrchestrationService` implements a new method `public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions, bool fetchInput = true)` that is NOT part of the IOrchestrationServiceClient interface. This was done to avoid making a change to the DurableTask.Core framework. The Durable Functions Extension directly calls this method.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -427,21 +427,27 @@ namespace DurableTask.AzureStorage.Tracking
         }
 
         /// <inheritdoc />
-        public override async Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions)
+        public override async Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions, bool ignoreInput)
         {
-            return new[] { await this.GetStateAsync(instanceId, executionId: null) };
+            return new[] { await this.GetStateAsync(instanceId, executionId: null, ignoreInput: ignoreInput) };
         }
 
         /// <inheritdoc />
-        public override async Task<OrchestrationState> GetStateAsync(string instanceId, string executionId)
+        public override async Task<OrchestrationState> GetStateAsync(string instanceId, string executionId, bool ignoreInput)
         {
             if (instanceId == null)
             {
                 throw new ArgumentNullException(nameof(instanceId));
             }
 
+            List<string> columnsToRetrieve = typeof(OrchestrationInstanceStatus).GetProperties().Select(prop => prop.Name).ToList();
+            if (ignoreInput && columnsToRetrieve.Contains(InputProperty))
+            {
+                columnsToRetrieve.Remove(InputProperty);
+            }
+
             var stopwatch = new Stopwatch();
-            TableResult orchestration = await this.InstancesTable.ExecuteAsync(TableOperation.Retrieve<OrchestrationInstanceStatus>(instanceId, ""));
+            TableResult orchestration = await this.InstancesTable.ExecuteAsync(TableOperation.Retrieve<OrchestrationInstanceStatus>(instanceId, "", columnsToRetrieve));
             stopwatch.Stop();
             this.stats.StorageRequests.Increment();
             this.stats.TableEntitiesRead.Increment(1);
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -427,23 +427,28 @@ namespace DurableTask.AzureStorage.Tracking
         }
 
         /// <inheritdoc />
-        public override async Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions, bool ignoreInput)
+        public override async Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions, bool fetchInput)
         {
-            return new[] { await this.GetStateAsync(instanceId, executionId: null, ignoreInput: ignoreInput) };
+            return new[] { await this.GetStateAsync(instanceId, executionId: null, fetchInput: fetchInput) };
         }
 
         /// <inheritdoc />
-        public override async Task<OrchestrationState> GetStateAsync(string instanceId, string executionId, bool ignoreInput)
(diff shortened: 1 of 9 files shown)


src/DurableTask.Core/IOrchestrationServiceClient.cs
```diff
@@ -80,8 +80,9 @@ public interface IOrchestrationServiceClient
         /// </summary>
         /// <param name="instanceId">Instance id</param>
         /// <param name="allExecutions">True if method should fetch all executions of the instance, false if the method should only fetch the most recent execution</param>
+        /// <param name="ignoreInput">Whether or not the Input property for the instance will be retrieved</param>
         /// <returns>List of OrchestrationState objects that represents the list of orchestrations in the instance store</returns>
-        Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions);
+        Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions, bool ignoreInput);
```
cgillum: As a general rule, we try to avoid making any changes to **DurableTask.Core** unless we have to since our team doesn't own that code. I suggest reverting this change (and the impacted files in **DurableTask.ServiceBus** and **DurableTask.Emulator** and having the Durable Functions extension code call directly into **DurableTask.AzureStorage** APIs.

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
```diff
@@ -75,14 +75,16 @@ interface ITrackingStore
         /// </summary>
         /// <param name="instanceId">Instance Id</param>
         /// <param name="allExecutions">True if states for all executions are to be fetched otherwise only the state for the latest execution of the instance is fetched</param>
-        Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions);
+        /// <param name="ignoreInput">Whether or not the Input property for the instance will be retrieved</param>
+        Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions, bool ignoreInput);
```
cgillum: If we're going with verbs that default to `false`, I would suggest `omitInput`. It more strongly suggests that inputs won't be fetched. `ignoreInput` suggests that the input is fetched but we don't do anything with it (we ignore it).

cgillum: Though it might be easier to just call it `fetchInput` and default it to `true`. It's not unprecedented. :)

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -427,21 +427,27 @@ public override async Task<IList<string>> RewindHistoryAsync(string instanceId,
...
         /// <inheritdoc />
-        public override async Task<OrchestrationState> GetStateAsync(string instanceId, string executionId)
+        public override async Task<OrchestrationState> GetStateAsync(string instanceId, string executionId, bool ignoreInput)
         {
             if (instanceId == null)
             {
                 throw new ArgumentNullException(nameof(instanceId));
             }
 
+            List<string> columnsToRetrieve = typeof(OrchestrationInstanceStatus).GetProperties().Select(prop => prop.Name).ToList();
```
cgillum: Reflection is a bit expensive so should be avoided if it's not necessary. In this case it's appropriate to use, but let's check to see if `ignoreInput` (or whatever we call it) is true before doing this reflection.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -596,7 +596,7 @@ public async Task TerminateInstanceAsync(OrchestrationInstance orchestrationInst
         /// <exception cref="InvalidOperationException">Thrown if instance store not configured</exception>
         public async Task<OrchestrationState> GetOrchestrationStateAsync(string instanceId)
         {
-            IList<OrchestrationState> state = await GetOrchestrationStateAsync(instanceId, false);
+            IList<OrchestrationState> state = await GetOrchestrationStateAsync(instanceId, allExecutions: false);
```
glennamanns: Accidentally left this in. I'll revert it in my next commit

src/DurableTask.Core/IOrchestrationServiceClient.cs
glennamanns: I reverted the changes to DurableTask.Core, including IOrchestrationServiceClient, which AzureStorageOrchestrationService implements. I added a separate GetOrchestrationStateAsync method to it that takes in a fetchInput bool that the DF extension calls directly. Is this what you had in mind?

I accidentally named a parameter in one of the Core files that wasn't named before, so I'll tack that reversion on to my next commit.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -427,21 +427,32 @@ public override async Task<IList<string>> RewindHistoryAsync(string instanceId,
...
+                if (columnsToRetrieve.Contains(InputProperty))
+                {
+                    // Retrieve all columns except the input column
+                    columnsToRetrieve.Remove(InputProperty);
+                }            
+            }
+
             var stopwatch = new Stopwatch();
-            TableResult orchestration = await this.InstancesTable.ExecuteAsync(TableOperation.Retrieve<OrchestrationInstanceStatus>(instanceId, ""));
+            TableResult orchestration = await this.InstancesTable.ExecuteAsync(TableOperation.Retrieve<OrchestrationInstanceStatus>(instanceId, "", columnsToRetrieve));
```
cgillum: Is null an acceptable input for this method?

glennamanns: Yes. That's the default value for TableOperation.Retrieve :)

PR #227, opened 2018-11-07
Add return value(s) to purge history

glennamanns:
**Objective:** PurgeInstanceHistoryAsync now supports returning the number of instances purged, as well as the number of rows purged. Without this information, users have no way of telling whether or not their query was successful, let alone metrics about it, without querying again or manually checking the DurableFunctionsHubHistory table themselves.

- Adds new dependency on System.ValueTuple in order to return more than one value from an async task.
- Only implemented in the case where users pass a time range and list of statuses to filter by. I could add a rowsDeleted output for the PurgeInstanceHistory that only takes an instance ID.

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -24,10 +24,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
-  </ItemGroup>
-
-  <ItemGroup>
     <ProjectReference Include="..\DurableTask.Core\DurableTask.Core.csproj" />
   </ItemGroup>
 </Project>
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -24,6 +24,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+  </ItemGroup>
+
+  <ItemGroup>
     <ProjectReference Include="..\DurableTask.Core\DurableTask.Core.csproj" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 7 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -23,6 +23,10 @@
     <PackageReference Include="WindowsAzure.Storage" version="8.6.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
cgillum: Arg, sorry, I've already told people that I don't want us to take a dependency on this. >_<  Can we instead create a new class which holds the properties we want?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -701,22 +711,25 @@ public override async Task PurgeInstanceHistoryAsync(string instanceId)
 
             if (orchestrationInstanceStatus != null)
             {
-                int storageRequests = await this.DeleteAllDataForOrchestrationInstance(orchestrationInstanceStatus);
+                var stats = await this.DeleteAllDataForOrchestrationInstance(orchestrationInstanceStatus);
```
cgillum: Let's not use the local variable name `stats` because we already have a class variable with that same name in various places.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -701,22 +711,25 @@ public override async Task PurgeInstanceHistoryAsync(string instanceId)
...
                     DateTime.MinValue.ToString(),
                     DateTime.MinValue.ToString(),
                     null,
-                    storageRequests,
+                    stats.storageRequests,
                     stopwatch.ElapsedMilliseconds,
                     Utils.ExtensionVersion);
             }
+
+            return new PurgeHistoryStats(0, 0, 0);
```
cgillum: Should we be returning the result of `DeleteAllDataForOrchestrationInstance` here? Seems like that value is being discarded.

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -64,4 +64,39 @@ static async Task InvokeThrottledAction<T>(T item, Func<T, Task> action, Semapho
...
+        {
+            this.storageRequests = storageRequests;
+            this.instancesDeleted = instancesDeleted;
+            this.rowsDeleted = rowsDeleted;
+        }
+
+        /// <summary>
+        /// Number of requests sent to Storage during this execution of purge history
+        /// </summary>
+        public int storageRequests { get; }
```
cgillum: In .NET, property names are supposed to be capitalized.

Issue #236, opened 2018-12-08
Unable to set OrchestrationInstance and IsReplaying on OrchestrationContext from implementation


lukak-msft:
I was trying to unit test some of the code I wrote using orchestrations, and I thought the easiest way would be to implement [OrchestrationContext ](https://github.com/Azure/durabletask/blob/f9cc450539b5e37c97c19ae393d5bb1564fda7a8/src/DurableTask.Core/OrchestrationContext.cs) class since it's abstract, and pass it to the orchestration and validate the behavior.
Unfortunately `IsReplaying` and `OrchestrationInstance` properties are non-virtual, and do not expose a protected setter.

cgillum:
This will go out with the v2.1.0 release

PR #237, opened 2018-12-08
Make OrchestrationInstance and IsReplaying internal protected

lukak-msft:
I could set them to internal protected to expose them to implementation of the abstract class, or I could make them both virtual.
It felt like `internal protected` is less intrusive, but I admit I do not fully understand why these two are not virtual.

#236


cgillum:
Seems fine to me. @simonporter what do you think?

simonporter:
Yeah this looks fine.

PR #242, opened 2019-01-08
DurableTask.AzureStorage: Fix out-of-order regression in 1.4.0

### LINKED ISSUE #241: DurableTask.AzureStorage: Orchestrator stuck with repeated ReceivedOutOfOrderMessage warnings
This affects the **v1.4.0** release of DurableTask.AzureStorage.  The linked issue is where this problem was originally discovered: https://github.com/Azure/azure-functions-durable-extension/issues/555.

The problem is this code here, which assumes that EventIds are always added in order, but in the linked case, they were not:

```
// This message is a response to a task. Make sure the event ID matches the range of the previously known
// scheduled events. We don't need to ensure the event types actually match because the core runtime
// will do that anyways and fail the orchestration if there is a non-matching event type.
HistoryEvent mostRecentTaskEvent = this.RuntimeState.Events.LastOrDefault(e => e.EventId >= 0);
if (mostRecentTaskEvent != null && taskScheduledId <= mostRecentTaskEvent.EventId)
{
    return false;
}
```
If the order of EventIds cannot be guaranteed, then this code will need to change to search the entire history rather than looking for the first non-negative EventId.

cgillum:
Resolves #241 

The fix is to do a more thorough search through the orchestration history to see if a particular response message matches a known request message. The previous "out-of-order" detection made some bad assumptions about the order of history events (premature optimization...), causing false-positives.

There is potentially a CPU performance impact for orchestrations that do huge fan-outs since we may end up searching through a larger portion of the orchestration history. This is acceptable, though, since it's not expected to be common and reliability is a priority.

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -1546,6 +1546,27 @@ namespace DurableTask.AzureStorage.Tests
             }
         }
 
+        /// <summary>
+        /// Tests an orchestration that does two consecutive fan-out, fan-ins.
+        /// This is a regression test for https://github.com/Azure/durabletask/issues/241.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public async Task DoubleFanOut(bool enableExtendedSessions)
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
+            {
+                await host.StartAsync();
+
+                var client = await host.StartOrchestrationAsync(typeof(Orchestrations.DoubleFanOut), null);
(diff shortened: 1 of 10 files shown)


kashimiz:
+1 to tracking possible performance issues.

cgillum:
> We should think about starting a list of performance things like this, where we've noted it "may be an issue" so we can start tracking them.

I agree in principal, but for this particular case I don't think it's severe enough to be noteworthy. I think it would be better for us to do a dedicated performance analysis, and then make fixes based on what profiling shows us. That will help us know we're fixing high-value perf issues and not spending time on issues that don't actually make much real-world difference.

cgillum:
Note that I've also squeezed in some ETW tracing improvements:
* Added task event IDs to our message-related traces. This is to help me diagnose https://github.com/Azure/azure-functions-durable-extension/issues/559.
* Fixed tracing bugs I discovered while updating the existing trace code.

Issue #243, opened 2019-01-09
AzureStorageOrchestrationService TokenProvider/MSI support


daksha121:
We plan to use Managed Identity in our project to access Azure Storage (https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/tutorial-vm-windows-access-storage#get-an-access-token-and-use-it-to-call-azure-storage) . Are there any plans to support a TokenProvider way to use AzureStorageOrchestrationService instead of using a storage connection string?

cgillum:
No plans currently, but we could certainly look into this as a very good feature request, or accept it as a PR.

cgillum:
This will go out with the v1.5.0 release.

PR #248, opened 2019-01-14
Accept StorageCredentials to create AzureStorageOrchestrationService

daksha121:
#243 
- Adding a new constructor for AzureStorageOrchestrationService to accept StorageCredentials. This will help support Managed Service Identity for DurableTask.AzureStorage ([https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/tutorial-vm-windows-access-storage#get-an-access-token-and-use-it-to-call-azure-storage]) 
- Currently Azure Storage does not accept TokenProvider authentication for TableClient. But I tested this by using account name and key to create Storage Credentials.
- No unit test updates yet as I was unable to use StorageCredentials (account name and key) to access Storage Emulator as I couldn't find a place to provide the blob, queue and table endpoints (to point them to the localhost emulator endpoint). Any help/suggestions appreciated for this.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -61,6 +61,7 @@ namespace DurableTask.AzureStorage
         readonly WorkItemQueue workItemQueue;
         readonly ConcurrentDictionary<string, ActivitySession> activeActivitySessions;
         readonly MessageManager messageManager;
+        readonly StorageAccountDetails storageAccountDetails;
 
         readonly ITrackingStore trackingStore;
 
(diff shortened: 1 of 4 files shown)


daksha121:
Thanks for reviewing! Updated the PR with your feedback.

I think we should still keep the connection string option as that helps override blob, table and queue endpoints. I am not aware of that being supported by StorageCredentials. Unless we accept BlobEndpoint, QueueEndpoint, TableEndpoint, FileEndpoint into StorageAccountDetails instead.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -99,7 +101,11 @@ public AzureStorageOrchestrationService(AzureStorageOrchestrationServiceSettings
 
             this.settings = settings;
             this.tableEntityConverter = new TableEntityConverter();
-            CloudStorageAccount account = CloudStorageAccount.Parse(settings.StorageConnectionString);
+            this.storageAccountDetails = settings.StorageAccountDetails;
```
cgillum: minor nit: It looks like `this.storageAccountDetails` is only used here in the constructor. Do we really need to create the instance variable to save the reference, or can we just use `settings.StorageAccountDetails` directly?

daksha121: Fixed

PR #252, opened 2019-01-26
Enabling to share instance/history table

TsuyoshiUshio:
Hi @cgillum ,

As we discussed, I create a PR for DurableTask.  This PR will help someone who want to simplify Zero Downtime deployment for Durable Functions. 

 This PR enable us to configure separate ConnectionString between queue and Storage Table, also, increate AppId column to the Instance table with Query.

Please have a review of this PR. If it is accepted, I'll try to test via Durable.


TsuyoshiUshio:
Hi @cgillum  Are you waiting for the PR of the Durable repo side? If so, I'll do it.

cgillum:
Thank you for making this PR. I think the impact of this feature could be very big. not only does this allow side-by-side deployments, but it could also be the first step to enable hybrid deployments, something I've been thinking about recently. For example, maybe an orchestrator running in Azure could schedule an activity in another "app" which is hosted on premise.

No need to make a Durable Functions PR yet. Please give me a little more time to think about this one.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1335,6 +1335,23 @@ public async Task<DurableStatusQueryResult> GetOrchestrationStateAsync(DateTime
...
+        /// </summary>
+        /// <param name="createdTimeFrom">CreatedTime of orchestrations. Fetch status grater than this value.</param>
+        /// <param name="createdTimeTo">CreatedTime of orchestrations. Fetch status less than this value.</param>
+        /// <param name="runtimeStatus">RuntimeStatus of orchestrations. You can specify several status.</param>
+        /// <param name="appIds">AppId of orchestration. You can specify several Ids.</param>
+        /// <param name="top">Top is number of records per one request.</param>
+        /// <param name="continuationToken">ContinuationToken of the pager.</param>
+        /// <param name="cancellationToken">Cancellation Token</param>
+        /// <returns>List of <see cref="OrchestrationState"/></returns>
+        public async Task<DurableStatusQueryResult> GetOrchestrationStateAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus, string appIds, int top, string continuationToken, CancellationToken cancellationToken = default(CancellationToken))
```
cgillum: Two things:
1. The `appIds` should not be a string since it can be multiple parameters. Instead, it should be an `IEnumerable<string>` or something similar.
2. The list of parameters is getting too long. I think we should consider using a class to represent the filter. Maybe we can use your `OrchestrationInstanceStatusQueryCondition` class?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -75,7 +75,8 @@ class AzureTableTrackingStore : TrackingStoreBase
             this.tableEntityConverter = new TableEntityConverter();
             this.taskHubName = settings.TaskHubName;
 
-            CloudStorageAccount account = CloudStorageAccount.Parse(settings.StorageConnectionString);
+            string storageConnectionString = settings.StorageTableConnectionString ?? settings.StorageConnectionString;
+            CloudStorageAccount account = CloudStorageAccount.Parse(storageConnectionString);
```
cgillum: I think we should make the same change for the `InstanceStoreBackedTrackingStore` class.

src/DurableTask.AzureStorage/Tracking/TrackingStoreBase.cs
```diff
@@ -69,6 +69,12 @@ public virtual Task<DurableStatusQueryResult> GetStateAsync(DateTime createdTime
             throw new NotSupportedException();
         }
 
+        /// <inheritdoc />
+        public virtual Task<DurableStatusQueryResult> GetStateAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus, string appIds, int top, string continuationToken, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            throw new NotSupportedException();
```
cgillum: Let's implement this for the other tracking store as well.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -69,6 +69,17 @@ public class AzureStorageOrchestrationServiceSettings
...
+        /// <summary>
+        /// Gets or sets the Instance/History Table Storage connection string.
+        /// In case of null, StorageConnectionString is applied.
+        /// </summary>
+        public string StorageTableConnectionString { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Application Id. 
+        /// </summary>
+        public string AppId { get; set; }
```
cgillum: I'm not sure about the `AppId` name. I'm thinking about naming this `Zone` or something similar. The idea is that within a single task hub, you can define multiple zones which contain a distinct set of orchestrations and activities. Side-by-side deployments for versioning is one possible scenario. Maybe they can run in different hosting environments as well. I'm going to chat with Jeff about this idea.

TsuyoshiUshio: Hi @cgillum  , I have a question about the concept of "Zone".  The fist idea was, just store the AppId (the name of FunctionApp) into the Instance table. Zone is a part of a single task hub. Then how they share the information with other Function Apps? And what is the example of the zone? then How we can retrieve the zone info since it can be multiple zones in a single task hub?

TsuyoshiUshio:
Hi @cgillum , As we discussed, I remove all the new query and new attribute for instance table, and add these feature

* Separate Connection String for Instance/History table
* Create Zone and make it prefix of the History/Instance tables. If no Zone configured, it uses TaskHubName.

TsuyoshiUshio:
@cgillum  I'll implement with AppId search again. If you think the AppId is proper, I'd happy to change it. (logic might be the same even if it change into TaskHub for example. )

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: I make it Condition object as you said.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: InstanceStoreBackedTrakingStore has only this constructor. 

```
        public AzureTableInstanceStore(string hubName, string tableConnectionString)
        {
            this.tableClient = new AzureTableClient(hubName, tableConnectionString);

            // Workaround an issue with Storage that throws exceptions for any date < 1600 so DateTime.Min cannot be used
            DateTimeUtils.SetMinDateTimeForStorageEmulator();
        }
```

Also, the references are just Testing. I guess, Durable Functions use the Constructor, then it should be done by there.  Am I Correct? @cgillum

src/DurableTask.AzureStorage/Tracking/TrackingStoreBase.cs
TsuyoshiUshio: You mean move the AzureTableTrackingStore.GetStateAsync(OrchestrationInstanceStatusQueryCondition condition, int top, string continuationToken, CancellationToken cancellationToken = default(CancellationToken))
    to this place? It also include other private method might come here. Do I understand correctly?

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
TsuyoshiUshio: I implement both Zone and AppId

TsuyoshiUshio:
@cgillum I have only one questions. Also, if you don't like AppId, please let me know.

TsuyoshiUshio:
Hi @cgillum, 

Could you review one more time? I rewrite whole.

cgillum:
Yes, I will take a look now.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -69,6 +69,24 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public string StorageConnectionString { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Instance/History Table Storage connection string.
+        /// In case of null, StorageConnectionString is applied.
+        /// </summary>
+        public string StorageTableConnectionString { get; set; }
```
cgillum: Let's rename this to `TrackingStoreConnectionString`.

cgillum:
Actually, I will just put my main feedback here.

I've been thinking about this PR over the last several weeks because I think it is important for a lot of reasons. I want to simplify the design and get it merged soon.  Here are my current thoughts:

* I don't want to introduce a new "AppId" concept. Instead, I think we should use `TaskHub` as both the filter and the column name.
* I also don't want to introduce a new "Zone" concept. Instead, let's have a property which is just a prefix for the table names. We can call it `RemoteTrackingStoreNamePrefix`. This value is **only** used when `StorageTableConnectionString` is specified. The default value for `RemoteTrackingStoreNamePrefix` can be `DurableTask`. This means that by default we will create a table named `DurableTaskHistory` and `DurableTaskInstances`.

So basically, we don't introduce any new concepts - we just reuse existing concepts. The reason I want to do this is because I worry about creating confusion between concepts that exist for Azure Storage but not for other providers. I also don't want to speculate about other theoretical future use cases, since that can be dangerous and lead to technical debt.

Let me know what you think. I will also add some comments to the review.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
cgillum: Sorry, I changed my mind again :(.  I think there should be just one new property called `RemoteTrackingStoreNamePrefix`.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: I see. If there are only test references, then we don't have to do anything here.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -22,24 +22,45 @@ namespace DurableTask.AzureStorage.Tracking
...
 
+        /// <summary>
+        /// CreatedTimeTo
+        /// </summary>
         public DateTime CreatedTimeTo { get; set; }
 
+        /// <summary>
+        /// Collection of AppId
+        /// </summary>
+        public IEnumerable<string> AppIds { get; set; }
```
cgillum: Per my other feedback, let's name this `TaskHubs`. I am hesitating about introducing new concepts.

src/DurableTask.AzureStorage/Tracking/TrackingStoreBase.cs
cgillum: Actually, let's not worry about it. This can be done later if/when somebody needs it.

TsuyoshiUshio:
Hi @cgillum , 

This is the AppId discussion.

Imagine that we have production/staging slot for the deployment. 
If we use TaskHub concept, we need to change the taskhub name for every deployment. 
Also, we need to store somewhere which is which.  People might think they want to query for 
the target FunctionApp. 

I don't care if it is AppId or others. However, we might need something to query for the different FunctionApp. TaskHub is OK, but the operation might be complex. 

or Durable Functions Extension side pass the TaskHubName with FunctionApp name. Hmm... It seems not cool. If you have any comment, please let me know.

cgillum:
I understand the concern. For the case that you mentioned, can customers use a combination of `%WEBSITE_SITE_NAME%` and `%WEBSITE_SLOT_NAME%` as the name of the task hub? That way it will change automatically.

In Durable 2.0, I want to make a breaking change where we derive the task hub name from the function app name by default. Let me know what you think.

TsuyoshiUshio:
@cgillum It is good idea! It sounds cool! I'll start fix this PR.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
cgillum: Actually, I just remembered one thing. Someone recently contributed a new way of specifying the storage account: `StorageAccountDetails`. This allows people to use MSI instead of putting connection credentials directly into their configuration. Could you change your implementation to use `StorageAccountDetails` instead of `string` for the tracing store connection? Maybe then we would call it `TrackingStoreConnectionDetails`.

TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Sure

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
TsuyoshiUshio: However, I change it to TaskHubNames for consistency of the other parts.

TsuyoshiUshio:
Hi @cgillum 

I've done it. However, something I need to tell you. 

1. I change the name of TaskHubs to TaskHubNames for the consistency
2. AzureTableTrackingStore seems change the constructor, it passes CloudStorageAccount object. I change the code because of the conflict. Test passes. Please review it, I'm ready to do more fix, If you have any comment, please let me know. I'll fix it today.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -67,6 +88,16 @@ string GetConditions()
                 }
             }
 
+            if (this.TaskHubNames != null)
+            {
+                string appIdCondition = this.TaskHubNames.Select(x => TableQuery.GenerateFilterCondition("TaskHubName", QueryComparisons.Equal, x.ToString()))
```
cgillum: I think this variable needs to be renamed to `taskHubCondition`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -72,6 +72,19 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public string StorageConnectionString { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Instance/History Table Storage connection string.
+        /// In case of null, StorageConnectionString is applied.
+        /// </summary>
+        public string TrackingStoreConnectionString { get; set; }
```
cgillum: I asked this in another comment already, but can we change this from `string` to `StorageAccountDetails`? Someone else introduced the `StorageAccountDetails` class, and it allows the user to configure with MSI.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -75,16 +76,19 @@ class AzureTableTrackingStore : TrackingStoreBase
             this.stats = stats;
             this.tableEntityConverter = new TableEntityConverter();
             this.taskHubName = settings.TaskHubName;
+            this.remoteTrackingStoreNamePrefix = settings.RemoteTrackingStoreNamePrefix;
 
-            this.storageAccountName = account.Credentials.AccountName;
+            CloudStorageAccount trackingStoreAccount = string.IsNullOrEmpty(settings.TrackingStoreConnectionString) ? account : CloudStorageAccount.Parse(settings.TrackingStoreConnectionString);
```
cgillum: Instead of changing this logic, can we change the caller of this constructor to pass in the correct `account` object? It doesn't make sense to pass an `account` parameter and then ignore it and create a new one.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
cgillum: See this PR for more details: https://github.com/Azure/durabletask/pull/248

TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done

TsuyoshiUshio:
Hi @cgillum, 

I've done all fix. 

According to your comment, I change TrakingStoreConnectionString from string to StorageAccountDetails. 

However, things is not simple for this time. As you commented on the #248 , you want to obsolete the StorageConnectionString. We need to add the StorageConnectionString to the StorageAccountDetails with Automatically decide if it is StorageConnectionString or Creadentials.

Also, according to the GRASP pattern, I simplify the AzureStorageOrchestrationServiceSettings. 

If you don't like this design, I'd happy to refactor that with the highest priority or you can fix by your self. if you have no time.

Test/DurableTask.AzureStorage.Tests/AzureStorageOrchestrationServiceSettingsTest.cs
```diff
@@ -0,0 +1,50 @@
...
+            Assert.AreEqual("fooInstances", settings.InstanceTableName);
+        }
+
+        [TestMethod]
+        public void TrackingStoreStorageAccountDetailsHasSet()
+        {
+            var settings = new AzureStorageOrchestrationServiceSettings()
+            {
+                TaskHubName = "foo",
+                RemoteTrackingStoreNamePrefix = "bar",
```
cgillum: I sincerely apologize, but I changed my mind a bit. I realized that I don't like the "Remote" part of this property name. Instead, I think it should just be `TrackingStoreNamePrefix` because "Remote" doesn't really mean anything in this context.

I'm really glad you wrote unit tests to make sure this behaves correctly. Thank you!

src/DurableTask.AzureStorage/StorageAccountDetails.cs
```diff
@@ -34,5 +34,10 @@ public sealed class StorageAccountDetails
         /// The storage account endpoint suffix
         /// </summary>
         public string EndpointSuffix { get; set; }
+
+        /// <summary>
+        /// The storage account connection string
+        /// </summary>
+        public string StorageAccountConnectionString { get; set; }
```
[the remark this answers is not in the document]
TsuyoshiUshio: Hi @cgillum , 

If so, we need TrakingStoreStorageAccountConnectionString.  I read your issue and I thought you want to put something in the StorageAccountDetails. That is why I add it here. Also I read you want to obsolete the StorageAccountConnectionString in the near future.  How can we set the ConnectionString for StorageAccountDetails? Don't need it?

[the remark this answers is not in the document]
TsuyoshiUshio: Self resolved. StorageCredentials class has a method of tryParseCredentials for ConnectionString. Sorry for that.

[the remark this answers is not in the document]
TsuyoshiUshio: Removed

Test/DurableTask.AzureStorage.Tests/AzureStorageOrchestrationServiceSettingsTest.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/StorageAccountDetailsExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+    using Microsoft.WindowsAzure.Storage;
+
+    /// <summary>
+    /// Extension methods for <see cref="StorageAccountDetails"/>
+    /// </summary>
+    public static class StorageAccountDetailsExtensions
```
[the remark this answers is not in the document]
TsuyoshiUshio: Remove this. Make it instance method of StorageAccountDetails.

src/DurableTask.AzureStorage/StorageAccountDetails.cs
```diff
@@ -36,8 +37,8 @@ public sealed class StorageAccountDetails
         public string EndpointSuffix { get; set; }
 
         /// <summary>
-        /// The storage account connection string
+        ///  CloudStorageAccount 
         /// </summary>
-        public string StorageAccountConnectionString { get; set; }
+        public CloudStorageAccount CloudStorageAccount => new CloudStorageAccount(this.StorageCredentials, this.AccountName, this.EndpointSuffix, true);
```
cgillum: Last thing: Generally speaking, .NET properties should not allocate like this. Customers will not expect this kind of behavior and it can lead to unexpected problems. Instead, make it a method, like `ToCloudStorageAccount()`.

TsuyoshiUshio:
Hi @cgillum 
I finish the change. However, I encounter the issue of testing. MultiWorkerLeaseMovement fails. However, there is no reason this one fails. I makes sure it works related Connection String. Could you run the test on your environment? I'll keep on investigate it.

src/DurableTask.AzureStorage/StorageAccountDetails.cs
TsuyoshiUshio: Done

PR #254, opened 2019-02-01
Support creation of events from within an orchestration

sebastianburckhardt:
With this change, orchestrations (not just clients) can create events. This allows message-passing patterns to be implemented between orchestrations.

The changes are pretty straightforward:
- add `CreateEvent` to the OrchestrationContext public interface, following same syntax as on client
- add an orchestrator action for `CreateEvent`
- add a history entry `EventCreatedEvent ` 
- include corresponding event messages in the `orchestratorMessages` list when calling `CompleteTaskOrchestrationWorkItemAsync`

I also added a little test to the DurableTask.AzureStorage.Tests.

For the `DurableTask.AzureStorage` backend, I believe the delivery guarantee is at-least-once only because orchestrator messages are not delivered transactionally (see comment in `AzureStorageOrchestrationService.CommitOutboundQueueMessages`).

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -165,27 +165,12 @@ namespace DurableTask.AzureStorage.Tests
             }
         }
 
-        [TestMethod]
-        public async Task EventConversation()
+        [DataTestMethod]
+        [DataRow(false)]
+        [DataRow(true)]
+        public async Task EventConversation(bool enableExtendedSessions)
         {
-            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: false))
-            {
-                await host.StartAsync();
-
-                var client = await host.StartOrchestrationAsync(typeof(Orchestrations.Greeter), "");
-                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
-
-                Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);
-                Assert.AreEqual("OK", JToken.Parse(status?.Output));
-
-                await host.StopAsync();
-            }
-        }
-
-        [TestMethod]
-        public async Task EventConversationNoReplay()
-        {
-            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
             {
                 await host.StartAsync();
 
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -2210,14 +2210,14 @@ namespace DurableTask.AzureStorage.Tests
                     var responderOrchestration = context.CreateSubOrchestrationInstance<string>(typeof(Responder), responderId, "Herkimer");
 
                     // send the id of this orchestration to the responder
-                    context.CreateEvent(responderId, channelName, context.OrchestrationInstance.InstanceId);
+                    context.SendEvent(responderId, channelName, context.OrchestrationInstance.InstanceId);
 
                     // wait for a response event 
                     var message = await tcs.Task;
                     Assert.AreEqual("hi from Herkimer", message);
 
                     // tell the responder to stop listening, then wait for it to complete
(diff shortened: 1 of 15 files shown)


cgillum:
FYI @adarsh1

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -324,6 +324,16 @@ public virtual Task<T> CreateSubOrchestrationInstance<T>(Type orchestrationType,
...
 
+
+        /// <summary>
+        ///     Raises an event for the specified orchestration instance, which eventually causes the OnEvent() method in the
+        ///     orchestration to fire.
+        /// </summary>
+        /// <param name="instanceId">Id of the instance in which to raise the event</param>
+        /// <param name="eventName">Name of the event</param>
+        /// <param name="eventData">Data for the event</param>
+        public abstract void CreateEvent(string instanceId, string eventName, object eventData);
```
cgillum: I wonder if `RaiseEvent` might be a better API name since it's consistent with `TaskHubClient.RaiseEventAsync`. I recognize that there is a conflict with how the history event gets named, however, since there is already an `EventRaised` history event. Maybe we could go with `SendEvent`? The main concern with `CreateEvent` is that the verb `Create` to me doesn't convey the idea that it's actually being sent somewhere.

src/DurableTask.Core/History/EventType.cs
```diff
@@ -89,7 +89,12 @@ public enum EventType
         OrchestratorCompleted,
 
         /// <summary>
-        /// External Event raised to orchestration event
+        /// Event was created by an orchestration
+        /// </summary>
+        EventCreated,
```
cgillum: Related to my other comment, I wonder if `EventSent` might be a better name than `EventCreated`.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -167,6 +167,27 @@ internal void ClearPendingActions()
             return this.dataConverter.Deserialize<T>(serializedResult);
         }
 
+        public override void CreateEvent(string instanceId, string eventName, object eventData)
+        {
+            if (string.IsNullOrWhiteSpace(instanceId))
+            {
+                throw new ArgumentException("instanceId");
```
cgillum: `nameof(instanceId)` is preferred over hardcoding `"instanceId"`.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -706,6 +711,32 @@ static bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
             return taskMessage;
         }
 
+        static TaskMessage ProcessCreateEventDecision(
+          CreateEventOrchestratorAction createEventAction,
+          OrchestrationRuntimeState runtimeState)
```
cgillum: nit: indentation should be 4 spaces.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -164,6 +165,41 @@ public async Task GetAllOrchestrationStatuses()
             }
         }
 
+        [TestMethod]
```
adarsh1: We should use DataTestMethod here with DataRows for enableExtendedSessions true and false instead of duplicating the test.

src/DurableTask.Core/OrchestrationContext.cs
adarsh1: +1

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -164,6 +165,41 @@ public async Task GetAllOrchestrationStatuses()
...
+        public async Task EventConversation()
+        {
```
adarsh1: Could we add a similar test for the ServiceBusProvider as well?

src/DurableTask.Core/OrchestrationContext.cs
sebastianburckhardt: Right.  I used `CreateEvent` because it aligns nicely with the other methods in the same API (`CreateOrchestration`, `CreateTimer`), but `SendEvent` is certainly more descriptive. `RaiseEvent` would be the most standard, but I agree that it is perhaps better to distinguish it clearly from the delivery of the event (which is already called `RaiseEvent`).

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
sebastianburckhardt: Done.

PR #258, opened 2019-03-02
ContinueAsNew fixes and test updates

cgillum:
* Fixed issues with fast ContinueAsNew (status and orchestration lifetime)
* Updated test framework dependencies
* Updated package versions
* Added some initial test categories for use with Azure Pipelines


cgillum:
@simonporter You'll notice we have the beginnings of a CI now. :)

image

PR #263, opened 2019-03-09
Allow connection string for configuring StorageAccountDetails

cgillum:
This PR makes it easier to configure the tracking store with a connection string.

@TsuyoshiUshio could you take a look? This is actually similar to what you tried originally. I thought maybe we shouldn't do this, but when I tried to use it in azure-functions-durable-extension, it was tricky and I realized trying to turn a connection string into a `StorageAccountDetails` object could be error prone in some cases. For that reason, I decided to add back the ability to configure with connection strings directly.

### DIFF
--- a/src/DurableTask.AzureStorage/StorageAccountDetails.cs
+++ b/src/DurableTask.AzureStorage/StorageAccountDetails.cs
@@ -37,11 +37,30 @@ namespace DurableTask.AzureStorage
         public string EndpointSuffix { get; set; }
 
         /// <summary>
+        /// The storage account connection string.
+        /// </summary>
+        /// <remarks>
+        /// If specified, this value overrides any other settings.
+        /// </remarks>
+        public string ConnectionString { get; set; }
+
+        /// <summary>
         ///  Convert this to its equivalent CloudStorageAccount.
         /// </summary>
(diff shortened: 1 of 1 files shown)


TsuyoshiUshio:
I understand the situation and it looks good. Also, the change is enclosed in this class, even if we come back to the ideal design, it will not big deal. 

What happens when you use the new API? StorageCredentails doesn't work well? (I just want to know)

cgillum:
StorageCredentials works fine. The problem is that Durable Functions only takes connection strings, and it's difficult to turn a connection string into StorageCredentials.

TsuyoshiUshio:
Hi @cgillum , 
I've done an experiment. How about this? 

```
            var ac = CloudStorageAccount.Parse("UseDevelopmentStorage=true");
            StorageCredentials sc = ac.Credentials;
            Console.WriteLine(sc.AccountName);
```

Result

```
devstoreaccount1
```

cgillum:
Sorry, I mispoke, I meant it's hard to accurately transform a connection string to a `StorageAccountDetails` object. Assigning the `StorageCredentials` and the `AccountName` is easy, but `EndpointSuffix` is not so easy because a connection string could specify different endpoints for different services.

PR #264, opened 2019-03-12
Release payload for DT.Core 2.1.0 and DT.AzureStorage 1.5.0

cgillum:
This PR is a collection of changes that were done in the `azure-storage` branch. All of the changes have been individually reviewed, but need to be moved into the `master` branch so that we can do an official release to nuget.org.

@adarsh1 @simonporter please do a quick scan of these changes to help make sure there's nothing going in that you didn't expect.

Our CI should also run automatically to verify that all tests are passing (minus the ones we explicitly filtered out).


adarsh1:
Looks good to me.

PR #267, opened 2019-03-15
Fix failing tests on the DurableTask.Emulator, fixing https://github.com/Azure/durabletask/issues/255

### LINKED ISSUE #255: Failing EmulatorFunctionalTests
Whilst working on #203 I've noticed that the PR #251 has introduced failures into the `EmulatorFunctionalTests`, namely `MockGenerationTest` and `MockRecreateOrchestrationTest`.

This manifests as a timeout, but the underlying issue is the orchestration is continually retrying due to a corrupt state where there are Multiple ExecutionCompletedEvents in the history.

I managed to trace the issue down to something to do with the `OrchestrationRuntimeState` being passed in to `CompleteTaskOrchestrationWorkItemAsync` inside `DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync` ([Line #402](https://github.com/Azure/durabletask/blob/master/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L402)  on `master` as of this issue). Perhaps the local orchestration service wasn't properly updated to deal with the new raised event fix?

I won't have time to look at this for another few days, so I thought I'd just post my findings here.

adarsh1:
This Fixes Failing 
MockRecreateOrchestration Test:
This was failing as the emulator was not clearing its session state for a non-running orchestration similar tot he way it was being done in the service bus service.

MockGenerationTest:
This was failing as the Continue as the new state of a previous generation was not being saved to the instance store along with the new started ContinueAsNew state.

This fixes #255

### DIFF
--- a/src/DurableTask.Emulator/LocalOrchestrationService.cs
+++ b/src/DurableTask.Emulator/LocalOrchestrationService.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.Emulator
 {
     using DurableTask.Core;
+    using DurableTask.Core.Common;
     using DurableTask.Core.Exceptions;
     using DurableTask.Core.History;
     using Newtonsoft.Json;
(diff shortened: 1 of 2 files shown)


adarsh1:
@simonporter @cgillum

adarsh1:
There seems to be an issue with the CI

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -439,49 +452,62 @@ public Task PurgeOrchestrationHistoryAsync(DateTime thresholdDateTimeUtc, Orches
...
+                }
+
                 if (state != null)
                 {
-                    if (!this.instanceStore.TryGetValue(workItem.InstanceId, out Dictionary<string, OrchestrationState> mapState))
-                    {
-                        mapState = new Dictionary<string, OrchestrationState>();
-                        this.instanceStore[workItem.InstanceId] = mapState;
-                    }
+                    CommitState(newOrchestrationRuntimeState, state).GetAwaiter().GetResult();
```
adarsh1: Have to Use GetAwaiter().GetResult(0 here since we have a lock here, might make sense to move over to a more async friendly construct in the future.

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -439,49 +452,62 @@ public Task PurgeOrchestrationHistoryAsync(DateTime thresholdDateTimeUtc, Orches
...
+                }
+            }
 
-                    mapState[workItem.OrchestrationRuntimeState.OrchestrationInstance.ExecutionId] = state;
+            return Task.FromResult(0);
+        }
 
-                    // signal any waiters waiting on instanceid_executionid or just the latest instanceid_
+        private Task CommitState(OrchestrationRuntimeState runtimeState, OrchestrationState state)
```
simonporter: nit: remove private modifier

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -439,49 +452,62 @@ public Task PurgeOrchestrationHistoryAsync(DateTime thresholdDateTimeUtc, Orches
...
+            // the high level orchestration is still ongoing
+            if (state.OrchestrationStatus != OrchestrationStatus.ContinuedAsNew
+                && this.orchestrationWaiters.TryGetValue(key1, out TaskCompletionSource<OrchestrationState> tcs1))
+            {
+                tasks.Add(Task.Run(() => tcs1.TrySetResult(state)));
+            }
+
+            if (tasks.Count > 0)
+            {
+                Task.WaitAll(tasks.ToArray());
```
simonporter: does this need to be blocking, also in a lock? should we use Task.WhenAll with the getawater to be consistent and when we flip to async this is correct

adarsh1: I don't believe you can do a GetAwaiter().GetResult() with a whenall since it doesn't return a result.
Moving it out of the lock should not have any side effects as far as I can tell as it is simply signaling the clients waiting for the orchestration, but would rather not risk it for this change. I may be more suited for when we do move things to async.

cgillum:
It looks like there is an issue in Azure DevOps where pull requests from external contributors are not able to actually execute the tests properly. So far I haven't been able to find a workaround, so I'm just going to go ahead and merge this, then run another build to validate that master is green.

cgillum:
The CI has verified that all tests are passing:
image

PR #269, opened 2019-03-21
Implement auto-start support in AzureStorageOrchestrationService

sebastianburckhardt:
This implements the auto-start feature needed for the durable actors project, for the AzureStorageOrchestrationService.

To avoid headaches with versioning, I made no changes to DurableTask.Core. Instead, this is just a semantic (functional) change to how the AzureStorageOrchestrationService back end deals with events delivered to  non-existing instances.

The new behavior is: when trying to deliver an event to a not-already-existing instance, instead of just throwing an error, first check if the instance id starts with the special character "@" . If so, insert an ExecutionStartedEvent for that instance, using an orchestration name equal to the instance id.

This could break existing code for users that are using their own instance ids and those instance ids start with '@', which is unlikely to be an issue but probably need to document it somewhere.

This PR is related to https://github.com/Azure/azure-functions-durable-extension/issues/21 and https://github.com/Azure/azure-functions-durable-extension/issues/22.

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AutoStartOrchestrationCreator.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AutoStartOrchestrationCreator.cs
@@ -1,0 +1,65 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


cgillum:
One of the other things we want to try to deliver soon is support for Durable to run against providers other than Azure Storage. For example, Connor is doing work so that we can run workloads against DurableTask.Emulator and [DurableTask.Redis](https://github.com/Azure/durabletask/tree/redis). Unfortunately this means that we won't get auto-start for these other providers if we only make it work for DurableTask.AzureStorage.

What would it take to get this PR implemented purely in DurableTask.Core?

BTW, don't worry about the CI failures. There is a known issue where they fail for external pull requests.

sebastianburckhardt:
I couldn't really figure out a way to confine the changes to DurableTask.Core because the providers will refuse to deliver an event to a non-existing instance. So, I think there is not really a good way to avoid changing the providers to support auto-start (there are bad ways, e.g. check if an instance exists prior to sending an event, which introduces a lot of extra latency and traffic).  

The good news is that the change is actually very simple, I don't anticipate any challenges in providing it in other providers.

sebastianburckhardt:
Can we merge this into azure-storage?

cgillum:
Yes, thanks for switching branches the branches.

PR #270, opened 2019-03-21
Update documentation to add details about Azure storage connection string for unit tests

grajdeanserghei:
Documentation in README file does mention only about Service Bus Connection string for unit tests. It would be good if Azure Storage configuration for unit tests is briefly described.

This pull request adds details about Azure Storage connection string configuration for unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,9 @@ https://www.nuget.org/packages/DurableTask
 
 To run unit tests, you must specify your Service Bus connection string for the tests to use. You can do this via the **ServiceBusConnectionString** app.config value in the test project, or by defining a **DurableTaskTestServiceBusConnectionString** environment variable. The benefit of the environment variable is that no temporary source changes are required. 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +44,8 @@ https://www.nuget.org/packages/DurableTask
 
 To run unit tests, you must specify your Service Bus connection string for the tests to use. You can do this via the **ServiceBusConnectionString** app.config value in the test project, or by defining a **DurableTaskTestServiceBusConnectionString** environment variable. The benefit of the environment variable is that no temporary source changes are required. 
 
+Unit tests also require Azure storage, but by default Azure Storage Emulator is used. You can change this behaviour via the **StorageConnectionString** app.config value in the test project, or by defining a **DurableTaskTestStorageConnectionString** environment variable.
```
cgillum: While this is true, one problem people will run into when using a real storage account is that many tests will fail with a 409 Conflict error. This is because tests delete and quickly recreate the same storage tables, and Azure Storage doesn't do well in these conditions. I think if we want to add these instructions, we should mentioned this caveat.

grajdeanserghei: I totally agree, it must be mentioned in documentation. I'll update the pull request.

grajdeanserghei: I've updated the description

PR #272, opened 2019-03-25
Use LastUpdatedTime in DurableTask.Emulator

ConnorMcMahon:
The Durable Functions tests have some assertions around the field LastUpdatedTime. When this field isn't set, the default values cause those assertions to fail when using DurableTask.Emulator as a storage provider.


cgillum:
@simonporter or @adarsh1 can you take a look? It's a simple one-line change.

PR #277, opened 2019-04-05
DurableTask.AzureStorage: Stop throwing ArgNullException on invalid status

cgillum:
This fix is required for https://github.com/Azure/azure-functions-durable-extension/issues/688

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -495,7 +495,8 @@ namespace DurableTask.AzureStorage.Tracking
             var orchestrationState = new OrchestrationState();
             if (!Enum.TryParse(orchestrationInstanceStatus.RuntimeStatus, out orchestrationState.OrchestrationStatus))
             {
-                throw new ArgumentException($"{orchestrationInstanceStatus.RuntimeStatus} is not a valid OrchestrationStatus value.");
+                // This is not expected, but could happen if there is invalid data in the Instances table.
+                orchestrationState.OrchestrationStatus = (OrchestrationStatus)(-1);
             }
 
             orchestrationState.OrchestrationInstance = new OrchestrationInstance


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -495,7 +495,8 @@ OrchestrationState ConvertFromAsync(OrchestrationInstanceStatus orchestrationIns
             var orchestrationState = new OrchestrationState();
             if (!Enum.TryParse(orchestrationInstanceStatus.RuntimeStatus, out orchestrationState.OrchestrationStatus))
             {
-                throw new ArgumentException($"{orchestrationInstanceStatus.RuntimeStatus} is not a valid OrchestrationStatus value.");
+                // This is not expected, but could happen if there is invalid data in the Instances table.
+                orchestrationState.OrchestrationStatus = (OrchestrationStatus)(-1);
```
ConnorMcMahon: So here we are casting a `short` value to an enum that doesn't have a value for that short. From my understanding, that just means that any OrchestrationStatus the customer tries to compare this against will fail, and if they try to log what the actual "status" was for debugging purposes, it will print a `-1` .

How exactly do we expect the customers to handle this case? This feels like a case where an exception is actually warranted and more clear for the end consumer of this API.

cgillum: The problem with exceptions is that there's nothing the customer can do to fix this situation (short of updating the storage table directly). The customer facing issue with the current implementation is that orchestration queries are failing due to one invalid row in the table. I don't think failing the entire query operation is warranted. Another operation is to remove the row from the results. However, I don't like this either because it hides the fact that an instance actually exists, even though the data somehow got messed up. I preferred an approach which allows the customer to see as much of their data as possible.

Ideally there would be an `Unknown` status instead of using `-1`, but such an enumeration doesn't exist in `DurableTask.Core` today. We could add it, but it will require approval from the DTFx core maintainers. For Durable Functions, I actually have another PR which adds an `Unknown` status which does map to `-1`, so those customers will see a real status value that they can handle.

ConnorMcMahon: I ended up seeing the new enum value in your other PR. That takes away most of my reservations with this. I was mainly concerned with returning some practically invisible magic value instead of returning some notion that something went wrong.

Issue #274, opened 2019-04-01
AzureStorageOrchestrationService.GetStatus returns blob references


sebastianburckhardt:
When calling GetStatus(), blob references appearing in the Input field are not automatically decompressed. 

This makes it difficult for clients to use since they can't (or shouldn't) decompress these references themselves. I think that either these references should be always decompressed automatically, or (perhaps better for performance) the clients should have some way to call into an API so they can decompress the references themselves.

PR #282, opened 2019-04-19
Provide method for downloading blobs in AzureStorageOrchestrationService

sebastianburckhardt:
When using `AzureStorageOrchestrationService`, large status fields (input or output) are compressed and stored in blobs, which is difficult for clients to access (see issue #274).

This PR solves this problem by adding a public method for decompressing and downloading the content of such blobs.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1523,12 +1523,11 @@ namespace DurableTask.AzureStorage
         }
 
         /// <summary>
-        /// Download status fields whose content was too large to be stored in the orchestration status,
-        /// such as the input or output, and for which the blob URI was stored instead.
+        /// Download content that was too large to be stored in table storage,
+        /// such as the input or output status fields, and for which the blob URI was stored instead.
         /// </summary>
-        /// <param name="blobUri"></param>
-        /// <returns></returns>
-        public Task<string> DownloadLargeStatusField(string blobUri)
+        /// <param name="blobUri">The URI of the blob.</param>
+        public Task<string> DownloadBlobAsync(string blobUri)
         {
             return this.messageManager.DownloadAndDecompressAsBytesAsync(new Uri(blobUri));
         }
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1522,6 +1522,17 @@ public Task PurgeOrchestrationHistoryAsync(DateTime thresholdDateTimeUtc, Orches
             return this.trackingStore.PurgeHistoryAsync(thresholdDateTimeUtc, timeRangeFilterType);
         }
 
+        /// <summary>
+        /// Download status fields whose content was too large to be stored in the orchestration status,
+        /// such as the input or output, and for which the blob URI was stored instead.
+        /// </summary>
+        /// <param name="blobUri"></param>
+        /// <returns></returns>
```
cgillum: I suggest removing the empty `<returns>` comment.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1522,6 +1522,17 @@ public Task PurgeOrchestrationHistoryAsync(DateTime thresholdDateTimeUtc, Orches
...
+        public Task<string> DownloadLargeStatusField(string blobUri)
```
cgillum: This method name needs to be suffixed with `Async`. Also, since it's not always a status field, consider a more generic name. `DownloadBlobAsync`?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: Absolutely.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: Sure, makes sense.

cgillum:
The CI failure can be safely ignored for now. There is an issue where we cannot run tests correctly when PRs are submitted from external contributors.

PR #288, opened 2019-04-30
Merge initial Redis Provider work into master

ConnorMcMahon:


src/DurableTask.Core/WorkItemDispatcher.cs
```diff
@@ -173,6 +173,7 @@ public async Task StopAsync(bool forced)
                 if (!forced)
                 {
                     var retryCount = 7;
+                    //TODO: race condition in graceful shutdown.
```
cgillum: I don't remember - what is this comment about?

cgillum:
I suggest we add a `README` which explains:

1. That this is currently "experimental" and has caveats (e.g. only a single worker is supported)
2. How to use it - i.e. a simple code snippet for how to setup the TaskHubWorker with the correct configuration + some notes about running Redis locally.

src/DurableTask.Core/WorkItemDispatcher.cs
ConnorMcMahon: Ah, this is something I found during testing. Sometimes `TaskHubWorker.StopAsync(isForced=false)` returns from the `await` before the `TaskHubWorker` actually stops due to what I believe is this line. This was causing some major issues in my test `SimpleFanOutOrchestration_DurabilityTest`, as I would create a new `TaskHubWorker` while the old one was still processing events.

PR #290, opened 2019-05-15
Prevent executing new events if their task has already completed

ConnorMcMahon:
This should at least partially mitigate [this issue](https://github.com/Azure/azure-functions-durable-extension/issues/724) for Durable Functions.

### DIFF
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -36,7 +36,7 @@ namespace DurableTask.Core
         /// </summary>
         public IList<HistoryEvent> NewEvents { get; }
 
-        private readonly ISet<int> completedEventIds;
+        readonly ISet<int> completedEventIds;
 
         /// <summary>
         /// Compressed size of the serialized state
@@ -214,15 +214,15 @@ namespace DurableTask.Core
             SetMarkerEvents(historyEvent);
         }
 
-        private bool IsDuplicateEvent(HistoryEvent historyEvent)
+        bool IsDuplicateEvent(HistoryEvent historyEvent)
         {
             if (historyEvent.EventId >= 0 &&
                 historyEvent.EventType == EventType.TaskCompleted &&
                 !completedEventIds.Add(historyEvent.EventId))
             {
                 TraceHelper.Trace(TraceEventType.Warning, 
-                    "AddEvent-DuplicateEvent", 
-                    "The orchestration {0} has already have seen a completed task with id {1}.",
+                    "OrchestrationRuntimeState-DuplicateEvent", 
+                    "The orchestration {0} has already seen a completed task with id {1}.",
                     this.OrchestrationInstance.InstanceId,
                     historyEvent.EventId);
                 return true;
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -35,6 +35,8 @@ namespace DurableTask.Core
         /// </summary>
         public IList<HistoryEvent> NewEvents { get; }
 
+        private ISet<int> SuccessfulEventIds { get; }
+
         /// <summary>
         /// Compressed size of the serialized state
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -195,10 +198,24 @@ public void AddEvent(HistoryEvent historyEvent)
...
+                     !SuccessfulEventIds.Contains(historyEvent.EventId) // It shares an ID with a successfully completed event
+                )
+                {
+                    NewEvents.Add(historyEvent);
+                    Events.Add(historyEvent);
+                }
+            }
+            else
+            {
+                Events.Add(historyEvent);
```
ConnorMcMahon: I should potentially also perform the dedupe check here, but I wasn't sure if we wanted to ignore a history event if it is already in the history table.

cgillum: My suggestion is to do this check regardless of `isNewEvent` and to also log a warning message. i.e. at the top of this method, do this:

```csharp
// We use the task's event ID to check for and filter out duplicate events
if (historyEvent.EventId >= 0 &&
    DeDupeEventType(historyEvent.EventType) &&
    !SeenEventIds.Add(historyEvent.EventId))
{
    // trace warning
    return;
}
```

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -35,6 +35,8 @@ public class OrchestrationRuntimeState
         /// </summary>
         public IList<HistoryEvent> NewEvents { get; }
 
+        private ISet<int> SuccessfulEventIds { get; }
```
cgillum: I don't think there is any concept of "success" at this layer. I would name this `SeenEventIds`. Also, since it's not public, I would make it a `readonly HashSet<int>` _field_ instead of an `ISet<T>` _property_. Going through an interface just adds unnecessary runtime overhead.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -195,10 +198,24 @@ public void AddEvent(HistoryEvent historyEvent)
...
+                )
+                {
+                    NewEvents.Add(historyEvent);
+                    Events.Add(historyEvent);
+                }
+            }
+            else
+            {
+                Events.Add(historyEvent);
+                if (historyEvent.EventType == EventType.TaskCompleted)
```
cgillum: Small nit, but I think it might be slightly more readable/maintainable if you put this check behind a method, like in my other code suggestion.

src/DurableTask.Core/OrchestrationRuntimeState.cs
ConnorMcMahon: For the naming, I'm a little hesitant to call it `SeenEventIds`, as there are some events that we add that we don't want to dedupe in order to still support Rewind behavior. `CompletedEventIds` might be the appropriate rename option.

cgillum:
Hi @simonporter , @adarsh1 - would you mind reviewing this small PR to DurableTask.Core? It helps address some customer issues we've been encountering recently.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -35,6 +36,8 @@ public class OrchestrationRuntimeState
         /// </summary>
         public IList<HistoryEvent> NewEvents { get; }
 
+        private readonly ISet<int> completedEventIds;
```
simonporter: nit: please remove 'private modifier' here and lower

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -204,6 +214,22 @@ void AddEvent(HistoryEvent historyEvent, bool isNewEvent)
...
 
+        private bool IsDuplicateEvent(HistoryEvent historyEvent)
+        {
+            if (historyEvent.EventId >= 0 &&
+                historyEvent.EventType == EventType.TaskCompleted &&
+                !completedEventIds.Add(historyEvent.EventId))
+            {
+                TraceHelper.Trace(TraceEventType.Warning, 
+                    "AddEvent-DuplicateEvent", 
+                    "The orchestration {0} has already have seen a completed task with id {1}.",
```
simonporter: wording: 'has already have seen' should be 'has already seen'

PR #291, opened 2019-05-21
AzureTableTrackingStore fix for RewindHistoryAsync query

Scooletz:
This PR fixes the query creation in `AzureTableTrackingStore.RewindHistoryAsync`. As far as I see there was a missing whitespace in there.

A note for the reviewer: I found not that much tests for AzureTableTrackingStore (only a few with a mocked `Instance` table) so maybe these are tested in some other way as well.


cgillum:
I verified that all the existing rewind tests are still passing with this change, so it looks like we're good to go with this PR. I'll go ahead and merge it and include it as part of the next release.

Scooletz:
@cgillum I tried to do the same yet I failed to find a test covering this case. It might be me, or is it possible that it is not covered at all?

PR #298, opened 2019-06-10
Fix race condition that creates stuck orchestrations in Azure Storage provider.

ConnorMcMahon:


ConnorMcMahon:
Targeted wrong branch with initial change. Will retarget.

PR #299, opened 2019-06-10
Fix race condition that creates stuck orchestrations in Azure Storage provider

ConnorMcMahon:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/774


src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -146,5 +149,10 @@ internal bool IsOutOfOrderMessage(MessageData message)
             // The message is out of order and cannot be handled by the current session.
             return true;
         }
+
+        internal bool IsNonexistantInstance()
```
cgillum: I'm thinking this should just be private instead of `internal`. (per the coding conventions for durabletask, we should just remove the `internal` modifier).

PR #303, opened 2019-06-21
Add support to automatically fetch large messages from blob storage

ConnorMcMahon:


src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -109,6 +110,24 @@ public async Task<string> SerializeMessageDataAsync(MessageData messageData)
...
 
+        /// <summary>
+        /// If the "message" of an orchestration state is actually a URI retrieves actual message from blob storage.
+        /// Otherwise returns the message as is.
+        /// </summary>
+        /// <param name="message">The message to be fetched if it is a url.</param>
+        /// <returns>Actual string representation of message.</returns>
+        public async Task<string> FetchLargeMessageIfNecessary(string message)
+        {
+            if (Uri.IsWellFormedUriString(message, UriKind.Absolute))
```
cgillum: What if the input/output is a URL but it's not actually a blob URL?

ConnorMcMahon: Hmm, did not think of that edge case...

A couple of options:

1. Try to detect if the URL is a blob url.
2. Wrap the "grab blob" logic in a try catch and return the URL if an exception is thrown.

The first one feels correct, but also a bit harder to do. Thoughts?

cgillum: I feel like there should be a more guaranteed way to do this because we didn't always return a URL like this in the past. We might need to look at the code history to understand the method we were using previously. I believe as part of deserializing table record objects we also downloaded the blobs.

ConnorMcMahon: I have added a test to verify that if the input/output is a URI, we don't run into any issues. The reason is because inputs and outputs are encoded as JTokens, so they are wrapped in an extra set of `""`.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -684,22 +678,19 @@ public async Task ActorOrchestrationDeleteAllLargeMessageBlobs(bool enableExtend
...
                 // TODO: Sleeping to avoid a race condition where multiple ContinueAsNew messages
                 //       are processed by the same instance at the same time, resulting in a corrupt
                 //       storage failure in DTFx.
                 await Task.Delay(10000);
                 await client.RaiseEventAsync("operation", "double");
+                finalMessage = finalMessage + new string(finalMessage.Reverse().ToArray());
                 counter *= 2;
                 await Task.Delay(10000);
                 await client.RaiseEventAsync("operation", "double");
-                counter *= 2;
```
ConnorMcMahon: Got rid of these because the test was running needlessly long already.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2014,7 +2089,7 @@ internal class CharacterCounter : TaskOrchestration<Tuple<string, int>, Tuple<st
                     {
                         case "double":
                             inputData = new Tuple<string, int>(
-                                $"{inputData.Item1}{inputData.Item1.Reverse()}",
+                                $"{inputData.Item1}{new string(inputData.Item1.Reverse().ToArray())}",
```
ConnorMcMahon: The second half of the string was being printed as `IEnumerable<char>` instead of the actual value. This fixes that.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -122,6 +122,12 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public bool ExtendedSessionsEnabled { get; set; }
 
+        /// <summary>
+        /// Gets or sets a flag indicating whether to automatically fetch large orchestration input and outputs
+        /// when it is stored in a compressed blob when retrieving orchestration state.
+        /// </summary>
+        public bool FetchLargeMessageDataEnabled { get; set; } = false;
```
cgillum: Shouldn't this be `true` by default?

src/DurableTask.AzureStorage/MessageManager.cs
cgillum: Okay, let's go ahead with this then.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
ConnorMcMahon: I was debating this. I am fine with defaulting it to `true` in Durable Functions when we construct the `AzureStorageOrchestrationServiceSettings`, but I wasn't sure if we have other users of DurableTask.AzureStorage that might be more surprised by this change.

cgillum: It's possible, but I'm willing to take my chances with this one given that it's mostly only used by Durable customers and the current behavior is quite surprising and unpleasant.

PR #304, opened 2019-06-26
Fix pending orchestration race condition in Azure Storage Provider

ConnorMcMahon:
Adds progressively longer visibility delays for abandoned messages to
give history table time to populate.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
@@ -216,7 +216,7 @@ namespace DurableTask.AzureStorage.Messaging
 
             try
             {
-                // We "abandon" the message by settings its visibility timeout to it's dequeue count + 1 seconds.
+                // We "abandon" the message by settings its visibility timeout to its dequeue count + 1 seconds.
                 // This allows it to be reprocessed on this node or another node.
                 await this.storageQueue.UpdateMessageAsync(
                     queueMessage,
--- a/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
@@ -196,6 +196,9 @@ namespace DurableTask.AzureStorage.Messaging
                 // The delay needs to be long enough to allow some progress to be made but
                 // short enough to allow for debugging.
                 visibilityDelay = TimeSpan.FromMinutes(10);
+            } else
+            {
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -196,6 +196,9 @@ public async Task AbandonMessageAsync(MessageData message, SessionBase session)
                 // The delay needs to be long enough to allow some progress to be made but
                 // short enough to allow for debugging.
                 visibilityDelay = TimeSpan.FromMinutes(10);
+            } else
```
cgillum: nit: Can you fix the formatting on this line?

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -196,6 +196,9 @@ public async Task AbandonMessageAsync(MessageData message, SessionBase session)
...
+            {
+                visibilityDelay = TimeSpan.FromSeconds(queueMessage.DequeueCount+1);
```
cgillum: I wonder if we should make these delays a tad longer. What if we did `Math.Pow(queueMessage.DequeueCount - 1, 2)`?

ConnorMcMahon: I think we should look at two goals when setting the visibility timer.

1. Ensuring that the second and possibly third attempts to process an out of order message don't take too long. This is because the race condition is almost always resolved very quickly (sub-second) and we don't want to wait 5+ seconds to try again in that case.
2. Ensuring that the sum of all of the attempts is long enough for the worst case scenario we imagine this race condition will last.

I chose DequeueCount + 1 seconds, as we get 2 retries within 3 seconds, and after all 4 retries we have given 10 seconds for the race condition to resolve (I did my math wrong in the comment in `IsOutOfOrderMessage`). 

I am open to other sequences of numbers, but the one you suggested strikes me as a bit aggressive, especially considering the maximum we set a poisoned message to is 10 minutes, and that is only after 100 dequeues.  I am a little nervous about using quadratic growth considering other code paths also abandon messages, and don't necessarily stop at 5 tries.

ConnorMcMahon: I have changed this to use exponential backoff. It should give us 3 tries in the first 3 seconds, and then falls off rapidly until it hits the maximum of 10 minutes. It has the advantage of being a tried and true approach for this sort of problem.

PR #317, opened 2019-09-18
[ServiceFabricProvider] Unify Service name across different ways to deploy service.  Favor interface based tasks.

kamesh-msft:
* Unify Service name across different ways to deploy service. 
* Introduce interface based tasks.


Test/TestFabricApplication/TestApplication.Common/OrchestrationTasks/UserTasks.cs
```diff
@@ -11,14 +11,21 @@
...
+    using System.Threading.Tasks;
 
-    public class GreetUserTask : TaskActivity<string, string>
+    public class UserTasks : IUserTasks
     {
-        protected override string Execute(TaskContext context, string user)
+        public async Task<string> GetUserAsync()
         {
+            await Task.Yield();
+            return "Gabbar";
```
shankarsama: Can't we just do Task.FromResult()?

Issue #319, opened 2019-09-26
AzureTableInstanceStore can only be initialized once on .Net Core 3.0


scabana:
Exception: ```{"Cannot set initonly static field 'MinDateTime' after type 'DurableTask.Core.Common.DateTimeUtils' is initialized."}``` is thrown if ```new AzureTableInstanceStore``` is called more than once. Seems like the reflection call at https://github.com/Azure/durabletask/blob/f9cc450539b5e37c97c19ae393d5bb1564fda7a8/src/DurableTask.Core/Common/DateTimeUtils.cs#L41-L47 is not blocked in .net core 3.0 if ran more than once. Because of this, it's not possible to initialize more than 1 AzureTableInstanceStore instance per application lifetime.

For reference: https://github.com/dotnet/coreclr/issues/21268#issuecomment-442675193

cgillum:
Interesting. So the reflection code works the first time but fails the second time?

Also, can you describe a scenario where you might have more than one `AzureTableInstanceStore` instance per application lifetime? I'm wondering whether this will impact Durable Functions after the Azure Functions host moves to .NET Core 3.

FYI @simonporter

scabana:
That's the behavior I'm getting, fails on second attempt.

As for why, I don't know. I'm migrating a service using durable tasks, didn't write the original code (which may not be the most optimal). For the moment, transformed the ```AzureTableInstanceStore``` as a singleton.

PR #320, opened 2019-09-27
Removed invalid .net core 3 reflection call.

scabana:
Resolves https://github.com/Azure/durabletask/issues/319. 

In .net core 3.0, validation was added so readonly fields could not be set with reflection.

### DIFF
--- a/src/DurableTask.Core/Common/DateTimeUtils.cs
+++ b/src/DurableTask.Core/Common/DateTimeUtils.cs
@@ -20,11 +20,6 @@ namespace DurableTask.Core.Common
     /// </summary>
     public static class DateTimeUtils
     {
-        static DateTimeUtils()
-        {
-            MinDateTime = DateTime.MinValue;
-        }
-
         /// <summary>
         /// Returns bool indicating is the datetime has a value set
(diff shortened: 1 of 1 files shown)


scabana:
It does indeed break people using it. I can suggest an alternative where the existing field remains but is only used as a publicly accessible value. The code within the class would use another field to keep the value private and encapsulated..

scabana:
I'm not sure I like the alternative all that much, since someone could think it's now ok to change the value. 

Also, changes would need to be more extensive since there are 15 references to the field from within the code and those places could be affected by a changed value. I think previous solution was better even though there's a breaking change than the added complexity of this new solution.

cgillum:
Could we just remove the `readonly` modifier to unblock .NET Core 3 usage? Obviously it's not "good" because the field becomes completely mutable, but it's at least not a breaking change.

It might also be interesting to see whether the latest versions of the storage emulator are actually still affected by the limitation mentioned in the comments. I know quite a few things have changed since this code was originally written.

scabana:
I don't mind just removing the readonly field. Even though I don't like it, I think it's the simplest solution.

PR #322, opened 2019-10-15
bugfix: RemoteOrchestrationService should parse response only if HttpStatusCode is successful.

kamesh-msft:


src/DurableTask.AzureServiceFabric/Extensions.cs
```diff
@@ -48,5 +50,21 @@ internal static void EnsureValidInstanceId(this string instanceId)
...
+            requestingUri = requestingUri ?? throw new ArgumentNullException(nameof(requestingUri));
+            return httpClient.GetStringResponseAsync(new Uri(requestingUri));
+        }
+
+        internal static async Task<string> GetStringResponseAsync(this HttpClient httpClient, Uri requestingUri)
+        {
+            httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+            requestingUri = requestingUri ?? throw new ArgumentNullException(nameof(requestingUri));
+            HttpResponseMessage response = await httpClient.GetAsync(requestingUri);
+            response.EnsureSuccessStatusCode();
```
shankarsama: >response.EnsureSuccessStatusCode(); [](start = 12, length = 35)

When it throws, does it include response content in the exception?

kamesh-msft: It doesn't include the body. It has http status code, reason phrase.

---
In reply to: [348185637](https://github.com/Azure/durabletask/pull/322#discussion_r348185637) [](ancestors = 348185637)

PR #324, opened 2019-10-16
New persistence order for azure storage

sebastianburckhardt:
Draft of new algorithm for persisting queue / history in opposite order than before (first persist history, then persist queue messages).

This requires adding more state to some of the events in the history, so it is possible to resend the exact same messages simply by looking at the history.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -614,7 +614,7 @@ namespace DurableTask.AzureStorage
                     List<MessageData> outOfOrderMessages = null;
                     foreach (MessageData message in session.CurrentMessageBatch)
                     {
-                        if (session.IsOutOfOrderMessage(message))
+                        if (!message.NewPersistenceOrder && session.IsOutOfOrderMessage(message))
                         {
                             if (outOfOrderMessages == null)
                             {
@@ -648,6 +648,8 @@ namespace DurableTask.AzureStorage
                         // This will also remove the messages from the current batch.
                         await this.AbandonMessagesAsync(session, outOfOrderMessages);
                     }
+
+                    await this.PreprocessEvents(session, linkedCts.Token);
 
                     if (session.CurrentMessageBatch.Count == 0)
                     {
@@ -859,19 +861,8 @@ namespace DurableTask.AzureStorage
             string instanceId = workItem.InstanceId;
             string executionId = runtimeState.OrchestrationInstance.ExecutionId;
 
-            // First, add new messages into the queue. If a failure happens after this, duplicate messages will
-            // be written after the retry, but the results of those messages are expected to be de-dup'd later.
-            // This provider needs to ensure that response messages are not processed until the history a few
-            // lines down has been successfully committed.
-            await this.CommitOutboundQueueMessages(
-                session,
-                outboundMessages,
-                orchestratorMessages,
-                timerMessages,
-                continuedAsNewMessage);
-
-            // Next, commit the orchestration history updates. This is the actual "checkpoint". Failures after this
-            // will result in a duplicate replay of the orchestration with no side-effects.
+            // First, commit the orchestration history updates. This is the actual "checkpoint". Failures after this
+            // will no longer change the history.
             try
             {
                 session.ETag = await this.trackingStore.UpdateStateAsync(runtimeState, workItem.OrchestrationRuntimeState, instanceId, executionId, session.ETag);
@@ -901,8 +892,24 @@ namespace DurableTask.AzureStorage
                 throw;
             }
 
+            // update the latest runtime state in the in-memory session object, if using in-memory sessions.
+            session.UpdateRuntimeState(runtimeState);
+
+            // Send outbound queue messages.
+            await this.CommitOutboundQueueMessages(
+                session,
+                outboundMessages,
+                orchestratorMessages,
+                timerMessages,
+                continuedAsNewMessage);
+
             // Finally, delete the messages which triggered this orchestration execution. This is the final commit.
             await this.DeleteMessageBatchAsync(session);
+
+            // If we fail before this final step, but after committing the history, the messages will become visible again; 
+            // but then they are identified as duplicates prior
+            // to processing, and any outbound messages (which may have not been durably enqueued) 
+            // are reconstructed from the saved history and resent.
         }
 
         async Task CommitOutboundQueueMessages(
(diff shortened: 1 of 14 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -956,6 +963,155 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
+                            timerMessages.Add(message);
+                            break;
+                    }
+                }
+            }
+
+            if (continuedAsNewMessage != null)
+            {
+                timerMessages.Clear();
+                outboundMessages.Clear();
```
ConnorMcMahon: It's not immediately obvious to me that we want to clear all outbound messages, and not just the ones after we encounter a `continuedAsNewMessage`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -956,6 +963,155 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
             await enqueueOperations.ParallelForEachAsync(
                 this.settings.MaxStorageOperationConcurrency,
                 op => op.Queue.AddMessageAsync(op.Message, session));
+
+            // continue as new must be enqueued after all other messages have been successfully enqueued,
+            // because once the continue as new is received, the previous history goes away, so it cannot
+            // be used to recover messages that failed during enqueueing
+            if (continuedAsNewMessage != null)
```
ConnorMcMahon: Shouldn't we remove the above block where the `continuedAsNewMessage` was added to the `enqueueOperations`?

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -107,6 +109,11 @@ public IEnumerable<MessageData> AddOrReplaceMessages(IEnumerable<MessageData> me
                 this.nextMessageBatch.Clear();
             }
 
+            if (this.CurrentMessageBatch.Count > 0)
+            {
+                await ((AzureStorageOrchestrationService)orchestrationService).PreprocessEvents(this, CancellationToken.None);
```
ConnorMcMahon: This is the second place I see us calling `PreprocessEvents`. Is there a reason we need to do it here as well as `LockNextTaskOrchestrationWorkItemAsync`?

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -150,6 +150,63 @@ public static T ReadObjectFromByteArray<T>(byte[] serializedBytes)
...
         }
 
+        /// <summary>
+        /// Compute an FNV hash from two strings and a long.
+        /// </summary>
+        /// <param name="str1">The first string.</param>
+        /// <param name="str2">The second string.</param>
+        /// <param name="number">The number.</param>
+        /// <returns></returns>
+        public static long ComputeLongHash(string str1, string str2, uint number)
```
ConnorMcMahon: This does not seem to be used elsewhere?

src/DurableTask.Core/History/EventRaisedEvent.cs
```diff
@@ -48,5 +48,13 @@ public EventRaisedEvent(int eventId, string input)
...
         [DataMember]
         public string Input { get; set; }
+
+        /// <summary>
+        /// Gets or sets the fingerprint of the event. The fingerprint is used to detect duplicate delivery of events
+        /// within an orchestration history. Collision probability is sufficiently low (less than 1 in a billion for
+        /// histories with less than 100,000 entries).
+        /// </summary>
+        [DataMember]
+        public long FingerPrint { get; set; }
```
ConnorMcMahon: Is there a reason to use a 64 bit long here instead of a ~128 bit string guid?

src/DurableTask.Core/IOrchestrationSession.cs
```diff
@@ -29,11 +29,6 @@ public interface IOrchestrationSession
         /// or until an internal wait period has expired. In either case, <c>null</c> can be returned
         /// and the dispatcher will shut down the session.
         /// </remarks>
-        Task<IList<TaskMessage>> FetchNewOrchestrationMessagesAsync(TaskOrchestrationWorkItem workItem);
-
-        /// <summary>
-        /// Updates the in-memory session info.
-        /// </summary>
-        void UpdateRuntimeState(OrchestrationRuntimeState runtimeState);
```
ConnorMcMahon: It doesn't seem like we can safely remove this method.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -956,6 +963,155 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
+                System.Diagnostics.Debug.Assert(targetedExecutionId == null || session.Instance.ExecutionId == targetedExecutionId);
+
+                // check if message is a duplicate of some history event
+                var events = session.RuntimeState.Events;
+                for (var pos = 0; pos < events.Count; pos++)
+                {
+                    var evt = events[pos];
+                    foreach (var msg in session.CurrentMessageBatch)
+                    {
+                        if (msg.TaskMessage.Event.EventType == evt.EventType
```
ConnorMcMahon: Let's move this check into `this.IsDuplicate()` so if another dev comes along and tries using the method before verifying the event types are the same we don't throw a casting exception.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -956,6 +963,155 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
+            if (toRemove.Count > 0)
+            {
+                foreach (var x in toRemove)
+                {
+                    session.CurrentMessageBatch.Remove(x);
+                }
+
+                if (earliestDuplicate < int.MaxValue)
+                {
+                    await this.ResendAllPossiblyLostMessages(session, earliestDuplicate);
```
ConnorMcMahon: Is there a reason we send all the messages after the first duplicate instead of just sending all of the duplicates?

sebastianburckhardt:
I realized a fundamental problem with this whole approach - it is not compatible with the "fast continueAsNew" optimization. Because the fastContinueAsNew throws away the histories of all but the latest executionId, so they won't be persisted, so they can't be used to recover lost messages.

sebastianburckhardt:
We are suspending work on this in favor of developing the new back-end. We can reconsider this choice if we see bugs manifesting in practice.

cgillum:
Closing for now.

PR #325, opened 2019-10-16
Added Options to Configure Behavior When Starting a New Orchestrator With an Existing Instance Id

amdeel:
Added options to configure which states you want to override an existing orchestrator instance id when attempting to start a new instance.

Fixes this issue in the azure-functions-durable-extension repository
https://github.com/Azure/azure-functions-durable-extension/issues/367

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1244,6 +1244,12 @@ namespace DurableTask.AzureStorage
                 return;
             }
 
+            if (existingInstance != null && !OverrideConditionsMet(existingInstance.OrchestrationStatus))
+            {
+                // An instance in a state that is configured to not be overridable already exists
+                return;
+            }
+
             ControlQueue controlQueue = await this.GetControlQueueAsync(creationMessage.OrchestrationInstance.InstanceId);
             MessageData internalMessage = await this.SendTaskOrchestrationMessageInternalAsync(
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1244,6 +1244,12 @@ public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, Orch
                 return;
             }
 
+            if (existingInstance != null && !OverrideConditionsMet(existingInstance.OrchestrationStatus))
```
cgillum: Could we achieve the same outcome if we simply depended on the existing `dedupeStatuses` parameter? I don't have an good answer for this that I can think of right now, but something to consider.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1244,6 +1244,12 @@ public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, Orch
...
+            {
+                // An instance in a state that is configured to not be overridable already exists
+                return;
```
cgillum: I think we should throw an `InvalidOperationException` in this case. Otherwise customers might not realize that their attempt to start an orchestration was ignored.

amdeel:
Closing this PR. Decided to use dedupe status instead and modify code in azure-functions-durable-extension. https://github.com/Azure/azure-functions-durable-extension/pull/987

PR #326, opened 2019-10-16
Make large message blob names unique

ConnorMcMahon:
We previously generated "unique" blob names using the SequenceNumber
field of the MessageData object. However, these sequence numbers are
generated by the client, This means is no uniqueness guarantee for a given
instance id, as any VM running the application is allowed to send
messages. This meant it was possible for large messages sent from
different VMs to the same instance could overwrite each other.

We now use activity ids, as they are random guids that should
not conflict with other messages, regardless of which client they are
sent from.


ConnorMcMahon:
Not quite sure how to reliably test this, considering I would have to emulate running a single instance two separate TaskHubWorkers, and prove the negative that they would never create identical blob names.

PR #329, opened 2019-10-24
DurableTask.AzureStorage: Orchestration message counting fixes

### LINKED ISSUE #328: Improper message counting leads to performance degradation
In **DurableTask.AzureStorage**, we prefetch orchestration messages as a performance optimization. Counting logic exists to ensure we don't prefetch too many messages. There is a bug in the prefetch message counting where we don't count messages which are assigned to an already-loaded orchestration session. This can lead to unbounded message pile-ups and huge amounts of duplicate message processing, which greatly impacts performance.

This issue was discovered while doing performance testing for the [Durable Entities](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-entities) scenario.

cgillum:
Resolves #328 

This PR greatly simplifies the message counting logic to avoid message-piling problems when many messages are sent to an active instance. I'm keeping a list of all message IDs that we read from the queue and removing from that list when we abandon or delete those messages. This solution should be much more accurate than the previous method while also fixing the pile-up problem.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
@@ -183,9 +183,12 @@ namespace DurableTask.AzureStorage.Messaging
 
         public override Task AbandonMessageAsync(MessageData message, SessionBase session)
         {
-            if (this.pendingMessageIds.Remove(message.OriginalQueueMessage.Id))
+            lock (this.pendingMessageIds)
             {
-                this.stats.PendingOrchestratorMessages.Decrement();
+                if (this.pendingMessageIds.Remove(message.OriginalQueueMessage.Id))
+                {
+                    this.stats.PendingOrchestratorMessages.Decrement();
+                }
             }
 
             return base.AbandonMessageAsync(message, session);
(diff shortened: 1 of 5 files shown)


src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -156,6 +181,26 @@ await batch.ParallelForEachAsync(async delegate (CloudQueueMessage queueMessage)
             }
         }
 
+        public override Task AbandonMessageAsync(MessageData message, SessionBase session)
+        {
+            if (this.pendingMessageIds.Remove(message.OriginalQueueMessage.Id))
```
ConnorMcMahon: I feel like locking on `this.pendingMessageIds` might be prudent here. It's unlikely that we would try to Delete/Abandon the same message concurrently, but if we did it could definitely mess with our `this.stats.PendingOrchestratorMessages` count.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -315,8 +287,6 @@ public async Task<OrchestrationSession> GetNextSessionAsync(CancellationToken ca
                                 ExecutionId = nextBatch.OrchestrationExecutionId,
                             };
 
-                        this.stats.PendingOrchestratorMessages.Increment(-nextBatch.Messages.Count);
```
ConnorMcMahon: Now that the only write operations to `this.stats.PendingOrchestratorMessages` are inside of `ControlQueue`, it would really be nice if we had some way to enforce that. It would probably involve a lot of coding effort to manage this, but I think we should at least document this.

cgillum: I agree, and I had the same thought about enforcement. However, I do think it's probably more work (and churn risk) than I'm willing to spend on this PR.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
cgillum: Hmm...originally I thought this was fine because processing a single orchestration happens on a single (logical) thread. However, your comment is making me realize that there could be multiple orchestrations running concurrently and messing with this HashSet. I don't think that a race condition would impact the pending orchestrator messages count, but it could certainly blow up the HashSet if two threads hit it at once. I'll look into locking and/or using a thread-safe data structure.

PR #330, opened 2019-10-24
Removed ToLower on blob storage path names

amdeel:
Removed ToLower on blob storage path names that caused blobs to be stored with a lowercase instanceId.

Fixes azure-functions-durable-extension issue #755: 
https://github.com/Azure/azure-functions-durable-extension/issues/755

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -269,7 +269,7 @@ namespace DurableTask.AzureStorage
             string eventType = message.TaskMessage.Event.EventType.ToString();
             string sequenceNumber = message.SequenceNumber.ToString("X16");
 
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -269,7 +269,7 @@ internal string GetNewLargeMessageBlobName(MessageData message)
             string eventType = message.TaskMessage.Event.EventType.ToString();
             string activityId = message.ActivityId.ToString("N");
 
-            return $"{instanceId}/message-{activityId}-{eventType}.json.gz".ToLowerInvariant();
+            return $"{instanceId}/message-{sequenceNumber}-{eventType}.json.gz";
```
cgillum: I think your commit may have undid a bug fix made by Connor. I think this line should actually be:
```csharp
return $"{instanceId}/message-{activityId}-{eventType}.json.gz";
```

PR #331, opened 2019-10-28
DurableTask.AzureStorage: Improved detection and handling of known race conditions

cgillum:
There are several race conditions in DurableTask.AzureStorage which can result in slow processing and duplicate executions. These were especially apparent when building the reminders feature of Durable Entities and running entity workloads under heavy load.

There is a new `README.md` file which describes what the race conditions are and how we are handling them. I recommend reading that first before diving into the code.

There should be no observeable performance change based on these updates, except that things may run faster due to much fewer duplicate executions.

This PR should be seen as an alternate to https://github.com/Azure/durabletask/pull/324, which is also about reliability but has a much higher performance impact. I validated that the current unblocks various tests in the reminders PR: https://github.com/Azure/azure-functions-durable-extension/pull/990


src/DurableTask.AzureStorage/MessageData.cs
```diff
@@ -69,11 +75,18 @@ public MessageData()
...
-        /// This value may be <c>0</c> if the orchestration instance that created
-        /// the message was started before episode numbers were tracked.
+        /// This value may be <c>null</c> if the orchestration instance that created
+        /// the message was started before episode numbers were tracked, or if the message
+        /// was created by a client.
         /// </remarks>
-        [DataMember]
-        public int Episode { get; private set; }
+        [DataMember(EmitDefaultValue = false)]
+        public int? Episode { get; private set; }
```
ConnorMcMahon: I understand that making this nullable is technically more accurate, but in all practicality, it seems we have just moved the burden of null checking from the callers of the constructor for `MessageData` to all of the consumers of `MessageData`, which means far more null handling overall, and more opportunities to have inconsistent values in our logs for null values of `Episode`.

Is there some approach we can take to make sure that all of our logging infrastructure uses the same fixed value for `null`?

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -122,23 +140,34 @@ internal bool IsOutOfOrderMessage(MessageData message)
...
-            // this task was scheduled. We don't have the luxery of transactions between queues and tables, so queue
-            // messages are always written before we update the history in table storage. This means that in some
-            // cases the response message could get picked up before we're ready for it.
-            HistoryEvent mostRecentTaskEvent = this.RuntimeState.Events.LastOrDefault(e => e.EventId == taskScheduledId);
-            if (mostRecentTaskEvent != null)
+            if (this.LastCheckpointTime > message.TaskMessage.Event.Timestamp)
             {
+                // LastCheckpointTime represents the time at which the most recent history checkpoint completed.
+                // The checkpoint is written to the history table only *after* all queue messages are sent.
+                // A message is out of order when its timestamp *preceeds* the most recent checkpoint timestamp.
```
ConnorMcMahon: This seems to be the opposite of true. Here we are saying the message _isn't_ out of order when the message's timestamp proceeds the last checkpoint time.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -122,23 +140,34 @@ internal bool IsOutOfOrderMessage(MessageData message)
...
+                // This logic only applies for messages sent by orchestrations.
+                // Orchestration checkpoint time information was added only after v1.6.4.
                 return false;
             }
 
+            ////// This message is a response to a task. Search the history to make sure that we've recorded the fact that
+            ////// this task was scheduled. We don't have the luxery of transactions between queues and tables, so queue
+            ////// messages are always written before we update the history in table storage. This means that in some
+            ////// cases the response message could get picked up before we're ready for it.
+            ////HistoryEvent mostRecentTaskEvent = this.RuntimeState.Events.LastOrDefault(e => e.EventId == taskScheduledId);
```
ConnorMcMahon: So here we are relying on our new "optimistic" concurrency approach, and no longer need our old validation logic, right? If so, we should probably remove this commented out section.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -407,11 +434,11 @@ public PendingMessageBatch(ControlQueue controlQueue, string instanceId, string
 
             public ControlQueue ControlQueue { get; }
             public string OrchestrationInstanceId { get; }
-            public string OrchestrationExecutionId { get; }
+            public string OrchestrationExecutionId { get; set; }
             public MessageCollection Messages { get; } = new MessageCollection();
             public OrchestrationRuntimeState OrchestrationState { get; set; }
             public string ETag { get; set; }
-            public bool IsLoadingState { get; set; }
```
ConnorMcMahon: Out of curiosity, what was this field used for? It looks like it was only ever set, but never referenced.

src/DurableTask.AzureStorage/README.md
```diff
@@ -0,0 +1,33 @@
+﻿# DurableTask.AzureStorage
+
+## Orchestration message processing
+
+### Prefetching
+
+A single task hub worker listens on zero or more control queues. Each control queue processes messages for potentially many different orchestration instances. As a performance optimization, these messages are prefetched and buffered in memory using a background thread. As many as `ControlQueueBufferThreshold` messages can be buffered in memory at a time on a single task hub worker (64 by default).
+
+Each message has an orchestration *instance ID* and *execution ID*. Prefetched messages are sorted by their instance and execution IDs into *batches*. For each batch, we also prefetch the orchestration runtime state from table storage. Once the runtime state has been fetched, the message batch is made available for processing. When the DTFx dispatcher requests the next work item, a batch of messages with their runtime state are converted into an orchestration *session*.
```
ConnorMcMahon: nit: "grouped" might be more correct than "sorted", as sorted as extra connotations.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -174,11 +191,19 @@ async void DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
-                            targetBatch = batch;
-                            break;
+                            if (executionId == null || batch.OrchestrationExecutionId == executionId)
+                            {
+                                targetBatch = batch;
+                                break;
+                            }
+                            else if (batch.OrchestrationExecutionId == null)
+                            {
+                                targetBatch = batch;
```
ConnorMcMahon: So here, if we have a batch in the same orchestration that doesn't yet have an execution id, it takes on the execution id of the the first message with a  matching instance id that has an execution id? That logic makes sense, but I guess I don't know what potential bugs/corner-cases this addresses.

src/DurableTask.AzureStorage/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Because orchestration messages are prefetched on a background thread, messages may arrive for an orchestration instance that has already started processing an earlier batch. When this happens, these messages are added to the existing orchestration session. If *extended sessions* are enabled, these messages will be eventually picked up by the orchestration when its ready for its next batch. Otherwise, these messages will be added back to the message buffer after the existing orchestration session is released.
```
ConnorMcMahon: I did not realize we added messages to in flight orchestration sessions. Does this ever cause race conditions?

I.e. what happens if an orchestration has already replayed an Activity that was not completed from its perspective, but before it finishes the session, it receives a message saying that activity has completed?

src/DurableTask.AzureStorage/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Race conditions
+
+The checkpoint order mentioned previously adds queue messages before writing history to Table storage. This behavior creates the possibility of interesting race conditions. For example, if an orchestration schedules an activity to run, its possible that the activity runs and returns a result before the orchestration has finished updating its history. This race could potentially cause a prefetch of the old orchestration history, which becomes invalid as soon as the still-running orchestration finishes writing the new history to Table storage!
+
+The message/history race condition is handled in two different ways:
+
+* If a message arrives for an *active* orchestration *on the same task hub worker*, instead of prefetching the history, we add that message to the active orchestration session. The session will then either process the message later or will add it back to the prefetch buffer after it finishes writing the history. This is the common case since orchestration messages are usually processed by the same task hub worker.
```
ConnorMcMahon: Adding a description of when we "process the message later" vs "adding it back to the prefetch buffer" might be useful to add.

src/DurableTask.AzureStorage/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+* If a lease was reassigned to a different task hub worker after an orchestration started running, it's possible that the orchestration instance will be active on one worker but messages will arrive on another. In this case, we cannot know whether the orchestration associated with a message is active or not. Instead, we have to use timestamps to detect race conditions. Each message has a timestamp and the orchestration has a "last checkpoint completed" timestamp. If the timestamp of a message comes after the timestamp of the checkpoint, then we know the orchestration is not yet ready to process the message. We handle this by putting the message back on the queue and tracing a warning message (`AbandoningMessage`). The orchestration's "last checkpoint completed" timestamp will be updated soon, so eventually the abandoned message can be processed successfully.
```
ConnorMcMahon: NIT: Technically, the timestamp workaround also applies for the corner case when the TaskHubWorker dies after sending messages but before it can checkpoint the history. We may want to specifically call out that case, as that is an important scenario.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -195,6 +204,17 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
                 executionId = expectedExecutionId;
             }
 
+            // Read the checkpoint completion time from the sentinel row, which should always be the last row.
+            // The only time a sentinel won't exist is if no instance of this ID has ever existed.
+            // The IsCheckpointCompleteProperty was newly added _after_ v1.6.4.
+            DateTime checkpointCompletionTime = DateTime.MinValue;
```
ConnorMcMahon: So this means that for 1.6.4 and previous orchestrations, we are effectively saying we never completed a checkpoint? That means we will rewrite the history, this time writing a value for `CheckpointCompletedTimestamp`, so that when we try to process it again we will have a value?

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -85,21 +95,37 @@ public static int GetEpisodeNumber(IEnumerable<HistoryEvent> historyEvents)
         }
 
         public static int GetTaskEventId(HistoryEvent historyEvent)
+        {
+            if (TryGetTaskScheduledId(historyEvent, out int taskScheduledId))
+            {
+                return taskScheduledId;
+            }
+
+            return historyEvent.EventId;
```
ConnorMcMahon: Is this always equivalent to `-1` like we were returning before, or is this a correction of some previous bug?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1211,9 +1245,10 @@ static string GetBlobName(DynamicTableEntity entity, string property)
             {
                 for (int i = tableResultList.Count - 1; i >= 0; i--)
                 {
-                    if (((DynamicTableEntity)tableResultList[i].Result).RowKey == SentinelRowKey)
+                    DynamicTableEntity resultEntity = (DynamicTableEntity)tableResultList[i].Result;
```
ConnorMcMahon: I realize that this is just a readability fix, so this code has been working already for a long time, but how strong is our guarantee that `tableResultList[i]` is actually a `DynamicTableEntity`? It might be safer to use the `as` operator here and use the `?.` operator in the line below, but if we have strong enough guarantees, then we don't necessarily need to add the extra null checking overhead.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -314,6 +314,7 @@ public void GeneralError(string Account, string TaskHub, string Details, string
             int RequestCount,
             long LatencyMs,
             string ETag,
+            DateTime LastCheckpointTime,
```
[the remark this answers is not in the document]
cgillum: Yes, that's the idea. I've done this before and it has worked out fine.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
cgillum: Yep, that's the idea - sorry, I forgot to delete this before publishing.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
cgillum: Yeah, I also noticed that it was never being read, which is why I removed it. As I was interating on the original design for prefetching I think one version of it required me to know whether a batch was done loading (I relied on polling logic rather than notification queues). It turns out I ultimately didn't need this, but apparently I didn't clean up all the code.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
cgillum: I added this behavior as a bug fix for Sebastian's reminder stress test. The scenario is that a `EventRaised` arrives with no execution ID, followed by a `TimerFired` event, which does have an execution ID. In that case, we have two different batches for the same instance. The result is a false split-brain because the two batches prefetch their own copies of the history, and one succeeds while the other fails.

There is a potential corner case/bug with this approach: if a random message arrives for an old, dead instance (for example, a canceled timer message for a completed orchestration) and if we attach `EventRaised` messages to that batch, then those `EventRaised` messages will get discarded with the old `TimerFired` event. To account for this, I modified our message discarding logic to *not* discard messages that don't have an execution ID unless *all* messages in the batch don't have an execution ID. Instead, those messages will be put back into the prefetch queue so that they can (eventually) be attached to an orchestration that is actually alive.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: I don't know how strong the guarantee is, but I would rather it fail with an `InvalidCastException` than risk a hard-to-find bug related to missing eTags. In fact, now that I'm looking at this code more closely, I would like to remove the surrounding `if (tableResultList != null)` check. I figure if the assumption is ever not valid, we'll find out about it sooner rather than later.

src/DurableTask.AzureStorage/Utils.cs
cgillum: There actually is no change in behavior. I'm simply separating out the existing logic into two methods to make the _intent of the code_ more correct. But yes, in practice I expect this will always be -1.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
ConnorMcMahon: I think looking at the code in `OrchestrationSession` again, I think I understand. We are essentially saying that by setting the `checkpointCompletionTime` to `DateTime.MinValue` for orchestrations created in 1.6.4 or previous, we are never using our new optimistic concurrency approach for these orchestrations.

However, is this a concern considering we removed our old validation approach?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -667,10 +668,42 @@ static TaskHubInfo GetTaskHubInfo(string taskHub, int partitionCount)
 
                     if (!this.IsExecutableInstance(session.RuntimeState, orchestrationWorkItem.NewMessages, out string warningMessage))
                     {
+                        // If all messages belong to the same execution ID, then all of them need to be discarded.
+                        // However, it's also possible to have messages for *any* execution ID batched together with messages
```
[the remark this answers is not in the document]
cgillum: What part exactly do you think is not necessarily true? `IsExecutableInstance()` can mean one of three things:

1. The instance is failed, completed, or terminated
2. The instance doesn't exist (was purged or never existed)
3. We have a history corruption issue in table storage (not expected)

An instance is identified by both the instance ID and the execution ID. All messages have instance IDs but some messages (external events, terminate, etc.) do not have an execution ID. Our message batching logic will arbitrarily pair messages with a particular instance ID and execution ID with messages that have the same instance ID and no execution ID.

This comment is explaining that in those cases, if the Instance ID/Execution ID resolves to a dead instance, we will try again to pair them with a non-dead instance, if one exists (i.e. one with the same instance ID but a different execution ID). If one doesn't exist, we'll discard it in a subsequent go-around.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
cgillum: Correct. This is why we're returning `false` here.

src/DurableTask.AzureStorage/MessageData.cs
cgillum: From a coding perspective, I get what you're saying. My concern is actually about correctness and avoiding misrepresenting something. I found that a value of 0 (the previous default) seemed a bit ambiguous and confusing to me. I made this nullable so I could better distinguish between an actual episode value and a non-value. arguably I could have passed -1 around, but I felt using `.HasValue` was more readable than `== -1` in various places.

src/DurableTask.AzureStorage/README.md
cgillum: It shouldn't cause any race conditions. We will attach the response message to the session. If the session has a chance to consume it, then great. If not, we will add it back to the prefetch buffer and release time so that it can be consumed in the next episode.

src/DurableTask.AzureStorage/README.md
cgillum: I changed "later" to "momentarily" and added a note that this applies to extended sessions.

src/DurableTask.AzureStorage/README.md
cgillum: I updated this a bit.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
cgillum: It turns out we still need this old code. I completely forgot about response messages for activity functions, which could be created before or after the history is updated. I put it back in my most recent iteration.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: I re-added the old validation approach as I found some tests that failed without it. Now we have _both_ the timestamp approach and the task event ID approach.

src/DurableTask.AzureStorage/MessageData.cs
ConnorMcMahon: As long as we only use `-1` when we need to (i.e. logging), I am fine with the new approach, as I agree about the correctness point.

I just want to make sure we don't have two notions of `null` floating around.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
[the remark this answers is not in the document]
ConnorMcMahon: I think your explanation helped me grasp it more.

I guess in case 3, it wasn't clear to me that there could be another execution of the same instance that we could potentially associate these messages with with.

In general, I would prefer to have IsExecutableInstance split into 3 separate checks so that it is more explicit what scenarios we are using this new approach in, but that is likely outside of the scope of this PR.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
ConnorMcMahon: That makes a lot more sense.  I was a bit confused about the logic about checking if all messages don't have an execution id.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
ConnorMcMahon: Correct for what?

I think that the comment is saying that the message is out of order, but the code is saying that the message is not out of order.

cgillum: Oh I see what you're saying now. Yes, this is confusing. I'm adding another sentence saying _"In this case, we see that the checkpoint came *after* the message, so there is no out-of-order concern."_ to make this less confusing.

PR #332, opened 2019-10-29
Refactoring monitor to allow for metrics collection and ad hoc scale decisions.

gzuber:
This is an initial draft of the changes needed for runtime scale in the durable extension. I haven't made any tests yet, just looking for initial feedback.

### DIFF
--- a/src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
+++ b/src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.AzureStorage.Monitoring
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Linq;
     using System.Text;
@@ -80,11 +81,29 @@ namespace DurableTask.AzureStorage.Monitoring
         internal QueueMetricHistory WorkItemQueueLatencies => this.workItemQueueLatencies;
 
         /// <summary>
-        /// Collects and returns a sampling of all performance metrics being observed by this instance.
+        /// Collects and returns a sampling of all performance metrics being observed by this instance as well as a scale
+        /// recommendation.
         /// </summary>
         /// <param name="currentWorkerCount">The number of workers known to be processing messages for this task hub.</param>
-        /// <returns>Returns a performance data summary or <c>null</c> if data cannot be obtained.</returns>
+        /// <returns>Returns a performance data summary with scale recommendation or <c>null</c> if data cannot be obtained.</returns>
         public virtual async Task<PerformanceHeartbeat> PulseAsync(int currentWorkerCount)
+        {
+            var heartbeatPayload = await this.PulseAsync();
+
+            if (heartbeatPayload != null)
+            {
+                heartbeatPayload.ScaleRecommendation = MakeScaleRecommendation(currentWorkerCount);
+            }
+
+            return heartbeatPayload;
+        }
+
+        /// <summary>
+        /// Collects and returns a sampling of all performance metrics being observed by this instance. Will not return a scale
+        /// recommendation.
+        /// </summary>
+        /// <returns>Returns a performance data summary with no scale recommendation or <c>null</c> if data cannot be obtained.</returns>
+        public virtual async Task<PerformanceHeartbeat> PulseAsync()
         {
             if (!await this.UpdateQueueMetrics())
             {
@@ -98,8 +117,7 @@ namespace DurableTask.AzureStorage.Monitoring
                 WorkItemQueueLength = this.currentWorkItemQueueLength,
                 WorkItemQueueLatencyTrend = this.WorkItemQueueLatencies.CurrentTrend,
                 ControlQueueLengths = this.currentControlQueueLengths,
-                ControlQueueLatencies = this.ControlQueueLatencies.Select(h => TimeSpan.FromMilliseconds(h.Latest)).ToList(),
-                ScaleRecommendation = this.MakeScaleRecommendation(currentWorkerCount),
+                ControlQueueLatencies = this.ControlQueueLatencies.Select(h => TimeSpan.FromMilliseconds(h.Latest)).ToList()
             };
 
             return heartbeatPayload;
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -14,10 +14,12 @@
 namespace DurableTask.AzureStorage.Monitoring
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Linq;
     using System.Text;
     using System.Threading.Tasks;
+    //using Dynamitey.DynamicObjects;
```
gzuber: didn't mean to leave this in, I'll be removing this using statement since it doesn't seem to be used and was causing build issues

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -262,17 +276,66 @@ protected virtual async Task<ControlQueueData> GetAggregateControlQueueLengthAsy
...
+
+        /// <summary>
+        /// Calculates a Scale Recommendation based on passed-in performance metrics.
+        /// </summary>
+        /// <param name="workerCount">The number of workers known to be processing messages for this task hub.</param>
+        /// <param name="performanceHeartbeats">Previously collected, chronologically-ordered performance metrics.</param>
+        /// <returns>Returns a scale recommendation</returns>
+        public virtual ScaleRecommendation MakeScaleRecommendation(int workerCount, PerformanceHeartbeat[] performanceHeartbeats)
+        {
+            if (performanceHeartbeats == null)
```
ConnorMcMahon: We can probably expand this to include the case where `performanceHeartbeats.Length` is 0. It prevents needing to check when we calculate partition count, and having an empty array also seems like the "no heartbeat metrics" scenario.

src/DurableTask.AzureStorage/Monitoring/PerformanceHeartbeat.cs
```diff
@@ -25,32 +25,32 @@ public class PerformanceHeartbeat
...
         /// Gets the number of partitions configured in the task hub.
         /// </summary>
-        public int PartitionCount { get; internal set; }
+        public int PartitionCount { get; set; }
 
         /// <summary>
         /// Gets the number of messages across all control queues.
         /// </summary>
-        public IReadOnlyList<int> ControlQueueLengths { get; internal set; }
+        public IReadOnlyList<int> ControlQueueLengths { get; set; }
```
ConnorMcMahon: The removal of `internal` on all of these `set` operations is because the Durable Task extension will need to set these values?

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -262,17 +276,66 @@ protected virtual async Task<ControlQueueData> GetAggregateControlQueueLengthAsy
...
+                return new ScaleRecommendation(ScaleAction.None, keepWorkersAlive: true, reason: "No heartbeat metrics");
+            }
+
+            int partitionCount = performanceHeartbeats.Length > 0 ? performanceHeartbeats.Last().PartitionCount : 0;
+            QueueMetricHistory workItemQueueLatencyHistory = new QueueMetricHistory(QueueLengthSampleSize);
+            List<QueueMetricHistory> controlQueueLatencyHistory = new List<QueueMetricHistory>();
+
+            foreach (PerformanceHeartbeat heartbeat in performanceHeartbeats)
+            {
+                workItemQueueLatencyHistory.Add((int)heartbeat.WorkItemQueueLatency.TotalMilliseconds);
```
ConnorMcMahon: Don't love the downcast from `long` to `int` here, but we likely have bigger problems if the latency of these queues hits 24+ days...

gzuber: Yeah I didn't feel great about this either, but I'm pretty sure [we're always downcasting in UpdateQueueMetrics](https://github.com/Azure/durabletask/blob/9fe0fe04a41b4f3e8d45124df3832fb6438abaf7/src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs#L153)

src/DurableTask.AzureStorage/Monitoring/PerformanceHeartbeat.cs
gzuber: Correct. We have to store metrics data in a table, so when we use that data to call MakeScaleRecommendation we have to create a new PerformanceHeartbeat. Open to suggestions -- we could create another data type, but this seemed easier/cleaner

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
mathewc: Agreed

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -262,17 +276,66 @@ protected virtual async Task<ControlQueueData> GetAggregateControlQueueLengthAsy
...
+
+                while (controlQueueLatencyHistory.Count < heartbeat.ControlQueueLatencies.Count)
```
mathewc: This is a strange while loop, and you're iterating through again below. Can't you just have a single for loop that iterates through all the heartbeat.ControlQueueLatencies and build what you need in one pass?

mathewc: Here you're just auto growing the history as needed based on the number of latencies in each heartbeat?

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -85,6 +87,19 @@ public DisconnectedPerformanceMonitor(string storageConnectionString, string tas
         /// <param name="currentWorkerCount">The number of workers known to be processing messages for this task hub.</param>
         /// <returns>Returns a performance data summary or <c>null</c> if data cannot be obtained.</returns>
         public virtual async Task<PerformanceHeartbeat> PulseAsync(int currentWorkerCount)
+        {
```
mathewc: Consider updating the summary text to indicate that it will also make a scale recommendation

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -85,6 +87,19 @@ public DisconnectedPerformanceMonitor(string storageConnectionString, string tas
...
+            var heartbeatPayload = await this.PulseAsync();
+
+            heartbeatPayload.ScaleRecommendation = MakeScaleRecommendation(currentWorkerCount);
+
+            return heartbeatPayload;
+        }
+
+        /// <summary>
+        /// Collects and returns a sampling of all performance metrics being observed by this instance.
```
mathewc: Clarify that this overload WONT be making a scale recommendation

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
gzuber: Yeah the rationale was to account for the number of control queues increasing between heartbeats, but it can be simplified/done in one loop

PR #333, opened 2019-10-30
Change Azure Storage SDK back to previous version

ConnorMcMahon:
A previous change upgraded the Azure Storage SDK from 8.6 to 9.3.3. This
changes the version to 9.3.1, as 9.3.3 broke compatibility between the Durable
Functions extension and the Azure Storage extension. This PR also adds a
comment next to the dependency in the csproj to warn against changing
this in the future.


cgillum:
I noticed that the latest version of [Microsoft.Azure.WebJobs.Extensions.Storage](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.Storage/3.0.10) has a dependency on [WindowsAzyre.Storage 9.3.1](https://www.nuget.org/packages/WindowsAzure.Storage/9.3.1), but you're changing our dependency back to 8.6. Do we not need to be on the same version?

ConnorMcMahon:
Tested 9.3.1 with the Storage extension and it worked, so I will stick to that version.

PR #338, opened 2019-11-07
Added Debugger Check to TimeoutHandler

### LINKED ISSUE #336: Disable deadlock detection when debugger is attached
I noticed that when I'm debugging, hit a breakpoint, and don't resume from that breakpoint for over 2 minutes, I might accidentally trigger the new FailFast logic that was added.

I suggest we use the `Debugger.IsAttached` static property to skip over the FailFast logic when the debugger is attached.

amdeel:
Added check to the TimeoutHandler to prevent FailFast when using a debugger.

resolves #336

### DIFF
--- a/src/DurableTask.AzureStorage/TimeoutHandler.cs
+++ b/src/DurableTask.AzureStorage/TimeoutHandler.cs
@@ -26,6 +26,12 @@ namespace DurableTask.AzureStorage
 
         public static async Task<T> ExecuteWithTimeout<T>(string operationName, string clientRequestId, string account, string taskHub, Func<Task<T>> operation)
         {
+            if (Debugger.IsAttached)
+            {
+                // ignore long delays while debugging
+                return await operation();
+            }
+
             using (var cts = new CancellationTokenSource())
             {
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/TimeoutHandler.cs
```diff
@@ -32,7 +33,7 @@ public static async Task<T> ExecuteWithTimeout<T>(string operationName, string c
 
                 Task completedTask = await Task.WhenAny(timeoutTask, operationTask);
 
-                if (Equals(timeoutTask, completedTask))
+                if (Equals(timeoutTask, completedTask) && !Debugger.IsAttached)
```
cgillum: I recommend we bypass all of this code if the debugger is attached, rather than just bypassing the FailFast part. Something like this at the very top of the method:

```csharp
if (Debugger.IsAttached)
{
    // ignore long delays while debugging
    return await operation();
}
```

This will reduce the complexity of the code, and will avoid corner case timing issues where we were debugging for a while and then detached right around the time of the two minute expiration.

amdeel: That makes sense, thanks!

PR #339, opened 2019-11-08
fixes for handling timers and extended sessions

sebastianburckhardt:
Fixes two bugs in how timers are handled. These fixes are necessary for implementing the entity reminder feature azure/azure-functions-durable-extension#716.

- the `DeferredMessages` mechanism introduced in #331 was not functioning correctly with extended sessions, as the deferred messages, even if ready to deliver, would just sit in the session until the session times out. I updated the code in `CompleteTaskOrchestrationWorkItemAsync` so that it redelivers deferred messages whenever the execution id of the session changed.

- I limited the autostart mechanism so it applies only to events. Otherwise, a timer message to a not-yet-existing execution of an entity can autostart a new one, causing unnecessary and harmful conflicts. This bug was already fixed in #327 but the latter PR is currently on ice, so I copied the fix into this PR.

Also, I slightly simplified the code by moving the `UpdateRuntimeState` call into `CompleteTaskOrchestrationWorkItemAsync`. This makes more sense to me and means we don't have to expose it through the `IOrchestrationSession` interface.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -906,9 +906,9 @@ namespace DurableTask.AzureStorage
                 if (session.DeferredMessages.Count > 0
                     && executionId != workItem.OrchestrationRuntimeState.OrchestrationInstance?.ExecutionId)
                 {
-                    var messages = session.DeferredMessages.ToArray();
+                    var messages = session.DeferredMessages.ToList();
                     session.DeferredMessages.Clear();
-                    this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, default(Guid), CancellationToken.None);
+                    this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, session.TraceActivityId, CancellationToken.None);
                 }
             }
             catch (Exception e)
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -801,7 +801,9 @@ namespace DurableTask.AzureStorage
             {
                 var instanceId = newMessages[0].OrchestrationInstance.InstanceId;
 
-                if (instanceId.StartsWith("@"))
+                if (instanceId.StartsWith("@")
+                    && newMessages[0].Event.EventType == EventType.EventRaised
+                    && newMessages[0].OrchestrationInstance.ExecutionId == null)
                 {
                     // automatically start this instance
                     var orchestrationInstance = new OrchestrationInstance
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -896,6 +898,18 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
                 session.ETag = await this.trackingStore.UpdateStateAsync(runtimeState, workItem.OrchestrationRuntimeState, instanceId, executionId, session.ETag);
+
+                // update the runtime state and execution id stored in the session
+                session.UpdateRuntimeState(runtimeState);
+
+                // if we deferred some messages, and the execution id of this instance has changed, redeliver them
+                if (session.DeferredMessages.Count > 0
+                    && executionId != workItem.OrchestrationRuntimeState.OrchestrationInstance?.ExecutionId)
+                {
+                    var messages = session.DeferredMessages.ToArray();
```
cgillum: In my previous PR, one of the issues I ran into was the fact that `ToArray()` creates a fixed-size `IList<T>`, but there are some places where we actually try to mutate the list of messages. The fix was to replace usage of `ToArray()` with `ToList()`. I suggest we do the same here to avoid these types of problems. `ToList()` is also [a bit faster than `ToArray`](https://stackoverflow.com/a/16323412/2069).

It also seems like we don't even need to create this copy. You could just pass `session.DeferredMessages` as the second parameter for `orchestrationSessionManager.AddMessageToPendingOrchestration` and clear the list afterwards. That would be the most efficient.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -896,6 +898,18 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
+                    session.DeferredMessages.Clear();
+                    this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, default(Guid), CancellationToken.None);
```
cgillum: I suggest we use `session.TraceActivityId` here instead of `default(Guid)`. It will help with debugging.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -426,8 +426,6 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
 
             workItem.OrchestrationRuntimeState = runtimeState;
 
-            workItem.Session?.UpdateRuntimeState(runtimeState);
```
cgillum: Will this cause extended sessions to work incorrectly for other providers who depend on it? I don't believe other providers currently do, but I'm curious to know if this PR effectively creates a modified contract for implementing it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: I do need to make a copy because when `AddMessageToPendingOrchestration` executes, it may add the message back into `session.DeferredMessages` (the very same list), so if I clear it afterwards the message is lost.

Using `ToList` instead of `ToArray` is fine. It does not really matter here. I thought the latter was faster, that's why I used it.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: Ok. I believe the value is actually never used (because the path inside `AddMessageToPendingOrchestration` will always find an existing session), but I agree it is preferrable to not have an invalid guid there.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
sebastianburckhardt: Yes, it modifies (improves) the contract. Implementations should always implicitly update the sessions when completing a work item. Relying on a separate call to do that is needlessly complicated.

I don't think this will break anything - AFAIK no other implementation uses it. I added it at some point to v2, so it wasn't even in master until very recently.

PR #345, opened 2019-12-06
Support for delayed entity messages and improved split brain handling

sebastianburckhardt:
This PR adds support for delayed entity messages, by recognizing a specially encoded delay parameter in event names, and setting the visibility timeout. This affects only entity messages, and does not require changes to core, only to the Azure table storage provider.

This PR also improves split brain handling in two cases:
- detects conflict on concurrent creation of an instance
- terminates extended session when conflict is detected

The latter required changing `DurableTask.Core` because the exception that terminates the extended session needs to be caught, so the work item scheduler does not back off.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
@@ -176,13 +176,14 @@ namespace DurableTask.AzureStorage.Messaging
             if (taskMessage.Event is EventRaisedEvent eventRaisedEvent
                 && taskMessage.OrchestrationInstance.InstanceId[0] == '@')
             {
-                // Since all events targeting entities are named by the runtime, not the user, we can follow a custom 
-                // naming convention to pass a time parameter.
+                // We assume that auto-started orchestrations (i.e. instance ids starting with '@')
+                // are used exclusively by durable entities; so we can follow
+                // a custom naming convention to pass a time parameter.
                 var eventName = eventRaisedEvent.Name;
-                if (eventName.Length >= 3 && eventName[2] == '@')
+                if (eventName.Length >= 3 && eventName[2] == '@'
+                    && DateTime.TryParse(eventRaisedEvent.Name.Substring(3), out var scheduledTime))
                 {
-                    var scheduledUtc = DateTime.Parse(eventRaisedEvent.Name.Substring(3)).ToUniversalTime();
-                    initialVisibilityDelay = scheduledUtc - DateTime.UtcNow;
+                    initialVisibilityDelay = scheduledTime.ToUniversalTime() - DateTime.UtcNow;
                     if (initialVisibilityDelay < TimeSpan.Zero)
                     {
                         initialVisibilityDelay = TimeSpan.Zero;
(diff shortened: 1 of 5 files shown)


src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -172,6 +172,24 @@ async Task<MessageData> AddMessageAsync(TaskMessage taskMessage, OrchestrationIn
                 }
             }
 
+            // Special functionality for entity messages with a delivery delay 
+            if (taskMessage.Event is EventRaisedEvent eventRaisedEvent
+                && taskMessage.OrchestrationInstance.InstanceId[0] == '@')
+            {
+                // Since all events targeting entities are named by the runtime, not the user, we can follow a custom 
+                // naming convention to pass a time parameter.
```
cgillum: This is not actually true, right? The code [here](https://github.com/Azure/durabletask/blob/7f63415ec285caad8eed903a26f07d35aa27114b/src/DurableTask.Core/TaskHubClient.cs#L514) implies that users can actually set this from a `TaskHubClient`.  I'm personally not that worried about others running into this, but we should update this comment.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -172,6 +172,24 @@ async Task<MessageData> AddMessageAsync(TaskMessage taskMessage, OrchestrationIn
...
+                var eventName = eventRaisedEvent.Name;
+                if (eventName.Length >= 3 && eventName[2] == '@')
+                {
+                    var scheduledUtc = DateTime.Parse(eventRaisedEvent.Name.Substring(3)).ToUniversalTime();
```
cgillum: I suggest we use `DateTime.TryParse` instead of `DateTime.Parse` to gracefully bail out of the visibility delay logic and further minimize the chance that some user's event name will conflict with this new logic.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1205,7 +1205,7 @@ static string GetBlobName(DynamicTableEntity entity, string property)
             }
             else
             {
-                historyEventBatch.InsertOrReplace(sentinelEntity);
+                historyEventBatch.Insert(sentinelEntity);
```
cgillum: I think you mentioned to me that this isn't actually needed, or am I mis-remembering?

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
sebastianburckhardt: Yeah, sure.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
sebastianburckhardt: ok.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: No, it is definitely needed for handling split-brain correctly (I found failing testcases that were caused by this issue).

What is not needed is some special logic to handle situations where users are replacing an old instance with a new instance. `Insert` appears to work just fine even in that case... the four 'RecreateInstance' tests are all passing.

PR #346, opened 2019-12-09
Added InstanceIdPrefix to OrchestrationInstanceStatusQueryCondition

amdeel:
Added InstanceIdPrefix to allow query of an instance only given the first part of the instanceId. For example, with Entities, this allows query of EntityName (where instance id is composed of EntityName and EntityKey).

Related to issue https://github.com/Azure/azure-functions-durable-extension/issues/997
In the azure-functions-durable-extension repository.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
+++ b/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
@@ -61,7 +61,8 @@ namespace DurableTask.AzureStorage.Tracking
             if (!((this.RuntimeStatus == null || (!this.RuntimeStatus.Any())) && 
                 this.CreatedTimeFrom == default(DateTime) && 
                 this.CreatedTimeTo == default(DateTime) &&
-                this.TaskHubNames == null))
+                this.TaskHubNames == null &&
+                InstanceIdPrefix == null))
             {
                 query.Where(this.GetConditions());
             }
--- a/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
+++ b/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
@@ -45,6 +45,11 @@ namespace DurableTask.AzureStorage.Tracking
         public IEnumerable<string> TaskHubNames { get; set; }
 
         /// <summary>
+        /// InstanceIdPrefix
+        /// </summary>
+        public string InstanceIdPrefix { get; set; }
+
+        /// <summary>
         /// Get the TableQuery object
         /// </summary>
         /// <typeparam name="T"></typeparam>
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -98,6 +104,19 @@ string GetConditions()
                 }
             }
 
+            if (this.InstanceIdPrefix != null)
```
cgillum: Prefer `if (!string.IsNullOrEmpty(this.InstanceIdPrefix))` instead of `if (this.InstanceIdPrefix != null)`. Semantically, I think the empty string should be treated the same as `null`. This also avoids the `IndexOutOfRangeException` that you would hit if someone passed in the empty string as the prefix.

amdeel: Good point!

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -98,6 +104,19 @@ string GetConditions()
...
+            {
+                int length = InstanceIdPrefix.Length - 1;
+                int incrementedLastChar = InstanceIdPrefix[length] + 1;
```
cgillum: nit: I suggest making this a `char` instead of an `int`. That will make the intent more clear.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -98,6 +104,19 @@ string GetConditions()
                 }
             }
 
+            if (this.InstanceIdPrefix != null)
+            {
+                int length = InstanceIdPrefix.Length - 1;
```
cgillum: code-style nit: prefix all calls to `InstanceIdPrefix` with `this.`.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -98,6 +104,19 @@ string GetConditions()
...
+                int incrementedLastChar = InstanceIdPrefix[length] + 1;
+
+                string greaterThanPrefix = InstanceIdPrefix.Substring(0, length) + (char)incrementedLastChar;
+
+                conditions.Add(TableQuery.CombineFilters(
+                    TableQuery.GenerateFilterCondition("InstanceId", QueryComparisons.GreaterThanOrEqual, InstanceIdPrefix), 
+                    TableOperators.And, 
+                    TableQuery.GenerateFilterCondition("InstanceId", QueryComparisons.LessThan, greaterThanPrefix)));
```
cgillum: Should this be `"Partitionkey"` instead of `"InstanceId"`? I don't think we have a column named InstanceId in any of our tables.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -98,6 +104,19 @@ string GetConditions()
                 }
             }
 
+            if (!string.IsNullOrEmpty(this.InstanceIdPrefix))
+            {
+                int length = this.InstanceIdPrefix.Length - 1;
+                char incrementedLastChar = (char)(this.InstanceIdPrefix[length] + 1);
```
cgillum: With this logic, is it possible for us to end up with a `char` for `incrementedLastChar` that is not legal?  I'm looking at [this ASCII table](https://www.cs.cmu.edu/~pattis/15-1XX/common/handouts/ascii.html) and I'm wondering if a failure or weird behavior could occur if we create a Table Storage query that ends with a `DEL` character (`(char)127`), for example.

amdeel: No, when char 126 (~) is incremented to 127 (DEL) the unicode representation of DEL will be used and the query will still work as expected.

PR #347, opened 2019-12-11
Fix instance status update race condition (for replace)

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/1020.

We were already handling the race condition for setting the "Pending" status on a new instance.  However, we did not correctly handle the case where we were replacing an existing instance with a "Pending". This resulted in cases where orchestration instances completed successfully but had a "Pending" status in the instances table.

This fix handles the latter race condition (replacing an existing instance). After this, both cases should be handled, and we should never have a completed instance in the "Pending" status.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -478,6 +478,19 @@ namespace DurableTask.AzureStorage.Tracking
         /// <inheritdoc />
         public override async Task<OrchestrationState> GetStateAsync(string instanceId, string executionId, bool fetchInput)
         {
+            InstanceStatus instanceStatus = await this.FetchInstanceStatusInternalAsync(instanceId, executionId, fetchInput);
+            return instanceStatus?.State;
+        }
+
+        /// <inheritdoc />
+        public override Task<InstanceStatus> FetchInstanceStatusAsync(string instanceId)
+        {
+            return this.FetchInstanceStatusInternalAsync(instanceId, executionId: null, fetchInput: false);
+        }
+
+        /// <inheritdoc />
+        async Task<InstanceStatus> FetchInstanceStatusInternalAsync(string instanceId, string executionId, bool fetchInput)
+        {
             if (instanceId == null)
             {
                 throw new ArgumentNullException(nameof(instanceId));
@@ -514,7 +527,8 @@ namespace DurableTask.AzureStorage.Tracking
                 return null;
             }
 
-            return await this.ConvertFromAsync(orchestrationInstanceStatus, instanceId);
+            OrchestrationState state = await this.ConvertFromAsync(orchestrationInstanceStatus, instanceId);
+            return state != null ? new InstanceStatus(state, orchestration.Etag) : null;
         }
 
         async Task<OrchestrationState> ConvertFromAsync(OrchestrationInstanceStatus orchestrationInstanceStatus, string instanceId)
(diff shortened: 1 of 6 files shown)


src/DurableTask.AzureStorage/Tracking/InstanceStatus.cs
```diff
@@ -0,0 +1,35 @@
...
+    class InstanceStatus
+    {
+        public InstanceStatus(OrchestrationState state)
+            : this(state, null)
+        { }
+
+        public InstanceStatus(OrchestrationState state, string eTag)
+        {
+            this.State = state ?? throw new ArgumentNullException(nameof(state));
+            this.ETag = eTag ?? "*";
```
ConnorMcMahon: Is "*" a specialized case for azure table eTags?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -814,16 +829,20 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTim
             Stopwatch stopwatch = Stopwatch.StartNew();
             try
             {
-                if (ignoreExistingInstances)
+                if (eTag == null)
```
ConnorMcMahon: So to clarify, the only time `eTag` is null is if there wasn't an existing eTag, because all existing orchestrations will have eTags (or at least the dummy value of `*` due to our constructor).

src/DurableTask.AzureStorage/Tracking/InstanceStatus.cs
cgillum: Yeah, it basically means "ignore the eTag check"

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: More specifically this means no row existed in the Instances table for the specified instance ID.

madhupriyatamPV:
Hey @ConnorMcMahon, we are facing the same issue as well. One of our orchestration instances is stuck in Pending state and the orchestration trigger is not letting us create another instance since one of them is pending. How can we handle this?

This is the code of our orchestration trigger:

`private static bool ShouldStartNewInstance(DurableOrchestrationStatus status)
        {
            return status == null ||
                status.RuntimeStatus == OrchestrationRuntimeStatus.Canceled ||
                status.RuntimeStatus == OrchestrationRuntimeStatus.Failed ||
                status.RuntimeStatus == OrchestrationRuntimeStatus.Terminated ||
                status.RuntimeStatus == OrchestrationRuntimeStatus.Completed ||
                status.RuntimeStatus == OrchestrationRuntimeStatus.Unknown;
        }

        private async Task<IActionResult> StartScheduler(
            IDurableOrchestrationClient starter,
            ILogger log,
            SchedulerConfiguration config,
            string functionName)
        {
            var instanceStatus = await starter.GetStatusAsync(config.SchedulerName, false, false, false);
            bool shouldStartNew = ShouldStartNewInstance(instanceStatus), isForceRestart = false;
            if (!shouldStartNew && config.ForceRestart)
            {
                await starter.TerminateAsync(config.SchedulerName, "user kill");
                shouldStartNew = true;
                isForceRestart = true;
                // Delay a while to make sure staus get updated
                await Task.Delay(5 * 1000);
                log.LogInformation($"Shutting down durable function by forceRestart {config.SchedulerName}");
            }

            if (shouldStartNew)
            {
                await starter.StartNewAsync(functionName, config.SchedulerName, config);
                log.LogInformation($"Started {functionName} with ID = '{config.SchedulerName} with force={isForceRestart}, config = '{JsonConvert.SerializeObject(config)}'.");
                return new OkObjectResult($"Started {functionName} with ID = '{config.SchedulerName} with force={isForceRestart}");
            }
            else
            {
                return new ConflictObjectResult(
                    $"An instance of ScheduleLowPriorityEvents with ID '{config.SchedulerName}' already exists. " +
                    $"Status: {instanceStatus.RuntimeStatus}");
            }
        }`

davidmrdavid:
Hi @madhupriyatamPV: it would be best to create a new, separate, issue for this.
If this is a Durable Functions issue, please create it here: https://github.com/Azure/azure-functions-durable-extension
If this a DurableTask-level issue, then you're in the right repo but please create a separate ticket :)

PR #348, opened 2019-12-11
Added constructor to TaskHubClient with JsonDataConverter input

amdeel:
Added second constructor to TaskHubClient to allow for a custom JsonDataConverter to allow for custom serialization settings.

Related to this issue in the azure-functions-durable-extension repository:
https://github.com/Azure/azure-functions-durable-extension/issues/65

### DIFF
--- a/src/DurableTask.Core/TaskHubClient.cs
+++ b/src/DurableTask.Core/TaskHubClient.cs
@@ -44,14 +44,14 @@ namespace DurableTask.Core
         }
 
         /// <summary>
-        ///     Create a new TaskHubClient with the given OrchestrationServiceClient
+        ///     Create a new TaskHubClient with the given OrchestrationServiceClient and JsonDataConverter.
         /// </summary>
         /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
-        /// <param name="dataConverter"></param>
+        /// <param name="dataConverter">The <see cref="JsonDataConverter"/> to use for message serialization.</param>
         public TaskHubClient(IOrchestrationServiceClient serviceClient, JsonDataConverter dataConverter)
         {
             ServiceClient = serviceClient ?? throw new ArgumentNullException(nameof(serviceClient));
-            this.defaultConverter = dataConverter;
+            this.defaultConverter = dataConverter ?? throw new ArgumentNullException(nameof(dataConverter));
         }
 
         /// <summary>
--- a/src/DurableTask.Core/TaskHubClient.cs
+++ b/src/DurableTask.Core/TaskHubClient.cs
@@ -44,6 +44,17 @@ namespace DurableTask.Core
         }
 
         /// <summary>
+        ///     Create a new TaskHubClient with the given OrchestrationServiceClient
+        /// </summary>
+        /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
+        /// <param name="dataConverter"></param>
+        public TaskHubClient(IOrchestrationServiceClient serviceClient, JsonDataConverter dataConverter)
+        {
+            ServiceClient = serviceClient ?? throw new ArgumentNullException(nameof(serviceClient));
+            this.defaultConverter = dataConverter;
+        }
+
+        /// <summary>
         ///     Create a new orchestration of the specified type with an automatically generated instance id
         /// </summary>
         /// <param name="orchestrationType">Type that derives from TaskOrchestration</param>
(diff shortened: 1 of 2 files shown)


tools/DurableTask.props
```diff
@@ -43,9 +43,9 @@
   <!-- Nuget Package Settings -->
   <PropertyGroup>
     <PackageOutputPath>..\..\build_output\packages</PackageOutputPath>
-    <AssemblyVersion>2.1.3</AssemblyVersion>
-    <FileVersion>2.1.3</FileVersion>
-    <Version>2.1.3</Version>
+    <AssemblyVersion>2.1.4</AssemblyVersion>
+    <FileVersion>2.1.4</FileVersion>
+    <Version>2.1.4</Version>
```
cgillum: Since this is a new feature (and since some of our other changes could be considered new features), it's best to increment the minor version number. So 2.2.0 instead of 2.1.4.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -43,6 +43,17 @@ public TaskHubClient(IOrchestrationServiceClient serviceClient)
...
 
+        /// <summary>
+        ///     Create a new TaskHubClient with the given OrchestrationServiceClient
+        /// </summary>
+        /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
+        /// <param name="dataConverter"></param>
+        public TaskHubClient(IOrchestrationServiceClient serviceClient, JsonDataConverter dataConverter)
+        {
+            ServiceClient = serviceClient ?? throw new ArgumentNullException(nameof(serviceClient));
+            this.defaultConverter = dataConverter;
```
cgillum:
```suggestion
            this.defaultConverter = dataConverter ?? throw new ArgumentNullException(nameof(dataConverter));;
```

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -43,6 +43,17 @@ public TaskHubClient(IOrchestrationServiceClient serviceClient)
             this.defaultConverter = new JsonDataConverter();
         }
 
+        /// <summary>
+        ///     Create a new TaskHubClient with the given OrchestrationServiceClient
+        /// </summary>
+        /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
+        /// <param name="dataConverter"></param>
```
cgillum:
```suggestion
        /// <param name="dataConverter">The <see cref="JsonDataConverter"/> to use for message serialization.</param>
```

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -43,6 +43,17 @@ public TaskHubClient(IOrchestrationServiceClient serviceClient)
             this.defaultConverter = new JsonDataConverter();
         }
 
+        /// <summary>
+        ///     Create a new TaskHubClient with the given OrchestrationServiceClient
```
cgillum:
```suggestion
        ///     Create a new TaskHubClient with the given OrchestrationServiceClient and JsonDataConverter
```

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -43,6 +43,17 @@ public TaskHubClient(IOrchestrationServiceClient serviceClient)
             this.defaultConverter = new JsonDataConverter();
         }
 
+        /// <summary>
+        ///     Create a new TaskHubClient with the given OrchestrationServiceClient and JsonDataConverter.
```
ConnorMcMahon: nit: I know this follows the pattern of the other constructors, but I don't love the tab at the beginning of the summary.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -43,6 +43,17 @@ public TaskHubClient(IOrchestrationServiceClient serviceClient)
...
+        /// </summary>
+        /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
+        /// <param name="dataConverter">The <see cref="JsonDataConverter"/> to use for message serialization.</param>
```
ConnorMcMahon: nit: Maybe change this to "The default <see cref="JsonDataConverter"/> to use for message serialization.", just so it is more clear that this is only the default, and not used for all message serialization.

src/DurableTask.Core/TaskHubClient.cs
cgillum: I suggest that we leave this as-is for the sake of consistency with the existing code base.

src/DurableTask.Core/TaskHubClient.cs
cgillum: I think the name `defaultConverter` in the existing DTFx code is actually misleading. It's not actually the "default". Rather, it's the _only_ converter used for serialization. The naming is probably some artifact of an aspiration they once had to make it extensible. In that light, I think Amanda's description is actually more correct as written.

PR #349, opened 2019-12-13
Added FetchInput property to OrchestrationInstanceStatusQueryCondition

amdeel:
Added FetchInput property to allow to specify if the Input column should be selected in the query created with OrchestrationInstanceStatusQueryCondition.

Related to https://github.com/Azure/azure-functions-durable-extension/issues/997
to allow query of entities without selecting the input column.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
+++ b/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
@@ -54,6 +54,28 @@ namespace DurableTask.AzureStorage.Tracking
         /// </summary>
         public bool FetchInput { get; set; } = true;
 
+        private List<string> columnsWithoutInput;
+
+        private List<string> ColumnsWithoutInput
+        {
+            get
+            {
+                if (this.columnsWithoutInput == null)
+                {
+                    this.columnsWithoutInput = typeof(OrchestrationInstanceStatus).GetProperties()
+                        .Where(prop => !prop.Name.Equals(nameof(OrchestrationInstanceStatus.Input)))
+                        .Select(prop => prop.Name)
+                        .ToList();
+                }
+
+                return this.columnsWithoutInput;
+            }
+            set
+            {
+                this.columnsWithoutInput = value;
+            }
+        }
+
         /// <summary>
         /// Get the TableQuery object
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -49,6 +49,11 @@ public class OrchestrationInstanceStatusQueryCondition
         /// </summary>
         public string InstanceIdPrefix { get; set; }
 
+        /// <summary>
+        /// If true, the input will be returned with the query
```
cgillum:
```suggestion
        /// If true, the input will be returned with the results. The default value is true.
```

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -64,6 +69,12 @@ public TableQuery<T> ToTableQuery<T>()
                 this.TaskHubNames == null &&
                 this.InstanceIdPrefix == null))
             {
+                if (!FetchInput)
```
cgillum:
```suggestion
                if (!this.FetchInput)
```

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -64,6 +69,12 @@ public TableQuery<T> ToTableQuery<T>()
...
+                {
+                    IList<string> columns = new List<string> { "ExecutionId", "Name", "Version", "Output", "CustomStatus", "CreatedTime", "LastUpdatedTime", "RuntimeStatus" };
```
cgillum: I'm really worried about hardcoding the list of columns like this because it would be really easy for us to forget things or for it to get out of sync as we evolve the code over time. Can you instead get the list of column names in a more dynamic way, like we do [here](https://github.com/Azure/durabletask/blob/51d16b4a92a06873d1f81d363bfc7076788608b8/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L489)?

```csharp
List<string> columns = typeof(OrchestrationInstanceStatus).GetProperties()
    .Where(prop => !prop.Name.Equals(nameof(OrchestrationInstanceStatus.Input)))
    .Select(prop => prop.Name)
    .ToList();
```

_(I actually like this proposal a little better than the existing code I linked to)_

ConnorMcMahon: nit: Since that approach does use reflection, it might be a good idea to only calculate the value of `columns` once and store it in a variable for later use. It's probably not a huge problem, given that this code path should not be hit frequently, but it is something to be aware of.

PR #351, opened 2019-12-17
Fixed capitalization mistake in OrchestrationInstanceStatusQueryCondition

amdeel:


src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -93,7 +93,7 @@ public TableQuery<T> ToTableQuery<T>()
             {
                 if (!this.FetchInput)
                 {
-                    query.Select(columnsWithoutInput);
+                    query.Select(ColumnsWithoutInput);
```
cgillum: Actually, I don't think we should use a property here. Instead, I think it should be a `static readonly` field, like this:

```csharp
private static readonly List<string> ColumnsWithoutInput = typeof(OrchestrationInstanceStatus).GetProperties()
    .Where(prop => !prop.Name.Equals(nameof(OrchestrationInstanceStatus.Input)))
    .Select(prop => prop.Name)
    .ToList();
```

That way we never have to generate it more than once.

PR #352, opened 2019-12-17
Improve speed of graceful shutdown, and remove race condition

ConnorMcMahon:
Improves graceful shutdown the following ways:

1. Stop the dispatchers and orchestration service in parallel to save time.
2. Reduce delay from 4 seconds to 1 second when waiting for inflight work items.
3. Fix race condition where we could accidentally mark a WorkItemDispatcher as stopped when there are in flight work items.

### DIFF
--- a/src/DurableTask.Core/TaskHubWorker.cs
+++ b/src/DurableTask.Core/TaskHubWorker.cs
@@ -156,9 +156,14 @@ namespace DurableTask.Core
             {
                 if (this.isStarted)
                 {
-                    await this.orchestrationDispatcher.StopAsync(isForced);
-                    await this.activityDispatcher.StopAsync(isForced);
-                    await orchestrationService.StopAsync(isForced);
+                    var serviceShutdowns = new Task[]
+                    {
+                        this.orchestrationDispatcher.StopAsync(isForced),
+                        this.activityDispatcher.StopAsync(isForced),
+                        orchestrationService.StopAsync(isForced)
+                    };
+
+                    await Task.WhenAll(serviceShutdowns);
 
                     this.isStarted = false;
                 }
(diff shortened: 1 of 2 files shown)


src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -156,9 +156,14 @@ public async Task StopAsync(bool isForced)
             {
                 if (this.isStarted)
                 {
-                    await this.orchestrationDispatcher.StopAsync(isForced);
-                    await this.activityDispatcher.StopAsync(isForced);
-                    await orchestrationService.StopAsync(isForced);
+                    var serviceShutdowns = new Task[]
```
ConnorMcMahon: These may have been sequential for a reason, but I don't see why they had to be.

src/DurableTask.Core/WorkItemDispatcher.cs
```diff
@@ -173,11 +173,10 @@ public async Task StopAsync(bool forced)
                 if (!forced)
                 {
                     var retryCount = 7;
-                    //TODO: race condition in graceful shutdown.
-                    while ((this.concurrentWorkItemCount > 0 || this.activeFetchers > 0) && retryCount-- >= 0)
```
ConnorMcMahon: Before, it was theoretically possible that between the two conditional checks, this.activeFetchers went from 1 to 0, but added more concurrent work items before hitting 0. I reordered the conditional in `AllWorkItemsCompleted()` to prevent this case.

src/DurableTask.Core/TaskHubWorker.cs
cgillum: I do wonder if it would be better to make this sequential, or at least partially sequential. I do worry about what could happen if the orchestration service shuts down before the dispatcher does.

ConnorMcMahon: Made this semi-sequential as suggested.

PR #353, opened 2019-12-18
Misc updates for tracing and packaging

cgillum:
- Added more detail to dequeue tracing
- Fix nuget warnings in package generation
- Upgrade to non-preview version of Microsoft.SourceLink.GitHub
- Minor code cleanup

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -19,7 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.*" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
--- a/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
+++ b/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
@@ -96,6 +96,8 @@ namespace DurableTask.AzureStorage.Messaging
                 return null; // timed-out
             }
 
+            this.StartNewLogicalTraceScope();
+
             lock (this.nextMessageBatch)
             {
                 this.CurrentMessageBatch = this.nextMessageBatch.ToList();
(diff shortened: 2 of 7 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -19,7 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.*" PrivateAssets="All" />
```
ConnorMcMahon: Any specific reason for this change in versioning?

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFrameworks>netstandard2.0;net451</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageId>Microsoft.Azure.DurableTask.Core</PackageId>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
```
ConnorMcMahon: Is this something we want to commit?

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -96,6 +96,8 @@ public void AddOrReplaceMessages(IEnumerable<MessageData> messages)
                 return null; // timed-out
             }
 
+            this.StartNewLogicalTraceScope();
```
ConnorMcMahon: This is just to make sure we aren't missing activity ids for logs here?

cgillum: Right. When I was debugging that performance issue, I noticed we were missing a bunch of Activity IDs for the extended sessions case. This line of code fixes that.

src/DurableTask.Core/DurableTask.Core.csproj
cgillum: Not sure, I kind of wish I'd left that out.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
cgillum: `1.0.0-*` mapped to preview versions, like `1.0.0-beta2-19554-01`. There is a `1.0.0` version out now so I wanted to update our reference to get that build.

PR #358, opened 2019-12-20
[AzureServiceFabric] Provide support to bind DTFX on Https port.

kamesh-msft:


shankarsama:
>                 Thread.Sleep(Timeout.Infinite);

"Finally" is never called and unbind will not happen on the port. When the process is killed by Service Fabric, we need to listen Control+break and unbind the port. Add a TODO here at least.

---
Refers to: Test/TestFabricApplication/TestApplication.StatefulService/Program.cs:50 in c6ad652. [](commit_id = c6ad652176e08d2c2b37c9fd2968e572c7d3697e, deletion_comment = False)

PR #359, opened 2020-01-03
Support for aborting orchestration and activity executions

cgillum:
This is a followup item to help fix https://github.com/Azure/azure-functions-durable-extension/issues/472 but it also directly applies to https://github.com/Azure/azure-functions-durable-extension/issues/1133.

### Summary

The idea is to take the existing `SessionAbortedException` (added by Sebastian) and make it work for abandoning in-flight orchestrations and activities. Basically, if an orchestration or activity function throws this exception, the dispatcher will abandon the work item rather than failing the activity/orchestration. The work item can then be picked up by a healthy instance and immediately resumed.

### Purpose

The primary purpose of this feature is to enable correctly handling host shutdown scenarios. For example, in Durable Functions, when the Functions host shuts down in the middle of an activity execution that is potentially running for a long time, we may see `ObjectDisposedException` from WebJobs framework when attempting to invoke the function. With this feature, we can detect host shutdown and throw a `SessionAbortedException` to abandon the current execution and let it restart gracefully on another instance.  Previously errors related to shutdown caused the orchestration to go into a failed state, which is not the correct behavior.

### Example

The following example shows how this could be used:

```csharp
public class LongRunningActivity : TaskActivity<string, string>
{
    protected override string Execute(TaskContext context, string input)
    {
        // do some work
        // ...

        while (!done)
        {
            if (HostIsShuttingDown())
            {
                // abort the execution so we can restart on another worker
                throw new SessionAbortedException();
            }

            // continue to do work
            // ...
        }

        return "done";
    }
}
```

### Validation tasks
- [x] Validate activity usage
- [x] Validate orchestration usage w/extended sessions
- [x] Validate orchestration usage w/out extended sessions
- [x] Validate split-brain handling still works as expected

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -16,9 +16,9 @@ namespace DurableTask.AzureStorage.Tests
     using System;
     using System.Collections.Generic;
     using System.Configuration;
+    using System.Diagnostics;
     using System.IO;
     using System.Linq;
-    using System.Reflection;
     using System.Runtime.Serialization;
     using System.Text;
     using System.Threading;
@@ -31,12 +31,13 @@ namespace DurableTask.AzureStorage.Tests
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Blob;
     using Microsoft.WindowsAzure.Storage.Table;
-    using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
 
     [TestClass]
     public class AzureStorageScenarioTests
     {
+        public static readonly TimeSpan StandardTimeout = Debugger.IsAttached ? TimeSpan.FromMinutes(5) : TimeSpan.FromSeconds(30);
+
         /// <summary>
         /// End-to-end test which validates a simple orchestrator function which doesn't call any activity functions.
         /// </summary>
@@ -50,7 +51,7 @@ namespace DurableTask.AzureStorage.Tests
                 await host.StartAsync();
 
                 var client = await host.StartOrchestrationAsync(typeof(Orchestrations.SayHelloInline), "World");
-                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
+                var status = await client.WaitForCompletionAsync(StandardTimeout);
 
                 Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);
                 Assert.AreEqual("World", JToken.Parse(status?.Input));
@@ -73,7 +74,7 @@ namespace DurableTask.AzureStorage.Tests
                 await host.StartAsync();
 
                 var client = await host.StartOrchestrationAsync(typeof(Orchestrations.SayHelloWithActivity), "World");
-                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
+                var status = await client.WaitForCompletionAsync(StandardTimeout);
 
                 Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);
                 Assert.AreEqual("World", JToken.Parse(status?.Input));
@@ -94,7 +95,7 @@ namespace DurableTask.AzureStorage.Tests
                 await host.StartAsync();
 
                 var client = await host.StartOrchestrationAsync(typeof(Orchestrations.Factorial), 10);
-                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
+                var status = await client.WaitForCompletionAsync(StandardTimeout);
 
                 Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);
                 Assert.AreEqual(10, JToken.Parse(status?.Input));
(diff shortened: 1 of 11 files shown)


Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1725,6 +1726,29 @@ private static async Task ValidateBlobUrlAsync(string taskHubName, string instan
             }
         }
 
+        /// <summary>
+        /// Tests the behavior of <see cref="SessionAbortedException"/> from orchestrations and activities.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public async Task AbortOrchestrationAndActivity(bool enableExtendedSessions)
```
ConnorMcMahon: I'm not sure I love testing `SessionAbortedException` behavior in both Activities and Orchestrators in the same test. It makes the `AbortSessionOrchestration` logic a bit harder to reason about, and if the test fails, it will require deeper investigation to figure out which of the two behaviors broke..

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -87,9 +89,16 @@ IEnumerable<OrchestratorAction> ExecuteCore(IEnumerable<HistoryEvent> eventHisto
...
-                                Debug.Assert(this.result.Exception != null);
-
-                                this.context.FailOrchestration(this.result.Exception.InnerExceptions.FirstOrDefault());
+                                Exception exception = this.result.Exception?.InnerExceptions.FirstOrDefault();
+                                Debug.Assert(exception != null);
+
+                                if (Utils.IsExecutionAborting(exception))
+                                {
+                                    // Let this exception propogate out to be handled by the dispatcher
+                                    this.result.GetAwaiter().GetResult();
```
ConnorMcMahon: It's not exactly clear to me that this will let the exception propagate like the comment states. Is there a reason this is preferred to something like the below?

`ExceptionDispatchInfo.Capture(exception).Throw();`

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -87,9 +89,16 @@ IEnumerable<OrchestratorAction> ExecuteCore(IEnumerable<HistoryEvent> eventHisto
...
                             {
-                                Debug.Assert(this.result.Exception != null);
-
-                                this.context.FailOrchestration(this.result.Exception.InnerExceptions.FirstOrDefault());
+                                Exception exception = this.result.Exception?.InnerExceptions.FirstOrDefault();
+                                Debug.Assert(exception != null);
+
+                                if (Utils.IsExecutionAborting(exception))
+                                {
+                                    // Let this exception propogate out to be handled by the dispatcher
```
ConnorMcMahon: typo: propogate -> propagate

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -110,74 +110,80 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
             {
-                if (isExtendedSession)
-                {
-                    TraceHelper.Trace(
-                        TraceEventType.Verbose,
-                        "OnProcessWorkItemSession-Release",
-                        $"Releasing extended session after {processCount} batch(es).");
-                    this.concurrentSessionLock.Release();
-                }
+                // Either the orchestration or the orchestration service abandoned the session as part of normal control flow.
```
ConnorMcMahon: I'm not quite sure what "normal control flow" means in this case. Based on reading the rest of the code, it seems like it means that the orchestration/orchestration service explicitly (or intentionally) abandoned the session.

src/DurableTask.Core/Exceptions/SessionAbortedException.cs
```diff
@@ -17,20 +17,21 @@ namespace DurableTask.Core.Exceptions
     using System.Runtime.Serialization;
 
     /// <summary>
-    /// Thrown when a session was aborted, for example in a split-brain situation
+    /// Thrown when an orchestration or activity session is aborted, for example in split-brain situation or host shutdown situations.
```
ConnorMcMahon: "host shutdown" seems too Azure Functions specific to include in DurableTask.Core documentation. Maybe something along the lines that consumers of the library can use it to retry processing a session due to transient errors?

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
cgillum: Since these are more like end-to-end tests and not unit tests, I prefer to have a smaller number of tests where possible, provided we get the same amount of code coverage.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
cgillum: It was a toss-up to me. I can change it if that makes it more clear.

ConnorMcMahon: I think I prefer the explicit throw, as then I don't need to read the comment to realize that this line will throw an exception.

Test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
ConnorMcMahon: Fair enough.

src/DurableTask.Core/Exceptions/SessionAbortedException.cs
cgillum: I think host is a generic-enough term and I don't think Azure Functions is implied in the comment. DTFx is a framework that needs to be self-hosted and any host the someone uses will have lifecycle startup & shutdown lifecycle events. I wanted to stick with well understood examples. Transient errors is a bit too broad. For example, an outbound HTTP call could fail due to a "transient error", but that's not necessarily an appropriate use of this exception.

PR #360, opened 2020-01-06
Azure Storage: Abandon prefetched orchestrator messages if the lease is lost

cgillum:
Resolves https://github.com/Azure/azure-functions-durable-extension/issues/382.

This is a small update to the orchestration work item polling logic in DurableTask.AzureStorage that proactively abandons messages if we've lost the lease for the source queue since fetching. The intent is to mitigate some of the split-brain problems see when partitions move around.

It won't fix 100% of split-brains caused by partition movement, but it should take care of the low-hanging fruit.

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
@@ -425,8 +425,9 @@ namespace DurableTask.AzureStorage.Tests
             await TestHelpers.WaitFor(
                 condition: () => !service.OwnedControlQueues.Any(),
                 timeout: TimeSpan.FromSeconds(10));
-            
-            // Small additional delay to account for tiny race condition.
+
+            // Small additional delay to account for tiny race condition between OwnedControlQueues being updated
+            // and LockNextTaskOrchestrationWorkItemAsync being able to react to that change.
             await Task.Delay(250);
 
             // STEP 3: Try to get an orchestration work item - a null value should be returned
--- a/Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
@@ -21,11 +21,14 @@ namespace DurableTask.AzureStorage.Tests
     using System.Threading.Tasks;
     using DurableTask.AzureStorage.Messaging;
     using DurableTask.AzureStorage.Monitoring;
+    using DurableTask.AzureStorage.Partitioning;
     using DurableTask.AzureStorage.Tracking;
     using DurableTask.Core;
+    using DurableTask.Core.History;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Blob;
+    using Microsoft.WindowsAzure.Storage.Queue;
     using Microsoft.WindowsAzure.Storage.Table;
 
     /// <summary>
@@ -172,11 +175,11 @@ namespace DurableTask.AzureStorage.Tests
             return results;
         }
 
-/// <summary>
-/// REQUIREMENT: Workers can be added or removed at any time and control-queue partitions are load-balanced automatically.
-/// REQUIREMENT: No two workers will ever process the same control queue.
-/// </summary>
-[TestMethod]
+        /// <summary>
+        /// REQUIREMENT: Workers can be added or removed at any time and control-queue partitions are load-balanced automatically.
+        /// REQUIREMENT: No two workers will ever process the same control queue.
+        /// </summary>
+        [TestMethod]
         public async Task MultiWorkerLeaseMovement()
         {
             const int MaxWorkerCount = 4;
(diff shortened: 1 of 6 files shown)


src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -179,9 +179,9 @@ async Task<MessageData> AddMessageAsync(TaskMessage taskMessage, OrchestrationIn
                 // We assume that auto-started orchestrations (i.e. instance ids starting with '@')
                 // are used exclusively by durable entities; so we can follow
                 // a custom naming convention to pass a time parameter.
-                var eventName = eventRaisedEvent.Name;
-                if (eventName.Length >= 3 && eventName[2] == '@'
-                    && DateTime.TryParse(eventRaisedEvent.Name.Substring(3), out var scheduledTime))
+                string eventName = eventRaisedEvent.Name;
+                if (eventName != null && eventName.Length >= 3 && eventName[2] == '@'
+                    && DateTime.TryParse(eventName.Substring(3), out DateTime scheduledTime))
```
cgillum: This is a fix for a null-ref issue I observed when developing my test. I don't think we would hit it under normal circumstances with Durable Entities since user's don't control the `Name` property of raised events, but it's theoretically possible for anyone using DurableTask.AzureStorage to hit this if they use our entity naming convention.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -610,6 +610,13 @@ static TaskHubInfo GetTaskHubInfo(string taskHub, int partitionCount)
                         return null;
                     }
 
+                    // Make sure we still own the partition. If not, abandon the session.
+                    if (session.ControlQueue.IsReleased)
+                    {
+                        await this.AbandonAndReleaseSessionAsync(session);
+                        return null;
+                    }
```
cgillum: This is the relevant change. It turned out that we were already setting `IsReleased` when we stop listening on a particular queue/lease. Here I'm just checking to see if a control queue was released before we try to process the messages.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -221,6 +221,7 @@ public override Task DeleteMessageAsync(MessageData message, SessionBase session
         public void Release()
         {
             this.releaseTokenSource.Cancel();
+            this.IsReleased = true;
```
cgillum: `IsReleased` would have been set to `true` anyways because of the cancellation token handling logic, but there can be a bit of a delay. I added this additional `IsReleased = true` here to speed up the process and make my new test succeed more quickly and reliably.

Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -371,6 +374,75 @@ public async Task TestInstanceAndMessageDistribution()
             }
         }
 
+        /// <summary>
+        /// If a partition is lost, verify that all pre-fetched messages associated
+        /// with that partition are abandoned and not processed.
+        /// </summary>
+        [TestMethod]
+        public async Task PartitionLost_AbandonPrefetchedSession()
```
cgillum: Like other tests in this class, we're testing methods on `AzureStorageOrchestrationService` rather than going through `TaskHubWorker`. We're still going through Azure Storage, but we have a bit more control using this test approach.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
ConnorMcMahon: Would love to see a comment that explains this in the code, as it may confuse us in the future otherwise.

Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -371,6 +374,75 @@ public async Task TestInstanceAndMessageDistribution()
...
+            //         The orchestration service should detect this and update its state.
+            BlobLease lease = (await service.ListBlobLeasesAsync()).Single();
+            await lease.Blob.ChangeLeaseAsync(
+                proposedLeaseId: Guid.NewGuid().ToString(),
+                accessCondition: AccessCondition.GenerateLeaseCondition(lease.Token));
+            await TestHelpers.WaitFor(
+                condition: () => !service.OwnedControlQueues.Any(),
+                timeout: TimeSpan.FromSeconds(10));
+            
+            // Small additional delay to account for tiny race condition.
```
ConnorMcMahon: Is the race condition that we may have gotten rid of the lease but not yet abandoned our work items yet?

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
cgillum: Sure, I can add that.

Test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
cgillum: No the work item is abandoned as part of the `LockNextTaskOrchestrationWorkItemAsync` call, so there is no racing there. The race being referred to here is the fact that `OwnedControlQueues` collection gets updated one instruction before the `ControlQueue.Release()` is called, which could mean that `TestHelpers.WaitFor` could theoretically complete and move onto the next step of the test too soon. It would be _extremely_ rare, but not impossible. I added this delay to avoid the possibility of flakiness in the test execution.

PR #363, opened 2020-01-29
Fix ETag retrieval for orchestrations with multiple executions

ConnorMcMahon:
It is possible for the history table to get into a state where the
latest execution history is first, followed by the previous execution
history, with the last entry for an instance id being the sentinal row.
The previous history parsing logic stopped processing as soon as it saw
a new execution id, meaning it failed to grab the ETag from the sentinal
row.

This, in combination with changes made when the ETag were not present,
caused UpdateState calls to fail, meaning queue messages for Activities
were being scheduled without recording that we did schedule them. This
leads to duplicate Activity executions until we retrieve the history
with an execution id (i.e. after an activity function successfully
executes).

This fix makes our ETag retrieval logic more robust.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -168,6 +168,7 @@ namespace DurableTask.AzureStorage.Tracking
             IList<HistoryEvent> historyEvents;
             string executionId;
             string eTagValue = null;
+            DateTime checkpointCompletionTime = DateTime.MinValue;
             if (tableEntities.Count > 0)
             {
                 // The most recent generation will always be in the first history event.
@@ -176,19 +177,26 @@ namespace DurableTask.AzureStorage.Tracking
                 // Convert the table entities into history events.
                 var events = new List<HistoryEvent>(tableEntities.Count);
 
-                bool areMultipleExecutions = false;
-                foreach (DynamicTableEntity entity in tableEntities)
+                for (int i = 0; i < tableEntities.Count; i++)
                 {
+                    DynamicTableEntity entity = tableEntities[i];
                     if (entity.Properties["ExecutionId"].StringValue != executionId)
                     {
                         // The remaining entities are from a previous generation and can be discarded.
-                        areMultipleExecutions = true;
                         break;
                     }
 
+                    // The sentinal row does not contain any history events, so ignore this row.
                     if (entity.RowKey == SentinelRowKey)
                     {
-                        eTagValue = entity.ETag;
+                        if (i != (tableEntities.Count - 1))
+                        {
+                            AnalyticsEventSource.Log.GeneralWarning(
+                                this.storageAccountName,
+                                this.taskHubName,
+                                $"The sentinal row should be the last row in history events query for instance id {instanceId}. It is row {i} of {tableEntities.Count}",
+                                Utils.ExtensionVersion);
+                        }
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -176,11 +176,13 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
                 // Convert the table entities into history events.
                 var events = new List<HistoryEvent>(tableEntities.Count);
 
+                bool areMultipleExecutions = false;
                 foreach (DynamicTableEntity entity in tableEntities)
                 {
                     if (entity.Properties["ExecutionId"].StringValue != executionId)
                     {
                         // The remaining entities are from a previous generation and can be discarded.
+                        areMultipleExecutions = true;
```
cgillum: To simplify this, rather than introducing a flag here, could we just assign `eTagValue` here?

ConnorMcMahon: Good idea. I realized I should also check that the sentinal's execution id matches what I expect as well, so I will do those both in the next iteration.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -176,17 +177,26 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
...
                         break;
                     }
 
+                    // The sentinal row does not contain any history events, so ignore this row.
                     if (entity.RowKey == SentinelRowKey)
                     {
-                        eTagValue = entity.ETag;
+                        if (i != (tableEntities.Count - 1))
+                        {
+                            AnalyticsEventSource.Log.GeneralWarning(
```
ConnorMcMahon: Debatably this should be an error and an exception should be thrown here. However, I am a bit hesitant throwing an exception, as I'm not sure how it is handled here.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -176,17 +177,26 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
...
+                                this.storageAccountName,
+                                this.taskHubName,
+                                $"The sentinal row should be the last row in history events query for instance id {instanceId}. It is row {i} of {tableEntities.Count}",
+                                Utils.ExtensionVersion);
+                        }
                         continue;
```
ConnorMcMahon: If we are going off the strict assumption that the Sentinel row should be the last row, this could be a break to be more semantically clear.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
ConnorMcMahon: I went with a slightly more intensive refactor, to move all of our sentinel processing into one place. I also added some checks to make sure the sentinel row is indeed the last row.

cgillum:
I see what you're trying to do but this last refactor seems unnecessarily complicated. Here is my suggestion, which I think is simpler and does not strictly require the sentinel to be the last row:

```csharp
IList<HistoryEvent> historyEvents;
string executionId;
DynamicTableEntity sentinel = null;
if (tableEntities.Count > 0)
{
    // The most recent generation will always be in the first history event.
    executionId = tableEntities[0].Properties["ExecutionId"].StringValue;

    // Convert the table entities into history events.
    var events = new List<HistoryEvent>(tableEntities.Count);

    foreach (DynamicTableEntity entity in tableEntities)
    {
        if (entity.Properties["ExecutionId"].StringValue != executionId)
        {
            // The remaining entities are from a previous generation and can be discarded.
            break;
        }

        if (entity.RowKey == SentinelRowKey)
        {
            sentinel = entity;
            continue;
        }

        // Some entity properties may be stored in blob storage.
        await this.DecompressLargeEntityProperties(entity);

        events.Add((HistoryEvent)this.tableEntityConverter.ConvertFromTableEntity(entity, GetTypeForTableEntity));
    }

    historyEvents = events;
}
else
{
    historyEvents = EmptyHistoryEventList;
    executionId = expectedExecutionId;
}

// Read the checkpoint completion time from the sentinel row, which should always be the last row.
// The only time a sentinel won't exist is if no instance of this ID has ever existed.
// The IsCheckpointCompleteProperty was newly added _after_ v1.6.4.
DateTime checkpointCompletionTime = DateTime.MinValue;
if (sentinel == null)
{
    sentinel = tableEntities.LastOrDefault(e => e.RowKey == SentinelRowKey);
}

string eTagValue = sentinel?.ETag;
if (sentinel != null &&
    sentinel.Properties.TryGetValue(CheckpointCompletedTimestampProperty, out EntityProperty timestampProperty))
{
    checkpointCompletionTime = timestampProperty.DateTime ?? DateTime.MinValue;
}

```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -184,9 +184,11 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
                         break;
                     }
 
+                    // The sentinal row does not contain any history events, so save it for later
```
sebastianburckhardt: sentinal => sentinel

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -208,9 +210,10 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
             // The only time a sentinel won't exist is if no instance of this ID has ever existed.
```
sebastianburckhardt: ... or it was removed via a history purge management operation?

cgillum:
Though it looks like the test run timed out. Let's make sure the CI is healthy before merging.

ConnorMcMahon:
I rebased onto master just to retrigger the build, and it appears to have worked this time. The timeout last time was in a Service Bus test, which I didn't touch here, so I am going to merge now.

Issue #368, opened 2020-02-20
NullReferenceException when Task returns void


mark-szabo:
A NullReferenceException id thrown when a Durable Function Task returns void like this:

```csharp
[FunctionName("MyFunctions_RunTaskAsync")]
public async Task RunTaskAsync([ActivityTrigger] string input)
{
    ...
}
```

```
Function 'MyFunctions (Orchestrator)' failed with an error.Reason: System.NullReferenceException: Object reference not set to an instance of an object.
at DurableTask.Core.TaskOrchestrationContext.ScheduleTaskToWorker[TResult](String name, String version, String taskList, Object[] parameters) in C:\source\durabletask\src\DurableTask.Core\TaskOrchestrationContext.cs:line 80
at DurableTask.Core.TaskOrchestrationContext.ScheduleTask[TResult](String name, String version, Object[] parameters) in C:\source\durabletask\src\DurableTask.Core\TaskOrchestrationContext.cs:line 70
at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableOrchestrationContext.CallDurableTaskFunctionAsync[TResult](String functionName, FunctionType functionType, Boolean oneWay, String instanceId, String operation, RetryOptions retryOptions, Object input, Nullable`1 scheduledTimeUtc) in d:\a\r1\a\azure-functions-durable-extension\src\WebJobs.Extensions.DurableTask\ContextImplementations\DurableOrchestrationContext.cs:line 607
at MyNamespace.MyFunctions.RunOrchestrator(IDurableOrchestrationContext context)
```

mark-szabo:
You should add something like this:

```csharp
if (result == null) throw new Exception("Task result is null, and cannot be casted to TResult.");
```

to line 79:
https://github.com/Azure/durabletask/blob/effb469a04480d35935c67bf07c50e902cb3e9be/src/DurableTask.Core/TaskOrchestrationContext.cs#L75-L81

PR #369, opened 2020-02-20
Fixing #368: NullReferenceException when Task returns void

mark-szabo:
Fixing #368: NullReferenceException when Task returns void


mark-szabo:
Ok, this is a bad solution to the issue. Closing now.

PR #372, opened 2020-02-28
Handle null continuation tokens properly in paginated GetInstances API

ConnorMcMahon:
Previously, when there is a null continuation token in the paginated
GetInstances API call for DurableTask.AzureStorage, we return the base64
encoded value of "null", instead of null.

We now return null properly in this case.


ConnorMcMahon:
It definitely should. This should only affect the terminal page, but I am willing to do a manual test to verify that the while-loop logic would work now.

ConnorMcMahon:
Glad I tested. Fixed a nasty bug that would cause an infinite loop and added extra test to handle pagination logic.

PR #373, opened 2020-03-02
Enabling Extended Session Distributed Tracing

TsuyoshiUshio:
I found that current implementation doesn't work well for extended session settings. 
This PR unlock the extended session with Distributed Tracing. 
Could you have a look, @cgillum ?

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1002,6 +1002,16 @@ namespace DurableTask.AzureStorage
 
             // Correlation
 
+            CorrelationTraceClient.Propagate(
+                () =>
+                {
+                    // In case of Extended Session, Emit the Dependency Telemetry. 
+                    if (workItem.IsExtendedSession)
+                    {
+                        TrackExtendedSessionDependencyTelemetry(session);
+                    }
+                });
+
             TraceContextBase currentTraceContextBaseOnComplete = null;
             CorrelationTraceClient.Propagate(
                 () =>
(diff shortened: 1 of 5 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -120,6 +120,16 @@ async Task OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem)
                 }
 
                 var isExtendedSession = false;
+
+                CorrelationTraceClient.Propagate(
+                    () =>
+                    {                
+                        // Check if it is extended session.
+                        isExtendedSession = this.concurrentSessionLock.Acquire();
```
TsuyoshiUshio: @cgillum Is there a better way to check if it is an extended session? 
I don't know why checking the lock availability becomes checking the extended session. It works, however, not cool.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1095,6 +1105,37 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
+                            {
+                                var orchestrationDependencyTraceContext = traceContext.OrchestrationTraceContexts.Pop();
+                                CorrelationTraceClient.TrackDepencencyTelemetry(orchestrationDependencyTraceContext);
+                            }
+
+                            break;
+
+                        default:
+                            // When internal error happens, multiple message could come, however, it should not be prioritized.
+                            break;
```
ConnorMcMahon: I'm a bit confused why we want to break here. Is there a reason we don't want to go through the rest of the messages to check if they indicate the execution has finished?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
ConnorMcMahon: There might be... I will get back to you on that.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
ConnorMcMahon: Disregard this, the `break` is for the switch statement, NOT the for-loop. I think it is fine as is!

PR #376, opened 2020-03-09
Added DataConverter property to OrchestrationContext to allow for custom serializer settings

amdeel:
related to https://github.com/Azure/azure-functions-durable-extension/issues/1155

### DIFF
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -17,6 +17,7 @@ namespace DurableTask.Core
     using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
+    using DurableTask.Core.Serializing;
     using ImpromptuInterface;
 
     /// <summary>
@@ -30,6 +31,11 @@ namespace DurableTask.Core
         /// </summary>
         [ThreadStatic]
         public static bool IsOrchestratorThread;
+
+        /// <summary>
+        /// JsonDataConverter for serialization settings
+        /// </summary>
+        public JsonDataConverter DataConverter { get; set; }
 
         /// <summary>
         /// Instance of the currently executing orchestration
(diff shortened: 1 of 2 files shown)


src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -31,6 +32,11 @@ public abstract class OrchestrationContext
         [ThreadStatic]
         public static bool IsOrchestratorThread;
 
+        /// <summary>
+        /// JsonDataConverter for serialization settings
+        /// </summary>
+        public JsonDataConverter DataConverter { get; set; }
```
ConnorMcMahon: If I remember correctly, don't we have a separate `JsonDataConverter` for message serialization vs error serialization in Durable Functions proper? I don't remember why we went with that approach, but we may want to include a second converter for errors.

amdeel: Technically our JsonDataConverter is our MessagePayloadDataConverter. It contains both Message and Error settings. It would make sense to split this up and pass both data converters to DTFx.

ConnorMcMahon: Yeah, it looks like if you use `MessagePayloadDataConverter` as an actual `JsonDataConverter`, it doesn't use either the `Message` nor the `Error` settings.

amdeel: It uses the Message settings. It initializes itself as a JsonDataConverter with the message settings. But I realized that I also have it return an ErrorSettings JsonDataConverter if I need it to, so I can still use it to return that data converter to DTFx. I think I'm going to change it so they return JsonDataConverters and not MessagePayloadDataConverters because thats kind of confusing.

amdeel: I'll make this work a separate PR

PR #379, opened 2020-03-26
Prevent value drift on PendingOrchestrationCount

ConnorMcMahon:
Currently, when a host lease is lost and reclaimed before its pending messages are processed, a new control queue is spun up, losing track of what messages it has already processed.

In an ideal solution, we would find some way to reclaim that old control queue. This commit tracks a more simple workaround in which we don't rely on the message id being present in the current control queues pending messages in order to decrement PendingOrchestrationCount. This may not fix all of the problems (i.e. duplicate messages may be still be double counted in the lease reaquire scenario). It would still be an improvement over the current scenario however.

### DIFF
--- a/src/DurableTask.Core/Stats/Counter.cs
+++ b/src/DurableTask.Core/Stats/Counter.cs
@@ -32,7 +32,10 @@ namespace DurableTask.Core.Stats
         /// </summary>
         public void Increment()
         {
-            Interlocked.Increment(ref this.counterValue);
+            lock (this)
+            {
+                this.counterValue++;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/DurableTask.Core/Stats/Counter.cs
```diff
@@ -41,15 +44,24 @@ public void Increment()
...
         /// <summary>
-        /// Decrements the counter by 1
+        /// Decrements the counter by 1. Ensures that counter cannot go below 0.
         /// </summary>
         public void Decrement()
         {
-            Interlocked.Decrement(ref this.counterValue);
+            lock (this)
+            {
+                if (this.counterValue > 0)
```
ConnorMcMahon: This is the key reason for changing Counter. I figured now that we don't have that guard against us having seen the pending message before decrementing on a `Delete` or `Abandon` operation, it *may* be possible for us to go negative. This would prevent that.

ConnorMcMahon:
As mentioned in the description, I don't think this workaround covers the scenario where we receive a duplicate message after reacquiring the lease. I think a more robust solution involving holding onto control queues after we release them might be required to fully fix the problem.

src/DurableTask.Core/Stats/Counter.cs
```diff
@@ -32,7 +32,10 @@ public class Counter
         /// </summary>
         public void Increment()
         {
-            Interlocked.Increment(ref this.counterValue);
+            lock (this)
```
cgillum: Note that `lock (this)` [is considered dangerous](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement#remarks). Consider using a dedicated private object for locking instead.

ConnorMcMahon: This is good to know, and makes a lot of sense!

ConnorMcMahon:
I like your suggestion about just having a global pendingOrchestrationMessages dictionary. It sounds straightforward, and I will definitely give it a shot in a timeboxed fashion. If it looks like it is too tricky to get correct, I will fix the locking I added in Counter and merge this.

ConnorMcMahon:
Closing in favor of new PR.

Issue #223, opened 2018-10-18
Any samples with Service Fabric


aRajeshKumar:
Do we have any samples with Service Fabric please?

PR #381, opened 2020-03-27
Added service fabric sample

WhitWaldo:
I put the existing sample into a directory called WinForms and added another directory called ServiceFabric.

In the latter, I added a simple Service Fabric solution with a single stateful service and many of the original activity samples, replacing all console logging with the diagnostic logging in SF.

It's in a state that someone can open the solution provided, hit F5 and view the orchestration in the Diagnostic Events.

This pull request also provides a resolution to #223


cgillum:
Hey @shankarsama, since this is related to Service Fabric, would you mind reviewing this PR?

Issue #280, opened 2019-04-19
DurableTask.AzureStorage: Duplicate activity execution caused by renew/delete race


cgillum:
It's been observed in some production workloads that message deletion can fail with "message not found" if it happens around the same instant that the message is renewed. The end result is that the activity function gets executed twice.

The nature of this problem is described in this StackOverflow post: https://stackoverflow.com/questions/46112828/pop-receipt-changed-on-update-message

Most likely some form of locking will need to be done to avoid this race.

PR #385, opened 2020-04-07
Implemented retry for DeleteMessageAsync to fix delete/renew race condition issue

amdeel:
resolves https://github.com/Azure/durabletask/issues/280

When a message is renewed right before a delete operation takes place, the delete will fail and the activity will run twice.

This retry on TaskHubQueue.DeleteMessageAsync will retry one time when CloudQueue.DeleteMessageAsync returns a 404.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
@@ -324,13 +324,13 @@ namespace DurableTask.AzureStorage.Messaging
                 }
                 catch (Exception e)
                 {
-                    this.HandleMessagingExceptions(e, message, $"Caller: {nameof(DeleteMessageAsync)}");
-
                     if (!haveRetried && IsMessageGoneException(e))
                     {
                         haveRetried = true;
                         continue;
                     }
+
+                    this.HandleMessagingExceptions(e, message, $"Caller: {nameof(DeleteMessageAsync)}");
                 }
                 finally
                 {
--- a/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
@@ -312,27 +312,44 @@ namespace DurableTask.AzureStorage.Messaging
                 message.SequenceNumber,
                 Utils.ExtensionVersion);
 
-            try
+            var haveRetried = false;
+            var successfulExecutions = 0;
+            while (successfulExecutions < 1)
             {
-                await this.storageQueue.DeleteMessageAsync(
-                    queueMessage,
-                    this.QueueRequestOptions,
-                    session.StorageOperationContext);
+                try
+                {
+                    await this.storageQueue.DeleteMessageAsync(
+                        queueMessage,
+                        this.QueueRequestOptions,
+                        session.StorageOperationContext);
+                }
+                catch (Exception e)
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -312,27 +312,44 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase se
...
+                finally
+                {
+                    successfulExecutions++;
+                    this.stats.StorageRequests.Increment();
+                }
             }
         }
 
-        void HandleMessagingExceptions(Exception e, MessageData message, string details)
+        private bool ExceptionIsRenewDeleteRace(Exception e)
```
cgillum, line 344: I don't like this method name because it's a bit misleading. A 404 status code means the message is gone. It doesn't _necessarily_ mean it's a renew/delete race. A more accurate name would be `IsMessageGoneException`.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -312,27 +312,44 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase se
...
-            try
-            {
-                await this.storageQueue.DeleteMessageAsync(
-                    queueMessage,
-                    this.QueueRequestOptions,
-                    session.StorageOperationContext);
-            }
-            catch (Exception e)
+            var haveRetried = false;
+            var successfulExecutions = 0;
```
cgillum, line 316: The use of `successfulExecutions` seems really confusing to me, probably because it is always incremented in the `finally` block, even if the execution actually failed. I suggest removing this variable and instead use `continue` and `break` to manage control flow.  For example:

```csharp
while (true)
{
  try { ... }
  catch
  {
    if (!hasRetried && ...)
    {
      hasRetries = true;
      continue;
    }

    ...
  }
  finally { ... }

  break;
}
```

amdeel: That makes sense, this is much cleaner.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -312,27 +312,44 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase se
...
+                try
+                {
+                    await this.storageQueue.DeleteMessageAsync(
+                        queueMessage,
+                        this.QueueRequestOptions,
+                        session.StorageOperationContext);
+                }
+                catch (Exception e)
+                {
+                    this.HandleMessagingExceptions(e, message, $"Caller: {nameof(DeleteMessageAsync)}");
```
cgillum, line 327: Last suggestion: Could we move the call to `HandleMessagingException` below the `if` block? I think there are a couple reasons to do this:
1. If the message is actually gone, we'll report two "message gone" warnings, which is a little confusing.
2. If we really did hit the renew/delete race and we fixed it with a retry, it will be confusing to see a "message gone" warning in the traces. Only someone who is intimately aware of this workaround will be able to understand that this warning can be safely ignored.

If we instead only trace _after_ confirming that a retry didn't help, then we'll see a single warning message which we know is relevant. The downside is that we won't be able to clearly see in the logs when we encounter the race, but I think that is acceptable.

PR #386, opened 2020-04-09
Fix casing of directories

kfarnung:
On case-sensitive platforms, like Linux, there is both "Test" and
"test". This causes issues with actually building the code on Linux.

The fix is to move all files into the "test" directory and fix some
casing issues in the solution file.


kfarnung:
Rebased on the latest master and pushed again. Any update on getting this merged?

ConnorMcMahon:
@kfarnung, sorry for the delay on this. I will get this merged shortly. Just want to make sure the tests still run.

PR #398, opened 2020-05-22
Add optional parameter to remove some telemetry for performance reasons

ConnorMcMahon:
Add a flag to TaskHubWorker and some of its internal classes in order to
flag that we prefer performance over perfectly detailed telemetry.

Currently the only telemetry this field affects is the runtime state
dump. This is a linear operation with the size of the event history,
and even if you filter out trace/information level events, there is
still the computational cost of creating the dump. In efficient
telemetry mode, we only emit the number of events rather than the events
themselves.

### DIFF
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -257,28 +257,40 @@ namespace DurableTask.Core
         /// <summary>
         /// Gets a statedump of the current list of events
         /// </summary>
+        /// <param name="abridgedDump">A full dump can be expensive, so this optional parameter just grabs the count of events.</param>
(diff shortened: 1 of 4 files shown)


ConnorMcMahon:
@sebastianburckhardt, I definitely had the same thought, but we don't have a clean way to do this without a breaking change like you mentioned.

The other options is I just get rid of the expensive telemetry option altogether, but I don't know if anyone currently relies on it, and it is really only expensive when you have a large history.

sebastianburckhardt:
I think I like your suggestion to remove it altogether. If we get complaints we can revisit and add the feature back, and perhaps spend a bit more time on getting the design right. If we don't get complaints then the code just got simpler and faster for everybody.

PR #399, opened 2020-05-22
Improve performance of the runtime state dump

ConnorMcMahon:
The actions of dumping the orchestration state for the telemetry event
type TaskOrchestrationDispatcher-ExecuteUserOrchestration-Begin is
expensive. It scales linearly with the number of events, meaning
orchestrations with large histories start to spend most of their time
writing this verbose telemetry item, which is not ideal.

Instead, we now just log the number of events, as this is a constant
time operation.

### DIFF
--- a/src/DurableTask.Core/OrchestrationRuntimeStateDump.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeStateDump.cs
@@ -13,22 +13,20 @@
 
 namespace DurableTask.Core
 {
-    using DurableTask.Core.History;
-    using System.Collections.Generic;
-
     /// <summary>
     /// A snapshot / state dump of an OrchestrationRuntimeState's events
     /// </summary>
     public class OrchestrationRuntimeStateDump
     {
         /// <summary>
-        /// List of all history events for this runtime state dump
+        /// The number of history events.
         /// </summary>
-        public IList<HistoryEvent> Events;
+        public int EventCount { get; set; }
 
         /// <summary>
-        /// List of new events added during an execution for this runtime state dump
+        /// The number of new events.
         /// </summary>
-        public IList<HistoryEvent> NewEvents;
+        public int NewEventsCount { get; set; }
+
     }
 }
--- a/src/DurableTask.Core/OrchestrationRuntimeStateDump.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeStateDump.cs
@@ -13,20 +13,32 @@
 
 namespace DurableTask.Core
 {
+    using DurableTask.Core.History;
+    using System.Collections.Generic;
+
     /// <summary>
     /// A snapshot / state dump of an OrchestrationRuntimeState's events
     /// </summary>
     public class OrchestrationRuntimeStateDump
     {
         /// <summary>
-        /// The number of history events.
+        /// The number of all history events for this runtime state dump.
         /// </summary>
         public int EventCount { get; set; }
 
         /// <summary>
-        /// The number of new events.
+        /// The number of new events added during an execution for this runtime state dump.
         /// </summary>
         public int NewEventsCount { get; set; }
 
+        /// <summary>
+        /// List of all history events for this runtime state dump
+        /// </summary>
+        public IList<HistoryEvent> Events;
+
+        /// <summary>
+        /// List of new events added during an execution for this runtime state dump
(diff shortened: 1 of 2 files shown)


cgillum:
Wow, I have known about this verbose tracing for a while but had not gone the next step to see whether it might be impactful! Thanks for finding and making this fix.

To avoid impacting anyone who might depend on this behavior, I suggest we employ conditional complication. For `#if DEBUG`, let's keep the behavior as-is. Otherwise, we can use your new trace. Does that sound like a good idea?

src/DurableTask.Core/OrchestrationRuntimeStateDump.cs
```diff
@@ -13,22 +13,20 @@
...
     /// A snapshot / state dump of an OrchestrationRuntimeState's events
     /// </summary>
     public class OrchestrationRuntimeStateDump
     {
         /// <summary>
-        /// List of all history events for this runtime state dump
+        /// The number of history events.
         /// </summary>
-        public IList<HistoryEvent> Events;
+        public int EventCount { get; set; }
```
cgillum, line 24: Careful: this is technically a breaking API change since the class and properties are `public`. Maybe we can add the new fields without removing the old ones?

sebastianburckhardt:
Should the verbose tracing for DurableTask.Core even enabled by default? I am not sure we need it. If it is not enabled by default, this whole problem has an easy solution: check the verbosity level before doing expensive formatting, e.g. as in

```csharp
if (TraceHelper.IsVerboseEnabled)
{
        if (TraceHelper.TraceInstance(
                        TraceEventType.Verbose,
                        "TaskOrchestrationDispatcher-ExecuteUserOrchestration-Begin",
                        runtimeState.OrchestrationInstance,
                        "Executing user orchestration: {0}",
                        DataConverter.Serialize(runtimeState.GetOrchestrationRuntimeStateDump(this.efficientTelemetry), true));
}
```

ConnorMcMahon:
@sebastianburckhardt,

Unfortunately, it looks like ETW doesn't give you very good data on whether the listener is enabled unless you have a specific listener enabled. See [here](https://stackoverflow.com/questions/51611353/how-does-system-diagnostics-tracing-eventsource-isenabled-work). Since TraceHelper's `Is<log-level>Enabled` methods rely on the `EventSource.IsEnabled()` method, I'm not sure we can rely on it.

I think that as a quick workaround in the meantime, I will go with Chris's suggestion.

sebastianburckhardt:
As far as I understand the general idea (for both EventSource and ILogger) is that the log level is determined by the subscribed consumers, not by the producer. I think we should try to follow this standard pattern where possible.

In this particular case I would suggest that for our DF hosted scenarios, we stop collecting verbose tracing from DurableTask.Core. I don't think it gets stored in the Kusto logs anyway, so I am not sure why we should generate those events.

cgillum:
I believe you are correct that this behavior is determined by the consumer. The problem I think we're facing is that we do not control the consumer. Rather, the consumer is a piece of infrastructure that runs on the Azure VMs (we call it "MDS"). We explicitly ask it to listen to `DurableTask-Core` event traces so that we can capture specific event IDs, but I suspect this infrastructure is subscribing to all trace levels. We are able to have more controlled filters before pushing the data to Kusto, but I don't know if we have the ability to control the trace levels on the ETW consumer side. It's something we'd need to look into and I'm not confident that the MDS configuration allows us this level of control.

sebastianburckhardt:
I see. In that case perhaps the next best solution is to extend the TraceHelper so a lower limit on the tracing level can be specified by a configuration parameter. This is what I did for the EventSourced backend. We don't have a nice way to pass configuration parameters right now though, I was discussing that with @ConnorMcMahon last week.

oising:
@sebastianburckhardt Sorry to jump in here (hi @cgillum), but any chance of a private preview/alpha for Microsoft.Azure.DurableTask.EventSourced with durable entities? I'm trying to make a critical path decision over moving to Orleans or not in the next 6 months... :P

oising:
> @sebastianburckhardt Sorry to jump in here (hi @cgillum), but any chance of a private preview/alpha for Microsoft.Azure.DurableTask.EventSourced with durable entities? I'm trying to make a critical path decision over moving to Orleans or not in the next 6 months... :P

Switched to Orleans.

PR #403, opened 2020-05-29
Fix bug where execution started events were treated as out-of-order

ConnorMcMahon:
The idea here is to apply the same fix from PR #299 to entity calls. It currently does not apply to messages sent/received from outside of the orchestration context, so it excludes entity calls.

The idea behind this fix is to apply the "message from the future" retry logic for all non "ExecutionStarted" events if the orchestration instance does not exist yet.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
+++ b/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
@@ -132,14 +132,15 @@ namespace DurableTask.AzureStorage.Messaging
 
         internal bool IsOutOfOrderMessage(MessageData message)
         {
-            if (this.IsNonexistantInstance() && message.OriginalQueueMessage.DequeueCount > 5)
+            if (this.IsNonexistantInstance() && 
+                message.TaskMessage.Event.EventType != EventType.ExecutionStarted)
             {
                 // The first five times a message for a nonexistant instance is dequeued, give the message the benefit
                 // of the doubt and assume that the instance hasn't had its history table populated yet. After the 
                 // fifth execution, ~30 seconds have passed and the most likely scenario is that this is a zombie event. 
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -132,14 +132,15 @@ internal void DiscardMessage(MessageData data)
 
         internal bool IsOutOfOrderMessage(MessageData message)
         {
-            if (this.IsNonexistantInstance() && message.OriginalQueueMessage.DequeueCount > 5)
+            if (this.IsNonexistantInstance() &&
```
ConnorMcMahon, line 135: I am not  convinced this is the right fix, as the internal customer testing this ran into increased latency and still ended up with stuck orchestrations. This may be due to the fact that they are using eternal orchestrations + entities, but I still need to look at the logs to see what is going on.

cgillum: Interesting...I wouldn't have expected that to happen and I worry about this particular fix. External events are not supposed to have an ExecutionId, which means that they can be handled by any generation of an eternal orchestration (see [here](https://github.com/Azure/durabletask/blob/c0fe23a82627a539b55665d4ec51e7b166ce603c/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs#L159)). I agree it would be good to confirm what's going on in the logs. Things to look for:
* Did the response event arrive before or after the instance history was saved?
* Was the response event correctly abandoned so that it could be retried?
* What happened to the retried message (which should have been processed after the orchestration persisted)?

cgillum:
Closing this for now since it appears to be stale. Feel free to re-open if you still think we need to take action on this.

PR #416, opened 2020-07-13
Use DownloadToStreamAsync(Stream) to download blob contents

bachuv:
This change resolves [issue #1406](https://github.com/Azure/azure-functions-durable-extension/issues/1406) from the azure-function-durable-extension repo.

The current implementation to download contents from a blob (`OpenReadAsync()`) has a chance to result in a race condition, but switching to using the `DownloadToStreamAsync(Stream)` API would solve this issue.


cgillum:
Hey @ConnorMcMahon, it looks like there is something wrong with the DTFx CI. A couple of PR builds have failed due to what looks like SDK issues. Can you help take a look?

cgillum:
Hey @bachuv can you update your PR with the latest from master and push again? That should resolve the CI issue for this PR.

PR #420, opened 2020-07-16
Storage SDK deadlock protection for UploadHistoryBatch

cgillum:
Resolves https://github.com/Azure/azure-functions-durable-extension/issues/1377.

Based on my investigation, the only reasonable explanation for the stuck orchestration that I can think of is that this is a storage SDK deadlock when checkpointing orchestration history. I'm adding deadlock protection for this code path in **AzureStorageTrackingStore.cs**.

I also made a very small logging fix in **MessageManager.cs** related to large messages that I discovered while investigating the issue.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -145,10 +145,14 @@ namespace DurableTask.AzureStorage
                     decompressedMessage,
                     this.taskMessageSerializerSettings);
                 envelope.MessageFormat = MessageFormatFlags.StorageBlob;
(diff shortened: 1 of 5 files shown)


src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -145,10 +145,14 @@ public async Task<MessageData> DeserializeQueueMessageAsync(CloudQueueMessage qu
                     decompressedMessage,
                     this.taskMessageSerializerSettings);
                 envelope.MessageFormat = MessageFormatFlags.StorageBlob;
+                envelope.TotalMessageSizeBytes = Encoding.Unicode.GetByteCount(decompressedMessage);
```
ConnorMcMahon, line 148: Thank you for this fix!

PR #425, opened 2020-07-24
Structured and extensible logging

cgillum:
# Please Read!
This PR adds basic support for [.NET Core's ILogger infrastructure](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/) to **DurableTask.Core** and **DurableTask.AzureStorage**. The motivation for this work is primarily for the benefit of Durable Functions, but there are significant benefits for any consumer of DTFx. The stated goals are as follows:

1. Enable **DurableTask.Core** logs to be more easily surfaced in Antares Kusto clusters (internal to Azure App Service)
2. Enable correlating **DurableTask.Core** logs with **DurableTask.AzureStorage** traces using simple Kusto queries
3. Enable users to export **DurableTask.Core** and **DurableTask.AzureStorage** logs to Application Insights automatically when running in Azure Functions.
4. Enable users to capture logs in any way they choose by supporting `ILogger` extensibility (this is critical for supporting Durable Functions on Linux platforms).
5. Make it easy for future transaction store providers (like **DurableTask.SqlServer**) to leverage the same logging infrastructure.

Before reviewing the code, I highly recommend you review [this README.md file](https://github.com/Azure/durabletask/blob/c80e6b5d96d27543202dc9078d1a9edb4e5708a4/src/DurableTask.Core/Logging/README.md) that I added. This will help you understand the design.

### Change summary
* **Upgraded all projects from `net451` to `net461`**. This was necessary to support `ILogger` without conditional compilation. All projects should now be `netstandard2.0` compatible. This will require any upstream projects that target `net451` to upgrade if they want to keep using DTFx.
* **Added a NEW `DurableTask-Core` EventSource implementation**. The old one will still be there (for now) just in case we find issues with the new implementation. The new one was meant to be compatible with our `DurableTask-AzureStorage` event source provider to make it easier to query in Kusto.
* **Added optional `ILoggerFactory` configuration** to both `DurableTask.Core` and `DurableTask.AzureStorage`. For Azure Functions, we will add the Azure Functions `ILoggerFactory` to these DTFx classes so that Functions users can have direct access to these logs. We can also use this to do JSON logging on Linux platforms (that part is not in scope for this PR, however).
* **Added logs of new structured logging code to `DurableTask.Core`**. This project needed an overhaul of its logging to make it more structured and query-able. I've basically added a bunch of new log events to help us with production debugging.
* **Replaced EventSource tracing with new LogHelper abstraction** in `DurableTask.AzureStorage`. This was a very mechanical change and should not impact existing Event Source logs. However, it allows us to plug into `ILogger` and get all those benefits.

### Pending changes:
- [x] Add log messages for DurableTask.AzureStorage log events
- [x] Add "AppName" log event field to make it even easier to correlate traces with Azure Functions traces.
- [x] Debug why "EventName" is not showing up in Application Insights logs (see https://github.com/Azure/azure-webjobs-sdk/pull/2544)

### Manual testing:
- [x] Capture ETW events and ensure everything flows correctly across `DurableTask-Core` and `DurableTask-AzureStorage`
- [x] Capture logs from a Function app and make sure they appear in Application Insights (when configured to do so)

### DIFF
--- a/samples/DurableTask.Samples/DurableTask.Samples.csproj
+++ b/samples/DurableTask.Samples/DurableTask.Samples.csproj
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -16,6 +16,7 @@ namespace DurableTask.AzureStorage
     using System;
     using System.Diagnostics.Tracing;
     using System.Threading;
+    using DurableTask.AzureStorage.Logging;
 
     /// <summary>
     /// ETW Event Provider for the DurableTask.AzureStorage provider extension.
@@ -41,7 +42,7 @@ namespace DurableTask.AzureStorage
             SetCurrentThreadActivityId(activityId);
         }
 
-        [Event(101, Level = EventLevel.Informational, Opcode = EventOpcode.Send, Task = Tasks.Enqueue, Version = 5)]
+        [Event(EventIds.SendingMessage, Level = EventLevel.Informational, Opcode = EventOpcode.Send, Task = Tasks.Enqueue, Version = 5)]
         public void SendingMessage(
             Guid relatedActivityId,
             string Account,
@@ -59,7 +60,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEventWithRelatedActivityId(
-                101,
+                EventIds.SendingMessage,
                 relatedActivityId,
                 Account,
                 TaskHub,
@@ -76,7 +77,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(102, Level = EventLevel.Informational, Opcode = EventOpcode.Receive, Task = Tasks.Dequeue, Version = 5)]
+        [Event(EventIds.ReceivedMessage, Level = EventLevel.Informational, Opcode = EventOpcode.Receive, Task = Tasks.Dequeue, Version = 5)]
         public void ReceivedMessage(
             Guid relatedActivityId,
             string Account,
@@ -96,7 +97,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEventWithRelatedActivityId(
-                102,
+                EventIds.ReceivedMessage,
                 relatedActivityId,
                 Account,
                 TaskHub,
@@ -115,7 +116,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(103, Level = EventLevel.Informational, Version = 4)]
+        [Event(EventIds.DeletingMessage, Level = EventLevel.Informational, Version = 4)]
         public void DeletingMessage(
             string Account,
             string TaskHub,
@@ -129,7 +130,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                103,
+                EventIds.DeletingMessage,
                 Account,
                 TaskHub,
                 EventType,
@@ -142,7 +143,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(104, Level = EventLevel.Warning, Version = 5)]
+        [Event(EventIds.AbandoningMessage, Level = EventLevel.Warning, Version = 5)]
         public void AbandoningMessage(
             string Account,
             string TaskHub,
@@ -157,7 +158,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                104,
+                EventIds.AbandoningMessage,
                 Account,
                 TaskHub,
                 EventType,
@@ -171,31 +172,31 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(105, Level = EventLevel.Warning, Message = "An unexpected condition was detected: {2}")]
+        [Event(EventIds.AssertFailure, Level = EventLevel.Warning, Message = "An unexpected condition was detected: {2}")]
         public void AssertFailure(
             string Account,
             string TaskHub,
             string Details,
             string ExtensionVersion)
         {
-            this.WriteEvent(105, Account, TaskHub, Details, ExtensionVersion);
+            this.WriteEvent(EventIds.AssertFailure, Account, TaskHub, Details, ExtensionVersion);
         }
 
-        [Event(106, Level = EventLevel.Warning, Version = 3)]
+        [Event(EventIds.MessageGone, Level = EventLevel.Warning, Version = 4)]
         public void MessageGone(
             string Account,
             string TaskHub,
+            string EventType,
+            int TaskEventId,
             string MessageId,
             string InstanceId,
             string ExecutionId,
             string PartitionId,
-            string EventType,
-            int TaskEventId,
             string Details,
             string ExtensionVersion)
         {
             this.WriteEvent(
-                106,
+                EventIds.MessageGone,
                 Account,
                 TaskHub,
                 MessageId,
@@ -208,16 +209,18 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(107, Level = EventLevel.Error)]
+        [Event(EventIds.GeneralError, Level = EventLevel.Error)]
         public void GeneralError(string Account, string TaskHub, string Details, string ExtensionVersion)
         {
-            this.WriteEvent(107, Account, TaskHub, Details, ExtensionVersion);
+            this.WriteEvent(EventIds.GeneralError, Account, TaskHub, Details, ExtensionVersion);
         }
 
-        [Event(108, Level = EventLevel.Warning, Version = 2, Message = "A duplicate message was detected. This can indicate a potential performance problem. Message ID = '{2}'. DequeueCount = {3}.")]
+        [Event(EventIds.DuplicateMessageDetected, Level = EventLevel.Warning, Version = 3, Message = "A duplicate message was detected. This can indicate a potential performance problem. Message ID = '{4}'. DequeueCount = {8}.")]
         public void DuplicateMessageDetected(
             string Account,
             string TaskHub,
+            string EventType,
+            int TaskEventId,
             string MessageId,
             string InstanceId,
             string ExecutionId,
@@ -226,9 +229,11 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                108,
+                EventIds.DuplicateMessageDetected,
                 Account,
                 TaskHub,
+                EventType,
+                TaskEventId,
                 MessageId,
                 InstanceId,
                 ExecutionId ?? string.Empty,
@@ -237,10 +242,12 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(109, Level = EventLevel.Warning, Message = "A poison message was detected! Message ID = '{2}'. DequeueCount = {6}.")]
+        [Event(EventIds.PoisonMessageDetected, Level = EventLevel.Warning, Message = "A poison message was detected! Message ID = '{4}'. DequeueCount = {8}.")]
         public void PoisonMessageDetected(
             string Account,
             string TaskHub,
+            string EventType,
+            int TaskEventId,
             string MessageId,
             string InstanceId,
             string ExecutionId,
@@ -249,9 +256,11 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                109,
+                EventIds.PoisonMessageDetected,
                 Account,
                 TaskHub,
+                EventType,
+                TaskEventId,
                 MessageId,
                 InstanceId,
                 ExecutionId ?? string.Empty,
@@ -260,7 +269,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(110, Level = EventLevel.Informational, Version = 3)]
+        [Event(EventIds.FetchedInstanceHistory, Level = EventLevel.Informational, Version = 3)]
         public void FetchedInstanceHistory(
             string Account,
             string TaskHub,
@@ -275,7 +284,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                110,
+                EventIds.FetchedInstanceHistory,
                 Account,
                 TaskHub,
                 InstanceId,
@@ -289,7 +298,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(111, Level = EventLevel.Informational, Version = 4)]
+        [Event(EventIds.AppendedInstanceHistory, Level = EventLevel.Informational, Version = 4)]
         public void AppendedInstanceHistory(
             string Account,
             string TaskHub,
@@ -306,7 +315,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                111,
+                EventIds.AppendedInstanceHistory,
                 Account,
                 TaskHub,
                 InstanceId,
@@ -322,7 +331,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(112, Level = EventLevel.Informational)]
+        [Event(EventIds.OrchestrationServiceStats, Level = EventLevel.Informational)]
         public void OrchestrationServiceStats(
             string Account,
             string TaskHub,
@@ -339,7 +348,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                112,
+                EventIds.OrchestrationServiceStats,
                 Account,
                 TaskHub,
                 StorageRequests,
@@ -355,7 +364,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(113, Level = EventLevel.Informational, Version = 2)]
+        [Event(EventIds.RenewingMessage, Level = EventLevel.Informational, Version = 2)]
         public void RenewingMessage(
             string Account,
             string TaskHub,
@@ -369,7 +378,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                113,
+                EventIds.RenewingMessage,
                 Account,
                 TaskHub,
                 InstanceId,
@@ -382,7 +391,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(114, Level = EventLevel.Error, Version = 2)]
+        [Event(EventIds.MessageFailure, Level = EventLevel.Error, Version = 2)]
         public void MessageFailure(
             string Account,
             string TaskHub,
@@ -396,7 +405,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                114,
+                EventIds.MessageFailure,
                 Account,
                 TaskHub,
                 MessageId,
@@ -409,7 +418,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(115, Level = EventLevel.Error)]
+        [Event(EventIds.OrchestrationProcessingFailure, Level = EventLevel.Error)]
         public void OrchestrationProcessingFailure(
             string Account,
             string TaskHub,
@@ -419,7 +428,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                115,
+                EventIds.OrchestrationProcessingFailure,
                 Account,
                 TaskHub,
                 InstanceId,
@@ -428,7 +437,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(116, Level = EventLevel.Informational, Version = 2)]
+        [Event(EventIds.PendingOrchestratorMessageLimitReached, Level = EventLevel.Informational, Version = 2)]
         public void PendingOrchestratorMessageLimitReached(
             string Account,
             string TaskHub,
@@ -437,7 +446,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                116,
+                EventIds.PendingOrchestratorMessageLimitReached,
                 Account,
                 TaskHub,
                 PartitionId,
@@ -445,7 +454,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(117, Level = EventLevel.Informational)]
+        [Event(EventIds.WaitingForMoreMessages, Level = EventLevel.Informational)]
         public void WaitingForMoreMessages(
             string Account,
             string TaskHub,
@@ -453,14 +462,14 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                117,
+                EventIds.WaitingForMoreMessages,
                 Account,
                 TaskHub,
                 PartitionId,
                 ExtensionVersion);
         }
 
-        [Event(118, Level = EventLevel.Warning, Version = 3)]
+        [Event(EventIds.ReceivedOutOfOrderMessage, Level = EventLevel.Warning, Version = 3)]
         public void ReceivedOutOfOrderMessage(
             string Account,
             string TaskHub,
@@ -475,7 +484,7 @@ namespace DurableTask.AzureStorage
             string ExtensionVersion)
         {
             this.WriteEvent(
-                118,
+                EventIds.ReceivedOutOfOrderMessage,
                 Account,
                 TaskHub,
                 InstanceId,
@@ -489,7 +498,7 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
-        [Event(120, Level = EventLevel.Informational, Version = 2)]
+        [Event(EventIds.PartitionManagerInfo, Level = EventLevel.Informational, Version = 2)]
         public void PartitionManagerInfo(
             string Account,
             string TaskHub,
@@ -498,10 +507,17 @@ namespace DurableTask.AzureStorage
             string Details,
             string ExtensionVersion)
         {
-            this.WriteEvent(120, Account, TaskHub, WorkerName ?? string.Empty, PartitionId ?? string.Empty, Details, ExtensionVersion);
+            this.WriteEvent(
+                EventIds.PartitionManagerInfo,
+                Account,
+                TaskHub,
+                WorkerName ?? string.Empty,
+                PartitionId ?? string.Empty,
+                Details,
+                ExtensionVersion);
         }
 
-        [Event(121, Level = EventLevel.Warning, Version = 2)]
+        [Event(EventIds.PartitionManagerWarning, Level = EventLevel.Warning, Version = 2)]
         public void PartitionManagerWarning(
             string Account,
             string TaskHub,
@@ -510,7 +526,14 @@ namespace DurableTask.AzureStorage
             string Details,
             string ExtensionVersion)
         {
-            this.WriteEvent(121, Account, TaskHub, WorkerName ?? string.Empty, PartitionId ?? string.Empty, Details ?? string.Empty, ExtensionVersion);
+            this.WriteEvent(
+                EventIds.PartitionManagerWarning,
+                Account,
+                TaskHub,
+                WorkerName ?? string.Empty,
+                PartitionId ?? string.Empty,
+                Details ?? string.Empty,
+                ExtensionVersion);
         }
 
         [NonEvent]
@@ -525,7 +548,7 @@ namespace DurableTask.AzureStorage
             this.PartitionManagerError(account, taskHub, workerName, partitionId, exception.ToString(), ExtensionVersion);
         }
 
-        [Event(122, Level = EventLevel.Error, Version = 2)]
+        [Event(EventIds.PartitionManagerError, Level = EventLevel.Error, Version = 2)]
         public void PartitionManagerError(
             string Account,
             string TaskHub,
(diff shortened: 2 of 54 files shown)


src/DurableTask.AzureStorage/Logging/EventIds.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace DurableTask.AzureStorage.Logging
+{
+    static class EventIds
```
cgillum, line 3: NOTE: This class was auto-generated using a tool that reflects over `AnalyticsEventSource` and wrote out all the distinct event IDs and their names. I don't expect there to be any issues with event IDs being out of sync. Reviewers shouldn't bother reviewing this in too much detail.

src/DurableTask.AzureStorage/Logging/LogEvents.cs
```diff
@@ -0,0 +1,2382 @@
+﻿namespace DurableTask.AzureStorage.Logging
+{
+    using System;
+    using DurableTask.Core.Logging;
+    using Microsoft.Extensions.Logging;
+
+    static class LogEvents
```
cgillum, line 7: 99% of this code was auto-generated using a tool that reflects over `AnalyticsEventSource` and creates a class for every ETW event we were logging. I feel reasonably confident that the code accurately represents our existing ETW tracing. Reviewers shouldn't bother reviewing this in too much detail.

src/DurableTask.AzureStorage/Logging/LogHelper.cs
```diff
@@ -0,0 +1,826 @@
...
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage.Logging
+{
+    using System;
+    using DurableTask.Core.Logging;
+    using Microsoft.Extensions.Logging;
+
+    class LogHelper
```
cgillum, line 20: 99% of this code was auto-generated using a custom tool that reflects over `AnalyticsEventSource`. I'm fairly confident that it is accurate in how it accepts parameters and passes them to the log event class it creates. Reviewers shouldn't bother reviewing this in too much detail.

src/DurableTask.Core/OrchestrationInstance.cs
```diff
@@ -22,6 +22,12 @@ namespace DurableTask.Core
     [DataContract]
     public class OrchestrationInstance : IExtensibleDataObject
     {
+        internal readonly static OrchestrationInstance Empty = new OrchestrationInstance
```
davidmrdavid, line 25: Do we want this to be a valid instantiation? What will be the use of this?

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -131,7 +144,15 @@ OrchestrationStatus[] dedupeStatuses
         /// <returns>OrchestrationInstance that represents the orchestration that was created</returns>
         public Task<OrchestrationInstance> CreateOrchestrationInstanceAsync(string name, string version, object input)
         {
-            return InternalCreateOrchestrationInstanceWithRaisedEventAsync(name, version, null, input, null, null, null, null);
+            return this.InternalCreateOrchestrationInstanceWithRaisedEventAsync(
```
davidmrdavid, line 147: any chance we could comment in-place what all these `null`-ish values refer to :)

src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -56,6 +60,21 @@ public TaskHubWorker(IOrchestrationService orchestrationService)
         {
         }
 
+
+        /// <summary>
+        ///     Create a new TaskHubWorker with given OrchestrationService
```
davidmrdavid, line 65: nit: removing the whitespace in this summary :)

src/DurableTask.Core/WorkItemDispatcherContext.cs
```diff
@@ -26,24 +26,29 @@ public class WorkItemDispatcherContext
         /// <param name="dispatcherId">The context dispatcher id</param>
         public WorkItemDispatcherContext(string name, string id, string dispatcherId)
         {
-            Name = name;
-            Id = id;
-            DispatcherId = dispatcherId;
+            this.Name = name;
+            this.Id = id;
```
davidmrdavid, line 30: Is this `Id` the `WorkItemId`? Any thoughts on making this stand out a little more, considering that there's already another Id parameter here?

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -0,0 +1,1341 @@
...
+namespace DurableTask.Core.Logging
+{
+    using System;
+    using System.Text;
+    using DurableTask.Core.Command;
+    using DurableTask.Core.Common;
+    using DurableTask.Core.History;
+    using Microsoft.Extensions.Logging;
+
+    static class LogEvents
```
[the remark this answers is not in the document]
davidmrdavid: Whoops, didn't catch Chris' comments that this was autogenerated. Still though, since it's a part of the codebase, I'm thinking there's no harm in at least writing a top-class comment explaining
1. that it was originally autogenerated
2. what it is
perhaps this should be true about all these autogenerated files, which maybe can be autocommented too :)

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -434,5 +435,45 @@ public static Task DelayWithCancellation(TimeSpan timeout, CancellationToken can
...
+        {
+            if (TryGetTaskScheduledId(historyEvent, out int taskScheduledId))
+            {
+                return taskScheduledId;
+            }
+
+            return historyEvent.EventId;
+        }
+
+        static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int taskScheduledId)
```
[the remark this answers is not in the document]
cgillum: The reason for not throwing an exception is that not all `HistoryEvent` types actually have a valid event ID. I tried to indicate this in the method comment for `GetTaskEventId`. This is not "exceptional" which is why we don't throw here. In fact, I use this exact logic in DurableTask.SqlServer and go down different code paths depending on the outcome. I'll make this method public and document it to make this more clear. Note that the convention in .NET is to prefix methods like this with `Try` to indicate that there is a bool output that needs to be checked, so I think there is sufficient clarity for a .NET audience.

src/DurableTask.Core/Logging/LogEvents.cs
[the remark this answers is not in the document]
cgillum: This one was not auto-generated - I unfortunately had to write all of these from scratch. Only the one in DurableTask.AzureStorage was autogenerated. I'll add comments to both files accordingly.

src/DurableTask.Core/Logging/README.md
```diff
@@ -0,0 +1,58 @@
...
+| Event Source Type | Provider Name | Provider GUID |
+|-|-|-|
+| Structured | [DurableTask-AzureStorage](src/DurableTask.AzureStorage/AnalyticsEventSource.cs) | 4C4AD4A2-F396-5E18-01B6-618C12A10433 |
+| Structured | [DurableTask-SqlServer](https://github.com/cgillum/durabletask-sqlserver/blob/main/src/DurableTask.SqlServer/Logging/DefaultEventSource.cs) | 4BA38912-E64F-5FD2-170D-68AC65B1E58D |
+
+## Structured logging
+
+There are a large number of log events emitted by DurableTask.Core. The full list, including their event IDs, verbosity, and schemas can be found in [LogEvents.cs](src/DurableTask.Core/Logging/LogEvents.cs).
+
+Each log event is represented as a class that derives from [StructuredLogEvent](src/DurableTask.Core/Logging/StructuredLogEvent.cs). Each data field of the event is declared as a property with the [StructuredLogField](src/DurableTask.Core/Logging/StructuredLogFieldAttribute.cs) attribute. If a log event supports being written to Event Source, then it also implements the [IEventSourceEvent](src/DurableTask.Core/Logging/IEventSourceEvent.cs) interface.
```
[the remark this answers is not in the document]
cgillum: My bad. It looks like the links are supposed to be relative to the current directory. I'll fix these.

src/DurableTask.Core/Logging/README.md
```diff
@@ -0,0 +1,58 @@
...
+var worker = new TaskHubWorker(orchestrationService, loggerFactory);
+```
+
+## Event Source configuration
+
+Event Source logging is always enabled and does not have any _explicit_ configuration. However, there are currently two flavors of event source logging, _structured_ and _unstructured_. By default, Event Source logs are _unstructured_.
+
+Starting in **DurableTask.Core v2.4.0**, a more modern _structured_ Event Source logging is available. To enable the structured Event Source logging, a non-null `loggerFactory` must be configured. Even using [NullLoggerFactory.Instance](https://docs.microsoft.com/dotnet/api/microsoft.extensions.logging.abstractions.nullloggerfactory.instance) can be used to enabled structured event source logging.
+
+The difference between the legacy and structured event source providers is the provider GUID values, as shown in the following table.
```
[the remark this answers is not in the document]
cgillum: Apologies for the confusion. I reworded a few things here to make it hopefully less confusing.

src/DurableTask.Core/OrchestrationInstance.cs
cgillum: I'm not sure I understand what _valid instantiation_ means, but it looks like I'm not actually using this anywhere and can safely remove it. I can't remember exactly why I added this in the first place, but it was probably as a performance optimization somewhere at some point.

src/DurableTask.Core/TaskHubWorker.cs
cgillum: I'd prefer it be removed everywhere, but I was hesitant to create style inconsistencies within a single file, or to fix all occurrences of this because of how much it would inflate an already large PR (this codebase largely predates me).

src/DurableTask.Core/OrchestrationInstance.cs
davidmrdavid: By _valid instantiation_ I meant if it made sense to have an `OrchestrationInstance` object floating around that had everything as "empty" and whether or not that should be an allowed object in the first place.

But yeah, sounds like we can safely remove it :)

src/DurableTask.Core/WorkItemDispatcherContext.cs
cgillum: Unfortunately it's not even a work item ID. It's very non-obvious why there are two different "Id" properties here (this code also predates me) but I'm hesitant to change it as part of this PR. A lot of the existing DurableTask.Core could use a good refactoring.  :(

src/DurableTask.Core/TaskHubClient.cs
cgillum: I'll just update all these to use named parameters. That's a bit more future-proof.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -862,7 +793,6 @@ public void GeneralWarning(string Account, string TaskHub, string Details, strin
             string ETag,
             string ExtensionVersion)
         {
-            EnsureLogicalTraceActivityId();
```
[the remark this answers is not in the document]
cgillum: Unfortunately the answer to these questions are long and complicated. The short version is that this refactor allowed me to consolidate them into a single place, `LoggingExtensions.LogDurableEvent`.

The longer answer is that these EventSource APIs depend on thread static state for logging trace activity IDs (GUIDs that correlate multiple trace messages together as one logical "activity" - not to be confused with Durable Task "activities" or "activity functions"). These are the values you'll see in the `ActivityId` column in all (or most) of our Kusto tables. Thread-static variables don't work at all when writing async code, so I needed to invent another way of propagating this "activity" information. If you look at the top of this file, you'll see a `AsyncLocal<Guid> ActivityIdState` variable which is a more modern version of the thread-static variable concept that is compatible with async code. I'm using that to keep track of ambient trace activity IDs. The `EnsureLogicalTraceActivityId` was used simply to copy the `AsyncLocal<Guid>` value into the thread-static variable so that the call to `WriteEvent` had the correct value. So we still do this, but the refactor allowed me to do it in a single place rather than calling this method inside every one of our trace methods.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -470,19 +522,17 @@ public Task<OrchestrationInstance> CreateOrchestrationInstanceAsync(string name,
...
-                new TaskMessage
-                {
-                    OrchestrationInstance = orchestrationInstance,
-                    Event = startedEvent
-                }
+                OrchestrationInstance = orchestrationInstance,
+                Event = startedEvent
             };
 
+            TaskMessage eventMessage = null;
```
ConnorMcMahon, line 531: Can we move this logic into the if statement currently at line 556?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -213,9 +219,10 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
 
             OrchestrationState instanceState = null;
 
-            if (!ReconcileMessagesWithState(workItem))
+            if (!this.ReconcileMessagesWithState(workItem))
             {
                 // TODO : mark an orchestration as faulted if there is data corruption
+                this.logHelper?.DroppingOrchestrationWorkItem(workItem, "Received result for a deleted orchestration");
```
ConnorMcMahon, line 225: I notice that sometimes we use `?.` to reference `this.logHelper` and othertimes we don't. I would prefer if we don't need to, but either way we should probably be consistent.

samples/DurableTask.Samples/DurableTask.Samples.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFramework>net461</TargetFramework>
```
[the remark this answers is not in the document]
cgillum: IIRC, semver doesn't really have recommendations about how to version a library with respect to its dependencies so I thought of this as more of a judgement call. Note that .NET 4.6 has been available for over 5 years now.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -228,22 +223,16 @@ static CloudQueue GetQueueInternal(CloudQueueClient queueClient, string taskHub,
         }
 
         static BlobLeaseManager GetBlobLeaseManager(
-            string taskHub,
-            string workerName,
+            AzureStorageOrchestrationServiceSettings settings,
```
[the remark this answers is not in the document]
cgillum: Yes, this was my main motivation...

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -62,8 +63,6 @@ public sealed class AzureStorageOrchestrationService :
 
         readonly ITrackingStore trackingStore;
 
-        readonly TableEntityConverter tableEntityConverter;
```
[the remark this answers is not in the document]
cgillum: Yep, VS2019 pointed this out to me.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -764,18 +748,33 @@ static TaskHubInfo GetTaskHubInfo(string taskHub, int partitionCount)
...
-            // and on the logical call context. AnalyticsEventSource will use this 
-            // activity ID for all trace operations.
-            Guid traceActivityId = Guid.NewGuid();
+            // Starting in DurableTask.Core v2.4.0, a new trace activity will already be
+            // started and we don't need to start one ourselves.
+            // TODO: When distributed correlation is merged, use that instead.
+            Guid traceActivityId;
+            if (useExisting && EventSource.CurrentThreadActivityId != Guid.Empty)
+            {
+                traceActivityId = EventSource.CurrentThreadActivityId;
```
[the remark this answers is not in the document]
cgillum: Yeah, I'm actually not sure that this is needed at all because all traces we write now grab the `AsyncLocal<T>` value from `StructuredEventSource` in DurableTask.Core. I plan to revisit this after merging Tsuyoshi's changes in the `correlation` branch.

src/DurableTask.AzureStorage/Logging/EventIds.cs
cgillum: We don't have any legacy ETW events in DurableTask.AzureStorage. We're keeping all of them. But in general, I think the reason to add a comment about code being auto-generated is if we don't want future maintainers modifying that code directly. Since this is the new source of truth, future edits should be done here w/out any more code generation.

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -145,11 +168,10 @@ internal virtual async Task<bool> UpdateQueueMetrics()
             catch (StorageException e) when (e.RequestInformation?.HttpStatusCode == 404)
             {
                 // The queues are not yet provisioned.
-                AnalyticsEventSource.Log.GeneralWarning(
+                this.settings.Logger.GeneralWarning(
```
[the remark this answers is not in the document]
cgillum: I'll admit this is a riskier change, but even a "dummy" `AzureStorageOrchestrationServiceSettings` object can correctly initialize the `Logger` property.

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -80,6 +90,19 @@ public DisconnectedPerformanceMonitor(string storageConnectionString, string tas
 
         internal QueueMetricHistory WorkItemQueueLatencies => this.workItemQueueLatencies;
 
+        static AzureStorageOrchestrationServiceSettings GetSettings(
```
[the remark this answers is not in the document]
cgillum: Agreed - this is the one place where the cleanliness of my solution kinda fell apart. One minor advantage of this part of the refactor, however, is that it allows users to pass in their own `AzureStorageOrchestrationServiceSettings` if they want. But I also agree that passing the logger everywhere isn't an ideal alternative either, so I'll stick with this approach as long as we feel like it functionally solves the problem.

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -0,0 +1,567 @@
...
+        }
+
+        [Event(EventIds.ProcessingOrchestrationMessage, Level = EventLevel.Verbose, Version = 1)]
+        internal void ProcessingOrchestrationMessage(
+            string InstanceId,
+            string ExecutionId,
+            string EventType,
+            int TaskEventId)
+        {
+            if (this.IsEnabled(EventLevel.Verbose))
```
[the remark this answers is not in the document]
cgillum: Right, this benefits non-Antares hosting. I think I added these filters because similar filters existing in the legacy EventSource implementation.

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -0,0 +1,567 @@
...
+            {
+                // TODO: Use WriteEventCore for better performance
```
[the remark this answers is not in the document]
cgillum: I think I might have to open an issue to track this if I want to finish this PR anytime soon. Using `WriteEventCore` is very manual and can be quite dangerous, so it might be another opportunity to use code generation...

src/DurableTask.Core/Logging/StructuredLogEvent.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// Abstract base class for all structured log events. This base class implements
+    /// <see cref="IReadOnlyDictionary{String, Object}"/>using reflection to make it easy
+    /// to support structured log events for providers like Application Insights.
+    /// </summary>
+    public abstract class StructuredLogEvent : ILogEvent, IReadOnlyDictionary<string, object>
+    {
+        static readonly ConcurrentDictionary<Type, ConcurrentDictionary<string, PropertyInfo>> SharedPropertiesCache =
+            new ConcurrentDictionary<Type, ConcurrentDictionary<string, PropertyInfo>>();
+
+        ConcurrentDictionary<string, PropertyInfo> Properties => GetProperties(this.GetType());
```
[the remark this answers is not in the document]
cgillum: Yes, I was definitely concerned about performance. A few things worth noting about my implementation:

* We reflect over the properties only once per type and save the results into a static dictionary (`SharedPropertiesCache`). Subsequent log events of the same type will use the cached property list.
* We only go through this reflection code path if there is an `ILogger` implementation which asks for it (e.g. Application Insights plus a log level configuration of "Informational"). Default scenarios should not invoke this reflection code at all, so it's "pay-for-play" in the classic sense.
* We still pay the cost of calls to `PropertyInfo.GetValue(this)` every time we actually need to read any value. I'm not sure how we can do better than this other than doing build-time code generation (.NET 5) or replacing reflection with hardcoded property references inside each LogEvent implementation, which I thought created too much of a maintenance burden.

Since this shouldn't impact the default code path, I personally feel okay with it. Agreed we should run performance testing against this to make sure we're not creating any major regressions.

One thing I'm considering adding on top of this is caching the result of `GetLogMessage()`. We'll call that once for every `ILogger` in the pipeline, which seems pretty wasteful in terms of string allocations.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -103,6 +111,9 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
 
                 if (taskMessage.Event.EventType != EventType.TaskScheduled)
                 {
+                    this.logHelper.TaskActivityDispatcherError(
+                        workItem, 
+                        $"The activity worker received an event of type '{taskMessage.Event.EventType}' but only '{EventType.TaskScheduled}' is supported.");
```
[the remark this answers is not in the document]
cgillum: I sure hope so! 😅

src/DurableTask.Core/TaskHubClient.cs
cgillum: Yes we can! I'm not sure why this method was previously so complicated...

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Oops, I thought I removed all those. Will fix and make sure that `logHelper` is never null.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -90,11 +57,11 @@ private static void EnsureLogicalTraceActivityId()
             string TargetExecutionId,
             long SequenceNumber,
             int Episode,
+            string AppName,
```
cgillum, line 60: I've added `AppName` columns to every log so that we can much more easily find these traces when searching in Kusto.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -245,65 +216,75 @@ private static void EnsureLogicalTraceActivityId()
...
-        public void GeneralError(string Account, string TaskHub, string Details, string ExtensionVersion)
+        [Event(EventIds.GeneralError, Level = EventLevel.Error, Version = 2)]
+        public void GeneralError(string Account, string TaskHub, string Details, string AppName, string ExtensionVersion)
         {
-            EnsureLogicalTraceActivityId();
-            this.WriteEvent(107, Account, TaskHub, Details, ExtensionVersion);
+            this.WriteEvent(EventIds.GeneralError, Account, TaskHub, Details, AppName, ExtensionVersion);
         }
 
-        [Event(108, Level = EventLevel.Warning, Version = 2, Message = "A duplicate message was detected. This can indicate a potential performance problem. Message ID = '{2}'. DequeueCount = {3}.")]
```
cgillum, line 258: I removed the `Message` property for these events because we don't actually use them and they easily get out of date.

src/DurableTask.AzureStorage/Messaging/Session.cs
```diff
@@ -72,8 +74,7 @@ public void TraceProcessingMessage(MessageData data, bool isExtendedSession)
                 Math.Max(0, (int)DateTimeOffset.UtcNow.Subtract(queueMessage.InsertionTime.Value).TotalMilliseconds),
                 data.SequenceNumber,
                 data.Episode.GetValueOrDefault(-1),
-                isExtendedSession,
-                Utils.ExtensionVersion);
```
cgillum, line 76: You'll notice that I got rid of `Utils.ExtensionVersion` in these method calls. We still log this information, but we just do it in a different place now inside LogEvents.cs. We only log it for ETW events. There is no actual behavior change here.

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -153,29 +150,30 @@ public async Task CreateLeaseIfNotExistAsync(string partitionId)
...
-                    this.storageAccountName,
-                    this.taskHubName,
-                    this.workerName,
-                    partitionId,
-                    string.Format(
-                        CultureInfo.InvariantCulture,
-                        "CreateLeaseIfNotExistAsync - leaseContainerName: {0}, consumerGroupName: {1}, partitionId: {2}, blobPrefix: {3}, exception: {4}.",
-                        this.leaseContainerName,
-                        this.consumerGroupName,
+                if (se.RequestInformation.HttpStatusCode != 409)
```
cgillum, line 159: I added this check simply because we were getting a lot of unnecessary noise in the logs on startup.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -766,51 +763,54 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string
...
                     instanceId,
                     DateTime.MinValue.ToString(),
                     DateTime.MinValue.ToString(),
-                    null,
-                    deletionStats.StorageRequests,
-                    stopwatch.ElapsedMilliseconds,
-                    Utils.ExtensionVersion);
+                    string.Empty,
+                    result.StorageRequests,
+                    result.InstancesDeleted,
```
cgillum, line 776: While auditing the logs we write, I noticed that we don't actually mention how many instances were deleted. I thought this information would be useful, so I added it to this log event.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -865,15 +865,14 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTim
...
             int currentEpisodeNumber = 0;
 
-            AnalyticsEventSource.Log.InstanceStatusUpdate(
+            this.settings.Logger.InstanceStatusUpdate(
                 this.storageAccountName,
                 this.taskHubName,
                 executionStartedEvent.OrchestrationInstance.InstanceId,
                 executionStartedEvent.OrchestrationInstance.ExecutionId,
-                executionStartedEvent.EventType.ToString(),
+                OrchestrationStatus.Pending,
```
cgillum, line 873: I don't know why I didn't make this change sooner. The intent was always to communicate the runtime status, but for some reason I was always instead logging the event type, which is not something a viewer of these logs would intuitively understand as it relates to the orchestration instance status.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1244,24 +1241,23 @@ static string GetBlobName(DynamicTableEntity entity, string property)
...
                 if (ex.RequestInformation.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
                 {
-                    AnalyticsEventSource.Log.SplitBrainDetected(
+                    this.settings.Logger.SplitBrainDetected(
                         this.storageAccountName,
                         this.taskHubName,
                         instanceId,
                         executionId,
-                        historyEventBatch.Count,
+                        historyEventBatch.Count - 1, // exclude sentinel from count
```
cgillum, line 1256: This is an old "off by one" error I noticed while inspecting the new logs that were being emitted.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -41,6 +42,23 @@ public static class Utils
...
+        /// <summary>
+        /// Gets or sets the name of the app, for use when writing structured event source traces.
+        /// </summary>
+        /// <remarks>
+        /// The default value comes from the WEBSITE_SITE_NAME environment variable, which is defined
+        /// in Azure App Service. Other environments can use DTFX_APP_NAME to set this value.
+        /// </remarks>
+        public static string AppName { get; set; } = 
+            Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME") ??
+            Environment.GetEnvironmentVariable("DTFX_APP_NAME") ??
```
cgillum, line 59: I felt bad for adding Antares-specific environment variable names (`WEBSITE_SITE_NAME`) in this code, but it was the easiest way to add the AppName column to our ETW traces for Kusto. In an attempt to make this useful outside of Antares, I also added support for a `DTFX_APP_NAME` environment variable, which is not tied to Antares.

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -156,21 +156,24 @@ public async Task CreateLeaseIfNotExistAsync(string partitionId)
...
+                {
+                    // eat any storage exception related to conflict
```
ConnorMcMahon, line 161: small nit: This comment makes a bit more sense above the if statement's conditional imo. I got confused reading it down here, as I assumed it was saying we were eating the exception that we are logging.

src/DurableTask.Core/Common/Utils.cs
ConnorMcMahon: Another possible way to address this is to add it AzureStorageOrchestrationServiceSettings as a parameter, and leave it up to Durable Functions to provide a value. The problem is that then you either need to pass around AzureStorageOrchestrationServiceSettings everywhere, or you need to set this static variable when you instantiate your AzureStorageOrchestrationService.

While the code is a bit messier, it keeps Antares code out of DurableTask.AzureStorage, and gives consumers a bit more flexibility in how they determine their app name.

cgillum: I considered this originally, but ultimately changed my mind because it's a lot of work for questionable benefit. AppName is only used for ETW logging and I expect that very few folks, if any, will actually care about it.

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -0,0 +1,567 @@
...
+            this.WriteEvent(
+                EventIds.FetchingInstanceState,
+                InstanceId,
+                ExecutionId ?? string.Empty);
+        }
+
+        [Event(EventIds.FetchingInstanceHistory, Level = EventLevel.Informational, Version = 1)]
+        internal void FetchingInstanceHistory(string InstanceId, string ExecutionId)
+        {
+            this.WriteEvent(
```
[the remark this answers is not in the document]
cgillum: Added remaining guards in the latest iteration.

PR #426, opened 2020-07-28
Fix: Correlation branch test is ignored

TsuyoshiUshio:
On the correlation branch, the testing framework ignore the testing. I did experiment on this branch. 

# What I did was: 

* Add Net451 test 
* Configure Binding Redirect NOTE: [Binding Redirect](https://stackoverflow.com/questions/43365736/assembly-binding-redirect-how-and-why), https://github.com/microsoft/ApplicationInsights-dotnet/issues/1672
* Upgrade Testing Framework

# What was the issue? 

When we add `net451` it returns this error.  That only happens when we run the test.  This issue is solved by using Binding Redirect. 

```
DurableTask.AzureStorage.Tests.Correlation.CorrelationScenarioTest.ContinueAsNewAsync threw exception:
        System.IO.FileNotFoundException: Could not load file or assembly 'System.Memory, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
      Stack Trace:
        Activity.TrySetTraceIdFromParent()
        Activity.GenerateW3CId()
        Activity.Start()
        DependencyTrackingTelemetryModule.PrepareFirstActivity()
        DependencyTrackingTelemetryModule.Initialize(TelemetryConfiguration configuration)
        TelemetryActivator.SetUpTelemetryClient(Action`1 onSend, String instrumentationKey) line 86
        TelemetryActivator.Initialize(Action`1 onSend, String instrumentationKey) line 39
        <ExtractTelemetry>d__7.MoveNext() line 650
        --- End of stack trace from previous location where exception was thrown ---
        TaskAwaiter.ThrowForNonSuccess(Task task)
        TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
        TaskAwaiter.GetResult()
        <ExecuteOrchestrationAsync>d__5.MoveNext() line 619
```
Then the test works perfectly on my local machine. 

However, AzDO doesn't detect the test. So I upgrade the Test framework, then AzDO successfully start the test. 

# Currently we test only on .NetFramework since we want to use parallel test feature. 
https://docs.microsoft.com/en-us/azure/devops/pipelines/test/parallel-testing-vstest?view=azure-devops
https://github.com/microsoft/azure-pipelines-tasks/issues/9880

# TODO 
I only test for Azure Storage. I didn't upgrade the test for Service Bus and Emulator that is supported on the pipeline. It also might required.


test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -2,15 +2,17 @@
...
   </PropertyGroup>
 
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
     <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" version="3.2.1" />
     <PackageReference Include="Moq" Version="4.10.0" />
-    <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
-    <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="2.1.2" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.1.2" />
```
cgillum, line 15: Interesting! So it was really outdated. We should try to make the same change for all test projects.

Issue #410, opened 2020-06-19
Improve DurableTask.AzureStorage partition management


cgillum:
## Overview
The **DurableTask.AzureStorage** provider uses partitioning and blob leases to ensure that an orchestration instance runs on only one worker instance at a time. More details on this design can be found [here](https://github.com/Azure/durabletask/tree/master/src/DurableTask.AzureStorage#partitioning). When the app is scaled out to additional workers, the new workers will attempt to steal partition leases to properly balance the distribution of partitions across all workers.

This stealing can be problematic because it is not cooperative. "Split-brain" can occur if a worker is executing an orchestration instance that is assigned to a particular partition and that partition is stolen by another worker mid-execution. In many cases, the result is often duplicate activity execution. When sub-orchestrations are involved, the impact of duplicate execution could be even more severe.

This issue tracks changing the partition management to remove the split-brain risks associated with lease stealing.

## Proposed design
One of the primary design goals is to try to _extend_ the existing design to minimize risks associated with code churn. This is the proposal I came up with, which I briefly discussed with @sebastianburckhardt a while back, in bullet-list form:

- There are _two_ lease blobs for each partition: an **intent** blob and an **ownership** blob.
- A worker will only poll for messages in a queue if it owns **both** the _intent_ and _ownership_ blob leases.
- All nodes try to acquire _intent_ lease blobs. If there is an imbalance, _intent_ leases can be stolen to create balance.
- For the most part, _intent_ lease management uses the exact same algorithm as the existing lease manager.
- Nodes poll for _ownership_ leases only if they own the corresponding _intent_ lease. _Ownership_ leases are never stolen.
- If an _intent_ lease is stolen, the holder of the corresponding _ownership_ lease must stop processing new messages for that lease. Once all messages for that lease are drained. They will release the ownership lease so that it can be acquired by the new owner of the _intent_ lease.

### Suggested polling frequencies (TBD)

  | Polling frequency | Renewal frequency | Lease lifetime
-- | -- | -- | --
Intent blob |   |   |  
Ownership blob |   |   |  

## Benefits
- Eliminates split-brain in the happy path. Lease transitions are cooperative.
- No additional overhead in steady state compared to current model.
- The design is relatively simple and easy to understand.

## Disadvantages
- Lease transitions are slower. This can result in lower scale-out speed for entities and orchestrations.

olitomlinson:
> Lease transitions are slower. This can result in lower scale-out speed for entities and orchestrations.

[...]

PR #427, opened 2020-07-28
Make partition management in DurableTask.AzureStorage more conservative

ConnorMcMahon:
Our current approach to partition management frequently allows split
brain to occur. This is mainly due to our lease rebalancing with lease
stealing.

The new approach goes with two separate sets of leases: intent leases
and owenership leases. Intent leases follow the old logic, with lease
stealing, while ownership leases do not allow this. A worker must own an
intent lease before attempting to obtain the ownership lease, and if the
worker loses either the ownership or intent lease, it will stop
retrieving new items. This means that when another worker is stealing a
partition, the previous worker has time to drain its current items
before the new worker starts processing items, greatly reducing the
chance of split brain.

Addresses #410


src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -20,15 +20,17 @@ namespace DurableTask.AzureStorage.Messaging
     using System.Threading;
     using System.Threading.Tasks;
     using DurableTask.AzureStorage.Monitoring;
+    using ImpromptuInterface;
```
cgillum, line 23: The `ImpromptuInterface` using statement is not being used.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -20,15 +20,17 @@ namespace DurableTask.AzureStorage.Messaging
...
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Queue;
 
     class ControlQueue : TaskHubQueue, IDisposable
     {
         static readonly List<MessageData> EmptyMessageList = new List<MessageData>();
 
-        readonly CancellationTokenSource releaseTokenSource;
-        readonly CancellationToken releaseCancellationToken;
+
```
cgillum, line 31: nit: remove extra newline

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -204,6 +206,12 @@ public void Release()
             this.IsReleased = true;
         }
 
+        public void Reaquire()
+        {
+            this.releaseTokenSource = new CancellationTokenSource();
```
cgillum, line 211: Do we need to dispose the old `releaseTokenSource` before overwriting it with a new one?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -525,7 +585,20 @@ void ReportStats()
                 Utils.ExtensionVersion);
         }
 
-        async Task IPartitionObserver<BlobLease>.OnPartitionAcquiredAsync(BlobLease lease)
+        Task OnIntentLeaseAquiredAsync(BlobLease lease)
+        {
+            this.orchestrationSessionManager.TryReaquireQueue(lease.PartitionId);
```
cgillum, line 590: Naming: The method name `TryReacquireQueue` implies that we previously acquired a queue and are now trying to acquire it again. But it seems to me that we're actually acquiring it for the first time. If that's true, then why `TryReacquireQueue` instead of simply `TryAcquireQueue`? I'm having doubts about whether I'm understanding the design correctly because the method name doesn't seem to match the behavior.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -85,11 +85,75 @@ public void RemoveQueue(string partitionId)
...
+                }
+            }
+            else
+            {
+                AnalyticsEventSource.Log.PartitionManagerWarning(
+                    this.storageAccountName,
+                    this.settings.TaskHubName,
+                    this.settings.WorkerId,
+                    partitionId,
+                    $"Attempted to reaquire control queue {partitionId}, which wasn't being watched'!",
```
cgillum, line 139:
```suggestion
                    $"Attempted to reaquire control queue {partitionId}, which wasn't being watched!",
```

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -85,11 +85,75 @@ public void RemoveQueue(string partitionId)
...
+
+        public bool IsControlQueueReceivingMessages(string partitionId)
+        {
+            return this.ownedControlQueues.TryGetValue(partitionId, out ControlQueue controlQueue)
+                && !controlQueue.IsReleased;
+        }
+
+        public bool IsControlQueueProcessingMessages(string partitionId)
+        {
+            return this.activeOrchestrationSessions.Values.Where(session => string.Equals(session.ControlQueue.Name, partitionId)).Any();
```
cgillum, line 154:
```suggestion
            return this.activeOrchestrationSessions.Values.Any(session => string.Equals(session.ControlQueue.Name, partitionId));
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
ConnorMcMahon: So this case is when we have lost the intent lease and stopped fetching new message on this partition due to cancelling the cancellation tokens when we "released" the queue, but have reacquired the intent lease before we have lost the ownership lease.

The most clear name would be something like "ResumeFetchingMessagesIfStillOwnPartition" or something like that.

cgillum: If that's indeed the behavior then the current name is probably fine. It seemed to me that the very first attempt to acquire the lease would result in this call, so I think I'm probably still not understanding the flow correctly. I'll probably need to run this through the debugger and look through the logs to help me understand.

ConnorMcMahon: This case should happen really rarely, it is essentially just to avoid a situation where the worker has both leases and is not processing new messages, because that is a nasty scenario where the partition is essentially dead until it loses the ownership lease.

ConnorMcMahon:
One potential option for this PR, considering I am mostly reusing existing code, is to make this mode optional (maybe even opt-in) at first. It would not be too hard to hide this behavior behind a switch. I don't think that will be necessary at this time, but if we find that this is blocking Durable 2.3.0 release, then it is a definite option to get it in peoples hands even if we don't think it is 100% ready.

cgillum:
It's not a bad idea to include a kill switch in case we find regressions after shipping this. I would prefer opt-out vs. opt-in though since this is a reliability improvement and not intended to be a breaking change.

My dream would be the keep the same interfaces we had before but just swap implementations based on this flag, but I realize that could be a big departure from where we are now and may not be practical at this point.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: I built your PR locally and ran some automated tests. I'm seeing "reacquire" warnings at the start of every test. Is this expected?
```
2020-07-30T13:58:46.349Z: [0000, 0000] PartitionManagerWarning: Account=devstoreaccount1, TaskHub=test, WorkerName=CGILLUM-DEV, PartitionId=test-control-00, Details=Attempted to reaquire control queue test-control-00, which wasn't being watched'!, ExtensionVersion=1.7.8
2020-07-30T13:58:46.349Z: [0000, 0000] PartitionManagerWarning: Account=devstoreaccount1, TaskHub=test, WorkerName=CGILLUM-DEV, PartitionId=test-control-01, Details=Attempted to reaquire control queue test-control-01, which wasn't being watched'!, ExtensionVersion=1.7.8
2020-07-30T13:58:46.349Z: [0000, 0000] PartitionManagerWarning: Account=devstoreaccount1, TaskHub=test, WorkerName=CGILLUM-DEV, PartitionId=test-control-02, Details=Attempted to reaquire control queue test-control-02, which wasn't being watched'!, ExtensionVersion=1.7.8
2020-07-30T13:58:46.349Z: [0000, 0000] PartitionManagerWarning: Account=devstoreaccount1, TaskHub=test, WorkerName=CGILLUM-DEV, PartitionId=test-control-03, Details=Attempted to reaquire control queue test-control-03, which wasn't being watched'!, ExtensionVersion=1.7.8
```

cgillum:
BTW, another thing I noticed in local testing is that the partition manager logs don't always distinguish between intent and ownership leases. For example, when the host starts up, we get the following log twice in the first 5 seconds:

```
2020-07-30T14:12:20.763Z: [0000, 0000] LeaseAcquisitionStarted: Account=devstoreaccount1, TaskHub=test, WorkerName=CGILLUM-DEV, PartitionId=test-control-00, ExtensionVersion=1.7.8, formattedMessage=Host 'CGILLUM-DEV' attempting to take lease for PartitionId 'test-control-00'.
2020-07-30T14:12:25.953Z: [0000, 0000] LeaseAcquisitionStarted: Account=devstoreaccount1, TaskHub=test, WorkerName=CGILLUM-DEV, PartitionId=test-control-00, ExtensionVersion=1.7.8, formattedMessage=Host 'CGILLUM-DEV' attempting to take lease for PartitionId 'test-control-00'.
```

I think we'll want to include information that indicates the type of lease we're operating on. The `PartitionManagerInfo` trace includes this in the `consumerGroupName` field, but most other traces do not. For example, I also noticed that `PartitionManagerWarning` does not include this information.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
ConnorMcMahon: This is currently a bug in the tracing level I would say. Almost all of the time, when we run "Reacquire" we expect there is no lease to reacquire, due to the fact that we would have had to lost the intent lease and then got the lease back before we disposed of the control queue. Really, this should be at most a verbose log.

ConnorMcMahon:
> My dream would be the keep the same interfaces we had before but just swap implementations based on this flag, but I realize that could be a big departure from where we are now and may not be practical at this point.

In regards to this, I didn't keep the old exact same interface due to the fact that we have different leaseacquire/leaselost behavior for intent and ownership queues, meaning we would need two implementations of ILeaseManager<T>. I started to do that by having two empty subclasses of BlobLease so that I could have these two separate implementations, but it really started to make the code around BlobLeaseManager ugly to support two dummy classes needlessly, hence the refactor.

That being said, putting this behind flags would not be that difficult, as the overall logic of the old `PartitionManager` is essentially just `LeaseCollectionManager` with the default parameter.

That being said, I am noticing was the need to add new methods and make method calls across `AzureStorageOrchestrationService`, `OrchestrationSessionManager`, `LeaseCollectionManager`, and `ControlQueue` which indicates to me that these classes may be too tightly coupled. I think things could be made cleaner if the relationship between these was refactored a bit, but that would increase the scope/risk of this PR.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionManager.cs
```diff
@@ -20,16 +20,19 @@ namespace DurableTask.AzureStorage.Partitioning
     using System.Threading;
     using System.Threading.Tasks;
 
-    sealed class PartitionManager<T> where T : Lease
+    sealed class LeaseCollectionManager<T> where T : Lease
```
amdeel, line 23: Why did you decide to change the name of this class? I think it makes it less apparent what this class actually does.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -65,8 +64,10 @@ public sealed class AzureStorageOrchestrationService :
         readonly TableEntityConverter tableEntityConverter;
 
         readonly ResettableLazy<Task> taskHubCreator;
-        readonly BlobLeaseManager leaseManager;
-        readonly PartitionManager<BlobLease> partitionManager;
+        readonly BlobLeaseManager intentLeaseManager;
+        readonly LeaseCollectionManager<BlobLease> intentLeaseCollectionManager;
+        readonly BlobLeaseManager ownershipLeaseManager;
+        readonly LeaseCollectionManager<BlobLease> ownershipLeaseCollectionManager;
```
amdeel, line 70: Did you consider a design where the LeaseCollectionManager managed both intent and ownership BlobLeaseManagers?

ConnorMcMahon: I definitely considered it. The problem would be that I would need to duplicate much of the state in LeaseCollectionManager for both sets of leases, and that the core logic of that class is really only designed to handle one set of leases.

I think a better option would be to add a new `PartitionManager` class that handles the `LeaseCollectionManager` for both intent leases and blob leases to prevent `AzureStorageOrchestrationService` from having to manage it. This becomes especially important if I allow customers to use the legacy partition management approach, as I could have a different PartitionManager implementation for that case.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionManager.cs
ConnorMcMahon: Hopefully the new name is more clear.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -202,6 +202,14 @@ public void Release()
             this.IsReleased = true;
         }
 
+        public void ResumeListening()
```
ConnorMcMahon, line 205: I actually don't think this is enough to resume listening, I think I need to spin off a new dequeue loop.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -182,6 +182,12 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public bool ThrowExceptionOnInvalidDedupeStatus { get; set; } = false;
 
+        /// <summary>
+        /// Use the legacy partition management strategy, which improves performance at the cost of being less resiliant
+        /// to split brain.
+        /// </summary>
+        public bool UseLegacyPartitionManagement { get; set; } = false;
```
ConnorMcMahon, line 189: Without a bit more testing, I'm not sure I trust the new code path's performance enough to enable it by default. We will see how much time I have for testing before 2.3.0 release now that I am on call.

src/DurableTask.AzureStorage/Partitioning/IPartitionManager.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
cgillum, line 1: Style consistency: each new file needs a copyright header. Also, the `using` statements need to go inside the `namespace` block.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -224,14 +225,15 @@ static CloudQueue GetQueueInternal(CloudQueueClient queueClient, string taskHub,
...
+            string leaseType,
             CloudStorageAccount account,
             AzureStorageOrchestrationServiceStats stats)
         {
             return new BlobLeaseManager(
                 settings,
                 leaseContainerName: settings.TaskHubName.ToLowerInvariant() + "-leases",
                 blobPrefix: string.Empty,
-                consumerGroupName: "default",
+                consumerGroupName: leaseType,
```
cgillum, line 236: The `consumerGroupName` parameter name is actually something that was copied from the Event Hubs message processor code and doesn't have any meaningful application in our code. Feel free to rename it so that it more accurately reflects this new use case.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -64,9 +63,8 @@ public sealed class AzureStorageOrchestrationService :
         readonly ITrackingStore trackingStore;
 
         readonly ResettableLazy<Task> taskHubCreator;
-        readonly BlobLeaseManager leaseManager;
-        readonly PartitionManager<BlobLease> partitionManager;
         readonly OrchestrationSessionManager orchestrationSessionManager;
+        readonly IPartitionManager partitionManager;
```
cgillum, line 67: So far I really like this new refactoring you did. The code is a lot easier to follow!

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
cgillum: Can I interpret this to mean you're going to change this in a future iteration?

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -71,6 +71,7 @@ public async Task DeleteTaskHub()
                 TaskHubName = taskHubName,
                 StorageConnectionString = storageConnectionString,
                 WorkerId = workerId,
+                UseLegacyPartitionManagement = true
```
cgillum, line 74: I see this is currently hardcoded to `true`, but it also seems like you have code that handles the case where it is set to `false`. Are you planning to have any automated tests configured to run with the "safe" partition management strategy?

src/DurableTask.AzureStorage/Partitioning/IPartitionManager.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DurableTask.AzureStorage.Partitioning
+{
+    internal interface IPartitionManager
```
cgillum, line 8: nit: you can drop the `internal` modifier since this codebase doesn't use default visibility modifiers.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancerOptions.cs
```diff
@@ -38,30 +38,23 @@ class PartitionManagerOptions
         public TimeSpan LeaseInterval { get; set; } = TimeSpan.FromSeconds(30);
 
         /// <summary>
-        /// Maximum number of receiver clients created for each host instance. Once the max is reached host will start rebalancing partitions 
-        /// among receiver clients already created.
+        /// Determines whether or not this set of leases should utilize lease stealing logic for rebalancing
+        /// during scale-out operations.
         /// </summary>
-        public int MaxReceiveClients { get; set; } = 4;
```
cgillum, line 44: Thanks for removing these unused settings. I should have done a more thorough audit of these after I originally incorporated this code into our repo.

src/DurableTask.AzureStorage/Partitioning/SafePartitionManager.cs
```diff
@@ -0,0 +1,159 @@
...
+            await Task.WhenAll(
+                this.intentLeaseCollectionManager.SubscribeAsync(
+                    this.service.OnIntentLeaseAquiredAsync,
+                    this.service.OnIntentLeaseReleasedAsync),
+                this.ownershipLeaseCollectionManager.SubscribeAsync(
+                    this.service.OnOwnershipLeaseAquiredAsync,
+                    this.service.OnOwnershipLeaseReleasedAsync)
+            );
+
+
```
cgillum, line 136: nit: extra newline

src/DurableTask.AzureStorage/Partitioning/SafePartitionManager.cs
```diff
@@ -0,0 +1,159 @@
...
+            this.ownershipLeaseCollectionManager = new LeaseCollectionBalancer<BlobLease>(
+                "ownership",
+                this.settings,
+                storageAccountName,
+                this.ownershipLeaseManager,
+                new LeaseCollectionBalancerOptions
+                {
+                    AcquireInterval = TimeSpan.FromSeconds(5),
+                    RenewInterval = TimeSpan.FromSeconds(10),
+                    LeaseInterval = TimeSpan.FromSeconds(15),
```
cgillum, line 78: What are your thoughts on making this configurable? Would these hardcoded values cause any problems if the configured values for the intent lease manager are changed from their defaults?

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
ConnorMcMahon: So currently the tests are very dependent on the legacy partition management. 

I would love to get tests in for the new approach before merging; at the very least I want a test before we make this the new default.

src/DurableTask.AzureStorage/Partitioning/SafePartitionManager.cs
ConnorMcMahon: I'm open to making this configurable, but I am a bit concerned about naming convention, and overall config pollution.

In general, I believe that due to the fact that a worker won't stop renewing it's ownership lease after losing the intent lease until it's done processing events, and we won't steal ownership leases, these intervals _shouldn't_ be configurable in a way that causes split brain.

The only potential exception would be if the lease interval is less than the renew interval, which we should probably have some validation.

I view 4 different approaches we can take here:
1. Fixed values that we deem appropriate
2. Eat the maintenance cost of adding 2-3 new config value fields.
3. Use the same values that we use for intent leases, for simplicity of configuration.
4. Derive the values from the configured values for intent leases, and eat the documentation cost of explaining this.

If we go with approach 4, my hunch says that due to the properties of the ownership lease (not stealing, and renewing until we are done processing events), we will want to be more aggressive with our intervals to compensate for the potential performance losses, but I don't have numbers to back this up right now.

cgillum: I'll leave it up to you to decide, but I agree that it would be nice to avoid config explosion.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
ConnorMcMahon: It turns out the simplest way to handle this is to "remove" the queue and add a new one.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
ConnorMcMahon: For now I set to this to true, as some tests seem to be failing with the other mode being default. I think each worker is taking a bit longer to start up, meaning 30 seconds isn't sufficient, but it will need some investigation.

Issue #428, opened 2020-08-05
Update nuget package configuration to resolve build warnings


cgillum:
The common nuget package configuration uses the `<PackageLicenseUrl>` and `<PackageIconUrl>` configuration elements, which are deprecated. This results in the following warnings:

* Error NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.
* NU5048: The 'PackageIconUrl'/'iconUrl' element is deprecated. Consider using the 'PackageIcon'/'icon' element instead. Learn more at https://aka.ms/deprecateIconUrl

This issue tracks fixing the underlying problem so that the nuget package generation follows up-to-date best practices.

PR #429, opened 2020-08-05
Fix various build issues

cgillum:
### Problem
The previous [commit to master ](https://github.com/Azure/durabletask/pull/425) broke the DurableTask.AzureStorage.Test project because it introduced an incompatibility with .NET Framework. This went undetected in the CI because of the switch to `net461` that caused the CI to skip running all tests (the CI was configured to look for test projects compiled specifically for `net451`).

### Fix
The CI has already been reconfigured to run tests compiled against `net461`. In addition to that, this PR fixes the compatibility issue so that the .NET Framework tests can run successfully.

Also bundled in this change are fixes to build warnings that have existed for a while, but were creating unnecessary noise in the build logs. In particular, we are suppressing some [nuget warnings](https://github.com/Azure/durabletask/issues/428) that have existed for a while, and also removing an unnecessary binding redirect that was causing build warnings.


test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1286,7 +1286,18 @@ public async Task FanOutToTableStorage(bool enableExtendedSessions)
...
+#if NETCOREAPP
             EventSourceAnalyzer.InspectAll(AnalyticsEventSource.Log);
+#else
+            try
+            {
+                EventSourceAnalyzer.InspectAll(AnalyticsEventSource.Log);
+            }
+            catch (FormatException)
+            {
+                Assert.Inconclusive("Known issue with .NET Framework, EventSourceAnalyzer, and DateTime parameters");
```
cgillum, line 1298: Unfortunately there is a bug in `EventSourceAnalyzer` that causes it to break and throw a `FormatException` when there is a DateTime parameter. This means we're only getting partial results running tests against full framework. I didn't want to block the CI on this, however, so I'm just asserting inconclusive.

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -7,26 +7,25 @@
 
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
```
cgillum, line 10: It seems that .NET Framework 4.6.1 is not compatible with `Microsoft.Extensions.Logging` 3.x, so I'm using 2.2 instead for .NET Framework (this was the guidance I found when doing online searches).

test/DurableTask.AzureStorage.Tests/app.config
```diff
@@ -4,12 +4,4 @@
...
     <add key="TaskHubName" value="test" />
   </appSettings>
-  <runtime>
-    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
-      <dependentAssembly>
-        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
-        <bindingRedirect oldVersion="6.0.0.0-7.0.0.0" newVersion="7.0.0.0" />
-      </dependentAssembly>
-    </assemblyBinding>
-  </runtime>
```
cgillum, line 14: This change was to fix some build warnings about redundant assembly redirects. It seems the build process is actually generating these instructions at build time, so the app.config updates are not necessary.

tools/DurableTask.props
```diff
@@ -6,6 +6,8 @@
     <DebugSymbols>True</DebugSymbols>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <LangVersion>7.3</LangVersion>
+    <!-- See https://github.com/Azure/durabletask/issues/428 -->
+    <NoWarn>NU5125,NU5048</NoWarn>
```
cgillum, line 10: I'm silencing these build warnings since they are known issues.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -209,12 +209,12 @@ public static void SetLogicalTraceActivityId(Guid activityId)
                 EventIds.MessageGone,
                 Account,
                 TaskHub,
+                EventType,
```
ConnorMcMahon, line 212: This change wasn't mentioned in the release notes. Were they intentional?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
ConnorMcMahon: To clarify, this is an issue in `EventSourceAnalyzer` only on .NET Framework, and not the version used in .NET Core.

I struggled to find where this method/class is even defined, as it appears to be in this library, which isn't immediately obvious how this dependency is included in our csproj? https://docs.microsoft.com/en-us/previous-versions/msp-n-p/dn134369(v=pandp.60)

cgillum: Yes, there are two different versions of this nuget package, one for .NET Framework and one for .NET Core. It's a transient dependency from DurableTask.Core.Tests: https://github.com/Azure/durabletask/blob/ac52351af71e68ef0758270c1e5371cb8e067d43/test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj#L9

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
cgillum: Yes, the event source verifier test was legitimately failing because of this. Sorry for forgetting to mention this one... 😓

PR #430, opened 2020-08-06
Fix:  correlation test fail except for external event with extended session

TsuyoshiUshio:
Hi @cgillum 

I found the root cause why the correlation test fails. 
`TaskHubClient` changes the behavior. `CorrelationTraceClient.Propagate(() => CreateAndTrackDependencyTelemetry(requestTraceContext));` Should be coming before a method that eventually call the `TaskHubQueue`.  

This is the old code.  
https://github.com/TsuyoshiUshio/durabletask/blob/correlation/final/src/DurableTask.Core/TaskHubClient.cs#L497-L501

## TODO 
Only the External Event with non-extended session will fail. I'm not fully understand the new behavior, however, it is getting very close. 

image

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -657,11 +657,11 @@ namespace DurableTask.AzureStorage
                     // Create or restore Correlation TraceContext
 
                     TraceContextBase currentRequestTraceContext = null;
-                    await CorrelationTraceClient.PropagateAsync(
-                        async () =>
+                    CorrelationTraceClient.Propagate(
+                        () =>
                         {
                             var isReplaying = session.RuntimeState.ExecutionStartedEvent?.IsPlayed ?? false;
-                            TraceContextBase parentTraceContext = await GetParentTraceContextAsync(session);
+                            TraceContextBase parentTraceContext = GetParentTraceContext(session);
                             currentRequestTraceContext = GetRequestTraceContext(isReplaying, parentTraceContext);
                         });
 
@@ -763,7 +763,7 @@ namespace DurableTask.AzureStorage
             }
         }
 
-        async Task<TraceContextBase> GetParentTraceContextAsync(OrchestrationSession session)
+        TraceContextBase GetParentTraceContext(OrchestrationSession session)
         {
             var messages = session.CurrentMessageBatch;
             TraceContextBase parentTraceContext = null;
@@ -808,8 +808,9 @@ namespace DurableTask.AzureStorage
                 {
                     if (message.TaskMessage.Event is EventRaisedEvent)
                     {
-                        var history = await this.trackingStore.GetHistoryEventsAsync(session.Instance.InstanceId, session.Instance.ExecutionId);
-                        var traceContextString = history.Events.OrderByDescending(p => p.Timestamp).Where(p => p.Correlation != null).FirstOrDefault().Correlation;
+                        var history = session.RuntimeState.Events;
+
+                        var traceContextString = history.Where(p => p.EventType == EventType.ExecutionStarted).Select(p => (ExecutionStartedEvent)p).FirstOrDefault().Correlation;
                         parentTraceContext = TraceContextBase.Restore(traceContextString);
                     }
                 }            
@@ -834,12 +835,6 @@ namespace DurableTask.AzureStorage
 
             return false;
         }
-
-        //async Task<TraceContextBase> GetTraceContextFromTrackingStoreAsync(OrchestrationSession session)
-        //{
-        //    var history = await this.trackingStore.GetHistoryEventsAsync(session.Instance.InstanceId, session.Instance.ExecutionId);
-          
-        //}
 
         static TraceContextBase GetRequestTraceContext(bool isReplaying, TraceContextBase parentTraceContext)
         {
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -657,11 +657,11 @@ namespace DurableTask.AzureStorage
                     // Create or restore Correlation TraceContext
 
                     TraceContextBase currentRequestTraceContext = null;
-                    CorrelationTraceClient.Propagate(
-                        () =>
+                    await CorrelationTraceClient.PropagateAsync(
+                        async () =>
                         {
                             var isReplaying = session.RuntimeState.ExecutionStartedEvent?.IsPlayed ?? false;
-                            TraceContextBase parentTraceContext = GetParentTraceContext(session.CurrentMessageBatch);
+                            TraceContextBase parentTraceContext = await GetParentTraceContextAsync(session);
                             currentRequestTraceContext = GetRequestTraceContext(isReplaying, parentTraceContext);
                         });
 
@@ -763,8 +763,9 @@ namespace DurableTask.AzureStorage
             }
         }
 
-        static TraceContextBase GetParentTraceContext(IList<MessageData> messages)
+        async Task<TraceContextBase> GetParentTraceContextAsync(OrchestrationSession session)
         {
+            var messages = session.CurrentMessageBatch;
             TraceContextBase parentTraceContext = null;
             foreach(var message in messages)
             {
@@ -803,7 +804,15 @@ namespace DurableTask.AzureStorage
 
                             break;
                     }                   
-                }               
+                } else
+                {
+                    if (message.TaskMessage.Event is EventRaisedEvent)
+                    {
+                        var history = await this.trackingStore.GetHistoryEventsAsync(session.Instance.InstanceId, session.Instance.ExecutionId);
+                        var traceContextString = history.Events.OrderByDescending(p => p.Timestamp).Where(p => p.Correlation != null).FirstOrDefault().Correlation;
+                        parentTraceContext = TraceContextBase.Restore(traceContextString);
+                    }
+                }            
             }
 
             return parentTraceContext ?? TraceContextFactory.Empty;
@@ -825,6 +834,12 @@ namespace DurableTask.AzureStorage
 
             return false;
         }
+
+        //async Task<TraceContextBase> GetTraceContextFromTrackingStoreAsync(OrchestrationSession session)
+        //{
+        //    var history = await this.trackingStore.GetHistoryEventsAsync(session.Instance.InstanceId, session.Instance.ExecutionId);
+          
+        //}
 
         static TraceContextBase GetRequestTraceContext(bool isReplaying, TraceContextBase parentTraceContext)
         {
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -810,7 +810,7 @@ namespace DurableTask.AzureStorage
                     {
                         var history = session.RuntimeState.Events;
 
-                        var traceContextString = history.Where(p => p.EventType == EventType.ExecutionStarted).Select(p => (ExecutionStartedEvent)p).FirstOrDefault().Correlation;
+                        string traceContextString = session.RuntimeState.ExecutionStartedEvent?.Correlation;
                         parentTraceContext = TraceContextBase.Restore(traceContextString);
                     }
                 }            
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -984,7 +984,7 @@ namespace DurableTask.AzureStorage.Tracking
                         CorrelationTraceClient.Propagate(() =>
                         {
                             historyEntity.Properties["Correlation"] = new EntityProperty(executionStartedEvent.Correlation);
-                            estimatedBytes += System.Text.ASCIIEncoding.ASCII.GetByteCount(executionStartedEvent.Correlation);
+                            estimatedBytes += Encoding.Unicode.GetByteCount(executionStartedEvent.Correlation);
                         });
 
                         this.SetInstancesTablePropertyFromHistoryProperty(
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -946,7 +946,7 @@ namespace DurableTask.AzureStorage.Tracking
                     ["LastUpdatedTime"] = new EntityProperty(newEvents.Last().Timestamp),
                 }
             };
-            
+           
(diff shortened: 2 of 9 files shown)


TsuyoshiUshio:
## The change I introduced
External Event test was failed. The reason was, the old version as false positive. ExecutionStarted and EventRaiseEvent is coming at the same time, It works. However, ExecutionStated and EventRaiseEvnet coming as different timing, EventRaiseEvent's one can not find the parent trace context. I put the traceContext on ExecutionStarted to trackingStore. If the EventRaiseEvent happens, and the ParentTraceContext is empty, I restored the context from the tracking Store. 

* Save the TraceContext to the trackingStore when the orchestration Started
* Restore the TraceContext when the EventRaiseEvent and ParentTraceContext is null
* Add CorrelationTraceClinet.PropageteAsync for the Async execution of the Correlation Logic
* Add MultipleParentMultilayer test for testing ChildOrchestration with RaiseEvent.

## Point of the Review
What I'm not sure how an Orchestrator with RaiseEvent find the your orchestrator's ExecutionStarted history? 
Currently, I just sort the history and pick up the latest. However, I'm not sure if this logic is proper or not. 
https://github.com/TsuyoshiUshio/durabletask/blob/tsushi/fixtest/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L812

image

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -803,7 +804,15 @@ static TraceContextBase GetParentTraceContext(IList<MessageData> messages)
 
                             break;
                     }                   
-                }               
+                } else
+                {
+                    if (message.TaskMessage.Event is EventRaisedEvent)
+                    {
+                        var history = await this.trackingStore.GetHistoryEventsAsync(session.Instance.InstanceId, session.Instance.ExecutionId);
```
cgillum, line 811: Unfortunately this will not be acceptable from a performance perspective. However, I don't think you need to fetch the history events at all. We have already fetched all of them in memory earlier than this. You can find them in the `session.RuntimeState.Events` property.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -980,6 +980,12 @@ public override Task StartAsync()
                         instanceEntity.Properties["Version"] = new EntityProperty(executionStartedEvent.Version);
                         instanceEntity.Properties["CreatedTime"] = new EntityProperty(executionStartedEvent.Timestamp);
                         instanceEntity.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Running.ToString());
+                        CorrelationTraceClient.Propagate(() =>
+                        {
+                            var traceContext = CorrelationTraceContext.Current;
+                            historyEntity.Properties["Correlation"] = new EntityProperty(traceContext.SerializableTraceContext);
```
cgillum, line 986: How large do you think this data will be?  I think you should also update `estimatedBytes` here so that we don't overflow the maximum batch size for Azure Tables.

Also, please update the `AzureTableTrackingStore.VariableSizeEntityProperties` array by adding your new "Correlation" property name so that we can correctly keep track of whether we need to compress the data and upload it to blob storage.

src/DurableTask.Core/CorrelationTraceClient.cs
```diff
@@ -113,6 +114,23 @@ public static void Propagate(Action action)
...
+        /// It suppresses the execution when <see cref="CorrelationSettings"/>.DisablePropagation is true.
+        /// </summary>
+        /// <param name="func"></param>
+        /// <returns></returns>
+        public static Task PropagateAsync(Func<Task> func)
+        {
+            if (CorrelationSettings.Current.EnableDistributedTracing)
+            {
+                return func();
+            } else
```
cgillum, line 128: Please fix the code formatting to put `else` on a line by itself. But actually, I think you can delete this `PropogateAsync` method.

src/DurableTask.Core/History/HistoryEvent.cs
```diff
@@ -88,5 +88,11 @@ protected HistoryEvent(int eventId)
         /// Implementation for <see cref="IExtensibleDataObject.ExtensionData"/>.
         /// </summary>
         public ExtensionDataObject ExtensionData { get; set; }
+
+        /// <summary>
+        /// Gets the Correlation
```
cgillum, line 93:
```suggestion
        /// Gets or sets the serialized end-to-end correlation state.
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -826,6 +835,12 @@ static bool IsActivityOrOrchestrationFailedOrCompleted(IList<MessageData> messag
             return false;
         }
 
+        //async Task<TraceContextBase> GetTraceContextFromTrackingStoreAsync(OrchestrationSession session)
+        //{
+        //    var history = await this.trackingStore.GetHistoryEventsAsync(session.Instance.InstanceId, session.Instance.ExecutionId);
+          
+        //}
```
cgillum, line 842: Don't forget to remove this unused code.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -657,11 +657,11 @@ static TaskHubInfo GetTaskHubInfo(string taskHub, int partitionCount)
...
 
                     TraceContextBase currentRequestTraceContext = null;
-                    CorrelationTraceClient.Propagate(
-                        () =>
+                    await CorrelationTraceClient.PropagateAsync(
+                        async () =>
                         {
                             var isReplaying = session.RuntimeState.ExecutionStartedEvent?.IsPlayed ?? false;
-                            TraceContextBase parentTraceContext = GetParentTraceContext(session.CurrentMessageBatch);
+                            TraceContextBase parentTraceContext = await GetParentTraceContextAsync(session);
```
cgillum, line 664: See my other comment about getting the correlation data directly from `session.RuntimeState.ExecutionStartedEvent`.

src/DurableTask.Core/History/HistoryEvent.cs
```diff
@@ -88,5 +88,11 @@ protected HistoryEvent(int eventId)
...
+        /// </summary>
+        [DataMember]
+        public string Correlation { get; set; }
```
cgillum, line 96: I think this property should be only defined on `ExecutedStartedEvent`.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: Actually, since you added the `Correlation` property to the history event class in DurableTask.Core, I don't think you need to do this at all. Just assign it to `ExecutionStartedEvent.Correlation` and we will use reflection to put it into table storage. In fact, would it make more sense to assign the Correlation property in the DurableTask.Core project? It seems like you don't actually need to do anything specific in DurableTask.AzureStorage for this.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done for `estimatedBytes` and `AzureTableTrackingStore.VariableSizeEntityProperties` However, I couldn't figure out where should I updated `ExecutionStartedEvent.Correlation` It is not here, right? Today is times up, I'll try tomorrow, however, If you know where it is, please let me know.

src/DurableTask.Core/CorrelationTraceClient.cs
TsuyoshiUshio: Done

src/DurableTask.Core/History/HistoryEvent.cs
TsuyoshiUshio: Removed from here. Your suggestion is added to the `EventStartedEvent`

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done that. Please check it it is what you expected.

src/DurableTask.Core/History/HistoryEvent.cs
TsuyoshiUshio: Done

TsuyoshiUshio:
image

## What I've done

* Move Correlation from `History` to `EventStartedEvent`
* Read History from session
* update `estimatedBytes` and `AzureTableTrackingStore.VariableSizeEntityProperties`

## TODO
Where should I insert the `Correlation`. I keep this logic.   I need to figure out where to put the assignment code for `ExecutionStartedEvent.Correlation`. Today is time up. Tomorrow, I'll try again, however, if you already knew it, please let me know. 
```
 historyEntity.Properties["Correlation"] = new EntityProperty(traceContext.SerializableTraceContext);
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Hi @cgillum 
I assign the ExecutionStartedEvent.Correlation on the DurableTask.Core side. Do  you mean like this? 
https://github.com/Azure/durabletask/pull/430/commits/11725f8ae70ba8d60e27d4a48b743b102957ecac
All test has passed. 
image

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -803,7 +804,16 @@ static TraceContextBase GetParentTraceContext(IList<MessageData> messages)
 
                             break;
                     }                   
-                }               
+                } else
```
cgillum, line 807: formatting: Please put `else` on it's own line.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -803,7 +804,16 @@ static TraceContextBase GetParentTraceContext(IList<MessageData> messages)
...
+                {
+                    if (message.TaskMessage.Event is EventRaisedEvent)
+                    {
+                        var history = session.RuntimeState.Events;
+
+                        var traceContextString = history.Where(p => p.EventType == EventType.ExecutionStarted).Select(p => (ExecutionStartedEvent)p).FirstOrDefault().Correlation;
```
cgillum, line 813: Why do you need to search through this history? Can't you do this?
```suggestion
                        string traceContextString = session.RuntimeState.ExecutionStartedEvent?.Correlation;
```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -980,6 +981,12 @@ public override Task StartAsync()
                         instanceEntity.Properties["Version"] = new EntityProperty(executionStartedEvent.Version);
                         instanceEntity.Properties["CreatedTime"] = new EntityProperty(executionStartedEvent.Timestamp);
                         instanceEntity.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Running.ToString());
+                        CorrelationTraceClient.Propagate(() =>
+                        {
+                            historyEntity.Properties["Correlation"] = new EntityProperty(executionStartedEvent.Correlation);
```
cgillum, line 986: Is this required? I thought this would be done automatically by `this.tableEntityConverter.ConvertToTableEntity(historyEvent)` on line 956.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -980,6 +981,12 @@ public override Task StartAsync()
...
+                            estimatedBytes += System.Text.ASCIIEncoding.ASCII.GetByteCount(executionStartedEvent.Correlation);
```
cgillum, line 987: For table storage, I think Unicode is always used.
```suggestion
                            estimatedBytes += Encoding.Unicode.GetByteCount(executionStartedEvent.Correlation);
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -392,6 +392,13 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                         }
                     }
 
+                    // correlation
+                    CorrelationTraceClient.Propagate(() => {
+                        if (runtimeState.ExecutionStartedEvent != null)
+                            runtimeState.ExecutionStartedEvent.Correlation = CorrelationTraceContext.Current.SerializableTraceContext;
+                     });
```
cgillum, line 399: Please use the C# formatting that is consistent with the rest of this project:
* The `{` should go on its own line.
* `if` statements should have both `{` and `}` on their own line.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -416,6 +423,11 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                                 workItem.InstanceId,
                                 "Updating state for continuation");
 
+                            // correlation
+                            CorrelationTraceClient.Propagate(() => {
```
cgillum, line 427: Please fix the formatting here too by putting `{` on its own line.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
TsuyoshiUshio: Done

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
TsuyoshiUshio: Done

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
TsuyoshiUshio: Done

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
TsuyoshiUshio: Done

TsuyoshiUshio:
@cgillum 
I accept all your change request. And It passes correlation tests.
Let's execute all the test on the ci when you merge correlation branch to master.

PR #431, opened 2020-08-06
Make logging dependencies consistent

cgillum:
Two fixes:

* There were inconsistencies in my `Microsoft.Extensions.Logging.Abstractions` dependencies across projects. This was resulting in runtime failures in `WebJobs.Extensions.DurableTask` when using Functions V1 bits. I've updated these dependencies to make things consistent.
* I accidentally switched on `<GeneratePackageOnBuild>` in a previous PR. Turning it back off in this one.

I also tested this on Functions v1 host (net461) and verified that things seem to be working correctly (required a lot of `MethodNotFound` error debugging due to compatible assembly version issues).

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
     <PackageReference Include="Newtonsoft.Json" Version="7.0.1" />
     <PackageReference Include="WindowsAzure.Storage" version="7.2.1" />
   </ItemGroup>
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -19,16 +19,17 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.5" />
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/Logging/NoOpLoggerFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage.Logging
+{
+    using System;
+    using Microsoft.Extensions.Logging;
+
+    // There is a NullLoggerFactory defined in Microsoft.Extensions.Logging.Abstractions v2.x and above, but we 
+    // cannot rely on it because Azure Functions v1 is pinned to Microsoft.Extensions.Logging.Abstractions v1.x.
+    sealed class NoOpLoggerFactory : ILoggerFactory
```
ConnorMcMahon, line 21: I actually prefer the name of this anyways, though it is a bit annoying.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -19,16 +19,17 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.5" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.*" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
```
ConnorMcMahon, line 26: I'm curious what best practices are around having these dependencies explicitly mentioned in the csproj vs just inheriting them transitively?

src/DurableTask.AzureStorage/Logging/NoOpLoggerFactory.cs
cgillum: Yeah, `NullLoggerFactory` is a weird name because nothing is actually "null".

PR #432, opened 2020-08-06
Added AppLease feature

amdeel:
This feature allows for Geo-Redundant applications that can support failover in case of region failure.

related to https://github.com/Azure/azure-functions-durable-extension/issues/1128

This feature adds a lease on the task hub container. An app must acquire this lease before proceeding in AzureStorageOrchestrationService StartAsync. If an app does not acquire the lease because another app using the same task hub already acquired it, it will delay on a configurable timing interval and continuously attempt to acquire the lease in a loop. If/once the first app gets shut down, the second app will be able to acquire the lease and continue execution where the first app left off.


src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,272 @@
...
+using System;
+using System.Threading.Tasks;
+using System.Threading;
+using DurableTask.AzureStorage.Monitoring;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace DurableTask.AzureStorage.Partitioning
```
amdeel, line 23: Maybe we should change this folder to LeaseManagement instead of Partitioning

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -443,6 +459,16 @@ public async Task StartAsync()
             this.shutdownSource = new CancellationTokenSource();
             this.statsLoop = Task.Run(() => this.ReportStatsLoop(this.shutdownSource.Token));
 
+            if (this.settings.UseAppLease)
+            {
+                while (!await appLeaseManager.TryAquireAppLeaseAsync())
```
cgillum, line 464: One thing we need to test before we can merge this is what happens to a function app that is blocked from starting because of this check inside the `StartAsync()` method. I don't know for sure what will happen, but it will prevent the task hub worker from starting which will block the durable extension triggers from initializing (see [here](https://github.com/Azure/azure-functions-durable-extension/blob/86567461d150f01ef15f6400d53b71c462ef1263/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L1042) and [here](https://github.com/Azure/azure-functions-durable-extension/blob/86567461d150f01ef15f6400d53b71c462ef1263/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs#L67)). This _might_ cause the function app to go into some kind of failure loop which, if that is the behavior, could be alarming for users when they are deploying their geo-redundant apps. If we determine that this is a problem, we might need to move this `TryAcquireAppLeaseAsync()` somewhere else (at the very start of [PartitionManager.LeaseTakerAsync()](https://github.com/Azure/durabletask/blob/0098fffd8fb86c3875763db28e04972e10144b24/src/DurableTask.AzureStorage/Partitioning/PartitionManager.cs#L227) is one idea we could try, if necessary).

src/DurableTask.AzureStorage/Partitioning/AppLeaseOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </summary>
+        public TimeSpan LeaseInterval { get; set; } = TimeSpan.FromSeconds(60);
+
+        /// <summary>
+        /// Creates an instance of <see cref="AppLeaseOptions"/> with following default values:
+        ///     a) RenewInterval = 25 seconds
+        ///     b) AcquireInterval = 300 seconds, or 5 minutes
+        ///     c) DefaultLeaseInterval = 60 seconds
+        /// </summary>
+        public static AppLeaseOptions DefaultOptions
```
cgillum, line 48: Tip: if you want to use a one-liner C# syntax that does the same thing, you can do so like this:
```suggestion
        public static AppLeaseOptions DefaultOptions => new AppLeaseOptions();
```

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,272 @@
...
+            try
+            {
+                AnalyticsEventSource.Log.AppLeaseAcquisitionStarted(
+                    this.accountName,
+                    this.taskHub,
+                    this.workerName,
+                    this.appId,
+                    Utils.ExtensionVersion);
+
+                appId = await taskHubContainer.AcquireLeaseAsync(this.options.LeaseInterval, appId);
```
cgillum, line 115: Is it possible for AcquireLeaseAsync(...) to return a new value for `appId`? I was hoping we could make this instance variable `readonly`.

cgillum: Also, let's prefix member access with `this.`
```suggestion
                await taskHubContainer.AcquireLeaseAsync(this.options.LeaseInterval, this.appId);
```

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,272 @@
...
+        private readonly string taskHub;
+        private readonly string workerName;
+        private readonly CloudBlobClient storageClient;
+        private readonly string taskHubContainerName;
+        private readonly string appName;
+        private readonly AppLeaseOptions options;
+        private readonly AzureStorageOrchestrationServiceStats stats;
+
+        private CloudBlobContainer taskHubContainer;
+        private string appId;
```
cgillum, line 37: Would it make sense to rename this `appLeaseId`? Looking through the code, it looks like this is a lease ID, but it wasn't clear to me from the name.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,272 @@
...
+                {
+                    renewed = false;
+
+                    AnalyticsEventSource.Log.AppLeaseRenewalFailed(
+                    this.accountName,
+                    this.taskHub,
+                    this.workerName,
+                    this.appId,
+                    errorMessage,
+                    Utils.ExtensionVersion);
```
cgillum, line 226: nit: this argument list needs to be indented.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,272 @@
...
+            try
+            {
+                AnalyticsEventSource.Log.ReleasingAppLease(
+                    this.accountName,
+                    this.taskHub,
+                    this.workerName,
+                    this.appId,
+                    Utils.ExtensionVersion);
+
+                AccessCondition accessCondition = new AccessCondition() { LeaseId = appId };
```
cgillum, line 263:
```suggestion
                AccessCondition accessCondition = new AccessCondition() { LeaseId = this.appId };
```

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,272 @@
...
+                await taskHubContainer.ReleaseLeaseAsync(accessCondition);
```
cgillum, line 264:
```suggestion
                await this.taskHubContainer.ReleaseLeaseAsync(accessCondition);
```

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -963,6 +963,171 @@ public void GeneralWarning(string Account, string TaskHub, string Details, strin
...
+           string WorkerName,
+           string Token,
+           string Details,
+           string ExtensionVersion)
+        {
+            EnsureLogicalTraceActivityId();
+            this.WriteEvent(142, Account, TaskHub, WorkerName ?? string.Empty, Token ?? string.Empty, Details, ExtensionVersion);
+        }
+
+        [Event(143, Level = EventLevel.Warning, Version = 2)]
```
cgillum, line 979: These are all new events, so you can set them to `Version = 1`, or leave out the `Version` property altogether.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -963,6 +963,171 @@ public void GeneralWarning(string Account, string TaskHub, string Details, strin
...
+            this.WriteEvent(
+                145,
+                Account,
+                TaskHub,
+                WorkerName ?? string.Empty,
+                Token ?? string.Empty,
+                ExtensionVersion);
+        }
+
+        [Event(146, Level = EventLevel.Verbose)]
```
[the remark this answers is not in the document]
amdeel: While my initial reaction was to create separate logs, I realize these logs are basically identical, so I just reused the existing PartitionManager events.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: I really don't know. I assume it would always be the leaseId I input here, but my analyzer has made me overly cautious, I suppose.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -431,11 +446,26 @@ public async Task StartAsync()
             this.shutdownSource = new CancellationTokenSource();
             this.statsLoop = Task.Run(() => this.ReportStatsLoop(this.shutdownSource.Token));
 
+            await Task.Factory.StartNew(() => this.LeaseManagerStarter());
+
+            this.isStarted = true;
+        }
+
+        private async void LeaseManagerStarter()
```
cgillum, line 454: Since this method returns `void` and since there is no code that awaits on it, if there is any exception thrown from this method, it will go "unobserved", meaning there won't be any code to handle it or log it. Also, I think the app would get stuck in this partially started state forever since there isn't any code to retry it.

To improve the robustness and debuggability of this code, I suggest we wrap this startup logic in a while-loop with a try/catch inside and keep retrying any time there is a failure. Something like this might work:
```csharp
// Keep trying to start up until we get a signal that the worker is shutting down
while (!this.shutdownSource.Token.IsCancellationRequested)
{
    try
    {
        // appLeaseManager startup...
        // partitionManager startup...
    }
    catch (Exception e)
    {
        // log an error
        this.settings.Logger.PartitionManagerError(...);

        // wait for a few seconds and try again
        await Task.Delay(TimeSpan.FromSeconds(10), this.shutdownSource.Token);
        continue;
    }

    // success!
    break;
}
```

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -0,0 +1,329 @@
...
+                else
+                {
+                    await this.taskHubContainer.DeleteIfExistsAsync();
+                }
+            }
+            catch (StorageException ex)
+            { 
+                if (ex.RequestInformation.HttpStatusCode != 404)
+                {
+                    ExceptionDispatchInfo.Capture(ex).Throw();
```
cgillum, line 320: You can just use the `throw` statement by itself. No need for using `ExceptionDispatchInfo`. That's more for cases where you need to rethrow an exception outside of a catch block.
```suggestion
                    throw;
```

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -282,13 +282,10 @@ public async Task DeleteAsync(BlobLease lease)
...
-            try
+            IEnumerable<BlobLease> blobs = await ListLeasesAsync();
+            foreach(BlobLease blob in blobs)
             {
-                await this.taskHubContainer.DeleteIfExistsAsync();
-            }
-            finally
-            {
-                this.stats.StorageRequests.Increment();
+                await DeleteAsync(blob);
```
cgillum, line 288: I would prefer that we continue to delete the container. I read your comments about the app lease conflict, though. Would we be able to resolve the issue if we deleted the app lease manager _before_ deleting the blob lease manager in `AzureStorageOrchestrationService.DeleteAsync()`? And as part of deleting the app lease manager, maybe we should release the lease?

amdeel: The AppLeaseManager may or may not have a lease on the task hub container, depending on if the UseAppLease setting is true or false. I wanted the AppLeaseManager to decide if it needed to use a leaseId to delete the task hub container, instead of the AzureStorageOrchestrationService or the BlobLeaseManager.

I could have the BlobLeaseManager still attempt to delete the task hub container, but this didn't make a lot of sense to me. If I delete the task hub container through the AppLeaseManager before I call the DeleteAllAsync method on the BlobLeaseManager, the BlobLeaseManager will never succeed in deleting the task hub. It will always be deleted already. So it doesn't seem like there's a point to call it (Even with this change, there isn't really a point to call it, so I could still remove the attempt to call BlobLeaseManager.DeleteAllAsync here and would personally feel fine doing that).

The reason I decided to change BlobLeaseManager.DeleteAllAsync was because I felt uncomfortable implementing a method (BlobLeaseManager implements ILeaseManager) that doesn't work. If we ever wanted to call DeleteAllAsync for any other reason, maybe the partition manager decided it wanted to delete the blob or something, it just wouldn't work. Leaving a method that doesn't work didn't seem like a great solution.

One other thing is that the BlobLeaseManager is supposed to manage Blobs in the task hub container, and the AppLeaseManager is managing the lease on the task hub container. It makes sense, to me, to leave the BlobLeaseManager to manage the blobs and to not worry about deleting the task hub container.

PR #434, opened 2020-08-07
Fix edge case with Azure Storage large message implementation

ConnorMcMahon:
Azure Storage queues can only store 64 KiB. Before, we would put any raw
messages larger than 60 KiB in blob storage and send a reference to the
blob in the queue message. However, a customer has experienced an issue
in which a raw input between 60000 bytes and 61440 bytes created a queue
message larger than 64 KiB, likely due to the metadata we send along
with the input in the queue message.

This PR solves this issue with a two-pronged approach. First, we reduce
the raw input size limit before writing to blob storage from 60 KiB to
60000 bytes. Second, we check the message size after including our
metadata to ensure we are not sending anything greater than 64 KiB, and
if we are, we store the input in blob storage.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -34,7 +34,10 @@ namespace DurableTask.AzureStorage
     /// </summary>
     class MessageManager
     {
-        const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
+        // NOTE: the original raw payload limit of 60KiB was a bit too large 
+        // when serialized with metadata so we now keep to 60kB.
+        const int MaxRawPayloadStorageQueueSize = 60000; // 60kB
+        const int MaxTotalPayloadStorageQueueSize = 65536; // 64 KiB
         const int DefaultBufferSize = 64 * 2014; // 64KB
 
         const string LargeMessageBlobNameSeparator = "/";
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -34,7 +34,10 @@ namespace DurableTask.AzureStorage
     /// </summary>
     class MessageManager
     {
-        const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
+        // NOTE: the original raw payload limit of 60KiB was a bit too large 
+        // when serialized with metadata so we now keep to 60kB.
+        const int MaxRawPayloadStorageQueueSize = 60000; // 60kB
```
cgillum, line 39: The comment says 60kB but 60kB is `60 * 1024`, not `60000`. Which value did you intend?

ConnorMcMahon: So technically, 60kB is `60000`, where 60 KiB is `60*1024`. This is according to [the NIST metric system](https://physics.nist.gov/cuu/Units/binary.html), but it seems like [the industry](https://ux.stackexchange.com/a/13850) doesn't really follow that consistently. I'm fine changing the comments if this is confusing...

cgillum:
@ConnorMcMahon can we close this PR in favor of https://github.com/Azure/durabletask/pull/435?

PR #435, opened 2020-08-11
Reduce maximum queue message size from 60 KB to 45 KB

ConnorMcMahon:
This change accounts for the fact that the Azure Storage SDK encodes the
content of the queue message in base64 encoding, increasing the size by
~33%. We shrink our message size maximum accordingly.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -34,7 +34,9 @@ namespace DurableTask.AzureStorage
     /// </summary>
     class MessageManager
     {
-        const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
+        // Use 45 KB, as the Storage SDK will base64 encode the message,
+        // increasing the size by a factor of 4/3.
+        const int MaxStorageQueuePayloadSizeInBytes = 45 * 1024; // 45KB
         const int DefaultBufferSize = 64 * 2014; // 64KB
 
         const string LargeMessageBlobNameSeparator = "/";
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -117,7 +117,7 @@ async Task<MessageData> AddMessageAsync(TaskMessage taskMessage, OrchestrationIn
                     Utils.GetTaskEventId(taskMessage.Event),
                     sourceInstance.InstanceId,
                     sourceInstance.ExecutionId,
-                    Encoding.Unicode.GetByteCount(rawContent),
+                    Encoding.UTF8.GetByteCount(rawContent),
```
cgillum, line 120: This will fix the tracing for this particular event, but do we need to also fix other occurrences of `Encoding.Unicode`?

ConnorMcMahon: Wow, it's clearly been a long week for me...

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -33,7 +33,9 @@ namespace DurableTask.AzureStorage
     /// </summary>
     class MessageManager
     {
-        const int MaxStorageQueuePayloadSizeInBytes = 60 * 1024; // 60KB
+        // Use 45 KB, as the Storage SDK will base64 encode the message,
+        // increasing the size by a factor of 4/3.
+        const int MaxStorageQueuePayloadSizeInBytes = 45 * 1024; // 45KB
```
jump523steps, line 38: In extreme cases, the length of Utf16 encoding is twice that of UTF8. Consider the Azure Table is UTF-16 encoding with 64KiB limit, update to 32 * 1024 (Math.min(64/2, 64*3/4)) makes sense.

[Comparison_of_Unicode_encodings](https://en.wikipedia.org/wiki/Comparison_of_Unicode_encodings)

PR #440, opened 2020-08-18
Make AppLease deletion safer

ConnorMcMahon:
Some tests were failing due to trying to delete an app lease when
someone else owns the lease.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
@@ -316,24 +316,12 @@ namespace DurableTask.AzureStorage.Partitioning
                 }
                 else
                 {
-                    // If we are not the lease owner, attempt a best effort to delete
-                    // the container, but it is very likely we will not succeed as someone
-                    // else may own the lease.
-                    try
-                    {
-                        await this.appLeaseContainer.DeleteIfExistsAsync();
-                    }
-                    catch (StorageException)
-                    {
-                    }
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -316,7 +316,16 @@ public async Task DeleteContainerAsync()
...
                 {
-                    await this.appLeaseContainer.DeleteIfExistsAsync();
+                    // If we are not the lease owner, attempt a best effort to delete
+                    // the container, but it is very likely we will not succeed as someone
+                    // else may own the lease.
+                    try
+                    {
+                        await this.appLeaseContainer.DeleteIfExistsAsync();
+                    }
+                    catch (StorageException)
```
amdeel, line 326: We can probably just have one catch (StorageException) here instead of two. If another app has the AppLease (if both apps are using the same taskHub and one app attempts to delete the container without acquiring the AppLease) we don't need to throw an exception. We could log the error here, though.

ConnorMcMahon: I was tempted to just add a check for `this.isLeaseOwner` in the existing catch block, but there is theoretically the possibility that would change in between when we do the first check and when we perform the exception handling.

Is your suggestion to just remove the `throw ex` in the existing `catch` block? That does appear to be what `BlobLeaseManager` does in deletion, so I am amenable to that.

amdeel (approved):
LGTM!

PR #441, opened 2020-08-18
Updated JsonConverter default JsonSerializerSetting to include DateParseHandling.None

amdeel:
This change prevents strings that look like dates from being serialized into DateTime objects.

related to https://github.com/Azure/azure-functions-durable-js/issues/208


ConnorMcMahon:
I don't have a problem with it, as long as we change our default settings in the extension as well.

PR #442, opened 2020-08-19
Updated JToken.Parse/JArray.Parse for DateParseHandling

amdeel:
We recently updated our internal default DateParseHandling setting to DateParseHandling.None.

JToken.Parse and JArray.Parse use the incorrect default DateParseHandling setting. I took these methods out and replaced them in a way we're able to use this custom DateParseHandling setting for consistency.

### DIFF
--- a/src/DurableTask.Core/Common/Utils.cs
+++ b/src/DurableTask.Core/Common/Utils.cs
@@ -83,7 +83,7 @@ namespace DurableTask.Core.Common
 
         internal static JArray ConvertToJArray(string input)
         {
-            JArray jArray = new JArray();
+            JArray jArray;
             using (var stringReader = new StringReader(input))
             using (var jsonTextReader = new JsonTextReader(stringReader) { DateParseHandling = DateParseHandling.None })
             {
--- a/src/DurableTask.Core/Common/Utils.cs
+++ b/src/DurableTask.Core/Common/Utils.cs
@@ -26,6 +26,7 @@ namespace DurableTask.Core.Common
     using DurableTask.Core.Serializing;
     using DurableTask.Core.Tracing;
     using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
 
     /// <summary>
     /// Utility Methods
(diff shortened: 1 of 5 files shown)


src/DurableTask.Core/ReflectionBasedTaskActivity.cs
```diff
@@ -71,7 +73,13 @@ public override string Run(TaskContext context, string input)
         /// <returns>Serialized output from the execution</returns>
         public override async Task<string> RunAsync(TaskContext context, string input)
         {
-            JArray jArray = JArray.Parse(input);
+            JArray jArray;
+            using (var stringReader = new StringReader(input))
+            using (var jsonTextReader = new JsonTextReader(stringReader) { DateParseHandling = DataConverter.Settings.DateParseHandling })
+            {
+                jArray = JArray.Load(jsonTextReader);
+            }
```
cgillum, line 81: Would it make sense to factor this into a helper method in the `Utils` class?

src/DurableTask.Core/Serializing/DataConverter.cs
```diff
@@ -13,13 +13,19 @@
 
 namespace DurableTask.Core.Serializing
 {
+    using Newtonsoft.Json;
```
cgillum, line 16: This class is supposed to be agnostic to any particular serializer. For example, down the road me may want to create an implementation that uses the more modern/faster `System.Text.Json` serializer tech or even binary serialization. I don't think we should add any Newtonsoft code here.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -80,6 +81,18 @@ public static string Truncate(this string input, int maxLength)
             return input;
         }
 
+        internal static JArray ConvertToJArray(string input)
+        {
+            JArray jArray = new JArray();
```
cgillum, line 86: I suggest removing the initial allocation since we're going to throw it away down below.
```suggestion
            JArray jArray;
```

Issue #448, opened 2020-08-31
Table Storage exception due to exceeding max message size


seancolonello:
After upgrading to v1.8.0 I started getting exceptions related to Table storage. The issue is caused by pull request #435, the new max size calculation isn't correct.

This content (37,913 bytes UTF8)...
[rawContent.txt](https://github.com/Azure/durabletask/files/5152330/rawContent.txt)
...causes the problem when found here...
https://github.com/Azure/durabletask/blob/3a07d8d65b84fda6ebc8289cce9a6bec1269cad2/src/DurableTask.AzureStorage/MessageManager.cs#L90-L94

Below is trace output from Azure functions.
> [8/31/2020 4:18:28 PM] Microsoft.WindowsAzure.Storage.StorageException: Bad Request
> [8/31/2020 4:18:28 PM]    at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ExecuteAsyncInternal[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext, CancellationToken token)
> [8/31/2020 4:18:28 PM]    at DurableTask.AzureStorage.Tracking.AzureTableTrackingStore.SetNewExecutionAsync(ExecutionStartedEvent executionStartedEvent, String eTag, String inputStatusOverride) in C:\source\durabletask\src\DurableTask.AzureStorage\Tracking\AzureTableTrackingStore.cs:line 845
> [8/31/2020 4:18:28 PM]    at DurableTask.AzureStorage.AzureStorageOrchestrationService.CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[] dedupeStatuses) in C:\source\durabletask\src\DurableTask.AzureStorage\AzureStorageOrchestrationService.cs:line 1362
> [8/31/2020 4:18:28 PM]    at DurableTask.Core.TaskHubClient.InternalCreateOrchestrationInstanceWithRaisedEventAsync(String orchestrationName, String orchestrationVersion, String orchestrationInstanceId, Object orchestrationInput, IDictionary`2 orchestrationTags, OrchestrationStatus[] dedupeStatuses, String eventName, Object eventData, Nullable`1 startAt) in C:\source\durabletask\src\DurableTask.Core\TaskHubClient.cs:line 608
[...]

PR #449, opened 2020-09-14
Make large message handling more robust

ConnorMcMahon:
PR #435 attempted to fix large message inconsistencies in how it handled UTF8 vs UTF16 character counts, but it turns out that the code path that starts new orchestrations relied on the overcounting of characters that used to occur.

This PR fixes that case, and adds more testing with characters of varying UTF8 and UTF16 sizes to make sure we cover all of the edge cases.

Addresses #448

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
@@ -319,12 +319,8 @@ namespace DurableTask.AzureStorage.Partitioning
                     await this.appLeaseContainer.DeleteIfExistsAsync();
                 }
             }
-            catch (StorageException ex)
+            catch (StorageException)
             { 
-                if (ex.RequestInformation.HttpStatusCode != 404)
-                {
-                    throw;
-                }
             }
             finally
             {
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -836,8 +836,8 @@ namespace DurableTask.AzureStorage.Tracking
                 }
             };
 
-            // It is possible that the queue message was not large enough to be written to a blob,
-            // but is too large to be written to a table property.
+            // It is possible that the queue message was small enough to be written directly to a queue message,
+            // not a blob, but is too large to be written to a table property.
             await this.CompressLargeMessageAsync(entity);
 
             Stopwatch stopwatch = Stopwatch.StartNew();
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -46,7 +46,7 @@ namespace DurableTask.AzureStorage.Tracking
         const string CheckpointCompletedTimestampProperty = "CheckpointCompletedTimestamp";
 
         // See https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#property-types
-        const int MaxTablePropertySizeInBytes = 60 * 1024; // 60KB
+        const int MaxTablePropertySizeInBytes = 60 * 1024; // 60KB to give buffer
 
         static readonly string[] VariableSizeEntityProperties = new[]
         {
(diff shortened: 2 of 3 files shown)


src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -319,12 +319,8 @@ public async Task DeleteContainerAsync()
                     await this.appLeaseContainer.DeleteIfExistsAsync();
                 }
             }
-            catch (StorageException ex)
+            catch (StorageException)
```
ConnorMcMahon, line 322: This should probably be covered with this PR: https://github.com/Azure/durabletask/pull/440.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1194,7 +1196,21 @@ static string GetBlobName(DynamicTableEntity entity, string property)
...
-            string eventType = entity.Properties["EventType"].StringValue;
+            string eventType;
+            if (entity.Properties.ContainsKey("EventType"))
+            {
+                eventType = entity.Properties["EventType"].StringValue;
+            }
+            else if (property == "Input")
+            {
+                // This queue is not actually tracked as input, so just use any name.
+                eventType = "Input";
```
cgillum, line 1207: I'm not understanding what this comment is conveying. If I understand the code change, you're doing this because the Instances table entities don't have an `EventType` property. Is that correct?  If so, I feel like we should communicate that in the comments. Also, I'm not sure why this comment here is talking about queues, since I don't believe `AzureTableTrackingStore` knows anything about queues, nor is `DyanamicTableEntity` associated with queues. Should we instead be saying something like "Entities for the Instances table don't have an EventType property, so we are using a hardcoded value instead to represent the orchestration input.

ConnorMcMahon: I'm not exactly sure what I was intending to convey with this comment either, and I don't know how queue got in there at all. The reason is that the initial row does not have an eventType, but it still has an input, so we use a hardcoded value instead. I will update the comment to reflect this.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -836,6 +836,8 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string
                 }
             };
 
+            await this.CompressLargeMessageAsync(entity);
```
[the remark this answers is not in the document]
ConnorMcMahon: Before, we relied on the fact that the queue message was already serialized as a blob for these large messages, and just used the large blob name from the queue message for the input. This worked because we were counting with UTF16 instead of UTF8.

Now, it is possible that the blob message was able to fit, but it is still too large for the table, so we need to compress the message first.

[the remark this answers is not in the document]
cgillum: I think you meant that the _queue_ message was able to fit, but otherwise, this makes sense. Thanks!

cgillum (approved):
LGTM! Also, thanks for taking the time to add these new test!

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1443,19 +1530,39 @@ public async Task LargeTextMessagePayloads_FetchLargeMessages_QueryState(bool en
...
+            {
+                throw new InvalidOperationException($"No characters have byte size {utf16ByteSize} for UTF16");
+            }
+            else if (utf8ByteSize < 1 || utf8ByteSize > 4)
+            {
+                throw new InvalidOperationException($"No characters have byte size {utf8ByteSize} for UTF8.");
+            }
+            else if (utf8ByteSize == 1 && utf16ByteSize == 2)
+            {
+                // Use a character set that is small for UTF16 and large for UTF8
```
ChenglongLiu, line 1546: Are the UTF8 and UTF16 in the wrong order?

ConnorMcMahon: Yeah, this comment is incorrect. This looks to be a copy paste problem. Will resolve before merging.

PR #452, opened 2020-09-17
fix #447 emulator deadlock

### LINKED ISSUE #447: Emulator hangs indefinitely instead of throwing expected error
Using the LocalOrchestrationService, if I create an orchestration instance and allow it to finish, then try creating another instance using the same id, the call to `CreateOrchestrationInstanceAsync` ends up hanging indefinitely. The same code with the ServiceBusOrchestrationService plugged in instead throws the expected error message: "An orchestration with id 'x' already exists. It is in state Completed".

Interestingly, if I don't wait for the first instance to finish and start the second instance immediately, I see the expected error message using the LocalOrchestrationService: "An orchestration with id 'x' already exists. It is in state Pending".

Here's a gist of the code I used to create the issue: https://gist.github.com/maxnorth/a9a860f86425c4890e01f3deb9307b69

ChenglongLiu:
The mixed-use of sync async causes a deadlock in CommitState. Use sync code to fix it.

### DIFF
--- a/src/DurableTask.Emulator/LocalOrchestrationService.cs
+++ b/src/DurableTask.Emulator/LocalOrchestrationService.cs
@@ -49,8 +49,8 @@ namespace DurableTask.Emulator
 
         //Dictionary<string, Tuple<List<TaskMessage>, byte[]>> sessionLock;
 
-        readonly object thisLock = new object();
-        readonly object timerLock = new object();
+        readonly SemaphoreSlim thisLock = new SemaphoreSlim(1, 1);
+        readonly SemaphoreSlim timerLock = new SemaphoreSlim(1, 1);
 
         readonly ConcurrentDictionary<string, TaskCompletionSource<OrchestrationState>> orchestrationWaiters;
 
@@ -74,7 +74,8 @@ namespace DurableTask.Emulator
         {
             while (!this.cancellationTokenSource.Token.IsCancellationRequested)
             {
-                lock (this.timerLock)
+                await this.timerLock.WaitAsync();
+                try
                 {
                     foreach (TaskMessage tm in this.timerMessages.ToList())
                     {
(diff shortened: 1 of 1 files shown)


cgillum:
@ChenglongLiu thanks for submitting this fix! Can you explain a bit more what the deadlock is and why your change fixes it? It wasn't immediately obvious to me.

ChenglongLiu:
@cgillum In maxnorth's example, the WaitForOrchestrationAsync in the Main wait on [tcs](https://github.com/Azure/durabletask/blob/6814e4d5e1ea5de2136afa6041dc81119efabec9/src/DurableTask.Emulator/LocalOrchestrationService.cs#L320)  

On the other hand, when the server worker finish the orchestration and call
```
            if (tasks.Count > 0)
            {
                Task.WaitAll(tasks.ToArray());
            }
```
The tcs set result in **the task thread** first.
The main method continues execution, however, it's executing on **the task thread**. So the tcs1.TrySetResult blocks and the work still hold the lock. The main method calls CreateOrchestrationInstanceAsync again and tries to get the lock and not release the task thread. 

The worker holds **thisLock** waits for **the task thread** to do tcs1.TrySetResult, **the task thread** waits to lock **thisLock**.
So deadlock occurs。

cgillum:
Generally speaking, `Task.Run(() => tcs.TrySetResult(...))` is a good pattern to use because of the synchronous continuation. It seems correct to me in this case to avoid other types of deadlocks.

In this case, the use of `thisLock` and a mix of sync/async like you mentioned seems like a bigger problem with this code. For example [here](https://github.com/Azure/durabletask/blob/6814e4d5e1ea5de2136afa6041dc81119efabec9/src/DurableTask.Emulator/LocalOrchestrationService.cs#L464). It seems like we should be using `await` instead of blocking with `GetAwaiter().GetResult()` (which can be very dangerous). Wouldn't it be better to fix the emulator code by having it implement the async pattern correctly?

ChenglongLiu:
Mix sync/async code is a disaster. Replace sync code with async code is a better solution. We need to do the following work:

- Use await replace GetAwaiter().GetResult();
- Replace sync lock with async lock, maybe AsyncLock or something

I'd like to look into it.

PR #455, opened 2020-09-29
Make partition management strategy default

ConnorMcMahon:
Adds a small fix to new partition management strategy, and makes it the default for DurableTask.AzureStorage.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
@@ -16,6 +16,7 @@ namespace DurableTask.AzureStorage.Partitioning
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
@@ -179,6 +180,7 @@ namespace DurableTask.AzureStorage.Partitioning
                 string.Empty /* partitionId */,
                 $"Starting background renewal of {this.leaseType} leases with interval: {this.options.RenewInterval}.");
 
+            Trace.TraceInformation($"{this.workerName}: Starting lease renewal {this.leaseType} thread");
             while (this.isStarted == 1 || !shutdownComplete)
             {
                 try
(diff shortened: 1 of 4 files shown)


test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -257,7 +257,7 @@ public async Task AutoStart(bool enableExtendedSessions)
         [DataTestMethod]
         [DataRow(false)]
         [DataRow(true)]
-        public async Task ContinueAsNewThenTimer(bool enableExtendedSessions)
+         public async Task ContinueAsNewThenTimer(bool enableExtendedSessions)
```
ConnorMcMahon, line 260: this is probably an accident.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -199,6 +200,10 @@ async Task LeaseRenewer()
                                     failedToRenewLeases.Add(lease);
                                 }
                             }));
+                        } 
+                        else 
+                        {
+                            await this.leaseObserverManager.NotifyShardReleasedAsync(lease, CloseReason.LeaseLost);
```
cgillum, line 206: What is this change for?

ConnorMcMahon: I can add a comment in the code to elaborate, but this is the general gist.

For intent leases and the legacy partition leases, `this.shouldRenewLeaseDelegate()` should always return true, so this code path is not a concern. This is because eventually, if someone is going to steal the lease, the renew will fail and we will call `NotifyShardReleasedAsync()`. 

For ownership leases, `ShouldRenewLeaseDelegate()` returning false means that someone has stolen the intent lease, and we are done processing all of our messages for the partition, so we are functionally "released". However, since we won't have a failed renewal task (we just stop renewing), we need to explicitly call `NotifyShardReleasedAsync()` here.

In practice, this shouldn't have too much of an effect in prod, as we have already stopped listening to the queue, but it did matter to the test code which want's to make sure that the ownership lease was "offically released".

ConnorMcMahon: Now that I relook at this, it may be more clear if in the else case, I add the lease to the `failedToRenewLeases` list, and let it be handled in that way, then the code reads a bit more like "If I should not renew this lease, I should treat it as equivalent to failing to renew the lease"

cgillum (approved):

PR #459, opened 2020-10-13
[ServiceFabricProvider]: Fix a race condition while adding or updating the lockedSessions dictionary

shankarsama:
Orchestration Dispatcher is trying to Release the session (UnlockSession) while the Activity Dispatcher is trying to complete the activity (EnqueueSession). In the Service Fabric provider layer, we use a concurrent dictionary for session locks. EnqueueSession and UnlockSession actions tried to update the SessionLocks dictionary in non-atomic way which left the session in a state where it can't be made available for further processing.


aresowj:
LGTM.

vinaysurya (approved):

lchew7 (approved):

PR #460, opened 2020-10-21
[ServiceFabricProvider]: Fix a race condition while adding or updating the lockedSessions dictionary and other changes

shankarsama:
1. Fixed the race condition in SF provider
2. Fixed IsMaxMessageCountExceeded to enable more parallel processing
3. Fixed RemoteTaskHubClient to handle exceptions properly
4. Fixed multiple flaky tests
5. Locate the Service Fabric test application using relative path

### DIFF
--- a/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
+++ b/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
@@ -117,7 +117,9 @@ namespace DurableTask.AzureServiceFabric
 
         public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState)
         {
-            return true;
+            // TODO: To avoid accumualating too many messages, make a decision based on the activities in progress 
+            // This is needed in case of parall tasks mostly (like WhenAll)
+            return currentMessageCount > 1000;
         }
 
         public int GetDelayInSecondsAfterOnProcessException(Exception exception)
--- a/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
+++ b/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
@@ -117,9 +117,7 @@ namespace DurableTask.AzureServiceFabric
 
         public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState)
         {
-            // TODO: To avoid accumualating too many messages, make a decision based on the activities in progress 
-            // This is needed in case of parall tasks mostly (like WhenAll)
-            return currentMessageCount > 1000;
+            return false;
         }
 
         public int GetDelayInSecondsAfterOnProcessException(Exception exception)
--- a/src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
+++ b/src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
@@ -47,6 +47,11 @@ namespace DurableTask.AzureServiceFabric
         #region IOrchestrationServiceClient
         public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
         {
+            if (creationMessage.Event is ExecutionStartedEvent executionStarted && executionStarted.ScheduledStartTime.HasValue)
+            {
+                throw new NotSupportedException("Service Fabric storage provider for Durable Tasks currently does not support scheduled starts");
+            }
+
             creationMessage.OrchestrationInstance.InstanceId.EnsureValidInstanceId();
             ExecutionStartedEvent startEvent = creationMessage.Event as ExecutionStartedEvent;
             if (startEvent == null)
@@ -92,13 +97,6 @@ namespace DurableTask.AzureServiceFabric
                 throw new NotSupportedException($"DedupeStatuses are not supported yet with service fabric provider");
             }
 
-            if (creationMessage.Event is ExecutionStartedEvent executionStarted && executionStarted.ScheduledStartTime.HasValue)
-            {
-                throw new NotSupportedException("Service Fabric storage provider for Durable Tasks currently does not support scheduled starts");
-            }
-
-            creationMessage.OrchestrationInstance.InstanceId.EnsureValidInstanceId();
-
             return CreateTaskOrchestrationAsync(creationMessage);
         }
 
@@ -127,13 +125,30 @@ namespace DurableTask.AzureServiceFabric
                 throw new ArgumentException($"No execution id found for given instanceId {instanceId}, can only terminate the latest execution of a given orchestration");
             }
 
-            var taskMessage = new TaskMessage
+            if (reason?.Trim().StartsWith("CleanupStore", StringComparison.OrdinalIgnoreCase) == true)
             {
-                OrchestrationInstance = new OrchestrationInstance { InstanceId = instanceId, ExecutionId = latestExecutionId },
-                Event = new ExecutionTerminatedEvent(-1, reason)
-            };
+                using (var txn = this.stateManager.CreateTransaction())
+                {
+                    // DropSession does 2 things (like mentioned in the comments above) - remove the row from sessions dictionary
+                    // and delete the session messages dictionary. The second step is in a background thread and not part of transaction.
+                    // However even if this transaction failed but we ended up deleting session messages dictionary, that's ok - at
+                    // that time, it should be an empty dictionary and we would have updated the runtime session state to full completed
+                    // state in the transaction from Complete method. So the subsequent attempt would be able to complete the session.
+                    var instance = new OrchestrationInstance { InstanceId = instanceId, ExecutionId = latestExecutionId };
+                    await this.orchestrationProvider.DropSession(txn, instance);
+                    await txn.CommitAsync();
+                }
+            }
+            else
+            {
+                var taskMessage = new TaskMessage
+                {
+                    OrchestrationInstance = new OrchestrationInstance { InstanceId = instanceId, ExecutionId = latestExecutionId },
+                    Event = new ExecutionTerminatedEvent(-1, reason)
+                };
 
-            await SendTaskOrchestrationMessageAsync(taskMessage);
+                await SendTaskOrchestrationMessageAsync(taskMessage);
+            }
         }
 
         public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions)
--- a/src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
+++ b/src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
@@ -36,7 +36,7 @@ namespace DurableTask.AzureServiceFabric.Remote
     public class RemoteOrchestrationServiceClient : IOrchestrationServiceClient, IDisposable
     {
         private readonly IPartitionEndpointResolver partitionProvider;
-        private readonly int maxPollDealyInSecs = 10;
+        private readonly int maxPollDelayInSecs = 10;
         private HttpClient httpClient;
 
         /// <summary>
@@ -227,7 +227,7 @@ namespace DurableTask.AzureServiceFabric.Remote
             instanceId.EnsureValidInstanceId();
             var maxTime = DateTime.Now.Add(timeout);
             OrchestrationState state = null;
-            double pollDelayInSecs = maxPollDealyInSecs;
+            double pollDelayInSecs = maxPollDelayInSecs;
             if (timeout.TotalSeconds < pollDelayInSecs)
             {
                 pollDelayInSecs = timeout.TotalSeconds;
--- a/src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
+++ b/src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
@@ -36,7 +36,7 @@ namespace DurableTask.AzureServiceFabric.Remote
     public class RemoteOrchestrationServiceClient : IOrchestrationServiceClient, IDisposable
     {
         private readonly IPartitionEndpointResolver partitionProvider;
-        private readonly TimeSpan pollDelay = TimeSpan.FromSeconds(10);
+        private readonly int maxPollDealyInSecs = 10;
         private HttpClient httpClient;
 
         /// <summary>
@@ -78,11 +78,9 @@ namespace DurableTask.AzureServiceFabric.Remote
         /// <param name="creationMessage">Orchestration creation message</param>
         /// <exception cref="OrchestrationAlreadyExistsException">Will throw an OrchestrationAlreadyExistsException exception If any orchestration with the same instance Id exists in the instance store.</exception>
         /// <returns></returns>
-        public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
+        public Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
         {
-            creationMessage.OrchestrationInstance.InstanceId.EnsureValidInstanceId();
-            var uri = await ConstructEndpointUriAsync(creationMessage.OrchestrationInstance.InstanceId, GetOrchestrationFragment(), CancellationToken.None);
-            await this.PutJsonAsync(uri, new CreateTaskOrchestrationParameters() { TaskMessage = creationMessage });
+            return this.CreateTaskOrchestrationAsync(creationMessage, null);
         }
 
         /// <summary>
@@ -155,6 +153,11 @@ namespace DurableTask.AzureServiceFabric.Remote
         /// <returns>The OrchestrationState of the specified instanceId or null if not found</returns>
         public async Task<OrchestrationState> GetOrchestrationStateAsync(string instanceId, string executionId)
         {
+            if (string.IsNullOrWhiteSpace(executionId))
+            {
+                return (await this.GetOrchestrationStateAsync(instanceId, false)).FirstOrDefault();
+            }
+
             instanceId.EnsureValidInstanceId();
             var uri = await ConstructEndpointUriAsync(instanceId, GetOrchestrationFragment(instanceId), CancellationToken.None);
             var builder = new UriBuilder($"{uri}?executionId={executionId}");
@@ -224,16 +227,28 @@ namespace DurableTask.AzureServiceFabric.Remote
             instanceId.EnsureValidInstanceId();
             var maxTime = DateTime.Now.Add(timeout);
             OrchestrationState state = null;
-            while (!cancellationToken.IsCancellationRequested && DateTime.Now < maxTime)
+            double pollDelayInSecs = maxPollDealyInSecs;
+            if (timeout.TotalSeconds < pollDelayInSecs)
             {
+                pollDelayInSecs = timeout.TotalSeconds;
+            }
+
+            state = await this.GetOrchestrationStateAsync(instanceId, executionId);
+            if (state != null && state.OrchestrationStatus.IsTerminalState())
+            {
+                return state;
+            }
+
+            do
+            {
+                await Task.Delay(TimeSpan.FromSeconds(pollDelayInSecs), cancellationToken);
                 state = await this.GetOrchestrationStateAsync(instanceId, executionId);
                 if (state != null && state.OrchestrationStatus.IsTerminalState())
                 {
                     return state;
                 }
 
-                await Task.Delay(this.pollDelay);
-            }
+            } while (!cancellationToken.IsCancellationRequested && DateTime.Now < maxTime);
 
             // Either cancellation was requested or timedout, return the last known state.
             return state;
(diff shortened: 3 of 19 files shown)


src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -117,7 +117,9 @@ public Task DeleteAsync(bool deleteInstanceStore)
 
         public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState)
         {
-            return true;
+            // TODO: To avoid accumualating too many messages, make a decision based on the activities in progress
```
lchew7, line 120: accumulating

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -117,7 +117,9 @@ public Task DeleteAsync(bool deleteInstanceStore)
 
         public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState)
         {
-            return true;
+            // TODO: To avoid accumualating too many messages, make a decision based on the activities in progress 
+            // This is needed in case of parall tasks mostly (like WhenAll)
```
lchew7, line 121: parallel

src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
```diff
@@ -47,6 +47,11 @@ public FabricOrchestrationServiceClient(IReliableStateManager stateManager, Sess
         #region IOrchestrationServiceClient
         public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
         {
+            if (creationMessage.Event is ExecutionStartedEvent executionStarted && executionStarted.ScheduledStartTime.HasValue)
```
lchew7, line 50: Why this change?

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -224,16 +227,28 @@ public async Task<OrchestrationState> WaitForOrchestrationAsync(string instanceI
             instanceId.EnsureValidInstanceId();
             var maxTime = DateTime.Now.Add(timeout);
             OrchestrationState state = null;
-            while (!cancellationToken.IsCancellationRequested && DateTime.Now < maxTime)
+            double pollDelayInSecs = maxPollDealyInSecs;
```
lchew7, line 230: maxPollDelayInSecs

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
```diff
@@ -314,6 +309,26 @@ void TryEnqueueSession(string instanceId)
             }
         }
 
+        public void TryUnlockSession(OrchestrationInstance instance, bool abandon = false, bool isComplete = false)
+        {
+            ServiceFabricProviderEventSource.Tracing.TraceMessage(instance.InstanceId, $"Session Unlock Begin, Abandon = {abandon}");
```
lchew7, line 314: Should we also log the value of isComplete?

test/DurableTask.AzureServiceFabric.Integration.Tests/AssemblySetup.cs
```diff
@@ -44,17 +44,16 @@ static string TestApplicationRootPath
         {
             get
             {
-                var sourceRoot = Environment.GetEnvironmentVariable("SourceRoot") ?? string.Empty;
-                var applicationPath = Path.Combine(sourceRoot.Trim(), "Test", "TestFabricApplication", "TestApplication");
+                var applicationPath = Path.Combine(Directory.GetCurrentDirectory(), @"..\..\..\..\TestFabricApplication\TestApplication");
```
lchew7, line 47: Was this a test change?

test/DurableTask.AzureServiceFabric.Integration.Tests/AssemblySetup.cs
```diff
@@ -44,17 +44,16 @@ static string TestApplicationRootPath
...
                 if (!Directory.Exists(applicationPath))
                 {
-                    throw new Exception("Could not find test application path, define SourceRoot environment variable to the source path");
+                    throw new Exception($"Could not find test application path '{applicationPath}', define SourceRoot environment variable to the source path");
                 }
 
                 if (!Directory.Exists(Path.Combine(applicationPath, "pkg", "Debug")))
                 {
-                    throw new Exception("Could not find test application package, make sure the test application is built and package generated before running the tests");
+                    throw new Exception($"Could not find test application package in '{applicationPath}', make sure the test application is built and package generated before running the tests");
```
lchew7, line 56: I think you should log the value of Path.Combine(applicationPath, "pkg", "Debug"), and not the value of applicationPath.

test/DurableTask.AzureServiceFabric.Integration.Tests/StressTests.cs
```diff
@@ -119,13 +119,21 @@ public async Task SimplePerformanceExperiment()
...
-            testOrchestratorInput.ExecutionTimeout = TimeSpan.FromMinutes(2);
+                testOrchestratorInput.UseTimeoutTask = true;
+                testOrchestratorInput.ExecutionTimeout = TimeSpan.FromMinutes(2);
 
-            await RunTestOrchestrationsHelper(100, testOrchestratorInput);
+                await RunTestOrchestrationsHelper(100, testOrchestratorInput);
+            }
+            catch(Exception ex)
+            {
+                Console.WriteLine($"Exception {ex.ToString()}");
```
lchew7, line 134: Is this necessary?  I think the testing framework already shows you the exception if it's not handled.

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -263,9 +278,11 @@ private async Task PutJsonAsync(Uri uri, object @object)
             };
 
             HttpResponseMessage result = await this.HttpClient.PutAsync(uri, @object, mediaFormatter);
+
+            // TODO: Improve exception handling
```
lchew7, line 282: Could you provide more detail as to what you would like to improve?  Better yet, could you create a work item describing it?

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
```diff
@@ -295,12 +272,30 @@ public async Task<PersistentSession> GetSession(string instanceId)
...
-        void TryEnqueueSession(string instanceId)
+        protected override void AddItemInMemory(string key, PersistentSession value)
+        {
+            lock (this.lockedSessions)
+            {
+                this.TryEnqueueSessionInternal(key);
+            }
+        }
+
+        void TryEnqueueSessionInternal(string instanceId)
```
lchew7, line 298: If TryEnqueueSessionInternal should always be called with a lock, can you acquire the lock here so that callers don't have to remember?

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -224,16 +227,28 @@ public async Task<OrchestrationState> WaitForOrchestrationAsync(string instanceI
...
             var maxTime = DateTime.Now.Add(timeout);
             OrchestrationState state = null;
-            while (!cancellationToken.IsCancellationRequested && DateTime.Now < maxTime)
+            double pollDelayInSecs = maxPollDealyInSecs;
+            if (timeout.TotalSeconds < pollDelayInSecs)
+            {
+                pollDelayInSecs = timeout.TotalSeconds;
+            }
+
+            state = await this.GetOrchestrationStateAsync(instanceId, executionId);
```
lchew7, line 236: Can we avoid duplicating this code outside the do-while-loop?  Why can't we run the do-while-loop without this block?

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
lchew7, line 230: Why did you change the poll delay's type from TimeSpan to int?  It seems like most of the operations need a TimeSpan, so it would be simpler to also specify the delay as a TimeSpan.

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -155,6 +153,11 @@ public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string i
         /// <returns>The OrchestrationState of the specified instanceId or null if not found</returns>
         public async Task<OrchestrationState> GetOrchestrationStateAsync(string instanceId, string executionId)
         {
+            if (string.IsNullOrWhiteSpace(executionId))
```
lchew7, line 156: Was this a bug before, that did not consider a null executionId?

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
```diff
@@ -295,12 +272,30 @@ public async Task<PersistentSession> GetSession(string instanceId)
             }, uniqueActionIdentifier: $"Orchestration InstanceId = {instanceId}, Action = {nameof(SessionProvider)}.{nameof(GetSession)}");
         }
 
+        // TryEnqueueSessionInternal:
+        // Issue: Race condition, in-between TryAdd and TryUpdate if the item is removed by TryUnlockSession then TryUpdate will miss an important state.
+        // We tried to avoid locks by using ConcurrentDictionary's atomic operations but couldn't acheive the following:
+        // 1. Conditional Removal: TryUnlockSession is trying to remove the item and add it back if the value is LockState.NewMessagesWhileLocked
+        // 2. Conditional AddOrUpdate with follow up action: TryEnqueueSessionInternal is trying to Add with LockState.InFetchQueue and if add fails 
+        //      then Update it. TryAddOrUpdate didn't help because we need to insert a record into another queue if it was an Add operation on the dictionary.
+        // Solution: To maintain atomicity between the Add/Remove operations, 'lock' is used to protect the Add/Remove operations on the dictionary.
```
lchew7, line 281: Does this mean we always have to lock the lockedSessions dictionary?  If so, can we just use a regular dictionary instead of a concurrent one?

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
shankarsama: Fixed

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: Fixed

shankarsama: Lots of typos it seems :-)

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
shankarsama: I need to think of a better way later.

---
In reply to: [508997470](https://github.com/Azure/durabletask/pull/460#discussion_r508997470) [](ancestors = 508997470)

test/DurableTask.AzureServiceFabric.Integration.Tests/AssemblySetup.cs
shankarsama: Nope. That's the permanent change.

---
In reply to: [508991109](https://github.com/Azure/durabletask/pull/460#discussion_r508991109) [](ancestors = 508991109)

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
shankarsama: Currently logging is completely disabled for release build. So it doesn't matter. 

---
In reply to: [508991049](https://github.com/Azure/durabletask/pull/460#discussion_r508991049) [](ancestors = 508991049)

src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
shankarsama: I restructured the code. Previously this check was enabled in one code and missed on the other code path. So I enabled it for all code paths by restructuring it.

---
In reply to: [508989161](https://github.com/Azure/durabletask/pull/460#discussion_r508989161) [](ancestors = 508989161)

test/DurableTask.AzureServiceFabric.Integration.Tests/AssemblySetup.cs
shankarsama: It's just test code, we can always step thru the debugger.

---
In reply to: [508991222](https://github.com/Azure/durabletask/pull/460#discussion_r508991222) [](ancestors = 508991222)

test/DurableTask.AzureServiceFabric.Integration.Tests/StressTests.cs
shankarsama: Somehow it trimmed the callstack, I had to add it to capture. I will revert this change anyway

---
In reply to: [508991715](https://github.com/Azure/durabletask/pull/460#discussion_r508991715) [](ancestors = 508991715)

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: No idea for now. I will track them as part of improvement plan.

---
In reply to: [508992378](https://github.com/Azure/durabletask/pull/460#discussion_r508992378) [](ancestors = 508992378)

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: We want to first check the state before entering and while existing the loop.

---
In reply to: [508993985](https://github.com/Azure/durabletask/pull/460#discussion_r508993985) [](ancestors = 508993985)

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: Improvement. If we execution Id  is passed then we needed to use a different query.

---
In reply to: [508996125](https://github.com/Azure/durabletask/pull/460#discussion_r508996125) [](ancestors = 508996125)

aresowj (approved):
LGTM

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -117,7 +117,9 @@ public Task DeleteAsync(bool deleteInstanceStore)
...
+            return currentMessageCount > 1000;
```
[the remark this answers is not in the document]
shankarsama: Actually I removed that check. We dont need to slow down the processing. SF provider can handle it.

---
In reply to: [509436820](https://github.com/Azure/durabletask/pull/460#discussion_r509436820) [](ancestors = 509436820)

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -224,16 +227,28 @@ public async Task<OrchestrationState> WaitForOrchestrationAsync(string instanceI
...
+                pollDelayInSecs = timeout.TotalSeconds;
+            }
+
+            state = await this.GetOrchestrationStateAsync(instanceId, executionId);
+            if (state != null && state.OrchestrationStatus.IsTerminalState())
+            {
+                return state;
+            }
+
+            do
```
vinaysurya, line 242: >do [](start = 12, length = 2)

If you wanted to make sure the loop executes atleast once without checking for the cancellation token and exiting before it does it, your change for the do-while loop will already take care of that. That change alone with the previous code should be sufficient and will get the same result as adding a check before the do-while as well right?

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -263,9 +278,11 @@ private async Task PutJsonAsync(Uri uri, object @object)
             };
 
             HttpResponseMessage result = await this.HttpClient.PutAsync(uri, @object, mediaFormatter);
+
+            // TODO: Improve exception handling
             if (result.StatusCode == System.Net.HttpStatusCode.Conflict)
             {
-                throw new OrchestrationAlreadyExistsException("Orchestration already exists");
+                throw await result.Content?.ReadAsAsync<OrchestrationAlreadyExistsException>();
```
vinaysurya, line 285: >throw await result.Content?.ReadAsAsync<OrchestrationAlreadyExistsException>(); [](start = 16, length = 79)

if result.Content is null as looks expected here, throw await null will throw a Null reference exception instead of the desired behavior here, is that what you are expecting here?

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -117,7 +117,7 @@ public Task DeleteAsync(bool deleteInstanceStore)
 
         public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState)
         {
-            return true;
+            return false;
```
vinaysurya, line 120: >return false; [](start = 12, length = 13)

why this change Shankar, was this a bug before that by default it should have been false and we returned true erroneously before?

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
vinaysurya, line 285: >throw await result.Content?.ReadAsAsync<OrchestrationAlreadyExistsException>(); [](start = 16, length = 79)

I dont understand the full flow but based on the changes made, it looks like FabricOrchestrationServiceController was always returning OrchestrationAlreadyExistsException, but now that it can return NotSupportedException with the latest change, should this change here as well. Please clarify if those two changes are not exactly related.

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
```diff
@@ -295,12 +272,30 @@ public async Task<PersistentSession> GetSession(string instanceId)
...
+        {
+            lock (this.lockedSessions)
+            {
+                this.TryEnqueueSessionInternal(key);
+            }
+        }
+
+        void TryEnqueueSessionInternal(string instanceId)
         {
             if (this.lockedSessions.TryAdd(instanceId, LockState.InFetchQueue))
```
vinaysurya, line 300: >lockedSessions [](start = 21, length = 14)

Would it be too big a change to move this to a regular dictionary instead of concurrent dictionary now? If so, or in either case, can you please add either a comment or a to-do to say why we are using both fatlock and a concurrent dictionary. It will not be entirely obvious to the reader of the code, why we use both lock and concurrent dictionary and why one was not sufficient over the other. So having a comment as to either that being the future work or reason a fat lock was added, will retain some memory of why we are having to use both here.

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: That code was calling the Delay after checking the state. After finishing with the delay it doesn't check the state anymore.

---
In reply to: [509687710](https://github.com/Azure/durabletask/pull/460#discussion_r509687710) [](ancestors = 509687710)

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: It doesn't throw the exception. I checked it. It will return Exception anyway :-)

---
In reply to: [509689113](https://github.com/Azure/durabletask/pull/460#discussion_r509689113) [](ancestors = 509689113)

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
shankarsama: Yes, that was by mistake somebody flipped for testing it seems. It took me sometime to figure out.

---
In reply to: [509689993](https://github.com/Azure/durabletask/pull/460#discussion_r509689993) [](ancestors = 509689993)

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
shankarsama: We need to move to SF RPC model or come up with an exception contract over Http to translate to correct exceptions at the client. It is going to be a big work.

---
In reply to: [509695682](https://github.com/Azure/durabletask/pull/460#discussion_r509695682) [](ancestors = 509695682)

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
shankarsama: There is should be much difference, I tested with lock and without lock, no perf difference. So we are good for now.

---
In reply to: [509699917](https://github.com/Azure/durabletask/pull/460#discussion_r509699917) [](ancestors = 509699917)

PR #461, opened 2020-10-22
Allow Netherite back-end to be implemented externally

sebastianburckhardt:
This contains the new Netherite backend implementation, with the Faster storage provider and the EventHubs transport provider. In this first version autoscaling is not yet supported, but it does support manual scaling (it balances the partitions among all the started nodes).

Most of this PR adds new files and projects; There are two new projects, DurableTask.EventSourced and DurableTask.EventSourced.Tests.

There are also some small changes / fixes to existing files:
- suppress deprecation warnings when building NuGet packages
- move entity-related functionality from AzureStorage to DurableTask.Core.Common so it can be shared by the providers
- make the `ExecutionCompletedEvent` field of `OrchestrationRuntimeState` public
- add a method to `OrchestrationState` that allows cloning of the state
- pass workitem, instead of workitem cursor, to resume
- do not start a work item renewal task if work item does not expire
- move KnownTypes from attributes into method (so other places can access this list of types)
- fix missing instanceState update in TaskOrchestrationDispatcher (this is a bug)
- fix missing ScheduledStartTime in Utils.cs (this is a bug)

### DIFF
--- a/src/DurableTask.Core/Common/Entities.cs
+++ b/src/DurableTask.Core/Common/Entities.cs
@@ -1,0 +1,105 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+using DurableTask.Core.History;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DurableTask.Core.Common
+{
+    /// <summary>
+    /// Helpers for dealing with special naming conventions around auto-started orchestrations (entities)
+    /// </summary>
+    public static class Entities
+    {
+        /// <summary>
+        /// Determine if a given instance id is an entity (an auto-started orchestration)
+        /// </summary>
+        /// <param name="instanceId"></param>
+        /// <returns></returns>
+        public static bool IsEntityInstance(string instanceId)
+        {
+            return !string.IsNullOrEmpty(instanceId) && instanceId[0] == '@';
+        }
+
+        /// <summary>
+        /// Determine if a task message is an entity message with a time delay
+        /// </summary>
+        /// <param name="taskMessage">The task message</param>
+        /// <param name="due">The universal time at which this message should be delivered</param>
+        /// <returns>true if this is an entity message with a time delay</returns>
+        /// <remarks>
+        /// We assume that auto-started orchestrations (i.e. instance ids starting with '@') are
+        /// used exclusively by durable entities; so we can follow 
+        /// a custom naming convention to pass a time parameter.
+        /// </remarks>
+        public static bool IsDelayedEntityMessage(TaskMessage taskMessage, out DateTime due)
+        {
+            // Special functionality for entity messages with a delivery delay 
+            if (taskMessage.Event is EventRaisedEvent eventRaisedEvent
+                && IsEntityInstance(taskMessage.OrchestrationInstance.InstanceId))
+            {
+                string eventName = eventRaisedEvent.Name;
+                if (eventName != null && eventName.Length >= 3 && eventName[2] == '@'
+                    && DateTime.TryParse(eventName.Substring(3), out DateTime scheduledTime))
+                {
+                    due = scheduledTime.ToUniversalTime();
+                    return true;
+                }
+            }
+
+            due = default;
+            return false;
+        }
+
+        /// <summary>
+        /// Tests whether an instance should automatically start if receiving messages and not already being started, and inserts
+        /// a corresponding ExecutionStartedEvent taskMessage.
+        /// </summary>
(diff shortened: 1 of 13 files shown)


sebastianburckhardt:
This is now ready to review. I have removed the actual Netherite provider code; what's left are only the changes to the core (as listed in the bulleted list above).

src/DurableTask.Core/Common/Entities.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// a custom naming convention to pass a time parameter.
+        /// </remarks>
+        public static bool IsDelayedEntityMessage(TaskMessage taskMessage, out DateTime due)
+        {
+            // Special functionality for entity messages with a delivery delay 
+            if (taskMessage.Event is EventRaisedEvent eventRaisedEvent
+                && IsEntityInstance(taskMessage.OrchestrationInstance.InstanceId))
+            {
+                string eventName = eventRaisedEvent.Name;
+                if (eventName != null && eventName.Length >= 3 && eventName[2] == '@'
```
ConnorMcMahon, line 54: Out of curiosity, what do the first 2 characters of these messages look like?

src/DurableTask.Core/Common/Entities.cs
```diff
@@ -0,0 +1,105 @@
...
+                    && DateTime.TryParse(eventName.Substring(3), out DateTime scheduledTime))
+                {
+                    due = scheduledTime.ToUniversalTime();
```
ConnorMcMahon, line 57: I'm not sure that converting these to universal time is correct? Aren't the times passed here already in UTC, and running ToUniversalTime() may shift it based on the local time of the current machine?

src/DurableTask.Core/OrchestrationState.cs
```diff
@@ -113,6 +113,36 @@ public class OrchestrationState : IExtensibleDataObject
         [DataMember]
         public DateTime? ScheduledStartTime { get; set; }
 
+        /// <summary>
+        /// Clear input and/or output fields. Creates a shallow copy since
+        /// we do not want to modify the original copy.
+        /// </summary>
+        /// <returns></returns>
+        public OrchestrationState ClearFieldsImmutably(bool includeInput, bool includeOutput)
```
ConnorMcMahon, line 121: This API is a bit confusing to me. I would have expected the parameters to be the converse, i.e. `bool clearInput, bool clearOutput`.

ConnorMcMahon (commented):
Looks good for the most part, just want to make sure we iron out the UTC part before we merge.

src/DurableTask.Core/Common/Entities.cs
sebastianburckhardt: 'op', I think that was short for *operation*

src/DurableTask.Core/Common/Entities.cs
sebastianburckhardt: From my understanding, if they are already UTC, then ToUniversalTime() is a no-op (there is a bit in DateTime that can tell if it is UTC). So far this has worked correctly.

src/DurableTask.Core/OrchestrationState.cs
sebastianburckhardt: I agree, the names are not quite right. I will change that.

src/DurableTask.Core/Common/Entities.cs
ConnorMcMahon: I know that DateTime.Kind contains that information, but I'm not sure how well that is preserved with newtonsoft serialization. The reason I ask is because apparently that bit is not present on DurableOrchestrationContext.CurrentUtcDateTime according to [this issue](https://github.com/Azure/azure-functions-durable-extension/issues/1570). If we can validate that DateTime.Kind on `scheduledTime` is correctly set to UTC, then I am fine with this code as is.

sebastianburckhardt: Yeah, this boils down to how DF sets this timestamp:
```csharp
        public static string ScheduledRequestMessageEventName(DateTime scheduledUtc) => $"op@{scheduledUtc:o}";
```

We are not actually using Newtonsoft at all. Serializing DateTime using the ToString("o") format creates a 'Z' suffix which then, on deserialization, sets DateTime.Kind to DateTimeKind.Utc.

ConnorMcMahon: Super interesting! I'm curious if we should handle this a bit better in our default serialization of datetime objects, but I am always a bit concerned about changing serialization strategies for in-flight orchestrations.

This looks good to me now.

ConnorMcMahon (approved):
LGTM!

src/DurableTask.Core/Common/Entities.cs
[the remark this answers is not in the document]
sebastianburckhardt: The reason is historical; we added this feature to implement "scheduled signals" for durable entities. So it looks for the specific naming convention used by DF entity signals. 

We decided to leave this API a bit obscure because (a) it was not intended to be used by DT users directly, but only by our own DF entity implementation, and (b) we were not going to implement it in legacy DT backends, only for the backends that are intended to be used with DF.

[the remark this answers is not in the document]
tomachristian: I am asking because we are planning to bypass the `TaskOrchestrationExecutor` with our own implementation and use Netherite as a backend, and we need a way to schedule mesages back to the same orchestration. Timers are not an option because we want to "squash" the history of the orchestration by using "ContinueAsNew" (while a timer is running/message is scheduled) and timers won't work if the execution ID changes (just like entities).

Is this even a good ideea what we are trying to do? Given that this is not a "public" API to schedule messages, can we still use it without it breaking in the future? We cannot use Durable Entities, because we want to be able to host everything ourselves.

Issue #457, opened 2020-10-04
System.NullReferenceException in OrchestrationSessionManager, line 217


chrisw000:
The latest code for OrchestrationSessionManager, at line 208, looks like this:

```
                        if (executionId == null || session.Instance.ExecutionId == executionId)
                        {
                            List<MessageData> pendingMessages;
                            if (!existingSessionMessages.TryGetValue(session, out pendingMessages))
                            {
                                pendingMessages = new List<MessageData>();
                                existingSessionMessages.Add(session, pendingMessages);
                            }

                            pendingMessages.Add(data);
                        }
```

in the above snippet of code;
`
List<MessageData> pendingMessages;
`

will remain null, when:

`
if (!existingSessionMessages.TryGetValue(session, out pendingMessages))
`
is not fulfilled. 

Therefore leading to the NullReferenceException on
`
pendingMessages.Add(data);
`

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at DurableTask.AzureStorage.OrchestrationSessionManager.AddMessageToPendingOrchestration(ControlQueue controlQueue, IEnumerable`1 queueMessages, Guid traceActivityId, CancellationToken cancellationToken) in C:\source\durabletask\src\DurableTask.AzureStorage\OrchestrationSessionManager.cs:line 217
   at DurableTask.AzureStorage.OrchestrationSessionManager.DequeueLoop(String partitionId, ControlQueue controlQueue, CancellationToken cancellationToken) in C:\source\durabletask\src\DurableTask.AzureStorage\OrchestrationSessionManager.cs:line 166
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_1(Object state)
   at System.Threading.QueueUserWorkItemCallback.<>c.<.cctor>b__6_0(QueueUserWorkItemCallback quwi)
   at System.Threading.ExecutionContext.RunForThreadPoolUnsafe[TState](ExecutionContext executionContext, Action`1 callback, TState& state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

cgillum:
[...]

PR #463, opened 2020-10-29
Update async void to async Task

ChenglongLiu:
If the method with async void signatures throws an exception, it crashes the process. In issue https://github.com/Azure/durabletask/issues/457 unhandled exception in the "DequeueLoop" method crashes the process.

Refer https://github.com/Microsoft/vs-threading/blob/master/doc/analyzers/VSTHRD100.md


ConnorMcMahon (approved):
This looks very straightforward to me. Let's run the CI with this just to make sure there are no unintended consequences.

cgillum:
Let's check for build warnings on this one too. Changing the return value to `Task` might cause warnings on the lines of code where we call these methods (since they are ignoring the new return values).

ChenglongLiu:
Update the lines of code where we call these methods

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -130,7 +130,8 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
                     throw new TypeMissingException($"TaskActivity {scheduledEvent.Name} version {scheduledEvent.Version} was not found");
                 }
 
-                renewTask = Task.Factory.StartNew(
+                // start a task to run RenewUntil
+                renewTask = await Task.Factory.StartNew(
```
cgillum, line 134: This new `await` look like a behavior change. Is it even correct? It seems to me that this path will block indefinitely because we'll never call Cancel() on the `renewCancellationTokenSource`.

ChenglongLiu: The return type is `Task<Task>` here, the outer `Task` wait on the lambda 
```
() => this.RenewUntil(workItem, renewCancellationTokenSource.Token)
```
, the inner Task wait on RenewUntil's execution. So here just wait for the `RenewUntil` start. I think the equivalent but simple expression is as follows. Should we replace `Task.Factory.StartNew` code with the following?
```
renewTask = this.RenewUntil(workItem, renewCancellationTokenSource.Token);
```


We do call Cancel() on `renewCancellationTokenSource` in `finally` part of the function ` OnProcessWorkItemAsync`

https://github.com/Azure/durabletask/blob/429065f0feecd8b1760fad79eb3b889ec8b1e93d/src/DurableTask.Core/TaskActivityDispatcher.cs#L195-L199

ChenglongLiu: @cgillum ping

PR #464, opened 2020-10-29
Add started event null check, Fix#457

ChenglongLiu:
Drop message when the current generation instance hasn't started.


ChenglongLiu:
For https://github.com/Azure/durabletask/issues/457

cgillum (approved):
This looks correct to me! Thanks for the contribution! I'll make sure this passes our CI testing and then merge if everything looks good. It can go out to nuget.org as part of our next month's release.

PR #465, opened 2020-10-29
Preserve null values when serializing to storage

cgillum:
I noticed this issue when working on the DurableTask.SqlServer backend provider. I think it should be a non-breaking change that only impacts that value written to the backend storage provider. I don't think this impacts null handling in user code. in that sense, it is primarily a storage and potentially a slight performance optimization.

I also incremented the version number of DurableTask.Core in this PR.


ConnorMcMahon (approved):
I'm not sure I fully understand the implications of returning `null` from this method, but since all of the tests pass I think it should be fine.

PR #469, opened 2020-11-02
Update stale taskhub.json in DurableTask.AzureStorage

ConnorMcMahon:
The Azure Functions scaling infrastructure relies on the partitionCount
value in the taskhub.json. However, our logic to create this
taskhub.json only updated the blob if it did not exist. This means that
if the taskhub lease store was created with one value of partitionCount,
and the customer updated the partitionCount, the taskhub.json would not
be updated, and the scaling logic would continue to look only at the
original control queues.

This PR aims to update the taskhub.json partition count if it is out of
sync with the provided value in AzureStorageOrchestrationService, so
that customers don't need to update this value by hand to get correct
scaling behavior.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
@@ -84,7 +84,7 @@ namespace DurableTask.AzureStorage.Partitioning
             }
         }
 
-        public async Task<bool> CreateLeaseStoreIfNotExistsAsync(TaskHubInfo eventHubInfo)
+        public async Task<bool> CreateLeaseStoreIfNotExistsAsync(TaskHubInfo eventHubInfo, bool checkIfStale)
         {
             bool result = false;
             if (!this.skipBlobContainerCreation)
@@ -93,7 +93,7 @@ namespace DurableTask.AzureStorage.Partitioning
                 this.stats.StorageRequests.Increment();
             }
 
-            await this.CreateTaskHubInfoIfNotExistAsync(eventHubInfo);
+            await this.GetOrCreateTaskHubInfoAsync(eventHubInfo, checkIfStale: checkIfStale);
 
             return result;
         }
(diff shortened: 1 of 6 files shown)


src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -348,32 +348,27 @@ public async Task CreateTaskHubInfoIfNotExistAsync(TaskHubInfo taskHubInfo)
...
-        internal async Task<TaskHubInfo> GetOrCreateTaskHubInfoAsync(TaskHubInfo createdTaskHubInfo)
+        internal async Task<TaskHubInfo> GetOrCreateTaskHubInfoAsync(TaskHubInfo newTaskHubInfo, bool checkIfStale)
         {
             TaskHubInfo currentTaskHubInfo = await this.GetTaskHubInfoAsync();
             if (currentTaskHubInfo != null)
             {
+                if (checkIfStale && IsStale(currentTaskHubInfo, newTaskHubInfo))
+                {
+                    string serializedInfo = JsonConvert.SerializeObject(newTaskHubInfo);
+                    await this.taskHubInfoBlob.UploadTextAsync(serializedInfo, null, AccessCondition.GenerateEmptyCondition(), null, null);
```
cgillum, line 359: In the previous implementation we handled `StorageException` in a catch block because of potential concurrency conflicts. It doesn't look like we have this protection anymore with this new implementation. Is that something we need to be worried about?

cgillum, line 359: Would it be helpful to have a log statement that we write whenever we update this blob?

cgillum (commented):
The implementation looks good to me. I'm just wondering if we need to add any robustness to this code.

One concern I have is that changing the partition count could cause some disruption for in-flight instances. If the partition count is _decreased_ then orchestrations could get stuck because they have messages in queues that are no longer being polled. If the partition count is _increased or decreased_ then the hash function will map instance IDs to different control queues and cause (I think) temporary split brain issues. It might be possible to mitigate the latter by checking to make sure any message we dequeue was pulled from the correct queue based on the current hash mapping. If we pulled a message for an instance that belongs to a different partition, then we could potentially requeue the message in the correct queue/partition. That might make this PR quite a bit larger, though.
 
Thoughts?

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
ConnorMcMahon: Probably a good idea!

ConnorMcMahon: Yes. I am thinking of logging it as a warning at this time, considering changing the partition count for a taskhub can break in flight orchestrations as you mentioned.

ConnorMcMahon:
@cgillum, good point about partition count changes breaking in flight orchestrations. That being said, I know many people do reuse TaskHubs and just make sure to only make breaking changes when there are no inflight orchestrations.

I would keep the scope of this PR to making sure that updating the partition count doesn't break our scaling logic, with the assumption that customers are updating partition count for a taskhub only when it is safe to do so.

I think opening a separate ticket to explore whether we can make updating partition count on a taskhub a safe operation for in flight orchestrations would be worthwhile, but outside of the scope of what we are trying to accomplish with this PR.

cgillum:
@ConnorMcMahon I'm not super comfortable taking this PR as-is. I think the split-brain problem is actually worse than I described because it would be semi-permanent. If an existing instance has a message in a control queue that belongs to a different partition, then it will get stuck in a dequeue/abandon loop indefinitely, right? That would cause permanent performance problems and I worry about the potential impact on IPD. I'd feel more comfortable if we had some sort of mitigation or at least an error message. We'll also need to clearly document the impact of changing the partition count so that customers aren't left in the dark when this happens.

ConnorMcMahon:
@cgillum 

I understand that in general, changing the partition count of a TaskHub is a bad idea unless you are very careful. We should definitely add in documentation to reflect that. I just don't see how this PR makes the runtime situation any worse than it already is. 

This PR doesn't change the runtime behavior at all. It just updates the metadata that the scale controller uses. I suppose that it implicitly implies that it is ok to change partition count on a task hub, but I think there are some cases where customers can do it in a safe way. In fact, I think the most common case is that customers change the partition count when they are still actively developing the app and it doesn't have active traffic. Then when they send that TaskHub into production, it doesn't scale correctly. I have seen at least two customers hit this bug, and they encountered it without hitting the nasty side affects you mentioned, so I presume they made the change when they didn't have inflight orchestrations.

If we want to get into cheap mitigations to prevent customers from breaking their task hubs via changing partition counts with in flight executions, we could look into validating that all of the control queues are empty before updating the stale taskhub.json, and fail to start up the TaskHubWorker if that isn't the case (or to start up the TaskHubWorker with the old partition count instead). That obviously isn't perfect, as there are some race conditions it doesn't cover, but it would be better than the absolute zero detection we have of this case today.

cgillum (approved):
> This PR doesn't change the runtime behavior at all.

Ohhhh, my bad. I completely misunderstood the purpose and impact of this PR! For some reason I thought the worker was getting the partition count from the blob and that this would change existing behavior. I looked again after reading your comment and see that this actually isn't changing anything other than scale controller behavior.

No need to add anything more. I'm good with this PR. Sorry for the confusion.

PR #476, opened 2020-11-20
Increment various package versions for Durable Functions v2.4.0 release

ConnorMcMahon:
Incrementing the following versions:

- DurableTask.AzureStorage v1.8.1 -> v1.8.2
- DurableTask.Core/DurableTask.Emulator v2.4.2 -> v2.5.0
- DurableTask.Redis v0.1.5-alpha -> v0.1.6-alpha

### DIFF
--- a/tools/DurableTask.props
+++ b/tools/DurableTask.props
@@ -45,9 +45,9 @@
   <!-- Nuget Package Settings -->
   <PropertyGroup>
     <PackageOutputPath>..\..\build_output\packages</PackageOutputPath>
-    <AssemblyVersion>2.4.2</AssemblyVersion>
-    <FileVersion>2.4.2</FileVersion>
-    <Version>2.4.2</Version>
+    <AssemblyVersion>2.5.0</AssemblyVersion>
+    <FileVersion>2.5.0</FileVersion>
+    <Version>2.5.0</Version>
     <Company>Microsoft</Company>
     <Authors>Microsoft</Authors>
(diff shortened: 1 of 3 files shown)


tools/DurableTask.props
```diff
@@ -45,9 +45,9 @@
   <!-- Nuget Package Settings -->
   <PropertyGroup>
     <PackageOutputPath>..\..\build_output\packages</PackageOutputPath>
-    <AssemblyVersion>2.4.2</AssemblyVersion>
```
ConnorMcMahon, line 48: Incremented minor version instead of patch here because of entity changes + distributed tracing support.

Debatably should have with DurableTask.AzureStorage, but it feels weird to start getting close to double digits in a minor version. More of an aesthetic thing than a practical reason, so if anyone disagrees I can make that 1.9.0 instead.

davidmrdavid: I don't feel strongly. Although if we ever start deploying more frequently, which would be a nice thing, we should probably start feeling comfortable with double-digits in minor versions.

Regardless, I don't think it matters that much :)

davidmrdavid (approved):
LGTM! Please ping me about kicking the release, I'd love to either shadow it or try to get it out myself

tools/DurableTask.props
amdeel: That seems reasonable to me.

amdeel (approved):
LGTM!

Issue #477, opened 2020-11-23
Fix null custom status logic for DurableTask.AzureStorage


ConnorMcMahon:
As a part of a recent change to our serialization logic (see PR #465), we changed how null is serialized. This unforunately broke our logic for setting `null` CustomStatuses in DurableTask.AzureStorage.

This is because the `Merge` table operation we are performing ignores null parameters. The correct fix for this is to change from a `Merge` table operation to a `Replace` operation, but this will require a bit more work, as we would have to restructure how we generate the payload for this operation to ensure we don't break any other functionality.

In the meantime, we have incurred some technical debt by just sending along `"null"` when we detect a null value to preserve the original behavior.

PR #478, opened 2020-11-23
Mitigate null custom status regression in DurableTask.AzureStorage

ConnorMcMahon:
Implement a temporary mitigation for `null` custom status updates not
actually updating the state in Azure Table Storage due to the way that
merge operations happen.

A better long term mitigation is detailed at
https://github.com/Azure/durabletask/issues/477.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -977,15 +977,13 @@ namespace DurableTask.AzureStorage.Tracking
 
             OrchestrationStatus runtimeStatus = OrchestrationStatus.Running;
 
-            // TODO: This is temporary workaround. We should prioritize https://github.com/Azure/durabletask/issues/477
-            // so that this is no longer necessary.
-            newRuntimeState.Status = newRuntimeState.Status ?? "null"; 
-
             var instanceEntity = new DynamicTableEntity(instanceId, string.Empty)
             {
                 Properties =
                 {
-                    ["CustomStatus"] = new EntityProperty(newRuntimeState.Status),
+                    // TODO: Translating null to "null" is a temporary workaround. We should prioritize 
+                    // https://github.com/Azure/durabletask/issues/477 so that this is no longer necessary.
+                    ["CustomStatus"] = new EntityProperty(newRuntimeState.Status ?? "null"),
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -977,6 +977,10 @@ public override Task StartAsync()
 
             OrchestrationStatus runtimeStatus = OrchestrationStatus.Running;
 
+            // TODO: This is temporary workaround. We should prioritize https://github.com/Azure/durabletask/issues/477
+            // so that this is no longer necessary.
+            newRuntimeState.Status = newRuntimeState.Status ?? "null";
```
cgillum, line 982: Why did you decide to mutate `newRuntimeState.Status` instead of doing `["CustomStatus"] = new EntityProperty(newRuntimeState.Status ?? "null")` on line 988? Would that approach work?

davidmrdavid (approved):
A hack indeed, but I suspect this should do the trick for now.
I'm surprised that an input `"null"` gets evaluated later into a proper `null` datatype, but maybe that's just standard .NET serialization logic.
Thanks for the quick fix!

davidmrdavid:
Actually, I'll wait on my approval now that I'm seeing Chris' suggestion above

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
ConnorMcMahon: It felt a bit more clear this way what action the comment as referring to as a workaround, but I could just edit the comment a bit to make it clear that the null coalescing is the workaround. That would be safer than mutating `newRuntimeState.Status`

ConnorMcMahon: Fixed.

davidmrdavid (approved):
LGTM

cgillum (approved):
:shipit:

davidmrdavid:
Before release, let's make sure the release is enough to fix the `correlation` branch in `durable-extension`

ConnorMcMahon:
@davidmrdavid, I am going to do an offiical build with DurableTask.AzureStorage, and test that locally against my PR branch here, which I believe have the same errors: https://github.com/Azure/azure-functions-durable-extension/pull/1577.

Once that passes, I will publish on nuget.org

PR #480, opened 2020-12-09
Enforce that ExecutionStarted is the first event in a WorkItem

davidmrdavid:
Currently, there's a race condition where raising an event to a pending orchestrator can lead to a raiseEvent message being processed before an executionStarted one. This leads to the workItem batch being discarded and the orchestrator getting into a bad state.

This _draft_ PR adds a pre-processing step that re-orders the workItem messages when an ExecutionStarted event is not at the beginning of its `NewMessages` property. We do the re-ordering in-place, for performance.

If we believe the in-place re-ordering is a premature optimization, we can replace the re-ordering method with two lines: one calling the IList `insert` API, and the other calling the IList `delete` API.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -174,7 +174,6 @@ namespace DurableTask.Core
                 batch.RemoveAt(executionStartedIndex);
                 batch.Insert(targetPosition, executionStartedEvent);
             }
-
         }
 
         async Task OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem)
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -167,7 +167,7 @@ namespace DurableTask.Core
 
             // If we found an ExecutionStartedEvent, we place it either
             // (A) in the beginning or
-            // (B)  after the "right-most" event with non-null executionID that came before it.
+            // (B) after the "right-most" event with non-null executionID that came before it.
             int executionStartedIndex = index;
             if ((executionStartedEvent != null) && (executionStartedIndex != targetPosition))
             {
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -113,7 +113,14 @@ namespace DurableTask.Core
             return this.orchestrationService.LockNextTaskOrchestrationWorkItemAsync(receiveTimeout, cancellationToken);
         }
 
-        IList<TaskMessage> EnsureExecutionStartedIsFirst(IList<TaskMessage> batch)
+
+        /// <summary>
+        /// Ensure that the workItem batch has an ExecutionStarted event as its first element.
+        /// If not, move the first instance of an ExecutionStarted event to the beginning of the
+        /// list. Note that this method modifies its input in-place.
+        /// </summary>
+        /// <param name="batch">The batch of workitems to potentially re-order in-place</param>
+        void EnsureExecutionStartedIsFirst(IList<TaskMessage> batch)
         {
 
             // We look for *the first* instance of an ExecutionStarted event
@@ -134,29 +141,13 @@ namespace DurableTask.Core
 
             // We ensure that event is at the beginning of the list. If not,
             // We put it at the beginning but preserve the ordering of other
-            // elements. This block should have no effect if execStartedIndex
-            // is already 0.
+            // elements.
             int execStartedIndex = index;
-            if (foundEvent)
+            if (foundEvent && (execStartedIndex != 0))
             {
-                // Starting from the ExecutionStarted event, replace each element
-                // with its adjacent element "to the left". This shifts all elements
-                // by 1 index starting from the start up until and including the
-                // ExecutionStarted event.
-                for(int pos = execStartedIndex; pos <= 0; index--)
-                {
-                    // Replace element in current position with element in the
-                    // position to the "left"
-                    int prevPos = pos - 1;
-                    TaskMessage newValue = batch[prevPos];
-                    batch[pos] = newValue;
-                }
-
-                // Finally, set the first value to the ExecutionStarted event
-                batch[0] = execStartedEvent;
+                batch.RemoveAt(execStartedIndex);
+                batch.Insert(0, execStartedEvent);
             }
-
-            return batch;
 
         }
 
@@ -165,7 +156,7 @@ namespace DurableTask.Core
             // DTFx history replay expects that ExecutionStarted comes before other events.
             // If this is not already the case, due to a race-condition, we re-order the
             // messages to enforce this expectation.
-            workItem.NewMessages = EnsureExecutionStartedIsFirst(workItem.NewMessages);
+            EnsureExecutionStartedIsFirst(workItem.NewMessages);
 
             try
             {
@@ -591,10 +582,11 @@ namespace DurableTask.Core
         }
 
         /// <summary>
-        /// Assumes
+        /// Determines whether the workItem should be discarded as per the orchestration's state.
+        /// It assumes that the workItem batch has an ExecutionStartedEvent as its first element.
         /// </summary>
-        /// <param name="workItem"></param>
-        /// <returns></returns>
+        /// <param name="workItem">A batch of work item messages.</param>
+        /// <returns>True if workItem should be processed further. False otherwise.</returns>
         bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
         {
             foreach (TaskMessage message in workItem.NewMessages)
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -128,7 +128,6 @@ namespace DurableTask.Core
 
             // We look for *the first* instance of an ExecutionStarted event in the batch, if any.
             int index;
-            bool foundEvent = false;
             TaskMessage execStartedEvent = null;
             Dictionary<string, int> earliestExecIdIndex = new Dictionary<string, int>();
             for(index = 0; index < batch.Count; index ++)
@@ -146,7 +145,6 @@ namespace DurableTask.Core
                 if (message.Event.EventType == EventType.ExecutionStarted)
                 {
                     execStartedEvent = message;
-                    foundEvent = true;
                     break;
                 }
             }
@@ -155,9 +153,9 @@ namespace DurableTask.Core
             // (A) in the beginning or
             // (B) at the earliest position found for an event with a matching executionID
             int execStartedIndex = index;
-            earliestExecIdIndex.TryGetValue(execStartedEvent.OrchestrationInstance.ExecutionId, out int newPosition);
-            if (foundEvent)
+            if (execStartedEvent != null)
             {
+                earliestExecIdIndex.TryGetValue(execStartedEvent.OrchestrationInstance.ExecutionId, out int newPosition);
                 batch.RemoveAt(execStartedIndex);
                 batch.Insert(newPosition, execStartedEvent);
             }
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -113,8 +113,60 @@ namespace DurableTask.Core
             return this.orchestrationService.LockNextTaskOrchestrationWorkItemAsync(receiveTimeout, cancellationToken);
         }
 
+        IList<TaskMessage> EnsureExecutionStartedIsFirst(IList<TaskMessage> batch)
+        {
+
+            // We look for *the first* instance of an ExecutionStarted event
+            // in the batch, if any.
+            int index;
+            bool foundEvent = false;
+            TaskMessage execStartedEvent = null;
+            for(index = 0; index < batch.Count; index ++)
+            {
+                TaskMessage message = batch[index];
+                if (message.Event.EventType == EventType.ExecutionStarted)
+                {
+                    execStartedEvent = message;
+                    foundEvent = true;
+                    break;
+                }
+            }
+
+            // We ensure that event is at the beginning of the list. If not,
+            // We put it at the beginning but preserve the ordering of other
+            // elements. This block should have no effect if execStartedIndex
+            // is already 0.
+            int execStartedIndex = index;
+            if (foundEvent)
+            {
+                // Starting from the ExecutionStarted event, replace each element
+                // with its adjacent element "to the left". This shifts all elements
+                // by 1 index starting from the start up until and including the
+                // ExecutionStarted event.
+                for(int pos = execStartedIndex; pos <= 0; index--)
+                {
+                    // Replace element in current position with element in the
+                    // position to the "left"
+                    int prevPos = pos - 1;
+                    TaskMessage newValue = batch[prevPos];
+                    batch[pos] = newValue;
+                }
+
+                // Finally, set the first value to the ExecutionStarted event
+                batch[0] = execStartedEvent;
+            }
+
+            return batch;
+
+        }
+
         async Task OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem)
         {
+            // DTFx history replay expects that ExecutionStarted comes before other events.
+            // If this is not already the case, due to a race-condition, we re-order the
+            // messages to enforce this expectation.
+            workItem.NewMessages = EnsureExecutionStartedIsFirst(workItem.NewMessages);
+
             try
             {
                 if (workItem.Session == null)
@@ -232,6 +284,8 @@ namespace DurableTask.Core
 
             OrchestrationState instanceState = null;
 
+
+            // Assumes Execution Started is the first event in the workItem batch
             if (!this.ReconcileMessagesWithState(workItem))
             {
                 // TODO : mark an orchestration as faulted if there is data corruption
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -113,8 +113,60 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
+                TaskMessage message = batch[index];
+                if (message.Event.EventType == EventType.ExecutionStarted)
+                {
+                    execStartedEvent = message;
+                    foundEvent = true;
+                    break;
+                }
+            }
+
+            // We ensure that event is at the beginning of the list. If not,
```
ConnorMcMahon, line 135: The IList interface should be flexible enough that we can simplify this by just calling the below code

```csharp
batch.RemoveAt(index);
batch.Insert(0, execStartedEvent);
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -113,8 +113,60 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
+                    int prevPos = pos - 1;
+                    TaskMessage newValue = batch[prevPos];
+                    batch[pos] = newValue;
+                }
+
+                // Finally, set the first value to the ExecutionStarted event
+                batch[0] = execStartedEvent;
+            }
+
+            return batch;
```
ConnorMcMahon, line 159: Since we are modifying in place, I don't know if there is much value in returning a new `IList<>`. In fact, that may be misleading to callers, as if I receive a new list I will assume the original is untouched.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -536,6 +590,11 @@ Task ResumeOrchestrationAsync(TaskOrchestrationWorkItem workItem)
             });
         }
 
+        /// <summary>
+        /// Assumes
```
ConnorMcMahon, line 594: Let's either remove these comments or fill them out.

ConnorMcMahon (commented):
Overall approach looks good, and great turnaround on this.

A few small suggestions that can simplify this, and then we can try getting a preview package in the hands of affected customers.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: Yup, left this incomplete because I wanted to get some eyes on the approach. I should have made that more explicit :)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: Good point. I'll make it `void`

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: Yeah, I was thinking about that but I was worried that I would get pushback on the performance: I think these operations are not in-place so we end-up re-allocating memory a few times. 
Seems like I was wrong about pushback, so I'll go for this!

ConnorMcMahon: So these methods actually are in-place, if by in-place, you mean that they don't instantiate any memory (except for some implementation specific edge cases)

Your original implementation is very similar to the approach used by `List<>`, since it is internally backed by an array. However, if the `IList<>` passed to us is an `LinkedList<>`, using your original implementation would actually _hurt_ performance, due to the fact that linked lists aren't performant for random access.

In general, for code that is passed an interface like this, I try to use the simplest method provided by the public methods, as each implementation has likely optimized the internal implementations of these methods already. If the code lies in a hotpath, you can generally then optimize by looking at what operations you perform on the interface, and swap out the implementation of that interface with the one that has the best performance profile for those operations. If you _really_ need to eek out performance, then you should pick a very specific datatype and optimize your algorithms for those.

In this specific case, there are two pieces of performance we need to care about: the average case, and the edge case.

In the average case, we are adding O(n) cost (where n = number of elements in the batch) by iterating through the list to check for ExecutionStarted events. Since we are just iterating through the list, I don't believe any implementations perform significantly better for this.

In the edge case, we are performing 2 O(n) operations if this is a 'List<>`. If we really want to optimize for this edge case, we switch the datatype of the batch to be a `LinkedList`, in which case a simple reorder like this costs O(m) + O(1) (where m is the index of the "Execution Started" event). That would optimize this specific case, but I think this case is a fairly rare edge case, even for customers whose usage patterns make this happen "frequently".

I realize this is now a fairly long diatribe on the topic of performance optimizations, but I think its an important topic, as sometimes we can make the problem worse, either now, or down the line if someone changes the data structure on us.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -232,6 +275,8 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
 
             OrchestrationState instanceState = null;
 
+
+            // Assumes Execution Started is the first event in the workItem batch
```
ConnorMcMahon, line 279: We may want to clarify this comment (and the summary comment for `ReconcileMessagesWithState`). 

The comment should be either:

Assumes execution started is the first event in the work item's event history

**or**

Assumes that if the batch contains a new "ExecutionStarted" event, it is the first message in the batch.

I much prefer the latter, as the while the first is technically correct, for the purposes of this code, we are far more concerned about the "new messages" than the total history of the orchestration.

ConnorMcMahon (approved):
Approved assuming the CI is happy, and once the comments I mentioned are clarified, this looks good to me.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: I hear you and this makes sense, thanks for writing it up.

cgillum (commented):
I think this change is _probably_ fine, but one tiny worry I have is how this could impact ContinueAsNew flows. In certain configurations, ContinueAsNew is represented as an ExecutionStarted event and this reordering could potentially change runtime behavior in those situations. It won't be an issue for Durable Functions but it could potentially be an issue for some DTFx users. If you want to make this 100% robust, then you'll want to consider tweaking the reordering algorithm a bit to make sure you're not changing the order of ExecutionIds in the batch. The Execution ID of an orchestration instance represents a particular generation of the orchestration. An ExecutionStarted message with the same instance ID but a different execution ID represents a transition from one generation to the next.

davidmrdavid:
@cgillum: thanks for the feedback, that's really insightful. I can try to make this more robust.
However, I'm confused about the following-case. Let's say I have a list of workItems (represented as tuple of eventTypes and instanceIDs) that looks like this:

[ (RaiseEvent, 1), (ExecutionStarted, 1), (RaiseEvent, 2), (ExecutionStarted, 2)]

Should the result be:

[ (ExecutionStarted, 1), (RaiseEvent, 1),  (RaiseEvent, 2), (ExecutionStarted, 2)]

or:

[ (ExecutionStarted, 1), (RaiseEvent, 1),  (ExecutionStarted, 2),  (RaiseEvent, 2),]

In other words, do we need to put the executionStartedEvent at the beginning of each orchestrator generation?

davidmrdavid:
My latest commit assumes that the right answer to my question above was:

[ (ExecutionStarted, 1), (RaiseEvent, 1), (RaiseEvent, 2), (ExecutionStarted, 2)]

In other words: it assumes that we only need to re-order `ExecutionStarted` events for the first executionID found in the batch.

cgillum:
@davidmrdavid good question - so it turns out your example actually isn't valid for a couple reasons: 

* `RaiseEvent` actually doesn't have an ExecutionId (it's always set to `null`). RaiseEvent can therefore be associated with any generation. The same is true for `Terminate`.
* I don't expect there would be any valid scenario where there are two `ExecutionStarted` events for the same instance in the same message batch. That is because an orchestration must have already started before it can call ContinueAsNew. The case where this might happen is if a customer tries to start two orchestrations at the same time with the same ID, but that's a totally different story for which we don't have any proper handling.

If you're curious about some of the code I had to write to navigate these subtleties, have a look at [this code that sorts messages into these batches after reading them from the queue](https://github.com/Azure/durabletask/blob/0b187d843e6bca94a50e123778bac8c94f9a5c26/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs#L181-L288) (don't worry, the problem we're tackling here is actually quite a bit simpler).

Now that I've pasted a link to the code above where we do message sorting, I'm realizing that it might not even be possible to hit a case with two ExecutionStarted events because I would have moved them into separate batches. However, that's only the case for DurableTask.AzureStorage - I don't know about other backends. But if you want to be as safe as possible, I think the best thing to do is to find the _first_ `ExecutionStarted` event in the batch and just move it backwards until either you reach the beginning of the list or until you find an event with _a different, non-null ExecutionId_. That will ensure that the race condition of start + raise events works out fine and will also ensure that we don't inadvertently complicate things with ContinueAsNew.

davidmrdavid:
My latest commit grabs the _first_ `ExecutionStarted` event and moves it either:
(A) to the beginning of the list, or...
(B) to the position of the _earliest_ event with the same executionID (of the same generation)

I can imagine a scenario where this is not the same as Chris' advice of "_move it backwards until either you reach the beginning of the list or until you find an event with a different, non-null ExecutionId_" ... but I'd assume that, if the first event we process for an executionID is not `ExecutionStarted`, then we'd error out anyways.

davidmrdavid:
Alright, CI is green. Can I get one final LGTM since the code's in a non-trivial way. Please also read my last comment above ^

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -113,8 +113,62 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
+                }
+            }
+
+            // If we found an ExecutionStartedEvent, we place it either
+            // (A) in the beginning or
+            // (B) at the earliest position found for an event with a matching executionID
+            int execStartedIndex = index;
+            if (execStartedEvent != null)
+            {
+                earliestExecIdIndex.TryGetValue(execStartedEvent.OrchestrationInstance.ExecutionId, out int newPosition);
```
ConnorMcMahon, line 158: I'm not convinced that this solves our edge case anymore.

Our edge case looks like the following:

`("ExternalEvent", null), ("ExecutionStarted", "NewExecutionId")`


In that case, this is just going to remove the ExecutionStarted event and readd it right where it was.

ConnorMcMahon (changes requested):

rfeltis:
A secondary issue is the detection and handling of the orchestrator getting into a bad state. Will that be addressed at all? Right now, we are polling `OrchestrationRuntimeStatus.Pending` with `CreatedTimeTo = DateTime.UtcNow.AddMinutes(-5)` every 5 minutes. That allows us to detect these bad orchestrations, but we can't do much with them once we detect them. We can terminate and restart them, but if we are trying to reuse the same instance ID, it stays in this bad state.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: You're right, I should have caught this but forgot about the `null`-ness of ExternalEvent executionIDs.
I'm fixing this in my next commit

davidmrdavid:
@rfeltis, could we move that discussion to a GitHub ticket? Otherwise, this PR might be cluttered with two different conversations. This PR _should_ mitigate the error we're currently seeing. As for recovering if a similar error were to occur again, we can certainly discuss strategies but let's start a separate thread on a GitHub issue. Do you mind opening that up and tagging me? Maybe referencing this PR as well?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -127,37 +127,46 @@ void EnsureExecutionStartedIsFirst(IList<TaskMessage> batch)
...
                 }
+                index++;
             }
 
             // If we found an ExecutionStartedEvent, we place it either
             // (A) in the beginning or
-            // (B) at the earliest position found for an event with a matching executionID
+            // (B)  after the "right-most" event with non-null executionID that came before it.
             int execStartedIndex = index;
             if (execStartedEvent != null)
```
ConnorMcMahon, line 166: nit: Let's consider adding a check that `execStartedIndex != targetIndex`, to avoid the cost of the reorder operations if they aren't necessary.

ConnorMcMahon (commented):
@cgillum after some further thought, I am not sure if we want to preserve execution id ordering.

Imagine an orchestration that ends in the following psuedocode:

```csharp
// Due to Task.WhenAny(), some of these tasks may complete between the time we
// start the last replay, but before ContinueAsNew() creates a new ExecutionStarted event
await Task.WhenAny(seriesOfTasks);
context.ContinueAsNew();
```


I would anticipate that if an external event came at the right time, we could come to some history like this.

`("ExternalEvent", null), ("TaskCompleted", "ExecutionId1"), ("TaskCompleted", "ExecutionId1"), ("TaskCompleted", "ExecutionId1"), ("ExecutionStarted", "ExecutionId2")`. I'm afraid we would end up discarding the whole batch still.

cgillum:
@ConnorMcMahon the weird thing about your example, though, is that the orchestration calls `ContinueAsNew()` before knowing whether all outstanding tasks have completed. I'm not sure how to think about that in terms of a user's expectations.

That said, it occurs to me that we might be able to side-step the issue of ContinueAsNew by only reordering when `ExecutionStartedEvent.ParentInstance` is null. That way we're never confusing orchestration start scenarios with ContinueAsNew scenarios.

davidmrdavid:
My latest commit inspects the `ParentInstance` property of the events to decide whether or not to re-order. We still break after encountering the very first instance of an `ExecutionStarted` event.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -536,6 +609,12 @@ Task ResumeOrchestrationAsync(TaskOrchestrationWorkItem workItem)
             });
         }
 
+        /// <summary>
+        /// Determines whether the workItem should be discarded as per the orchestration's state.
```
cgillum, line 613: nit: this summary doesn't include the most important behavior of this method, which is to convert the new messages into history events that get appended to the existing orchestration state.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -113,8 +113,79 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
+        /// of its executionID history.
+        /// If this is not already the case, we move the first ExecutionStarted event "backwards"
+        /// until it either reaches the beginning of the list or reaches a different, non-null, executionID.
+        ///
+        /// Note that this method modifies its input in-place.
+        /// </summary>
+        /// <param name="batch">The batch of workitems to potentially re-order in-place</param>
+        void EnsureExecutionStartedIsFirst(IList<TaskMessage> batch)
+        {
+
```
cgillum, line 128: nit: please remove the blank lines at the start and end of this methods. Also, if you could remove the extra newline on line 152. I just want to make sure that we're being stylistically consistent with the rest of this file.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -113,8 +113,79 @@ protected Task<TaskOrchestrationWorkItem> OnFetchWorkItemAsync(TimeSpan receiveT
...
+            // We look for *the first* instance of an ExecutionStarted event in the batch, if any.
+            int index = 0;
+            string prevExecId = "";
```
cgillum, line 131: nit: in C# abbreviations like "prev" and "exec" are generally discouraged vs. spelling out the actual word. We also don't name variables like this generally in this codebase. It would be nice if we could try to follow the existing variable naming style/conventions to help ensure consistency.

cgillum (commented):
A few small stylistic nits

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: Sounds like we need a linter for this repo's CI :)

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: @cgillum: thanks for pointing that out. I actually was thinking about that while writing this blurb. Do we have an explicit style guide anywhere? I don't mind correcting these, of course, just wondering

davidmrdavid: Also, what do you mean by "name variables like _this_"? Do you mean that we discourage using names like `previousExecutionId` or that we discourage using shortened a name like `prevExecId`? If it's the former, do you have a suggestion?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: Haha, good point! I found the method to be a little obscure (I really wasn't sure what it's purpose was) but I figured it needed at least _some_ commenting since we found a bug in it. 

While we're at it, what do you think about renaming it? I'm not too sure that `ReconcileMessagesWithState` implies converting messages into history events. How about the more literal alternative: `ConvertMessagesIntoHistoryEvents`? I don't feel strongly, just a thought.

davidmrdavid:
My latest PR applies the stylistic advice provided above.

ConnorMcMahon (approved):
LGTM!

PR #488, opened 2020-12-15
Fix a null-ref in the core deserialization path

cgillum:
Resolves https://github.com/Azure/azure-functions-durable-extension/issues/1613.

The problem is that we're trying to cast a `null` value into `T`, which is only legal if `T` is a reference type. If it is a value type, like `int` or `Guid` then the cast fails with a `NullReferenceException`. I found this by testing Durable Entities using the SQL backend.

Also included in this PR is a small fix to our logging to handle a certain scenario where we might unintentionally log a null value.

I've also proactively incremented the version of DurableTask.Core in this PR to **2.5.1**.


ConnorMcMahon (approved):
LGTM!

Issue #491, opened 2020-12-16
DTFx.AzureStorage not honoring DataContract on OrchestrationInstance during serializing


jviau:
`OrchestrationInstance` is marked with `DataContract` and has an `ExtensionData` object on it. This is to allow round-trip serialization when the types do not exactly match on each end, without losing any data. Unfortunately, AzureStorage is using custom Newtonsoft Json serialization with `TypeNameHandling.Object` and an `ISerializationBinder` which drops assembly name. The result is that AzureStorage implementation cannot round-trip any OrchestrationInstance serialization that has a change in type name. This means that any refactor that changes OrchestrationInstance full name will break ongoing orchestrations, and it outright prevents any use of custom OrchestrationInstance implementations (which is possible with other providers, like ServiceBus).

Luckily Newtonsoft.Json does honor `DataContract` attributes already, the fix should be as simple as adding `[JsonProperty(TypeNameHandling = TypeNameHandling.None)]` to `MessageData.Sender`.

cgillum:
@jviau can you provide an example scenario that requires custom, serializable `OrchestrationInstance` types with DTFx? I was not involved in the project when this type was originally designed and therefore don't have a sense for how to prioritize fixing the current DTFx.AzureStorage behavior.

jviau:
@cgillum I use custom `OrchestrationInstance` to flow a general bag of metadata with a given orchestration. This is useful for setting something like a `CorrelationId` for an orchestration, all of its sub orchestrations, and activities. Here is an example of how I do it: https://github.com/jviau/durabletask-hosting/tree/master/samples/DurableTask.Samples/SessionData (all 3 files).

I am sure this is not an _officially_ supported scenario, but since `OrchestrationInstance` is marked with `DataContract`, it _should_ work so long as everyone honors that contract.

The other issue here, is if `OrchestrationInstance` is ever renamed, or has its namespace changed, all running orchestrations for AzureStorage will break when upgrading. This can be avoided by not having `TypeNameHandling.Objects`. I put up a PR to do a quick fix for just this one object, but as I mentioned there, AzureStorage may want to evaluate dropping that TypeNameHandling altogether or rethink it at least.

cgillum:
Thanks for this additional context. Your correlation scenario is quite interesting. I also agree with your point about the potential fragility around type renames (I recall that the Service Bus provider maintainers had to do a lot of work related to this at one point).

[...]

PR #492, opened 2020-12-16
TypeNameHandling.None on MessageData.Sender

jviau:
Fix for #491 

Overall, AzureStorage owners should re-evaluate their use of `TypeNameHandling.Object` while serializing here, as it makes the whole system very fragile to type renames and namespace changes. My suggestion is to use `TypeNameHandling.None` and to implement `IExtensibleDataObject` on `MessageData` to give this the ability to round trip between difference versions of DurableTask.AzureStorage and not lose any data.

This change should not break existing orchestrations, Newtonsoft.Json can handle mismatch between `$type` property and `TypeNameHandling` setting.


cgillum (approved):
Looks good to me! Thanks for submitting this!

PR #494, opened 2021-01-05
Public ReflectionBasedTaskActivity.DataConverter setter

jviau:
Small PR to make this property have a public setter, but also enforce non-null values. The justification is that it may be useful to override this value in the dispatcher middleware with a custom converter - for example our system has a DI-enabled converter.

### DIFF
--- a/src/DurableTask.Core/ReflectionBasedTaskActivity.cs
+++ b/src/DurableTask.Core/ReflectionBasedTaskActivity.cs
@@ -26,6 +26,8 @@ namespace DurableTask.Core
     /// </summary>
     public class ReflectionBasedTaskActivity : TaskActivity
     {
+        private DataConverter dataConverter;
+
         /// <summary>
         /// Creates a new ReflectionBasedTaskActivity based on an activity object and method info
         /// </summary>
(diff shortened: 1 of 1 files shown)


cgillum:
Just to make sure I understand, you want to create a public setter rather than add a constructor parameter because you want some middleware to be able to change this without knowing anything about the activity?

Could you provide a small code snippet that illustrates how you would use this?

jviau:
Having this as part of the public ctor would require also providing a custom `ObjectCreator<T>` for this, preventing the use of `TaskHubWorker.AddActivitiesFromInterface<T>(T impl)`. By having a public setter, you can still use that function and swap out the data converter after DTFx has already constructed the activity (but before execution).

Example middleware:

``` CSharp
Task MyMiddleware(DispatchMiddlewareContext context, Func<Task> next)
{
    TaskActivity taskActivity = context.GetProperty<TaskActivity>();
    if (taskActivity is ReflectionBasedTaskActivity reflectionActivity)
    {
        reflectionActivity.DataConverter = new MyDataConverter();
    }

    return next;
}
```

PR #496, opened 2021-01-07
Introduce INameVersionProvider

### LINKED ISSUE #495: Proposal: introduce `INameVersionProvider`
### Background
I am working on enabling open-generic orchestrations and activities in DTFx. This will allow for registering an open generic to the TaskHubWorker (ie: `MyTaskActivity<T>`) and being able to schedule closed generic versions of it (`context.ScheduleTask<string>(typeof(MyTaskActivity<string>), "some input");`). I have it working for the most part, just the overloads where DTFx takes in the type and uses `NameVersionHelper.GetDefaultName(object obj)` does not work - as this uses `Type.ToString()` when I need `Type.FullName`. The difference between the two is `FullName` will include assembly info in the generic arguments, which is essential for loading that type in the worker.

### Proposal
1. Introduce a new interface:

``` CSharp
public interface INameVersionProvider
{
    string GetName(object obj, bool useFullyQualifiedMethodNames = false);

    string GetVersion(object obj);
}
```

2. Plum that interface through all places where `NameVersionHelper` is used.
   - `TaskHubClient` and `TaskHubWorker` should accept this interface - either a ctor overload or a public property.
   - The provider supplied to the `TaskHubWorker` should flow in to the `OrchestrationContext` and also be used there.
   - Supplying this will be optional.
3. When not supplied, `NameVersionHelper` will be used instead.

The goal will be to introduce no break changes and make this interface entirely optional - so users will only see a different behavior if they opt to supply a custom `INameVersionProvider`.

I have a PR nearly ready for this - will put it up soon.

jviau:
This PR introduces the ability to customize how DTFx derives its default name/versions for orchestrations and activities. As explained in #495, the motivation in this is to use `Type.FullName` instead of `Type.ToString()` when getting the default name so that the type params for generics contains the assembly info, allowing for fetching of the type by string when processing the orchestration or activity.

Closes #495

### DIFF
--- a/src/DurableTask.Core/INameVersionProvider.cs
+++ b/src/DurableTask.Core/INameVersionProvider.cs
@@ -1,0 +1,36 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.Core
+{
+    /// <summary>
+    /// An interface for supplying names and versions for a given object.
+    /// </summary>
+    public interface INameVersionProvider
+    {
+        /// <summary>
+        /// Gets the name of an Object instance.
+        /// </summary>
+        /// <param name="obj">Object to get the name for</param>
+        /// <param name="useFullyQualifiedMethodNames">Boolean indicating whether to use fully qualified names or not</param>
+        /// <returns>Name of the object instance's type</returns>
+        string GetName(object obj, bool useFullyQualifiedMethodNames = false);
+
+        /// <summary>
+        /// Gets the default version for an object.
+        /// </summary>
+        /// <param name="obj">Object to get the version for</param>
+        /// <returns>The version as string</returns>
+        string GetVersion(object obj);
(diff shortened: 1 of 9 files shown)


src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -32,6 +32,12 @@ public abstract class OrchestrationContext
         [ThreadStatic]
         public static bool IsOrchestratorThread;
 
+        /// <summary>
+        /// INameVersionProvider for getting names and versions when not explicitly supplied.
+        /// When null, <see cref="NameVersionHelper.Provider"/> is used.
+        /// </summary>
+        public INameVersionProvider NameVersionProvider { get; set; }
```
cgillum, line 39: Does this property need to be public here? It seems counter-intuitive to me to expose it the orchestration logic. Could we instead make it internal?

src/DurableTask.Core/INameVersionProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public interface INameVersionProvider
+    {
+        /// <summary>
+        /// Gets the name of an Object instance.
+        /// </summary>
+        /// <param name="obj">Object to get the name for</param>
+        /// <param name="useFullyQualifiedMethodNames">Boolean indicating whether to use fully qualified names or not</param>
+        /// <returns>Name of the object instance's type</returns>
+        string GetName(object obj, bool useFullyQualifiedMethodNames = false);
```
cgillum, line 27: It seems a little odd to include `useFullyQualifiedMethodNames` as part of this method signature since that implies all implementations will use reflection.

cgillum (commented):
Sorry for the delay in reviewing this. It completely fell off my radar! Overall this looks good to me. I would like to get some clarity on whether we can make `OrchestrationContext.OrchestrationContext` internal. That's the only potential blocker for me.

jviau:
Thanks for looking at this! I am reconsidering if this is the correct approach. I'll try to find some time to look at this. The need still exists - customizing how DTFx derives default names/versions when not explicitly provided, but I wonder if there is a better approach than this PR.

PR #497, opened 2021-01-08
Remove possible nullPtr exception in OnProcessWorkItemAsync

davidmrdavid:
This PR keeps `instanceState` consistent with `runtimeState` in case `instanceState` has not been initialized before use in `OnProcessWorkItemAsync`.

Previously, this caused a nullPtr exception that would prevent orchestrators from checkpointing their progress, therefore getting stuck.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -544,9 +544,9 @@ namespace DurableTask.Core
             runtimeState.Status = runtimeState.Status ?? carryOverStatus;
 
             // If we entered the if-statement above, `instanceState` might be null
-            // In that case, we keep instanceState consistent with the runtimeState
             if (instanceState == null)
             {
+                // In that case, we keep instanceState consistent with the runtimeState
                 instanceState = Utils.BuildOrchestrationState(runtimeState);
             }
 
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -544,12 +544,11 @@ namespace DurableTask.Core
             runtimeState.Status = runtimeState.Status ?? carryOverStatus;
 
             // If we entered the if-statement above, `instanceState` might be null
-            if (instanceState == null)
+            if (instanceState != null)
             {
-                instanceState = new OrchestrationState();
+                instanceState.Status = runtimeState.Status;
             }
 
-            instanceState.Status = runtimeState.Status;
 
             await this.orchestrationService.CompleteTaskOrchestrationWorkItemAsync(
                 workItem,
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -298,6 +298,8 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
 
             OrchestrationRuntimeState originalOrchestrationRuntimeState = runtimeState;
 
+            // This is dangerous because instanceState might not get assigned before use.
+            // We should explore providing a non-null default value here.
```
cgillum, line 302: I don't really like these comments since they come across as judgmental and don't really help the reader understand the behavior or intentions of the logic. I think the right way to tackle this potential improvement is to leverage the C# nullable types compiler feature. That should be sufficient to remove the danger and doesn't require us to make actual logic changes.

cgillum (approved):

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: That's a good point @cgillum, I didn't mean to come off as judgemental so I will update my comments :)

davidmrdavid: Now it reads as: `" OrchestrationState instanceState = null; // A good nullable types candidate"`
I'll eagerly merge this since it got the LGTM. Please let me know if you'd prefer I remove the comment altogether and I'll make a follow-up PR.
Thanks for the suggestion!

davidmrdavid: Actually, I'll just remove the comment altogether because it does not actual clarify the logic or its intentions.

davidmrdavid: I originally added that comment as a kind of "reminder" to revisit this block of code, but that can be addressed better by adding a comment on the nullable-types tracking ticket instead of in the code.

ConnorMcMahon (approved):
LGTM!

We can publish this on myget tomorrow.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -542,7 +542,13 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
             workItem.OrchestrationRuntimeState = originalOrchestrationRuntimeState;
 
             runtimeState.Status = runtimeState.Status ?? carryOverStatus;
-            instanceState.Status = runtimeState.Status;
+
+            // If we entered the if-statement above, `instanceState` might be null
```
ConnorMcMahon, line 546: Nit: this comment doesn't feel very helpful to me., as the fact we are checking for null implies that it is possible and safe for `instanceState` to be null.

ConnorMcMahon (approved):
LGTM minus the comment.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
davidmrdavid: That's true, the comment is a left-over from the original implementation. Will remove

PR #500, opened 2021-01-12
Retry operations on Storage SDK deadlocks instead of app recycle

ConnorMcMahon:
The Azure Storage backend will now attempt to retry operations when a
storage deadlock is detected instead of just recycling the app. In order
to do this properly, we had to make sure to cancel the in flight
operation just in case it actually would finish, to avoid performing
duplicate work. We also still need to recycle the app after multiple
deadlock hits to avoid resource leaks.

This PR also properly wraps several missed storage calls in our helper method
so that they have this deadlock mitigation as well.

### DIFF
--- a/src/DurableTask.AzureStorage/TimeoutHandler.cs
+++ b/src/DurableTask.AzureStorage/TimeoutHandler.cs
@@ -13,6 +13,7 @@
 
 namespace DurableTask.AzureStorage
 {
+    using Microsoft.WindowsAzure.Storage;
     using System;
     using System.Diagnostics;
     using System.Threading;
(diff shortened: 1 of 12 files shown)


cgillum (approved):
I like the clean refactoring you did in this PR. This looks great - just a few minor comments.

src/DurableTask.AzureStorage/TimeoutHandler.cs
```diff
@@ -22,44 +23,69 @@ namespace DurableTask.AzureStorage
...
 
-                    // Delay to ensure the ETW event gets written
-                    await Task.Delay(TimeSpan.FromSeconds(3));
-                    Environment.FailFast(message);
+                            cts.Cancel();
+                            continue;
+                        }
+                        else
+                        {
+                            string taskHubName = settings?.TaskHubName;
```
amdeel, line 71: Just a bit of minor code duplication.

amdeel (approved):
Looks great!

PR #422, opened 2020-07-16
Merge Correlation branch into master

cgillum:
The private preview was a success. Let's take the next step and get this merged into the main branch so it can be shipped with our nuget releases.

We'll want to do another pass on these changes to make sure we're not taking any changes that we don't want shipping.

FYI @TsuyoshiUshio

PR #501, opened 2021-01-14
Add System.Reactive.Compatibility as a dependency to DurableTask.Core 

ConnorMcMahon:
The Distributed Tracing work required adding System.Reactive.Core as a
dependency. However, if the customer app was using a different major
version of System.Reactive, it could lead to ambiguous compilation
issues.

Adding this package (and upgrading to use v4 of System.Reactive.Core)
should resolve this issue.

Addresses https://github.com/Azure/azure-functions-durable-extension/issues/1620

### DIFF
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -24,6 +24,8 @@
 
   <ItemGroup>
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.6.0" />
+    <PackageReference Include="System.Reactive.Core" Version="4.4.1" />
+    <PackageReference Include="System.Reactive.Compatibility" Version="4.4.1" />
   </ItemGroup>
   
 </Project>
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -23,7 +23,6 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


davidmrdavid (commented):
Looking at the CI failures, it would appear that the correlation tracing work depends on `System.Reactive.Core`, no?

ConnorMcMahon:
Oddly it is building locally for me, but not in the CI, so I'm trying to understand why.

cgillum:
Where is the dependency on System.Reactive coming from?

davidmrdavid:
I can confirm that `System.Reactive.Core` was added in the Dist.Tracing PR: https://github.com/Azure/durabletask/pull/422/files

ConnorMcMahon:
To clarify, it looks like `CorrelationTraceClient` uses some delegates that require types defined in System.Reactive (you can also get those types in System.Reactive.Core). It looks like we originally took on a dependency on System.Reactive.Core (v3) to address this.

The customer who encountered an issue was using v4 of System.Reactive, which leads to a compilation error. This is a well-known issue with using different major versions of System.Reactive.Core and System.Reactive.

Our options to make sure that customers don't run into this are:

1. Remove our dependency on System.Reactive.Core if possible.
2. Upgrade our System.Reactive.Core dependency to v4 (this would then break customers using v3 of System.Reactive)
3. Include a reference to the System.Reactive.Compatibility package, which should resolve this issue.

I think either option 1 or 3 are viable, but since I am running into build issues with option 1, I'll probably just go with option 3.

davidmrdavid (approved):
LGTM! Glad there already exists a compatibility layer :)

PR #503, opened 2021-01-20
[AzureStorage] Add optional JsonConverter for OrchestrationInstance

### LINKED ISSUE #491: DTFx.AzureStorage not honoring DataContract on OrchestrationInstance during serializing

jviau:
Adds on optional JsonConverter to support forward compatibility with `OrchestrationInstance` in AzureStorage provider. This is enabled via a setting. I opted for the setting to minimize any impact on anyone upgrading to this version.

Closes #491

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationInstanceConverter.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationInstanceConverter.cs
@@ -1,0 +1,90 @@
+﻿using DurableTask.Core;
+using Newtonsoft.Json;
+using System;
+using System.IO;
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Json;
+
+namespace DurableTask.AzureStorage
+{
+    /// <summary>
+    /// This class bridges <see cref="DataContractJsonSerializer"/> with Newtonsoft.Json. This serializer
+    /// is slower, but it handles writing to <see cref="IExtensibleDataObject"/>, which Newtonsoft does not.
+    /// A drawback of <see cref="DataContractJsonSerializer"/> is that ExtensionData Namespaces are not populated,
+    /// meaning reading via the regular <see cref="DataContractSerializer"/> will not correctly hydrate extra fields
+    /// from ExtensionData. However, it can still be done by using <see cref="DataContractJsonSerializer"/> instead.
+    /// </summary>
+    /// <remarks>
+    /// If other types implementing <see cref="IExtensibleDataObject"/> need to use this converter, then
+    /// uncomment the variant #2 of CanConvert.
+    /// </remarks>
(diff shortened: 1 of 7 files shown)


src/DurableTask.AzureStorage/OrchestrationInstanceConverter.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (objectType == null)
+            {
+                throw new ArgumentNullException(nameof(objectType));
+            }
+
+            // Variant #2 incase this needs to support any extensible data contract.
+            //return objectType.GetCustomAttribute<DataContractAttribute>() != null
+            //    && typeof(IExtensibleDataObject).IsAssignableFrom(objectType);
+
+            return typeof(OrchestrationInstance).IsAssignableFrom(objectType);
```
cgillum, line 34: As far as I can tell, nothing in the serialization or deserialization logic depends on the object being an instance of `OrchestrationInstance`. It seems to me you could use this for any data type. Can we rename this class to `DataContractJsonConverter` and remove this type constraint? That would also make the serializer better match the name of the property that enables it (`UseDataContractSerialization`).

cgillum (approved):
LGTM!

jviau:
Making the change to convert more than just the `OrchestrationInstance` is causing some problems. I'll look into either fixing, or scoping it back down to just that type.

jviau:
Resolved the issues:
1. DateTime format was causing issues. This now uses the DateTimeFormat from the json serializer. This is not perfect though - if they have a custom converter for date time, it may cause issues. But then again, this should _replace_ netwonsoft json serialization for these types and not interop with it.
2. `GenericEvent` was missing from `HistoryEvent.KnownTypes()` - refactored that method to use reflection to load up the initial list of types (then it is cached).

cgillum (approved):
Latest changes look great! Thanks for doing the extra validation and making the necessary fixes.

PR #504, opened 2021-01-22
Merge master into main

amdeel:


ConnorMcMahon:
Since this is just one commit, it might be easier just to cherry-pick it, so we don't end up with unnecessary merge commits.

PR #512, opened 2021-02-12
[DurableTaskCore] Fix the issue "Multiple ExecutionCompletedEvent found, potential corruption in state storage"

shankarsama:
Multiple ExecutionCompletedEvents possible when orchestration termination and completion happens at the same time.


aresowj (approved):
LGTM. So it's now try to complete the orchestration in all cases right?

lchew7 (approved):

shankarsama:
@aresowj Yes. It will not have conflict anymore if there multiple events.

PR #513, opened 2021-02-26
Reduce storage IO costs for partition management

ConnorMcMahon:
2a4fc4e
This PR attempts to address
Azure/azure-functions-durable-extension#1651.
In order to accomplish this, we now only download blob data for
ownership leases when we already own the intent lease.

This change did not play nicely with the old abstractions, which were
largely copied from the EventHubs partition manager. Now that we have
sufficiently deviated from this code, it did not seem worth updating the
abstraction, since we are unlikely to move away from blob leases.


ConnorMcMahon:
The failing tests appears to be an artifact of a slower TaskHubWorker startup now that we are using less aggressive polling on the ownership leases. The fix is to either use legacy partition management for tests, or increase wait time on the tests.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -4,9 +4,9 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.8.4</FileVersion>
+    <FileVersion>1.8.5</FileVersion>
```
davidmrdavid, line 7: Just checking this was intentional, since we had a couple mishaps with version upgrades last time.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -4,9 +4,9 @@
...
     <AssemblyVersion>$(FileVersion)</AssemblyVersion>
-    <Version>$(FileVersion)</Version>
+    <Version>$(FileVersion)-costreduction2</Version>
```
davidmrdavid, line 9: Also checking if this was intentional

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
```diff
@@ -13,34 +13,101 @@
...
+        /// <value>The epoch year of the lease.</value>
+        public long Epoch { get; set; }
+
+        /// <summary>Determines whether the lease is expired.</summary>
+        /// <returns>true if the lease is expired; otherwise, false.</returns>
+        public bool IsExpired => this.Blob.Properties.LeaseState != LeaseState.Leased;
+
+
+        [JsonIgnore]
+        internal CloudBlockBlob Blob { get; private set; }
```
davidmrdavid, line 81: Why is this marked with a `JsonIgnore`? Can we document it in the code?

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
```diff
@@ -13,34 +13,101 @@
...
+
+        public async Task DownloadLeaseAsync()
         {
-            this.Blob = source.Blob;
+            string serializedLease = await this.Blob.DownloadTextAsync();
+            this.stats.StorageRequests.Increment();
+
+            // Workaround: for some reason storage client reports incorrect blob properties after downloading the blob
```
davidmrdavid, line 88: nit: I would replace the word "workaround" with "compensation".
The word workaround makes me think that we're doing a hack, whereas compensation implies, to me, we're doing something which should be redundant. I think the latter is more fitting here.

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
```diff
@@ -13,34 +13,101 @@
...
+        }
 
-        public override bool IsExpired()
+        public override int GetHashCode()
         {
-            return this.Blob.Properties.LeaseState != LeaseState.Leased;
+            int hashCode = 1696262853;
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(LeaseType);
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(PartitionId);
+            return hashCode;
```
davidmrdavid, line 110: Can we document where this numbers come from? If they're arbitrary, it would be good to document it as well :)

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -21,20 +21,20 @@ namespace DurableTask.AzureStorage.Partitioning
     using System.Threading;
     using System.Threading.Tasks;
 
-    sealed class LeaseCollectionBalancer<T> where T : Lease
+    sealed class LeaseCollectionBalancer
```
davidmrdavid, line 24: I don't fully follow what this class does. What is this balancing _over_? Is it over partitions?

davidmrdavid (commented):
Thank you for this PR. Left some comments, mostly questions :)

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
ConnorMcMahon: These methods were autogenerated by VS2019, but I can comment to reflect this.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
ConnorMcMahon: It balances a set of partition leases across workers.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
ConnorMcMahon: Was from testing, will remove.

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
ConnorMcMahon: I just copied this code from a different method, which was originally copied from another project entirely. Not sure if this even still applies anymore honestly, but I was hesitant to change it.

davidmrdavid: Hmm, I'm not sure I follow.
I didn't mean to say that this workaround should be removed. I was just suggesting that, in the comment, I would prefer if we called it a "compensation". It just took me a minute to understand the original comment so I thought the word-change could help clarify things. No strong feelings, just clarifying what I meant.

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
```diff
@@ -13,34 +13,130 @@
...
 {
+    using DurableTask.AzureStorage.Monitoring;
+    using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Blob;
     using Newtonsoft.Json;
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Text;
+    using System.Threading.Tasks;
```
cgillum, line 24: Consistency nit: the convention in this codebase is to put System namespaces above non-System namespaces. There is a setting in Visual Studio you can set to makes this the default sort-order.

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -249,7 +249,10 @@ public async Task MultiWorkerLeaseMovement(bool useLegacyPartitionManagement, in
                 while (sw.Elapsed < timeout)
                 {
                     Trace.TraceInformation($"Checking current lease distribution across {currentWorkerCount} workers...");
-                    var leases = (await services[0].ListBlobLeasesAsync())
+                    var blobLeases = services[0].ListBlobLeases();
+                    await Task.WhenAll(blobLeases.Select(lease => lease.DownloadLeaseAsync()));
```
cgillum, line 253: It seems we're repeating these two lines in multiple places. Any reason not to compile then into a single call, like `ListBlobLeasesAsync()` that internally does the download? It looks like that's what the original implementation was doing (except synchronously, which isn't great).

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -509,15 +496,6 @@ async Task<bool> RenewLeaseAsync(T lease)
                 }
             }
 
-            this.settings.Logger.LeaseRenewalResult(
```
cgillum, line 512: Are we still writing these logs anywhere? If not, should we delete these methods (and the associated EventSource methods)?

cgillum (commented):
Can you list out which changes you made that are reducing blob costs, or perhaps point to them in the PR? There are a lot of changes to sort through and it's not completely obvious to me which changes are responsible for the cost reduction.

src/DurableTask.AzureStorage/Partitioning/BlobLease.cs
ConnorMcMahon: I actually dug into this a bit more, and it turns out that this is actually a "workaround" of bad SDK design. All blob downloads call the GetBlob Storage API, which returns the blob attributes/headers. It turns out some of the C# APIs discard this metadata, while others don't . I switched to an API that doesn't discard this to avoid unnecessary storage calls.

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
ConnorMcMahon: Yes, I've struggled with this a bit. The issue is less the synchronous download, and more so the fact that when we acquire leases, I want to be able to validate that it makes sense to download the lease before we actually download it (i.e. for ownership blobs, it doesn't make sense to even download the blob lease if we know we don't own the intent lease).

However, in most other cases, we want to actually download the blob leases when we read them like this. I'm guessing the answer is to make this an optional parameter on ListBlobLeases() about whether we want to download or not.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
ConnorMcMahon: I'm not opposed to leaving these logs in necessarily. Though I do think we need to start looking at our log volume, as we are  a rather chatty table, and as usage grows costs may grow prohibitive.

ConnorMcMahon: In the interest of limiting the scope of this PR, I am going to restore these logs, but I think we need to look into how much logs we are storing and eliminate the least helpful but high volume log statements.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -332,20 +333,11 @@ async Task LeaseTakerAsync()
...
-                {
-                    this.settings.Logger.PartitionManagerInfo(
-                        this.accountName,
-                        this.taskHub,
-                        this.workerName,
-                        string.Empty /* partitionId */,
-                        $"Skiping {this.leaseType} lease aquiring for {lease.PartitionId}");
-                    continue;
-                }
+            var acquirableLeases = (await this.leaseManager.ListLeasesAsync(@do: false)).Where(lease => this.shouldAcquireLeaseDelegate(lease.PartitionId));
```
cgillum, line 336: What is this `@do` syntax? I've never seen this before. Is it a shortcut way of specifying the `downloadLease` parameter?

src/DurableTask.AzureStorage/Partitioning/ILeaseManager.cs
```diff
@@ -22,12 +22,10 @@ interface ILeaseManager<T> where T : Lease
 
         Task<bool> CreateLeaseStoreIfNotExistsAsync(TaskHubInfo eventHubInfo, bool checkIfStale = true);
 
-        Task<IEnumerable<T>> ListLeasesAsync();
+        Task<IEnumerable<T>> ListLeasesAsync(bool downloadLeases);
 
         Task CreateLeaseIfNotExistAsync(string partitionId);
 
-        Task<T> GetLeaseAsync(string partitionId);
```
ConnorMcMahon, line 29: This was unused, so I just deleted it as opposed to updating it's implementation.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -332,20 +333,11 @@ async Task LeaseTakerAsync()
...
+            await Task.WhenAll(acquirableLeases.Select(blobLease => blobLease.DownloadLeaseAsync()));
```
cgillum, line 337: In the previous line I think you set `downloadLeases` to false, but you're then downloading them on this line?

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -529,6 +512,15 @@ async Task<bool> RenewLeaseAsync(T lease)
...
             }
 
+            this.settings.Logger.LeaseRenewalResult(
+                this.accountName,
+                this.taskHub,
+                this.workerName,
+                lease.PartitionId,
+                renewed,
+                lease.Token,
+                errorMessage);
```
cgillum: Are you intentionally changing the order between the `LeaseRenewalFailed` trace and the `LeaseRewalResult` trace? I don't think it actually matters, but was just curious...

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
ConnorMcMahon: I've addressed this with a mandatory boolean parameter to specify whether to download or not.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
ConnorMcMahon: Just fixed this after I noticed that.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
ConnorMcMahon: This is a typo. Fixing.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
ConnorMcMahon: We have already filtered the blob leases based on whether we should acquire them or not at this point. This makes sure that workers that don't own the intent leases won't download ownership leases.

davidmrdavid: I would prefer if we documented this, it's not entirely obvious to me :)

ConnorMcMahon: I literally just pushed this change :P

davidmrdavid: yeah I saw the change populate as I posted the comment haha

davidmrdavid (approved):
LGTM!

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -400,19 +359,21 @@ private bool IsStale(TaskHubInfo currentTaskHubInfo, TaskHubInfo newTaskHubInfo)
...
-            this.leaseDirectory = this.taskHubContainer.GetDirectoryReference(leaseDirectoryName);
-
-            string taskHubInfoBlobFileName = string.IsNullOrWhiteSpace(this.blobPrefix)
-                ? TaskHubInfoBlobName
-                : this.blobPrefix + TaskHubInfoBlobName;
-
-            this.taskHubInfoBlob = this.taskHubContainer.GetBlockBlobReference(taskHubInfoBlobFileName);
+            this.leaseDirectory = this.taskHubContainer.GetDirectoryReference(this.leaseType);
+            this.taskHubInfoBlob = this.taskHubContainer.GetBlockBlobReference(TaskHubInfoBlobName);
+            this.blobLeases = new List<BlobLease>();
```
cgillum, line 365:
```suggestion
            this.blobLeases = new List<BlobLease>(this.settings.PartitionCount);
```

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -332,20 +333,14 @@ async Task LeaseTakerAsync()
...
-                        $"Skiping {this.leaseType} lease aquiring for {lease.PartitionId}");
-                    continue;
-                }
+            // We wait to download leases until after we have filtered out leases we should
+            // not acquire
+            var allLeases = await this.leaseManager.ListLeasesAsync(downloadLeases: false);
+            var acquirableLeases = allLeases.Where(lease => this.shouldAcquireLeaseDelegate(lease.PartitionId));
+            await Task.WhenAll(acquirableLeases.Select(blobLease => blobLease.DownloadLeaseAsync()));
 
+            foreach (T lease in acquirableLeases)
```
cgillum, line 342: `acquirableLeases` is an `IEnumerable<T>` and it looks like running it in this `foreach` will cause us to  duplicate the filtering work. Should we do a `.ToList()` at the end of line 339 to avoid redoing work?

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -415,7 +415,7 @@ public async Task PartitionLost_AbandonPrefetchedSession()
                 LeaseRenewInterval = TimeSpan.FromMilliseconds(500),
                 TaskHubName = TestHelpers.GetTestTaskHubName(),
                 StorageConnectionString = TestHelpers.GetTestStorageAccountConnectionString(),
-                ControlQueueBufferThreshold = 100,
+                ControlQueueBufferThreshold = 100
```
cgillum, line 418: nit: I noticed that you've been removing trailing commas. Most linters actually suggest that you keep them because it makes for cleaner diffs if someone later adds to the end of a list.

cgillum (approved):
Just a few small things.

Issue #487, opened 2020-12-14
System.NullReferenceException in AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync


thirus:
Version: Core 2.4.1, AzureStorate: 1.8.1

For our use case we use sub orchestrations and external events, however we observed random NPE and our main orchestration instance get stuck since the sub orchestration instance never gets created.

Exception:
```
DurableTask.Core[29]
      TaskOrchestrationDispatcher-1bc6efed64814fb897c9e58d6f22c01b-0: Unhandled exception with work item 'fb874f1e8645614ea67e2330b4c60fcd': System.NullReferenceException: Object reference not set to an instance of an object.
         at DurableTask.AzureStorage.AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync(TaskOrchestrationWorkItem workItem, OrchestrationRuntimeState newOrchestrationRuntimeState, IList`1 outboundMessages, IList`1 orchestratorMessages, IList`1 timerMessages, TaskMessage continuedAsNewMessage, OrchestrationState orchestrationState) in C:\source\durabletask\src\DurableTask.AzureStorage\AzureStorageOrchestrationService.cs:line 1005
         at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in C:\source\durabletask\src\DurableTask.Core\TaskOrchestrationDispatcher.cs:line 463
         at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem) in C:\source\durabletask\src\DurableTask.Core\TaskOrchestrationDispatcher.cs:line 194
         at DurableTask.Core.WorkItemDispatcher`1.ProcessWorkItemAsync(WorkItemDispatcherContext context, Object workItemObj) in C:\source\durabletask\src\DurableTask.Core\WorkItemDispatcher.cs:line 459
      
      Backing off for 10 seconds until 5 successful operations
```

While debugging locally we found this line gets `runtimeState.OrchestrationInstance` as null.
https://github.com/Azure/durabletask/blob/bcc97e874baba09c722a83c164bb3ab2245017ec/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L948

below logic in the same method assumes the `runtimeState.OrchestrationInstance` can be null but not that line.
example line that expects nullable instance
[...]

PR #517, opened 2021-03-06
Fix null reference exception in DurableTask.AzureStorage

ConnorMcMahon:
A null value for OrchestrationInstance on the runtimeState object in
AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync
causes rare null reference exceptions for some customers. Unfortunately,
due to the lack of helpful information when this exception is
encountered, we don't know what scenarios cause the orchestration
instance to be null.

This PR currently just adds null coalescing + a warning log. It's
possible that a null execution id that this results in will cause issues
down the line in this method, but this will at least make sure we know
which orchestration instance hit this code path, which makes it far
easier to diagnose the issue.

Attempts to address #487

### DIFF
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -910,9 +910,9 @@ namespace DurableTask.AzureStorage
         }
 
         [Event(EventIds.GeneralWarning, Level = EventLevel.Warning, Version = 2)]
-        public void GeneralWarning(string Account, string TaskHub, string Details, string AppName, string ExtensionVersion)
+        public void GeneralWarning(string Account, string TaskHub, string Details, string AppName, string ExtensionVersion, string InstanceId)
         {
-            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion);
+            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion, InstanceId);
         }
 
         [Event(EventIds.SplitBrainDetected, Level = EventLevel.Warning, Version = 2)]
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -912,7 +912,7 @@ namespace DurableTask.AzureStorage
         [Event(EventIds.GeneralWarning, Level = EventLevel.Warning, Version = 2)]
         public void GeneralWarning(string Account, string TaskHub, string Details, string AppName, string ExtensionVersion, string InstanceId)
         {
-            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion, InstanceId);
+            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion, InstanceId ?? string.Empty);
         }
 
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1063,10 +1063,16 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
...
 
             string instanceId = workItem.InstanceId;
-            string executionId = runtimeState.OrchestrationInstance.ExecutionId;
+            string executionId = runtimeState.OrchestrationInstance?.ExecutionId;
+            if (executionId == null)
+            {
+                this.settings.Logger.GeneralWarning(
+                    this.storageAccountName,
+                    this.settings.TaskHubName,
+                    $"{nameof(CompleteTaskOrchestrationWorkItemAsync)}: Could not find execution id for instance {instanceId}.");
```
cgillum, line 1072:
```suggestion
                    $"{nameof(CompleteTaskOrchestrationWorkItemAsync)}: Could not find execution id in runtime state for instance '{instanceId}'.");
```

cgillum: Also, is there any more specific warning we can use that includes an InstanceId column? Typically when we're debugging an issue we filter by InstanceId and general warnings often get missed unless we look for warnings across all instances.

ConnorMcMahon: We've got splitbrain + discarding message for warnings that include instance id as a column. In general, we don't have great general info/warning/error level logs for orchestration instances.

I could add a new dedicated message for this, but that felt heavyweight.

cgillum: Another option might be to add an InstanceId column to GeneralWarning and have it default to null/empty string?

ConnorMcMahon: I always get a little scared adding new fields to ETW, but in general it's safe as long as we add them to the end or change the Version field on the attribute, right?

cgillum: I don't think you even need to add it to the end, but yeah I generally haven't had problems with it.

davidmrdavid:
Thanks for this! Linking https://github.com/Azure/azure-functions-durable-extension/issues/1635 to increase hit count / to help motivate it once we get to it :)

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -910,9 +910,9 @@ public void GeneralError(string Account, string TaskHub, string Details, string
         }
 
         [Event(EventIds.GeneralWarning, Level = EventLevel.Warning, Version = 2)]
-        public void GeneralWarning(string Account, string TaskHub, string Details, string AppName, string ExtensionVersion)
+        public void GeneralWarning(string Account, string TaskHub, string Details, string AppName, string ExtensionVersion, string InstanceId)
         {
-            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion);
+            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion, InstanceId);
```
cgillum, line 915: Maybe just for defense-in-depth:
```suggestion
            this.WriteEvent(EventIds.GeneralWarning, Account, TaskHub, Details, AppName, ExtensionVersion, InstanceId ?? string.Empty);
```

cgillum (approved):

PR #519, opened 2021-03-15
Service Fabric provider: Overload for orchestration registration delegate, adding a TaskHubClient for querying state

NeetArt:
## Problem
When utilizing Service Fabric provider in durable task, during the initialization phase of `TaskHubWorker` artifacts, there is no available API to access the corresponding `TaskHubClient`. The only available option is to create a `TaskHubClient` with the `RemoteOrchestrationServiceClient` which routes the request via HTTP. This is not ideal due to the fact that we will be routing requests to the outside of the process back into the same process again.

## Proposal
There is a version of orchestration client named `FabricOrchestrationServiceClient` which is capable of querying the orchestration state within the process itself by querying the underlying reliable collections. However, this client is not publicly exposed. The proposed solution creates a new `TaskHubClient` based on the `FabricOrchestrationServiceClient` and passes it on to the clients initializing the `TaskHubWorker` as an overload. The clients may choose to use this overload to query the orchestration state.

### DIFF
--- a/src/DurableTask.AzureServiceFabric/Service/TaskHubProxyListener.cs
+++ b/src/DurableTask.AzureServiceFabric/Service/TaskHubProxyListener.cs
@@ -47,7 +47,7 @@ namespace DurableTask.AzureServiceFabric.Service
     /// </remarks>
     /// <param name="taskHubWorker">Instance of <see cref="TaskHubWorker"/></param>
     /// <param name="taskHubLocalClient">Instance of <see cref="TaskHubClient"/> connected to local instance of <see cref="TaskHubWorker"/></param>
-    public delegate void RegisterOrchestrationArtifacts(TaskHubWorker taskHubWorker, TaskHubClient taskHubLocalClient);
+    public delegate void RegisterOrchestrations2(TaskHubWorker taskHubWorker, TaskHubClient taskHubLocalClient);
 
     /// <summary>
     /// An instance of this class is created for each service replica by the Service Fabric runtime.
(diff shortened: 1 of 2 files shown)


cgillum:
@shankarsama would you mind taking a look at this PR?

src/DurableTask.AzureServiceFabric/Service/TaskHubProxyListener.cs
```diff
@@ -35,17 +35,33 @@ namespace DurableTask.AzureServiceFabric.Service
...
+    /// when registering delegates
+    /// </summary>
+    /// <remarks>
+    /// The provided <see cref="TaskHubClient"/> is only connected to the local instance of
+    /// <see cref="TaskHubWorker"/>. It cannot be used to query orchestrations running in other
+    /// worker instances.
+    /// </remarks>
+    /// <param name="taskHubWorker">Instance of <see cref="TaskHubWorker"/></param>
+    /// <param name="taskHubLocalClient">Instance of <see cref="TaskHubClient"/> connected to local instance of <see cref="TaskHubWorker"/></param>
+    public delegate void RegisterOrchestrationArtifacts(TaskHubWorker taskHubWorker, TaskHubClient taskHubLocalClient);
```
shankarsama, line 50: RegisterOrchestrations2

NeetArt: Addressed

shankarsama (approved):

PR #526, opened 2021-03-19
Service Fabric provider: Identify stale partition endpoints and force refresh SF resolver cache in such a scenario

NeetArt:
## Problem
Service Fabric's `ServicePartitionResolver` class maintains an internal cache of `ResolvedServicePartition` objects for each partition which is served by an internal reference of `FabricClient`. The cache is refreshed via 2 different processes:

-  A notification protocol through which the `FabricClient` notifies resolver that a given value is stale and needs to be updated.
-  A complaint based resolution, as part of which, the client passes in the previously retrieved `ResolvedServicePartition` object to the `ResolveAsync()` API call. This API call notifies the resolver that the previously provided object is stale and a new one should be fetched from the `FabricClient`

Service Fabric also states in their documentation that the notification mechanism cannot be completely relied upon since it is an internal network call which is prone to failures. This is the one of the reason for the existence of complaint based resolution. In our current implementation, we were only relying on notifications and thus occasionally ran into failures due to non-existent/badendpoints.

## Proposal
We will implement a new API in `IPartitionEndpointResolver` `RefreshPartitionEndpointAsync()` which will retrieve the existing `ResolvedServicePartition` object via `ResolveAsync()` and then call another overload of `ResolveAsync()` with the retrieved object. This will force the internal cache to be refreshed so that subsequent calls to `ResolveAsync()` would retrieve the newer version, if available.
This API will be invoked by `RemoteOrchestrationServiceClient` whenever it detects a stale endpoint.

#### Now, how do we detect stale endpoints?
The ideal situation would be if our endpoint would simply fail to respond which would clearly indicate a non-existent endpoint. However, this situation is complicated by the default http.sys behavior. If the previous service process failed to clean up the urlacl entry from http.sys, let's say due to a process crash, http.sys will still respond to requests for that endpoint. The response will be a HTTP 503.
Another situation which is possible, is that Service Fabric decides to reuse that port for another replica. Since we suffix a partition endpoint with the partition id, http.sys is going to return a HTTP 404 instead.
In order to distinguish these error codes from genuine service error codes, I'm adding a header which will be part of a response from the service. Presence of such a header will help us distinguish between http.sys response and our service response. To make the header slightly more useful, I'm adding a ActivityId header which will be unique for each request/response combination. We could potentially use this header in the future to trace this request within the service.

### DIFF
--- a/src/DurableTask.AzureServiceFabric/Remote/FabricPartitionEndpointResolver.cs
+++ b/src/DurableTask.AzureServiceFabric/Remote/FabricPartitionEndpointResolver.cs
@@ -68,6 +68,20 @@ namespace DurableTask.AzureServiceFabric.Remote
             return endpoints;
         }
 
+        /// <inheritdoc/>
+        public async Task RefreshPartitionEndpointAsync(string instanceId, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            long hash = await this.instanceIdHasher.GeneratePartitionHashCodeAsync(instanceId, cancellationToken);
+
+            // Retrieve the current ResolvedServicePartition object from the cache and then pass in the same object
+            // to the ResolveAsync() API call. This operation notifies the ServicePartitionResolver that the current
+            // object is most likely stale and that it needs to refresh its own cache.
+            // Further calls to ResolveAsync() would return the newer version if the cache was refreshed.
+            var previousRsp = await this.partitionResolver.ResolveAsync(this.serviceUri, new ServicePartitionKey(hash), cancellationToken);
+            await this.partitionResolver.ResolveAsync(previousRsp, cancellationToken);
+        }
+
         private async Task<Int64RangePartitionInformation[]> GetServicePartitionsListAsync()
         {
             if (this.servicePartitionsTaskCompletionSource == null)
--- a/src/DurableTask.AzureServiceFabric/Remote/FabricPartitionEndpointResolver.cs
+++ b/src/DurableTask.AzureServiceFabric/Remote/FabricPartitionEndpointResolver.cs
@@ -74,10 +74,22 @@ namespace DurableTask.AzureServiceFabric.Remote
             cancellationToken.ThrowIfCancellationRequested();
             long hash = await this.instanceIdHasher.GeneratePartitionHashCodeAsync(instanceId, cancellationToken);
 
-            // Retrieve the current ResolvedServicePartition object from the cache and then pass in the same object
-            // to the ResolveAsync() API call. This operation notifies the ServicePartitionResolver that the current
-            // object is most likely stale and that it needs to refresh its own cache.
-            // Further calls to ResolveAsync() would return the newer version if the cache was refreshed.
+            /*
+             * Service Fabric's ServicePartitionResolver class maintains an internal cache of ResolvedServicePartition objects for each partition
+             * which is served by an internal reference of FabricClient. The cache is refreshed via 2 different processes:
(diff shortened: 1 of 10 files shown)


src/DurableTask.AzureServiceFabric/Service/ProxyServiceExceptionHandler.cs
```diff
@@ -18,16 +18,10 @@ namespace DurableTask.AzureServiceFabric.Service
     using System.Web.Http.ExceptionHandling;
     using System.Web.Http.Results;
 
-    using DurableTask.AzureServiceFabric.Tracing;
-
     internal class ProxyServiceExceptionHandler : ExceptionHandler
     {
         public override void Handle(ExceptionHandlerContext context)
         {
-            ServiceFabricProviderEventSource.Tracing.LogProxyServiceError(context.Request.Method.ToString(),
```
NeetArt, line 27: `ActivityLoggingMessageHandler` already logs the exception. There is no need to log it here again.

src/DurableTask.AzureServiceFabric/Remote/FabricPartitionEndpointResolver.cs
```diff
@@ -68,6 +68,20 @@ public async Task<IEnumerable<string>> GetPartitionEndpointsAsync(CancellationTo
...
+        /// <inheritdoc/>
+        public async Task RefreshPartitionEndpointAsync(string instanceId, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            long hash = await this.instanceIdHasher.GeneratePartitionHashCodeAsync(instanceId, cancellationToken);
+
+            // Retrieve the current ResolvedServicePartition object from the cache and then pass in the same object
+            // to the ResolveAsync() API call. This operation notifies the ServicePartitionResolver that the current
+            // object is most likely stale and that it needs to refresh its own cache.
+            // Further calls to ResolveAsync() would return the newer version if the cache was refreshed.
```
shankarsama, line 80: Elaborate that corner case where notification may be lost and partition never refreshes.

src/DurableTask.AzureServiceFabric/Service/ProxyServiceExceptionHandler.cs
```diff
@@ -18,16 +18,10 @@ namespace DurableTask.AzureServiceFabric.Service
...
-                                                                          context.Request.RequestUri.AbsolutePath,
-                                                                          context.Exception);
```
shankarsama, line 29: Keep this logging as is. MessageHandlder is not invoked when there is an exception from the controller.

shankarsama (approved):
:shipit:

src/DurableTask.AzureServiceFabric/Service/ProxyServiceExceptionHandler.cs
NeetArt: It'll be a duplicate trace. `HttpControllerDispatcher` first calls into `IExceptionLogger` and then calls `IExceptionHandler` in case the controller throws an exception. Refer here [HttpControllerDispatcher#L164](https://github.com/aspnet/AspNetWebStack/blob/42991b3d2537b702736463f76a10a4fcf2ea44c9/src/System.Web.Http/Dispatcher/HttpControllerDispatcher.cs#L164)
```
            await ExceptionLogger.LogAsync(exceptionContext, cancellationToken);
            HttpResponseMessage response = await ExceptionHandler.HandleAsync(exceptionContext, cancellationToken);
```

Issue #527, opened 2021-03-29
DurableTask.AzureStorage: OrchestrationState.CompletedTime is never set


cgillum:
I'm not sure how this went undetected for so long, but this field is never getting set and always has it's default value.

bbenameur:
Hello, Any fix for OrchestrationState.CompletedTime please?

PR #528, opened 2021-03-30
DT.AzureStorage: Runtime de-dupe support for orchestration start events

cgillum:
The DT.AzureStorage backend has always been vulnerable to concurrent starts of orchestrations with the same instance ID. This PR finally fixes that, as well as a couple other issues found along the way.

Fixes https://github.com/Azure/azure-functions-durable-extension/issues/1730 (this was the main issue being fixed)
Fixes https://github.com/Azure/durabletask/issues/527 (found this while working on ^)

## Problem
Often time users want to start an orchestration with a specific instance ID if it's not already started. To do so, they'd write code that looks like this:

```
var existing = GetOrchestrationInstance(id)
if (existing is null or existing is stopped)
    StartNewOrchestration(id)
```

The problem with this logic is that it's not thread-safe. If two threads call this logic at the same time, on the same machine or on different machines, it will create to orchestration start messages for the same instance. DT.AzureStorage did not have ideal handling for this. One instance may execute normally while another executes partially, leading to inconsistencies or duplicate executions.

In the past, we'd asked customers to implement their own concurrency management. With this PR, we should be able to automatically de-duplicate these start messages, allowing users to not worry about writing distributed locks to mitigate this race condition.

## Solution

For every `ExecutionStarted` event we dequeue from a control queue, we query the _Instances_ table to see whether the `ExecutionId` field of the start message matches that same property of the corresponding _Instances_ table. If they match, then we process the start message normally. Otherwise, we discard the message as a duplicate. We can do this because updates to the _Instances_ table are thread-safe, so the `ExecutionId` field is an easy way to uniquely associate an instance to an `ExecutionStarted` message.

The actual implementation is a bit more complex because we actually enqueue the `ExecutionStarted` message in the control queue before updating the _Instances_ table. For example, by the time we read the `ExecutionStarted` message, the corresponding _Instances_ table record may not yet exist, in which case we put the message back onto the queue so we can try processing it again later. We do this up to 10 times.

The other corner case we handle is where a _completed_ instance is being intentionally overwritten, which is something we actively support. In that case, an old `ExecutionId` values will exist temporarily until the new one can be written. To account for this, we also compare the timestamp of the message with the timestamp of the _Instances.CreatedTime_ record to distinguish this case from the duplicate start case.

### Pros
This solution is robust without introducing too much complexity.

### Cons
There is now one more Azure Table I/O request for every orchestration start. This means more storage I/O and more end-to-end latency for a single orchestration execution. I tried to avoid this but couldn't come up with anything that was sufficiently robust against all kinds of corner cases. Because the new I/O is outside of the main execution path, it shouldn't have a big impact on max throughput.

As an optimization, we will try to batch multiple Azure Table I/O requests into one to minimize overhead if there are lots of concurrent orchestration starts.

## Other alternative designs considered
I initially tried to solve this by simply de-duplicating `ExecutionStarted` messages in-memory when we dequeued them. However, this generally assumes that we dequeue all these messages in the same batch, which is not always the case. For example, if there are 40 duplicate start messages, then they would arrive it at least two different dequeue batches.

There are other potential solutions as well, such as keeping a list of all execution start messages over a period of time. However, doing this in memory can be fragile because 1) leases may move in which case some messages will show up on one VM and some on another, and 2) the list of instance IDs also needs to be kept in sync with orchestration status updates - for example, we need to know if an instance ID in our list has completed so we can remove it from the list. This becomes complicated quickly, so I did not pursue it.

## Testing

To test this, I create 40 new instances concurrently. I chose 40 to be some number larger than 32, which is out dequeue batch size (other designs I had in mind didn't work for more than one dequeue batch). The test verifies that exactly one instance runs without any duplicate activity executions.

To ensure I didn't regress anything, and to exercise the optimization I mentioned earlier, I also added a version of the test that creates 40 _unique_ orchestrations concurrently.

**NOTE**: There is a lot of new test code that I borrowed from the DT.SqlServer project. It allows us to define orchestrations and activities inline as C# lambdas without writing any classes, which is very convenient. My hope is that we can use this as the basis of a reusable test library for custom DTFx backend implementations to use.

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/StressTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/StressTests.cs
@@ -31,9 +31,9 @@ namespace DurableTask.AzureStorage.Tests
         /// both in the case where they all share the same instance ID and when they have unique
         /// instance IDs.
         /// </summary>
-        [DataTestMethod]
-        [DataRow(true)]
-        [DataRow(false)]
+        ////[DataTestMethod]
+        ////[DataRow(true)]
+        ////[DataRow(false)]
         public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
         {
             // Set the minimum thread count to 64+ to make this test extra concurrent.
@@ -65,8 +65,7 @@ namespace DurableTask.AzureStorage.Tests
                     Func<int, string> inputGenerator;
                     if (useSameInstanceId)
                     {
-                        DateTime now = DateTime.Now;
-                        instanceIdGenerator = _ => $"ConcurrentInstance_{now:yyyyMMdd.hhmmss.fff}";
+                        instanceIdGenerator = _ => $"ConcurrentInstance_SINGLETON";
                         inputGenerator = _ => "World";
                     }
                     else
--- a/Test/DurableTask.AzureStorage.Tests/StressTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/StressTests.cs
@@ -1,0 +1,120 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage.Tests
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using DurableTask.Core;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Newtonsoft.Json;
+
+    [TestClass]
+    public class StressTests
+    {
+        /// <summary>
+        /// End-to-end test which validates a simple orchestrator function which doesn't call any activity functions.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
+        {
+            // Set the minimum thread count to 64+ to make this test extra concurrent.
+            ThreadPool.GetMinThreads(out int minWorkerThreads, out int minIoThreads);
+            ThreadPool.SetMinThreads(Math.Max(64, minWorkerThreads), Math.Max(64, minIoThreads));
+            try
+            {
+                using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(
+                    enableExtendedSessions: false,
+                    modifySettingsAction: settings => settings.ThrowExceptionOnInvalidDedupeStatus = false))
+                {
+                    await host.StartAsync();
+
+                    var results = new ConcurrentBag<string>();
+
+                    // We want a number sufficiently high that it results in multiple message batches
+                    const int MaxConcurrency = 40;
+
+                    TaskActivity activity = TestOrchestrationHost.MakeActivity(
+                        delegate(TaskContext ctx, string input)
+                        {
+                            string result = $"Hello, {input}!";
+                            results.Add(result);
+                            return result;
+                        });
+
+                    // Use the same instance name for all instances
+                    Func<int, string> instanceIdGenerator;
+                    Func<int, string> inputGenerator;
+                    if (useSameInstanceId)
+                    {
+                        DateTime now = DateTime.Now;
+                        instanceIdGenerator = _ => $"ConcurrentInstance_{now:yyyyMMdd.hhmmss.fff}";
+                        inputGenerator = _ => "World";
+                    }
+                    else
+                    {
+                        instanceIdGenerator = i => $"ConcurrentInstance_{i:00}";
+                        inputGenerator = i => $"{i:00}";
+                    }
+
+                    List<TestInstance<string>> instances = await host.StartInlineOrchestrations(
+                        MaxConcurrency,
+                        instanceIdGenerator,
+                        inputGenerator,
+                        orchestrationName: "SayHelloOrchestration",
+                        version: string.Empty,
+                        implementation: (ctx, input) => ctx.ScheduleTask<string>("SayHello", "", input),
+                        activities: ("SayHello", activity));
+
+                    Assert.AreEqual(MaxConcurrency, instances.Count);
+
+                    // All returned objects point to the same orchestration instance
+                    OrchestrationState[] finalStates = await Task.WhenAll(instances.Select(
+                        i => i.WaitForCompletion(timeout: TimeSpan.FromMinutes(2), expectedOutputRegex: @"Hello, \w+!")));
+
+                    if (useSameInstanceId)
+                    {
+                        // Make sure each instance is exactly the same
+                        string firstInstanceJson = JsonConvert.SerializeObject(finalStates[0]);
+                        foreach (OrchestrationState state in finalStates.Skip(1))
+                        {
+                            string json = JsonConvert.SerializeObject(state);
+                            Assert.AreEqual(firstInstanceJson, json, "Expected that all instances have the same data.");
+                        }
+                    }
+                    else
+                    {
+                        // Make sure each instance is different
+                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.OrchestrationInstance.InstanceId).Distinct().Count());
+                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.OrchestrationInstance.ExecutionId).Distinct().Count());
+                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.Input).Distinct().Count());
+                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.Output).Distinct().Count());
+                    }
+
+                    await host.StopAsync();
+                }
+            }
+            finally
+            {
+                // Reset the thread pool configuration
+                ThreadPool.SetMinThreads(minWorkerThreads, minIoThreads);
+            }
+        }
+    }
+}
--- a/Test/DurableTask.AzureStorage.Tests/StressTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/StressTests.cs
@@ -27,7 +27,9 @@ namespace DurableTask.AzureStorage.Tests
     public class StressTests
     {
         /// <summary>
-        /// End-to-end test which validates a simple orchestrator function which doesn't call any activity functions.
+        /// Starts a large'ish number of orchestrations concurrently and verifies correct behavior
+        /// both in the case where they all share the same instance ID and when they have unique
+        /// instance IDs.
         /// </summary>
         [DataTestMethod]
         [DataRow(true)]
--- a/Test/DurableTask.AzureStorage.Tests/StressTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/StressTests.cs
@@ -31,9 +31,9 @@ namespace DurableTask.AzureStorage.Tests
         /// both in the case where they all share the same instance ID and when they have unique
         /// instance IDs.
         /// </summary>
-        ////[DataTestMethod]
-        ////[DataRow(true)]
-        ////[DataRow(false)]
+        [DataTestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
         public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
         {
             // Set the minimum thread count to 64+ to make this test extra concurrent.
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -15,7 +15,7 @@ namespace DurableTask.AzureStorage
 {
     using System;
     using System.Diagnostics.Tracing;
-    using System.Threading;
+    using System.Runtime.CompilerServices;
     using DurableTask.AzureStorage.Logging;
 
     /// <summary>
@@ -27,8 +27,6 @@ namespace DurableTask.AzureStorage
     [EventSource(Name = "DurableTask-AzureStorage")]
     class AnalyticsEventSource : EventSource
     {
-        static readonly AsyncLocal<Guid> ActivityIdState = new AsyncLocal<Guid>();
-
         /// <summary>
         /// Singleton instance used for writing events.
         /// </summary>
@@ -37,9 +35,23 @@ namespace DurableTask.AzureStorage
         [NonEvent]
         public static void SetLogicalTraceActivityId(Guid activityId)
         {
-            // We use AsyncLocal to preserve activity IDs across async/await boundaries.
-            ActivityIdState.Value = activityId;
-            SetCurrentThreadActivityId(activityId);
+            try
+            {
+                SetCoreTraceActivityId(activityId);
+            }
+            catch (MissingMethodException)
+            {
+                // Best effort. This method is only available starting in DurableTask.Core v2.5.3.
+                // We catch to maintain backwards compatibility with previous versions.
+            }
+        }
+
+        // NoInlining ensures we can get a predictable MissingMethodException if the build of DurableTask.Core
+        // we're using doesn't define the LoggingExtensions.SetLogicalTraceActivityId() method.
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void SetCoreTraceActivityId(Guid activityId)
+        {
+            DurableTask.Core.Logging.LoggingExtensions.SetLogicalTraceActivityId(activityId);
         }
 
         [Event(EventIds.SendingMessage, Level = EventLevel.Informational, Opcode = EventOpcode.Send, Task = Tasks.Enqueue, Version = 6)]
--- a/src/DurableTask.AzureStorage/AsyncAutoResetEvent.cs
+++ b/src/DurableTask.AzureStorage/AsyncAutoResetEvent.cs
@@ -24,12 +24,14 @@ namespace DurableTask.AzureStorage
         readonly LinkedList<TaskCompletionSource<bool>> waiters = 
             new LinkedList<TaskCompletionSource<bool>>();
 
-        bool isSignaled;
-
         public AsyncAutoResetEvent(bool signaled)
         {
-            this.isSignaled = signaled;
+            this.IsSignaled = signaled;
         }
+
+        public bool IsSignaled { get; private set; }
+
+        public int QueueLength => this.waiters.Count;
 
         public Task<bool> WaitAsync(TimeSpan timeout)
         {
@@ -42,18 +44,17 @@ namespace DurableTask.AzureStorage
 
             lock (this.waiters)
             {
-                if (this.isSignaled)
+                if (this.IsSignaled)
                 {
-                    this.isSignaled = false;
+                    this.IsSignaled = false;
                     return true;
                 }
                 else if (timeout == TimeSpan.Zero)
                 {
-                    return this.isSignaled;
+                    return false;
                 }
                 else
                 {
-                    // If we ever upgrade to .NET 4.6, we should use TaskCreationOptions.RunContinuationsAsynchronously
                     tcs = new TaskCompletionSource<bool>();
                     this.waiters.AddLast(tcs);
                 }
@@ -83,30 +84,30 @@ namespace DurableTask.AzureStorage
             }
         }
 
-        public void Set()
+        public bool Set()
         {
             lock (this.waiters)
             {
                 if (this.waiters.Count > 0)
                 {
-                    // Signal the first task in the waiters list. This must be done on a new
-                    // thread to avoid stack-dives and situations where we try to complete the
-                    // same result multiple times.
+                    // Signal the first task in the waiters list.
                     TaskCompletionSource<bool> tcs = this.waiters.First.Value;
                     Task.Run(() => tcs.SetResult(true));
                     this.waiters.RemoveFirst();
                 }
-                else if (!this.isSignaled)
+                else if (!this.IsSignaled)
                 {
                     // No tasks are pending
-                    this.isSignaled = true;
+                    this.IsSignaled = true;
                 }
+
+                return !this.IsSignaled;
             }
         }
 
         public override string ToString()
         {
-            return $"Signaled: {this.isSignaled.ToString()}, Waiters: {this.waiters.Count.ToString()}";
+            return $"Signaled: {this.IsSignaled}, Waiters: {this.waiters.Count}";
         }
     }
 }
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.8.5</FileVersion>
+    <FileVersion>1.8.6</FileVersion>
     <AssemblyVersion>$(FileVersion)</AssemblyVersion>
     <Version>$(FileVersion)</Version>
     <IncludeSymbols>true</IncludeSymbols>
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -3,6 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <LangVersion>8.0</LangVersion>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
     <FileVersion>1.8.6</FileVersion>
     <AssemblyVersion>$(FileVersion)</AssemblyVersion>
--- a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
 namespace DurableTask.AzureStorage.Messaging
 {
     using System;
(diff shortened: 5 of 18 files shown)


src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -37,9 +35,23 @@ class AnalyticsEventSource : EventSource
...
+                // We catch to maintain backwards compatibility with previous versions.
+            }
+        }
+
+        // NoInlining ensures we can get a predictable MissingMethodException if the build of DurableTask.Core
+        // we're using doesn't define the LoggingExtensions.SetLogicalTraceActivityId() method.
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void SetCoreTraceActivityId(Guid activityId)
+        {
+            DurableTask.Core.Logging.LoggingExtensions.SetLogicalTraceActivityId(activityId);
```
cgillum, line 54: The change in this file fixes a regression I caused in the v2.3.0 release. During testing, I observed that we were no longer tracing the `ActivityId` ETW field correctly for `ReceivedMessage` and `FetchInstanceHistory` traces, which made it harder for me to figure out why some of my solutions were not working.

src/DurableTask.AzureStorage/AsyncAutoResetEvent.cs
```diff
@@ -24,13 +24,15 @@ class AsyncAutoResetEvent
...
-
         public AsyncAutoResetEvent(bool signaled)
         {
-            this.isSignaled = signaled;
+            this.IsSignaled = signaled;
         }
 
+        public bool IsSignaled { get; private set; }
+
+        public int QueueLength => this.waiters.Count;
```
cgillum, line 34: There are no behavioral changes in this file. I simply exposed more information as public properties to make setting conditional breakpoints in the debugger easier.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -3,8 +3,9 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <LangVersion>8.0</LangVersion>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.8.5</FileVersion>
+    <FileVersion>1.8.6</FileVersion>
```
cgillum, line 8: FYI, @ConnorMcMahon, I'm proactively incrementing the nuget package versions so that we don't have to do it separately as part of the release process, but let me know if you'd prefer we wait to do this.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -188,6 +188,11 @@ public override Task AbandonMessageAsync(MessageData message, SessionBase sessio
             return base.AbandonMessageAsync(message, session);
         }
 
+        public Task DeleteMessageAsync(MessageData message)
+        {
+            return this.DeleteMessageAsync(message, session: null);
```
cgillum, line 193: Needed this new overload to handle the case where we might delete messages immediately after dequeuing them (before we've had a chance to create a session object).

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
```
cgillum, line 13: I'm starting to add C# 8.0 nullable reference types to this codebase to help protect against sneaky null-ref exceptions. There is actually one spot in the code where I may have fixed an unknown bug because of this.

src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
```diff
@@ -29,6 +29,7 @@ class OrchestrationInstanceStatus : TableEntity
         public string CustomStatus { get; set; }
         public DateTime CreatedTime { get; set; }
         public DateTime LastUpdatedTime { get; set; }
+        public DateTime? CompletedTime { get; set; }
```
cgillum, line 32: This is related to the other bug fix where we somehow never returned the completed time of an orchestration. The new test code I introduced (borrowed from the SQL backend tests) uncovered this. Ruslan also ran into this issue when prototyping using the Azure Storage backend.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+                }
+                else
+                {
+                    otherMessages.Add(message);
+                }
+            }
+
+            if (executionStartedMessages?.Count > 0)
+            {
+                Task.Run(async () =>
```
cgillum, line 210: We do the de-dupe detection on a background thread to avoid delaying the processing of all the non-`ExecutionStarted` messages. This should ensure that we don't introduce latency for other orchestrations. I can add a comment to clarify this.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -475,19 +617,48 @@ public virtual void Dispose()
...
             public string OrchestrationInstanceId { get; }
-            public string OrchestrationExecutionId { get; set; }
+            public string? OrchestrationExecutionId
+            {
+                get => this.executionId;
+                set
+                {
+                    if (this.executionId != null)
+                    {
+                        throw new InvalidOperationException($"This batch already has an ExecutionId '{this.executionId}' assigned.");
```
cgillum, line 639: This is new logic to ensure these properties are immutable. Other implementations of this fix had some really confusing data corruption issues, so I added some defensive code to help me reason about the integrity of this data. There should be no actual behavior change.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -953,7 +1045,8 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string
                     ["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Pending.ToString()),
                     ["LastUpdatedTime"] = new EntityProperty(DateTime.UtcNow),
                     ["TaskHubName"] = new EntityProperty(this.settings.TaskHubName),
-                    ["ScheduledStartTime"] = new EntityProperty(executionStartedEvent.ScheduledStartTime)
+                    ["ScheduledStartTime"] = new EntityProperty(executionStartedEvent.ScheduledStartTime),
+                    ["ExecutionId"] = new EntityProperty(executionStartedEvent.OrchestrationInstance.ExecutionId),
```
cgillum, line 1049: For some reason we didn't include the `ExecutionId` property when initially creating the "Pending" record. We added it in other updates to this record, however. Now it will _always_ be set.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1178,7 +1273,7 @@ public override Task StartAsync()
                         executionId,
                         historyEventBatch,
                         newEventListBuffer,
-                        newEvents.Count,
+                        allEvents.Count,
```
cgillum, line 1276: This is another small bug I discovered while testing out different fixes.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1457,7 +1552,7 @@ static string GetBlobName(DynamicTableEntity entity, string property)
                 episodeNumber,
                 stopwatch.ElapsedMilliseconds,
                 estimatedBatchSizeInBytes,
-                string.Concat(eTagValue ?? "(null)", " -->", Environment.NewLine, newETagValue ?? "(null)"),
+                string.Concat(eTagValue ?? "(null)", " --> ", newETagValue ?? "(null)"),
```
cgillum, line 1555: Newlines in the output was really hurting my ability to `grep` through logs, so I decided to remove this newline.

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -49,7 +49,7 @@ internal static void EnsureLogicalTraceActivityId()
         bool IsEnabled(EventLevel level) => this.IsEnabled(level, EventKeywords.None);
 
         [Event(EventIds.TaskHubWorkerStarting, Level = EventLevel.Informational, Version = 1)]
-        public void TaskHubWorkerStarting(string AppName, string ExtensionVersion)
+        internal void TaskHubWorkerStarting(string AppName, string ExtensionVersion)
```
cgillum, line 52: Making all of these `internal` for consistency. This is not an important part of this PR.

test/DurableTask.AzureStorage.Tests/TestOrchestrationHost.cs
```diff
@@ -117,6 +119,133 @@ public void AddAutoStartOrchestrator(Type type)
             return new TestOrchestrationClient(this.client, orchestrationType, instance.InstanceId, creationTime);
         }
 
+        public Task<TestInstance<TInput>> StartInlineOrchestration<TOutput, TInput>(
```
cgillum, line 122: This is part of the new test code that I copied over from DurableTask.SqlServer.

src/DurableTask.Core/Logging/LoggingExtensions.cs
```diff
@@ -47,5 +47,14 @@ public static void LogDurableEvent(this ILogger logger, ILogEvent logEvent, Exce
                 eventSourceEvent.WriteEventSource();
             }
         }
+
+        /// <summary>
+        /// Sets the trace activity ID for the current logical thread.
+        /// </summary>
+        /// <param name="traceActivityId">The trace activity ID to set.</param>
+        public static void SetLogicalTraceActivityId(Guid traceActivityId)
```
cgillum, line 55: New public API in DurableTask.Core that can be consumed by custom backends to make it easier to set ETW activity IDs in cases where traces aren't necessarily associated with a normal orchestration execution context. The prefetching we do for DurableTask.AzureStorage is an example of where I need more control over the activity IDs included in the ETW traces.

cgillum (commented):
Adding some comments to my own PR to hopefully make it a little easier to understand.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+                    executionStartedMessages ??= new List<MessageData>(messages.Count);
+                    executionStartedMessages!.Add(message);
+                }
+                else
+                {
+                    otherMessages.Add(message);
+                }
+            }
+
+            if (executionStartedMessages?.Count > 0)
```
ConnorMcMahon, line 208: This is technically valid C#, but I had to look it up to make sure. Not sure if it is worth adding a comment or even a redundant null coalescing to 0, but I just wanted to call that out.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+                            // Happy path: The message matches the table status. Allow it to run.
+                            messagesToKeep.Add(message);
+                        }
+                        else if (this.IsScheduledAfterInstanceUpdate(message, remoteInstance))
+                        {
+                            // The message was inserted after the Instances table was updated.
+                            // The same machine will have generated both timestamps so time skew is not a factor.
+                            // We know almost certainly that this is a redundant message and can be safely discarded.
+                            messagesToDiscard ??= new List<MessageData>(executionStartedMessages.Count);
+                            messagesToDiscard!.Add(message);
```
ConnorMcMahon, line 239: `!` not required.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+                        }
+                        else if (message.OriginalQueueMessage.DequeueCount >= 10)
+                        {
+                            // We've tried and failed repeatedly to process this start message. Most likely it will never succeed, possibly because
+                            // the client failed to update the Instances table after enqueuing this message. In such a case, the client would
+                            // have observed a failure and will know to retry with a new message. Discard this one.
+                            messagesToDiscard ??= new List<MessageData>(executionStartedMessages.Count);
+                            messagesToDiscard!.Add(message);
```
ConnorMcMahon, line 247: same nit here.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+            if (cloudQueueMessage.DequeueCount <= 1 || !cloudQueueMessage.NextVisibleTime.HasValue)
+            {
+                // We can't use the initial insert time and instead must rely on a re-insertion time,
+                // which is only available to use after the first dequeue count.
+                return false;
+            }
+
+            // This calculation assumes that the value of ControlQueueVisibilityTimeout did not change
+            // in any meaningful way between the time the message was inserted and now.
+            DateTime latestReinsertionTime = cloudQueueMessage.NextVisibleTime.Value.Subtract(this.settings.ControlQueueVisibilityTimeout).DateTime;
```
ConnorMcMahon, line 319: This does have implications on changing `ControlQeueuVisibilityTimeout` now being a "breaking change", but the chances of someone hitting that bug seems very slim.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -188,6 +188,11 @@ public override Task AbandonMessageAsync(MessageData message, SessionBase sessio
             return base.AbandonMessageAsync(message, session);
```
ConnorMcMahon, line 188: Since we already enabled nullable references in the subclass, it may be helpful to enable it in this class assuming it's not too much work.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
ConnorMcMahon: We never really settled on when to update it, but now seems fine to me.

ConnorMcMahon (commented):
Have looked over most of the source code at this point. Overall looks good to me so far, mainly some nits.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -0,0 +1,120 @@
...
+    using System.Threading.Tasks;
+    using DurableTask.Core;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Newtonsoft.Json;
+
+    [TestClass]
+    public class StressTests
+    {
+        /// <summary>
+        /// End-to-end test which validates a simple orchestrator function which doesn't call any activity functions.
```
davidmrdavid, line 30: Can you expand, in the comment itself, what exactly is being validated / what the passing conditions would be?

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -273,7 +273,7 @@ public virtual async Task AbandonMessageAsync(MessageData message, SessionBase s
                     TimeSpan.FromSeconds(numSecondsToWait),
                     MessageUpdateFields.Visibility,
                     this.QueueRequestOptions,
-                    session.StorageOperationContext);
+                    session?.StorageOperationContext ?? new OperationContext());
```
davidmrdavid, line 276: When might we need to create a new OperationContext?

src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
davidmrdavid: I assume this is something we'll want to patch in the future, right?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
             }
         }
 
+        IReadOnlyList<MessageData> FilterOutExecutionStartedForDedupeValidation(
```
davidmrdavid, line 186: I think this helper is important enough to be documented. Could we please embed a header comment outlining what this achieves and why it's needed? I realize this is in the PR itself, just asking for it to be in the codebase as well :)

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
davidmrdavid: I think this is clear enough by itself, but I vote to err on the side of caution and commenting it anyways :)

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
davidmrdavid, line 247: Should we add some telemetry here to flag this edge-case in our logs?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+                            return controlQueue.DeleteMessageAsync(msg, session: null);
+                        });
+                    }
+                });
+            }
+
+            return otherMessages;
+        }
+
+        bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteInstance)
```
davidmrdavid, line 292: Requesting a header comment clarifying what an "update" refers to in this case :)

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+        {
+            if (remoteInstance == null)
+            {
+                // This is a new instance and we don't yet have a status record for it.
+                // We can't make a call for it yet.
+                return false;
+            }
```
davidmrdavid, line 299: Wondering if we should even allow `null` values to be passed-in as parameters. Does it make sense to just check for null-ness prior to calling this method instead?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+
+            if (remoteInstance.CreatedTime < msg.TaskMessage.Event.Timestamp)
+            {
+                // The message was inserted after the Instances table was updated.
+                // The same machine will have generated both timestamps so time skew is not a factor.
+                // We know almost certainly that this is a redundant message and can be safely discarded.
+                return true;
```
davidmrdavid, line 306: I know I've said this in other comments, but I find the phrasing of "The messages was inserted after the Instances table was updated" to be too vague. The reason why is that the word "update" is too general: a lot of things could have been updated and, even if we knew what was updated, we wouldn't know in what way the  updated value changed.

Judging by the if-statement condition, it seems that a more direct phrasing could be: "This ExecutionStarted event does not correspond to the instance's creation time, so it cannot be the original event." Could we opt for something like this?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -178,6 +183,143 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+            }
+
+            CloudQueueMessage cloudQueueMessage = msg.OriginalQueueMessage;
+            if (cloudQueueMessage.DequeueCount <= 1 || !cloudQueueMessage.NextVisibleTime.HasValue)
+            {
+                // We can't use the initial insert time and instead must rely on a re-insertion time,
+                // which is only available to use after the first dequeue count.
+                return false;
+            }
```
davidmrdavid, line 315: Not exactly sure what this block is doing, can is the "originalQueueMessage"?

davidmrdavid (commented):
Thank you for this work, it seems like we're making good progress on these ordering race conditions! Left some comments, requests, and very many questions!

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
cgillum: I spent more time than I probably should have thinking about how I could rewrite this but honestly, every attempt I make feels like I'm further obscuring the logic. The problem with "This ExecutionStarted event does not correspond to the instance's creation time, so it cannot be the original event" is that it's still not clear what that actually means ("correspond" is also vague). My intent is less about giving an abstract sense of semantics and more about providing an explanation about how and why the logic is correct and that all corner cases have been considered. In order to do that, my comments are intentionally describing the mechanics. Hopefully some of the other comments that I already added are helping provide some of the additional context.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
cgillum: There is a typo in your comment, but I'm guessing you're asking _what_ is "originalQueueMessage"? It's the message object that we received from the control queue using the .NET SDK. It has Azure Storage-specific metadata about dequeue counts, visible times, etc and is not specific to Durable Functions. The "MessageData" wrapper type is a Durable-specific deserialized version of this message.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
cgillum: Yeah, this is me acknowledging that there is a potential danger here but I don't think it's worth the investment required to make it safe 100% of the time vs. 99% of the time, especially since most people won't change this value ever, let alone a busy app.

cgillum (commented):
Thanks for the feedback. See my responses below.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -642,6 +643,97 @@ async Task<OrchestrationState> ConvertFromAsync(OrchestrationInstanceStatus orch
...
+
+                queryBuilder.Append("PartitionKey eq '").Append(id).Append('\'');
+                count++;
+            }
+
+            // The caller of this method only cares about the instance ID, execution ID, and created time.
+            var query = new TableQuery
+            {
+                SelectColumns = new[] { "ExecutionId", "CreatedTime" },
+                FilterString = queryBuilder.ToString()
```
ConnorMcMahon, line 673: Is there a limit on the size of this FilterString?

ConnorMcMahon (commented):
I have a few minor qualms about what we are actually validating with our tests, but overall, I think the PR looks solid.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -0,0 +1,120 @@
...
+                            return result;
+                        });
+
+                    // Use the same instance name for all instances
+                    Func<int, string> instanceIdGenerator;
+                    Func<int, string> inputGenerator;
+                    if (useSameInstanceId)
+                    {
+                        DateTime now = DateTime.Now;
+                        instanceIdGenerator = _ => $"ConcurrentInstance_{now:yyyyMMdd.hhmmss.fff}";
```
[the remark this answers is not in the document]
cgillum: This `instanceIdGenerator` returns the exact same string every time now matter how fast/slow the code runs. `DateTime.Now` is never recomputed since it's outside of the lambda. Sorry that this is confusing -  I should have just used a constant string.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                    Assert.AreEqual(MaxConcurrency, instances.Count);
+
+                    // All returned objects point to the same orchestration instance
+                    OrchestrationState[] finalStates = await Task.WhenAll(instances.Select(
+                        i => i.WaitForCompletion(timeout: TimeSpan.FromMinutes(2), expectedOutputRegex: @"Hello, \w+!")));
+
+                    if (useSameInstanceId)
+                    {
+                        // Make sure each instance is exactly the same
```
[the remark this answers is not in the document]
cgillum: This part of the test ensures that all 40 concurrent clients have the same view of the one orchestration. It's representative of real-world client behavior where multiple clients start the same orchestration with the same instance ID at the same time. The expectation is that all clients see the same thing even though 39/40 clients had their start messages quietly discarded.

src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
cgillum: Correct. I was also surprised that we don't return the CompletedTime in Durable Functions, but I guess customers seem to be fine with just LastUpdatedTime. :)

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: Not that I'm aware of. I believe this all ends up being part of the query string of the HTTP request to Azure Storage. At most we'll only send 32 instance IDs since we can dequeue no more than 32 messages at a time. I suppose the maximum length of the query would be based on the maximum length for instance IDs, multiplied by 32.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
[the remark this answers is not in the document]
ConnorMcMahon: I see. Depending on the speed of the orchestration, is it possible that the `ExecutionStarted` messages for the other 39 clients would not yet have been discarded yet, since they have to be dequeued 10 times?

Test/DurableTask.AzureStorage.Tests/StressTests.cs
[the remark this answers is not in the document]
ConnorMcMahon: Ah, I missed that `now` was a variable defined above, my brain read that as string format sugar. This is fine.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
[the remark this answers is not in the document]
cgillum: The 10 dequeue count handles the case where the client enqueues a start message but crashes before updating the Instances table. That's just for starting a single orchestration and isn't necessarily related to concurrent orchestration starts. When running this test locally and examining the ETW output, I observed that 90% of the duplicate messages are discarded immediately and the other 10% get discarded after the second dequeue. The timestamp comparison logic is used to ensure we don't have to wait for 10 dequeue counts.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
ConnorMcMahon: Looks like the [only limit nowadays](https://stackoverflow.com/questions/43959589/max-filter-comparisons-in-an-azure-table-query) is the URL limit, which is ~32KB, which would be hard to hit unless each instance id is roughly ~1KB, which seems unlikely.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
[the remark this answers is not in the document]
ConnorMcMahon: Ah, I forgot about the time check. Yes this checks out, thanks for the clarification.

ConnorMcMahon (approved):
LGTM!

henggua-ramsoft:
This looks like an issue we are running into. How do we determine if this fix has been released yet?

cgillum:
The fix for this has been released: https://github.com/Azure/durabletask/releases/tag/durabletask.azurestorage-v1.8.6. We actually have done another v1.8.7 release since then which covers another issue.

PR #529, opened 2021-04-01
Added support to swap app lease owner and resume attempting to acquire app lease when lease is lost

amdeel:
Added support to allow a swap of the applease without turning off an instance.

Related to this issue: https://github.com/Azure/azure-functions-durable-extension/issues/1503

Summary of changes:
- Added a blob to the applease container, appleaseinfo, that contains the current OwnerId and the DesiredSwapId.
- Added ForceChangeAppLeaseAsync API to AzureStorageOrchestrationService. This updates the appleaseinfo blob's DesiredSwapId to the current app and restarts the LeaseManagerStarter task (in AzureStorageOrchestrationService) , responsible for acquiring and starting the app lease.
- AppLeaseManager.TryAquireAppLeaseAsync will now get the current appleaseinfo blob information. If the appId matches the appleaseinfo blob's DesiredSwapId the lease will be changed to the current app. The blob will also be updated when the lease has been successfully acquired to reflect the new current OwnerId and, if necessary, delete the DesiredSwapId.
- The partition manager is now started by the AppLeaseManager.
- When the AppLeaseManager fails to renew its lease due to another app owning the lease, it will shut itself down, which will also stop the Partitions from being listened to.
- The LeaseManagerStarter task now resumes attempting to acquire the applease when the AppLeaseManager has been stopped using AppLeaseManager.AwaitUntilAppLeaseManagerStopped in a while loop.
- When the applease is forcibly changed both apps will be listening to the partitions until the app that lost the lease notices that it lost its lease in the AppLeaseManager RenewTask. This shouldn't cause any issues when using the SafePartitionManager. When the customer is using the LegacyPartitionManager a delay is added after the lease is acquired to avoid potential split brain issues when starting the partition manager.


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -447,37 +450,51 @@ public async Task StartAsync()
...
-                            await Task.Delay(this.settings.AppLeaseOptions.AcquireInterval);
-                        }
+                this.leaseManagerStarterTokenSource.Cancel();
+                this.leaseManagerStarterTokenSource.Dispose();
+            }
+            this.leaseManagerStarterTokenSource = new CancellationTokenSource();
 
-                        await appLeaseManager.StartAsync();
-                    }
+            _ = await Task.Factory.StartNew(() => this.LeaseManagerStarter(this.leaseManagerStarterTokenSource.Token), this.leaseManagerStarterTokenSource.Token);
```
davidmrdavid, line 467: if we're discarding the result, can we remove the `_ = ` at the beginning of this line?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1824,6 +1872,26 @@ public async Task RewindTaskOrchestrationAsync(string instanceId, string reason)
             }
         }
 
+        /// <summary>
+        ///  Forces this app to take the AppLease if this app doesn't already have it. To use this, must be using the AppLease feature by setting UseAppLease to true in host.json.
```
davidmrdavid, line 1876: nit: the final sentence appears to be missing a subject? :)

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -76,9 +87,36 @@ sealed class AppLeaseManager
...
+                try
+                {
+                    string serializedInfo = JsonConvert.SerializeObject(new AppLeaseInfo());
+                    await this.appLeaseInfoBlob.UploadTextAsync(serializedInfo);
+                }
+                catch (StorageException)
+                {
+                    // eat any storage exception related to conflict
+                    // this means the blob already exist
+                }
```
davidmrdavid, line 110: are we sure that _all_ `StorageException`s here correspond to duplicates? Just want to triple check

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -109,11 +150,98 @@ public async Task StopAsync()
...
+                    this.workerName,
+                    this.appLeaseContainerName);
+
+                AccessCondition accessCondition = new AccessCondition() { LeaseId = currentLeaseId };
+                await appLeaseContainer.ChangeLeaseAsync(this.appLeaseId, accessCondition);
+
+                var appLeaseInfo = new AppLeaseInfo()
+                {
+                    OwnerId = this.appLeaseId,
+                };
```
davidmrdavid, line 206: Is it safe to only assign one of the two fields in `AppLeaseInfo`?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1824,6 +1872,26 @@ public async Task RewindTaskOrchestrationAsync(string instanceId, string reason)
...
+        /// </summary>
+        /// <returns>A task that completes when the steal app message is written to storage and the LeaseManagerStarter Task has been restarted.</returns>
+        public async Task ForceChangeAppLeaseAsync()
```
davidmrdavid, line 1879: Where will this get called?

davidmrdavid (commented):
Thank you for this PR, left some comments and questions. I'm still a bit iffy about the logic of how the `DesiredSwapId` property is assigned, but I know we've discussed that in the past so that's my fault. Could you please point me to a reference where I could brush up on that?

As a final request, I'm hoping we can add header comments to `TryAquireAppLeaseAsync` and `ChangeLeaseAsync` since they appear to be guiding the core logic. Thanks!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: I think it's also more common/recommended nowadays to just use `Task.Run(() => ...)`.

src/DurableTask.AzureStorage/Partitioning/AppLeaseInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage.Partitioning
+{
+    class AppLeaseInfo
```
cgillum, line 16: Is this used anywhere other than AppLeaseManager.cs? If not, can we make this a private inner class of `AppLeaseManager` to avoid adding more files to the source code?

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -328,5 +467,74 @@ public async Task DeleteContainerAsync()
...
+            {
+                this.stats.StorageRequests.Increment();
+            }
+        }
+
+        public async Task<AppLeaseInfo> GetAppLeaseInfoAsync()
+        {
+            if (await this.appLeaseInfoBlob.ExistsAsync())
+            {
+                await appLeaseInfoBlob.FetchAttributesAsync();
```
ConnorMcMahon, line 512: Make sure to wrap these API calls in our wrapper function to prevent deadlocks.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
ConnorMcMahon: it's not immediately clear to me why we don't just `await this.LeaseManagerStarter()` directly?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -487,6 +504,36 @@ private async Task LeaseManagerStarter()
             }
         }
 
+        private async Task AppLeaseManagerStarter(CancellationToken token)
```
ConnorMcMahon, line 507: I don't love this current level of abstraction. I think that if we are going to have an `AppLeaseManager`  class, it should be responsible for the acquire/stop/resume-listening loop itself, and the AzureStorageOrchestrationService should just kick off that initial loop.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -33,38 +34,48 @@ sealed class AppLeaseManager
...
 
+        private readonly string appLeaseInfoBlobName;
+        private readonly CloudBlockBlob appLeaseInfoBlob;
+
         private bool isLeaseOwner;
         private int isStarted;
         private bool shutdownComplete;
         private Task renewTask;
         private CancellationTokenSource leaseRenewerCancellationTokenSource;
+        private TaskCompletionSource<bool> appLeaseTCS;
```
ConnorMcMahon, line 51: If we move the initial lease acquisition logic to this class, I think we can delete the need for this TaskCompletionSource, as we know whether we own the lease or not within this class, and don't need to signal it to outside consumers. In general, TaskCompletionSource [can be dangerous](https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/) if you aren't careful.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -109,11 +150,98 @@ public async Task StopAsync()
...
+            {
+                leaseAcquired = await this.TryAquireLeaseAsync();
+            }
+
+            this.isLeaseOwner = leaseAcquired;
+
+            return leaseAcquired;
+        }
+
+        private async Task<bool> ChangeLeaseAsync(string currentLeaseId)
```
ConnorMcMahon, line 181: `ChangeLeaseAsync()` is broad. I recommend `StealLeaseAsync()`.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -126,8 +254,9 @@ public async Task<bool> TryAquireAppLeaseAsync()
                     this.appLeaseContainerName);
 
                 await appLeaseContainer.AcquireLeaseAsync(this.options.LeaseInterval, this.appLeaseId);
+
+                await this.UpdateOwnerAppIdToCurrentApp();
                 leaseAcquired = true;
-                this.isLeaseOwner = true;
```
ConnorMcMahon, line 130: Is there a reason we removed this.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -328,5 +467,74 @@ public async Task DeleteContainerAsync()
                 this.stats.StorageRequests.Increment();
             }
         }
+
+        public async Task UpdateOwnerAppIdToCurrentApp()
```
ConnorMcMahon, line 471: Be careful with having multiple helper methods that each call `GetAppLeaseInfoAsync()`. I have already found at least one place that calls multiple of these helper methods, which results in duplicate calls to fetch the blob. Consider making these helper methods take an AppLeaseInfo parameter so you ensure you only fetch once.

ConnorMcMahon (commented):
Glad to see the initial work on this. I have some small refactor suggestions that I think would simplify this PR a bit and make it easier to maintain.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -76,9 +87,36 @@ sealed class AppLeaseManager
...
             return result;
         }
 
+        public async Task CreateAppLeaseInfoIfNotExistsAsync()
+        {
+            if (!await this.appLeaseInfoBlob.ExistsAsync())
+            {
+                try
+                {
+                    string serializedInfo = JsonConvert.SerializeObject(new AppLeaseInfo());
```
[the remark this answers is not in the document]
amdeel: That string actually looks like "{\"OwnerId\":null,\"DesiredSwapId\":null}" And if I'm deserializing it to an AppleaseInfo don't I need to serialize an AppLeaseInfo?

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -76,9 +87,36 @@ sealed class AppLeaseManager
...
+                    // eat any storage exception related to conflict
+                    // this means the blob already exist
+                }
+                finally
+                {
+                    this.stats.StorageRequests.Increment();
+                }
+            }
+
+            this.stats.StorageRequests.Increment();
```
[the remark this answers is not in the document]
amdeel: This increment actually goes with the ExistsAsync check. The increment in the finally block goes with UploadTextAsync in the try catch.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: Good point

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: I now am setting this in TryAquireAppLeaseAsync() depending on the result of ChangeLeaseAsync or TryAcquireLeaseAsync. Previously this was just one method, TryAcquireLeaseAsync and that's why, I think, you're seeing this removed.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
amdeel: @ConnorMcMahon we don't actually want to await this task. I want to start up a LeaseManagerStarter task and move on while that LeaseManagerStarter task is a loop that will continuously attempt to acquire the app lease.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: Yes, it's important not to arbitrarily assign a value to DesiredSwapId.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
amdeel: nit on your nit: Your sentence appears to be using the wrong punctuation. :) haha. Actually the subject of the sentence is the implied you.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: When I update these I want to make sure that I have the current/most up to date information in AppLeaseInfo to avoid overwriting with incorrect data. So I don't think that I want these methods to take in a parameter because it could lead to that case. You seem to be talking about TryAcquireAppLease using GetAppLeaseInfo and then calling TryAcquireLease which, if that succeeds, calling an UpdateOwnerId method which will call GetAppLeaseInfo again. I don't personally mind this in this one case.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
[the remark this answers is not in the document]
ConnorMcMahon: So to my knowledge of newtonsoft deserialization, if you use `JsonConvert.DeserializeObject<T>("{}");`, then every property on type `T` will evaluate to the default value, which is equivalent to `JsonConvert.Deserialize<AppLeaseInfo>("{"OwnerId":null,"DesiredSwapId":null}");'

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: I actually did update these methods to be private. However, I do still need to check the blob with GetAppLeaseInfoAsync before updating it because any other worker/app could be updating this blob in the customers storage account. For example, when the StealAppLease API is called.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: ChangeLeaseAsync is actually the name of the CloudBlobContainer API. I realize that I don't have to use this name (and also with my refactor PR coming up I could change these names to whatever I want) but someone familiar with the different Azure Storage lease interactions would understand what this does. Also It's private. So I don't intend to change it unless you have a reason for your suggestion. Any naming changes should happen in my refactor PR. So if you feel strongly about changing ChangeLeaseAsync to StealLeaseAsync or something similar, we should discuss and make that change there.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: What are you suggesting as an alternative to using TaskCompletionSource here? I think it's the best option to avoid constant polling.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+            this.workerName = settings.WorkerId;
+            this.appName = settings.AppName;
+            this.appLeaseIsEnabled = this.settings.UseAppLease;
             this.appLeaseContainer = this.storageClient.GetContainerReference(this.appLeaseContainerName);
+            this.appLeaseInfoBlob = this.appLeaseContainer.GetBlockBlobReference(this.appLeaseInfoBlobName);
 
             using (MD5 md5 = MD5.Create())
             {
                 byte[] hash = md5.ComputeHash(Encoding.Default.GetBytes(this.appName));
                this.appLeaseId = new Guid(hash).ToString();
```
ConnorMcMahon, line 85: nit: I know this isn't this PR, but there is a whitespace mismatch here.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+                    continue;
+                }
+            }
+        }
+
+        public async Task StopAsync()
+        {
+            if (this.appLeaseIsEnabled)
+            {
+                await this.StopAppLeaseAsync();
```
ConnorMcMahon, line 176: nit: It wasn't immediately clear to me whether `this.StopAppLeaseAsync()` would also call `this.partitionManager.StopAsync()`. I see that it does, but I had to hop to that method to verify. I have a slight preference for explicitly calling it here instead, so we have parallelism with the else clause.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+            this.starterTokenSource.Cancel();
+        }
+
+        public async Task ForceChangeAppLeaseAsync()
+        {
+            if (!this.appLeaseIsEnabled)
+            {
+                throw new InvalidOperationException("Cannot force change app lease. UseAppLease is not enabled.");
+            }
+            if (this.isLeaseOwner)
```
ConnorMcMahon, line 192: Not sure if we should throw an exception here, or just make this a no-op. I could go either way.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+            try
+            {
+                if (this.isLeaseOwner)
+                {
+                    AccessCondition accessCondition = new AccessCondition() { LeaseId = appLeaseId };
+                    await this.appLeaseContainer.DeleteIfExistsAsync(accessCondition, null, null);
+                }
+                else
+                {
+                    await this.appLeaseContainer.DeleteIfExistsAsync();
```
ConnorMcMahon, line 223: It seems quite likely that this case will always result in a storage exception if we aren't the lease owner. Is there a specific case where we still need to try deleting it even if we aren't the owner?

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+            finally
+            {
+                    this.stats.StorageRequests.Increment();
+            }
+        }
+
+        async Task StartAppLeaseAsync()
         {
-            if (Interlocked.CompareExchange(ref this.isStarted, 1, 0) != 0)
+            if (Interlocked.CompareExchange(ref this.appLeaseIsStarted, 1, 0) != 0)
```
ConnorMcMahon, line 255: It seems like we have 3 separate sources of truth of the current status of the app lease renewal lookup:

- `appLeaseIsStarted`
- `appLeaseShutdownComplete`
- `leaseRenewerCancellationTokenSource`

Is there a reason we can't use just `leaseRenewerCancellationTokenSource`? I think this would simplify the logic and reduce the chance of forgetting to update one of these values.

ConnorMcMahon (commented):
Overall, I like the refactor a lot. A few more nits/suggestions for how we can potentially simplify even further.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
ConnorMcMahon: Yeah, I think a rename would solve this for me. Maybe StopAppLeaseAndTaskHubProcessing()?

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: If we aren't using the app lease feature, if stop was called before delete, if we lost our lease for any reason aside from another app taking it from us. Us not being the lease owner doesn't always imply that someone else is.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: Right, on second thought I think no-op makes more sense here.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: Good question. So I copied the logic from LeaseCollectionBalancer when I initially created this class. I don't have context on the development of that class, but let me try to explain what I think was considered here and what these variables are for.

First, isStarted assures that the appLease is only started once and is set only in the Start and Stop methods. I think this is pretty straightforward.

ShutdownComplete is used in our LeaseRenewer tasks (LeaseCollectionBalancer line 183 and AppLeaseManager line 437). This is used to continue the LeaseRenewer tasks after shutdown has started, but before their tasks were actually cancelled. In AppLeaseManager this would have the app lease renew until the partitions shut down.

As for why we aren't just using leaseRenewerCancellationTokenSource as our check in our LeaseRenewer task, lets look at how this is currently used. It's used to cancel the Task.Delay at the bottom of the while loop for each of the LeaseRenewer tasks. The only reason I can think of that would prevent us from simply using this CancellationToken cancellation status as our while loop check instead of isStarted and shutdownComplete is because of the possibility that this CacellationTokenSource would be disposed by the time we check this. Although, if the CancellationToken was null we could just add a null check. I assume that using isStarted here instead just makes the logic more readable.

@cgillum Can you provide context on this design? Is this correct?

cgillum: This lease management code has changed significantly since it was first introduced, so I don't quite remember why things are the way they are now. I agree that `appLeaseIsStarted` seems generally useful to ensure we don't accidentally start a lease manager twice. However, it's not clear to me why we need both `appLeaseShutdownComplete` and `leaseRenewerCancellationTokenSource`, and why we can't just reduce it to `leaseRenewerCancellationTokenSource`.

> The only reason I can think of that would prevent us from simply using this CancellationToken cancellation status as our while loop check instead of isStarted and shutdownComplete is because of the possibility that this CacellationTokenSource (CTS) would be disposed by the time we check this.

The following code is actually a little strange:

```csharp
await Task.Delay(this.options.RenewInterval, this.leaseRenewerCancellationTokenSource.Token);
```

It's not normal practice in C# to have a background loop references the CTS directly. In fact, this feels like an anti-pattern. Instead, I think this method should be given the `CancellationToken` as a parameter. That avoids any issue of accessing a disposed object. So I propose we update the `LeaseRenewer` method to take a `CancellationToken` as a parameter, and then update the code like this (I'm looking at the old code, so apologies if this doesn't quite match your PR's version of the code):

```csharp
async Task LeaseRenewer(CancellationToken cancelToken)
{
    this.settings.Logger.PartitionManagerInfo(...);

    while (!cancelToken.IsCancellationRequested)
    {
        try
        {
            bool renewSucceeded = await RenewLeaseAsync();

            if (!renewSucceeded)
            {
                break;
            }

            await Task.Delay(this.options.RenewInterval, cancelToken);
        }
        catch (OperationCanceledException)
        {
            this.settings.Logger.PartitionManagerInfo(...);
        }
        catch (Exception ex)
        {
            this.settings.Logger.PartitionManagerError(...);
        }
    }

    this.settings.Logger.PartitionManagerInfo(...);
}
```

This proposal removes the usage of `appLeaseShutdownComplete`, since I'm not sure what benefit it actually adds. It's also more standard in terms of how background tasks in C# are normally implemented.

amdeel: Thanks Chris! I agree with that decision. It matches my previous understanding of cancelling tasks, which is how I implemented it with the LeaseManagerStarter/AppLeaseManagerStarter task in this PR, so that gives me some confidence with that decision as well. And I agree that the shutdown variable doesn't seem to actually add value.

To address your point 2, I think this is just to avoid an error message from being written if the Task.Delay gets cancelled.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+
+        async Task AppLeaseManagerStarter(CancellationToken token)
+        {
+            while (!token.IsCancellationRequested)
+            {
+                try
+                {
+                    while (!token.IsCancellationRequested && !await this.TryAquireAppLeaseAsync())
+                    {
+                        await Task.Delay(this.settings.AppLeaseOptions.AcquireInterval);
```
cgillum, line 150: I suggest we add the CancellationToken to the call to Task.Delay and add a no-op catch block for `OperationCanceledException`. Otherwise there's a possibility that when shutting down the host, the shutdown process is going to get blocked on waiting for this `Task.Delay` to complete. Ideally, delays would get stopped immediately so that shutdown can happen very rapidly.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,369 @@
...
+                catch (Exception e)
+                {
+                    this.settings.Logger.PartitionManagerError(
+                        this.accountName,
+                        this.settings.TaskHubName,
+                        this.workerName,
+                        null,
+                        $"Error in AppLeaseStarter task. Exception: {e}");
+
+                    await Task.Delay(TimeSpan.FromSeconds(10), token);
```
cgillum, line 166: I'm a little worried that a token cancellation inside this `catch` block could result in an unhandled exception. Could we wrap this in another try/catch to avoid that, or refactor this so that we can catch it as part of another existing `catch (OperationCanceledException)` block (see my other comment)?

cgillum (commented):
A couple small comments related to cancellation.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: I don't even think we need to delay here. I removed this in my most recent update.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
amdeel: Youre right, not including the cancellationToken here was a mistake. I fixed this and added a catch block for the OperationCanceledException.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -13,107 +13,359 @@
...
 
             if (this.renewTask != null)
             {
                 this.leaseRenewerCancellationTokenSource.Cancel();
                 await this.renewTask;
             }
 
+            if (appLeaseTCS != null && !appLeaseTCS.Task.IsCompleted)
+            {
+                appLeaseTCS.SetResult(true);
```
cgillum, line 280: You can probably reduce this down to one line:
```csharp
appLeaseTCS?.TrySetResult(true);
```

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -305,28 +563,68 @@ private async Task ReleaseLeaseAsync()
...
+                return JsonConvert.DeserializeObject<AppLeaseInfo>(serializedEventHubInfo);
+            }
+
+            this.stats.StorageRequests.Increment();
+            return null;
+        }
+
+        Task AwaitUntilAppLeaseStopped()
+        {
+            this.appLeaseTCS = new TaskCompletionSource<bool>();
```
cgillum, line 620: I'm nervous about this use of `TaskCompletionSource` and how it could result in deadlocks if not managed correctly. For example, if we overwrite a non-null instance of `appLeaseTCS` here, then it will be impossible to ever complete any task created by the overwritten instance. Similarly, if through some corner case we call `StopAppLeaseAsync` before the starter loop has a chance to call AwaitUntilAppLeaseStopped, then it seems like we could potentially deadlock because nobody will every signal this task (it was previously null so we skipped the signaling).

A potentially safer alternative would be to use a `AsyncManualResetEvent`. This class already exists in this project. It's an asynchronous version of [ManualResetEvent](https://docs.microsoft.com/en-us/dotnet/api/system.threading.manualresetevent) which is often used for these kinds of coordination situations. My suggestion would be to do the following:

* Initialize a new `AsyncManualResetEvent` in the constructor so it's always non-null. Maybe we name it `shutdownCompletedEvent` and call `shutdownCompletedEvent.Set()` to put it in the "signaled" state.
* Call `shutdownCompletedEvent.Reset()` in the beginning of `StopAppLeaseAsync()` (after the `Interlocked` call) to put it into the "unsignaled" state.
* Call `shutdownCompletedEvent.WaitAsync(TimeSpan.FromSeconds(10), cancellationToken)` here (or in AppLeaseManagerStarter) to wait for shutdown to complete.

I think that should address the deadlock situations I had in mind. It also has the benefit allowing us to immediately break out of waiting if the cancellation token gets triggered. The timeout also ensures that we'll always exist out of this await, even if something goes horribly wrong.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -208,6 +456,15 @@ async Task LeaseRenewer()
...
                 "Background app lease renewer task completed.");
+
+            this.settings.Logger.PartitionManagerInfo(
+                this.accountName,
+                this.taskHub,
+                this.workerName,
+                this.appLeaseContainerName,
+                "Lease renewer task completing. Stopping AppLeaseManager.");
+
+            _ = Task.Run(() => this.StopAppLeaseAsync());
```
cgillum, line 467: Any particular reason why this is called as a background thread? The downside of using background threads like this is that exceptions can go unobserved. If there is a reason, then we should add a comment here in the code.

cgillum (commented):
A few very minor things and then one potential blocker that we can discuss.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -14,106 +14,358 @@
...
+        readonly CloudBlobContainer appLeaseContainer;
+        readonly CloudBlockBlob appLeaseInfoBlob;
+        readonly string appLeaseId;
+
+        bool isLeaseOwner;
+        int appLeaseIsStarted;
+        Task renewTask;
+        CancellationTokenSource starterTokenSource;
+        CancellationTokenSource leaseRenewerCancellationTokenSource;
+        AsyncManualResetEvent shutdownCompletedEvent;
```
cgillum, line 53: Can this be made `readonly`?

cgillum (approved):
The latest update looks good to me.

ConnorMcMahon (approved):
LGTM after final end-to-end validation.

Issue #530, opened 2021-04-09
Race condition between orchestration (re)start and external events cause orchestrations to get stuck in Pending status


cgillum:
This is the root cause of https://github.com/Azure/azure-functions-durable-extension/issues/1715.

The problematic scenario is the following:

* Restarting an instance of a completed orchestration
* At the same time, sending an external event to the restarting instance

The history prefetch behavior of DT.AzureStorage loads the old instance history because it sees the external event message first. It then adds the ExecutionStarted message to the same batch, thus associating it with the old history. Once the batch is executed as a session, both the ExecutionStarted and external event messages are discarded because the orchestration runtime state that got loaded is in the "Completed" state. This leaves the instance status (in the Instances table) stuck in Pending.

What should have happened is that we process the ExecutionStarted message separately from the event raised messages. This would allow us to load the correct (empty) history and start the new orchestration generation successfully. The relevant code that needs to be tweaked is [here](https://github.com/Azure/durabletask/blob/18962b891cf9693509bd476ff8cb0bd9de2633ad/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs#L408-L430).

The workaround is to restart the orchestration again with another ExecutionStarted message.

davidmrdavid:
Fixed in: https://github.com/Azure/durabletask/pull/531

PR #531, opened 2021-04-10
Avoid race-condition in Singletons with external events

davidmrdavid:
Addresses: https://github.com/Azure/durabletask/issues/530

This PR aims to avoid the race-condition in which Singletons load an older history when RaiseEvent events are received close to a new ExecutionStarted event. I also improve the documentation of the relevant method for maintainability, and my own learning.

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/StressTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/StressTests.cs
@@ -26,6 +26,24 @@ namespace DurableTask.AzureStorage.Tests
     [TestClass]
     public class StressTests
     {
+        int originalMinWorkerThreads;
+        int originalMinIoThreads;
+
+        [TestInitialize()]
+        public void Startup()
+        {
+            // Set the minimum thread count to 64+ to make these tests extra concurrent.
+            ThreadPool.GetMinThreads(out this.originalMinWorkerThreads, out this.originalMinIoThreads);
+            ThreadPool.SetMinThreads(Math.Max(64, this.originalMinWorkerThreads), Math.Max(64, this.originalMinIoThreads));
+        }
+
+        [TestCleanup()]
+        public void Cleanup()
+        {
+            // Reset the thread pool configuration
+            ThreadPool.SetMinThreads(this.originalMinWorkerThreads, this.originalMinIoThreads);
+        }
+
         /// <summary>
         /// Starts a large'ish number of orchestrations concurrently and verifies correct behavior
         /// both in the case where they all share the same instance ID and when they have unique
@@ -36,188 +54,116 @@ namespace DurableTask.AzureStorage.Tests
         [DataRow(false)]
         public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
         {
-            // Set the minimum thread count to 64+ to make this test extra concurrent.
-            ThreadPool.GetMinThreads(out int minWorkerThreads, out int minIoThreads);
-            ThreadPool.SetMinThreads(Math.Max(64, minWorkerThreads), Math.Max(64, minIoThreads));
-            try
-            {
-                using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(
-                    enableExtendedSessions: false,
-                    modifySettingsAction: settings => settings.ThrowExceptionOnInvalidDedupeStatus = false))
+            using TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(
+                enableExtendedSessions: false,
+                modifySettingsAction: settings => settings.ThrowExceptionOnInvalidDedupeStatus = false);
+
+            await host.StartAsync();
+
+            var results = new ConcurrentBag<string>();
+
+            // We want a number sufficiently high that it results in multiple message batches
+            const int MaxConcurrency = 40;
+
+            TaskActivity activity = TestOrchestrationHost.MakeActivity(
+                delegate(TaskContext ctx, string input)
                 {
-                    await host.StartAsync();
+                    string result = $"Hello, {input}!";
+                    results.Add(result);
+                    return result;
+                });
 
-                    var results = new ConcurrentBag<string>();
+            // Use the same instance name for all instances
+            Func<int, string> instanceIdGenerator;
+            Func<int, string> inputGenerator;
+            if (useSameInstanceId)
+            {
+                instanceIdGenerator = _ => $"ConcurrentInstance_SINGLETON";
+                inputGenerator = _ => "World";
+            }
+            else
+            {
+                instanceIdGenerator = i => $"ConcurrentInstance_{i:00}";
+                inputGenerator = i => $"{i:00}";
+            }
 
-                    // We want a number sufficiently high that it results in multiple message batches
-                    const int MaxConcurrency = 40;
+            List<TestInstance<string>> instances = await host.StartInlineOrchestrations(
+                MaxConcurrency,
+                instanceIdGenerator,
+                inputGenerator,
+                orchestrationName: "SayHelloOrchestration",
+                version: string.Empty,
+                implementation: (ctx, input) => ctx.ScheduleTask<string>("SayHello", "", input),
+                activities: ("SayHello", activity));
 
-                    TaskActivity activity = TestOrchestrationHost.MakeActivity(
-                        delegate(TaskContext ctx, string input)
-                        {
-                            string result = $"Hello, {input}!";
-                            results.Add(result);
-                            return result;
-                        });
+            Assert.AreEqual(MaxConcurrency, instances.Count);
 
-                    // Use the same instance name for all instances
-                    Func<int, string> instanceIdGenerator;
-                    Func<int, string> inputGenerator;
-                    if (useSameInstanceId)
-                    {
-                        instanceIdGenerator = _ => $"ConcurrentInstance_SINGLETON";
-                        inputGenerator = _ => "World";
-                    }
-                    else
-                    {
-                        instanceIdGenerator = i => $"ConcurrentInstance_{i:00}";
-                        inputGenerator = i => $"{i:00}";
-                    }
+            // All returned objects point to the same orchestration instance
+            OrchestrationState[] finalStates = await Task.WhenAll(instances.Select(
+                i => i.WaitForCompletion(timeout: TimeSpan.FromMinutes(2), expectedOutputRegex: @"Hello, \w+!")));
 
-                    List<TestInstance<string>> instances = await host.StartInlineOrchestrations(
-                        MaxConcurrency,
-                        instanceIdGenerator,
-                        inputGenerator,
-                        orchestrationName: "SayHelloOrchestration",
-                        version: string.Empty,
-                        implementation: (ctx, input) => ctx.ScheduleTask<string>("SayHello", "", input),
-                        activities: ("SayHello", activity));
-
-                    Assert.AreEqual(MaxConcurrency, instances.Count);
-
-                    // All returned objects point to the same orchestration instance
-                    OrchestrationState[] finalStates = await Task.WhenAll(instances.Select(
-                        i => i.WaitForCompletion(timeout: TimeSpan.FromMinutes(2), expectedOutputRegex: @"Hello, \w+!")));
-
-                    if (useSameInstanceId)
-                    {
-                        // Make sure each instance is exactly the same
-                        string firstInstanceJson = JsonConvert.SerializeObject(finalStates[0]);
-                        foreach (OrchestrationState state in finalStates.Skip(1))
-                        {
-                            string json = JsonConvert.SerializeObject(state);
-                            Assert.AreEqual(firstInstanceJson, json, "Expected that all instances have the same data.");
-                        }
-                    }
-                    else
-                    {
-                        // Make sure each instance is different
-                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.OrchestrationInstance.InstanceId).Distinct().Count());
-                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.OrchestrationInstance.ExecutionId).Distinct().Count());
-                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.Input).Distinct().Count());
-                        Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.Output).Distinct().Count());
-                    }
-
-                    await host.StopAsync();
+            if (useSameInstanceId)
+            {
+                // Make sure each instance is exactly the same
+                string firstInstanceJson = JsonConvert.SerializeObject(finalStates[0]);
+                foreach (OrchestrationState state in finalStates.Skip(1))
+                {
+                    string json = JsonConvert.SerializeObject(state);
+                    Assert.AreEqual(firstInstanceJson, json, "Expected that all instances have the same data.");
                 }
             }
-            finally
+            else
             {
-                // Reset the thread pool configuration
-                ThreadPool.SetMinThreads(minWorkerThreads, minIoThreads);
+                // Make sure each instance is different
+                Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.OrchestrationInstance.InstanceId).Distinct().Count());
+                Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.OrchestrationInstance.ExecutionId).Distinct().Count());
+                Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.Input).Distinct().Count());
+                Assert.AreEqual(MaxConcurrency, finalStates.Select(s => s.Output).Distinct().Count());
             }
+
+            await host.StopAsync();
         }
 
-        internal class Approval : TaskOrchestration<string, TimeSpan, bool, string>
+        [TestMethod]
+        public async Task RestartOrchestrationWithExternalEvents()
         {
-            TaskCompletionSource<bool> waitForApprovalHandle;
+            using TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: false);
+            await host.StartAsync();
 
-            public override async Task<string> RunTask(OrchestrationContext context, TimeSpan timeout)
+            // This is the local method we'll use to continuously recreate the same instance.
+            // It doesn't matter that the orchestration name is different as long as the instance ID is the same.
+            Task<TestInstance<string>> CreateInstance(int i) => host.StartInlineOrchestration(
+                input: "Hello, world!",
+                instanceId: "Singleton",
+                orchestrationName: $"EchoOrchestration{i}",
+                implementation: (ctx, input) => Task.FromResult(input));
+
+            // Start the first iteration and wait for it to complete.
+            TestInstance<string> instance = await CreateInstance(0);
+            await instance.WaitForCompletion();
+
+            int restartIterations = 10;
+            int externalEventCount = 10;
+
+            // We'll keep track of the execution IDs to make 100% sure we get new instances every iteration.
+            var executionIds = new HashSet<string> { instance.ExecutionId };
+
+            // Simultaneously send a bunch of external events as well as a "recreate" message. Repeat this
+            // several times. Need to ensure that the orchestration can always be restarted reliably. It
+            // doesn't matter whether the orchestration handles the external events or not.
+            for (int i = 1; i <= restartIterations; i++)
             {
-                Task<bool> approvalTask = this.GetWaitForApprovalTask();
+                List<Task> concurrentTasks = Enumerable.Range(0, externalEventCount).Select(j => instance.RaiseEventAsync("DummyEvent", j)).ToList();
+                Task<TestInstance<string>> recreateInstanceTask = CreateInstance(i);
+                concurrentTasks.Add(recreateInstanceTask);
+                await Task.WhenAll(concurrentTasks);
 
+                // Wait for the newly created instance to complete.
+                instance = await recreateInstanceTask;
+                await instance.WaitForCompletion();
 
-                await Task.WhenAny(approvalTask);
-                // The timer must be cancelled or fired in order for the orchestration to complete.
-
-                bool approved = approvalTask.Result;
-                return approved ? "Approved" : "Rejected";
-            }
-
-            async Task<bool> GetWaitForApprovalTask()
-            {
-                this.waitForApprovalHandle = new TaskCompletionSource<bool>();
-                bool approvalResult = await this.waitForApprovalHandle.Task;
-                this.waitForApprovalHandle = null;
-                return approvalResult;
-            }
-
-            public override void OnEvent(OrchestrationContext context, string name, bool approvalResult)
-            {
-                Assert.AreEqual("approval", name, true, "Unknown signal recieved...");
-                if (this.waitForApprovalHandle != null)
-                {
-                    this.waitForApprovalHandle.SetResult(approvalResult);
-                }
-            }
-        }
-
-        [DataTestMethod]
-        public async Task SingletonsLoadCorrectHistoryWithRaiseEvent()
-        {
-
-            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
-            {
-                await host.StartAsync();
-                int numExternalEvents = 50;
-                int maxIterations = 20;
-                int currentIteration = 0;
-                do
-                {
-                    // (1) Start a singleton orchestration that waits for a single event (wait for it to finish starting)
-                    TestOrchestrationClient client = await host.StartOrchestrationAsync(typeof(Approval), input: null, instanceId: "mySingleton");
-                    DateTime startTime = DateTime.UtcNow;
-                    OrchestrationState existingInstance = null;
-                    while (existingInstance == null || !existingInstance.OrchestrationStatus.Equals(OrchestrationStatus.Running))
-                    {
-                        existingInstance = await client.GetStatusAsync();
-                        if (DateTime.UtcNow - startTime > TimeSpan.FromMinutes(2))
-                        {
-                            Assert.Fail($"Singleton did not complete within 2 minutes with status {existingInstance.OrchestrationStatus}  at iteration {currentIteration}");
-                        }
-                    }
-
-                    // (2) Send a large number of external events AND start a new instance of the same orchestration concurrently.
-                    List<Task> externalEvents = new List<Task>();
-                    for (var i = 0; i < numExternalEvents; i++)
-                    {
-                        externalEvents.Add(client.RaiseEventAsync(eventName: "approval", eventData: true));
-                    }
-
-                    async Task getStartNewSingletonTask()
-                    {
-                        existingInstance = null;
-                        while (existingInstance == null || !existingInstance.OrchestrationStatus.Equals(OrchestrationStatus.Completed))
-                        {
-                            await client.RaiseEventAsync(eventName: "approval", eventData: true);
-                            existingInstance = await client.GetStatusAsync();
-                        }
-                        await host.StartOrchestrationAsync(typeof(Approval), input: null, instanceId: "mySingleton");
-                    };
-
-                    Task startNewSingletonTask = getStartNewSingletonTask();
-                    externalEvents.Add(startNewSingletonTask);
-
-                    await Task.WhenAll(externalEvents);
-
-                    // (3) check orchestration is not pending
-                    startTime = DateTime.UtcNow;
-                    existingInstance = null;
-                    while (existingInstance == null || !existingInstance.OrchestrationStatus.Equals(OrchestrationStatus.Completed))
-                    {
-                        await client.RaiseEventAsync(eventName: "approval", eventData: true);
-                        existingInstance = await client.GetStatusAsync();
-                        if (DateTime.UtcNow - startTime > TimeSpan.FromMinutes(2))
-                        {
-                            Assert.Fail($"Singleton did not complete within 2 minutes with status {existingInstance.OrchestrationStatus} at iteration {currentIteration}");
-                        }
-                    }
-
-
-                }
-                while (currentIteration++ < maxIterations);
-                Assert.IsTrue(true);
-                await host.StopAsync();
+                // Ensure that this is a new execution ID.
+                Assert.IsTrue(executionIds.Add(instance.ExecutionId));
             }
         }
     }
--- a/Test/DurableTask.AzureStorage.Tests/StressTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/StressTests.cs
@@ -117,5 +117,124 @@ namespace DurableTask.AzureStorage.Tests
                 ThreadPool.SetMinThreads(minWorkerThreads, minIoThreads);
             }
         }
+
+        internal class Approval : TaskOrchestration<string, TimeSpan, bool, string>
+        {
+            TaskCompletionSource<bool> waitForApprovalHandle;
+            public static bool shouldFail = false;
+
+            public override async Task<string> RunTask(OrchestrationContext context, TimeSpan timeout)
+            {
+                DateTime deadline = context.CurrentUtcDateTime.Add(timeout);
+
+                using (var cts = new CancellationTokenSource())
+                {
+                    Task<bool> approvalTask = this.GetWaitForApprovalTask();
+                    Task timeoutTask = context.CreateTimer(deadline, cts.Token);
+
+                    if (shouldFail)
+                    {
+                        throw new Exception("Simulating unhanded error exception");
+                    }
+
+                    if (approvalTask == await Task.WhenAny(approvalTask, timeoutTask))
+                    {
+                        // The timer must be cancelled or fired in order for the orchestration to complete.
+                        cts.Cancel();
+
+                        bool approved = approvalTask.Result;
+                        return approved ? "Approved" : "Rejected";
+                    }
+                    else
+                    {
+                        return "Expired";
+                    }
+                }
+            }
+
+            async Task<bool> GetWaitForApprovalTask()
+            {
+                this.waitForApprovalHandle = new TaskCompletionSource<bool>();
+                bool approvalResult = await this.waitForApprovalHandle.Task;
+                this.waitForApprovalHandle = null;
+                return approvalResult;
+            }
+
+            public override void OnEvent(OrchestrationContext context, string name, bool approvalResult)
+            {
+                Assert.AreEqual("approval", name, true, "Unknown signal recieved...");
+                if (this.waitForApprovalHandle != null)
+                {
+                    this.waitForApprovalHandle.SetResult(approvalResult);
+                }
+            }
+        }
+
+        [DataTestMethod]
+        public async Task SingletonsLoadCorrectHistoryWithRaiseEvent()
+        {
+
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
+            {
+                await host.StartAsync();
+                int numExternalEvents = 50;
+                int maxIterations = 20;
+                int currentIteration = 0;
+                do
+                {
+                    // (1) Start a singleton orchestration that waits for a single event (wait for it to finish starting)
+                    TestOrchestrationClient client = await host.StartOrchestrationAsync(typeof(Approval), input: null, instanceId: "mySingleton");
+                    DateTime startTime = DateTime.UtcNow;
+                    OrchestrationState existingInstance = null;
+                    while (existingInstance == null || !existingInstance.OrchestrationStatus.Equals(OrchestrationStatus.Completed))
+                    {
+                        existingInstance = await client.GetStatusAsync();
+                        if (DateTime.UtcNow - startTime > TimeSpan.FromMinutes(2))
+                        {
+                            Assert.Fail($"Singleton did not complete within 2 minutes with status {existingInstance.OrchestrationStatus}  at iteration {currentIteration}");
+                        }
+                    }
+
+                    // (2) Send a large number of external events AND start a new instance of the same orchestration concurrently.
+                    List<Task> externalEvents = new List<Task>();
+                    for (var i = 0; i < numExternalEvents; i++)
+                    {
+                        externalEvents.Add(client.RaiseEventAsync(eventName: "approval", eventData: ""));
+                    }
+
+                    Func<Task> getStartNewSingletonTask = async () =>
+                    {
+                        existingInstance = null;
+                        while (existingInstance == null || !existingInstance.OrchestrationStatus.Equals(OrchestrationStatus.Completed))
+                        {
+                            existingInstance = await client.GetStatusAsync();
+                        }
+                        await host.StartOrchestrationAsync(typeof(Approval), input: null, instanceId: "mySingleton");
+                    };
+
+                    Task startNewSingletonTask = getStartNewSingletonTask.Invoke();
+                    externalEvents.Add(startNewSingletonTask);
+
+                    await Task.WhenAll(externalEvents);
+
+                    // (3) check orchestration is not pending
+                    startTime = DateTime.UtcNow;
+                    existingInstance = null;
+                    while (existingInstance == null || !existingInstance.OrchestrationStatus.Equals(OrchestrationStatus.Completed))
+                    {
+                        existingInstance = await client.GetStatusAsync();
+                        if (DateTime.UtcNow - startTime > TimeSpan.FromMinutes(2))
+                        {
+                            Assert.Fail($"Singleton did not complete within 2 minutes with status {existingInstance.OrchestrationStatus} at iteration {currentIteration}");
+                        }
+                    }
+
+
+                }
+                while (currentIteration++ < maxIterations);
+                Assert.IsTrue(true);
+                await host.StopAsync();
+            }
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -413,11 +413,30 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
+                            // A null executionId represents either a management operation (ex: terminate),
+                            // or a pending / new orchestration generation.
+                            if (executionId == null)
+                            {
+                                // We process execution started events separately
+                                if (data.TaskMessage.Event.EventType == EventType.ExecutionStarted)
+                                {
+                                    // Set targetBatch to null to avoid accidentally loading
+                                    // and older history in the case of a singleton
+                                    targetBatch = null;
```
cgillum, line 430: If I'm not mistaken, `targetBatch = null` is the default state and we break out of the loop immediately after assigning it. Thus, assigning it to null here is redundant. I think it's best to avoid this redundant assignment because doing so confuses the expectations for this local variable - i.e. a reader might assume that there is a case where it transitions from non-null to null, and that the logic might be more complicated than it really is, which isn't the case.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -413,11 +413,30 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
                             {
                                 targetBatch = batch;
                                 break;
                             }
+                            // A null executionId represents either a management operation (ex: terminate),
+                            // or a pending / new orchestration generation.
+                            if (executionId == null)
+                            {
+                                // We process execution started events separately
+                                if (data.TaskMessage.Event.EventType == EventType.ExecutionStarted)
```
cgillum, line 426: I don't expect this will ever be true. ExecutionStarted messages always have an execution ID.

cgillum (commented):
Thanks for proactively looking into this! I have some comments inline, but they might be moot since I think there might be a simpler approach: What if instead of changing our if/else logic like you are, we simply skip over the whole `while (node != null)` loop body if we know it's an ExecutionStarted message? That way we will just always create new message batches for ExecutionStarted messages.

davidmrdavid:
Thank you for your comments, and I see your point. I'll go with your suggestion on my next commit. However, I'm still iffy about the algorithmic details of this method so I would like to add some comments to this method to help clarify its behavior to me, and to future maintainers of it. 

In addition to your suggestion, I've also added several comments that reflect my best guess as to what the method is doing. Some of them are contradictory, and some of them are direct questions to you (or anyone reading this PR) for things that I don't know how to interpret. Thank you in advance for giving them a look and clarifying them, @cgillum

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -364,14 +371,18 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
 
                 foreach (MessageData data in queueMessages)
                 {
+                    // The instanceID identifies the orchestration across replays and, in the case of singletons, across generations.
+                    // The executionID identifies an invocation of the orchestration, such as a replay.
```
cgillum, line 375: > The executionID identifies an invocation of the orchestration, such as a replay.

Not quite. The execution ID identifies a "generation" of an orchestration instance. It's randomly generated whenever a new orchestration instance is created or when an existing instance calls ContinueAsNew. It doesn't change between invocations or replays.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -364,14 +371,18 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
                     string instanceId = data.TaskMessage.OrchestrationInstance.InstanceId;
                     string executionId = data.TaskMessage.OrchestrationInstance.ExecutionId;
 
+                    // TODO: What is this if-statement checking for? What does a non-null session represent?
```
cgillum, line 379: > What is this if-statement checking for? What does a non-null session represent?

This dictionary holds a reference to every orchestration instance that is actively executing in memory on this worker. It's what we refer to as a "session". It's never null. It will be removed from this dictionary as soon as it's unloaded from memory.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -364,14 +371,18 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
                     if (this.activeOrchestrationSessions.TryGetValue(instanceId, out OrchestrationSession session))
                     {
                         // If the target orchestration is already running we add the message to the session
-                        // directly rather than adding it to the linked list. A null executionId value
-                        // means that this is a management operation, like RaiseEvent or Terminate, which
+                        // directly rather than adding it to the linked list. <Why?>
```
cgillum, line 383: > Why?

Extended sessions. The linked list of message batches represents the list of pending orchestrations that we want to run but haven't started running yet. These pending instances need to have their history loaded and executed from scratch. With extended sessions, if we detect that an orchestration is already active in memory, we can post messages directly to the active instance.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -403,40 +414,24 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
                         continue;
                     }
 
+                    // TODO - What is this block achieving? And can we define what a  `targetBatch` means?
+                    // We skip this block when the event is ExecutionStarted. This allows us to generate a new
+                    // history instead of loading an older one.
+                    PendingMessageBatch? targetBatch = null;
```
cgillum, line 420: Background: DTFx allows you to process multiple events for a single orchestration as a batch. For example, if you fan out to 10 activities and all activity results land in the control queue at the same time, we can dequeue them all at once and execute them together. This way we don't have to do a full history replay for every single result. Rather, we do it once for the entire batch.

Messages are dequeued from control queues in batches of up to 32 (this is the limit imposed by Azure Storage). The 32 messages could be for a single orchestration or for many different orchestrations. We have no idea until we actually read and deserialize the messages. The DTFx contract requires us to feed it batches of messages that all belong to the same orchestration instance, so this code was written to sort all the dequeued messages into "batches" based on their instance ID and execution ID (which uniquely identifies an orchestration instance). `targetBatch` represents the "correct" batch in which to place a particular message.

The linked list of batches is intended to emulate a combination of a dictionary (where the key is the instance ID/execution ID pair) and a FIFO queue. FIFO is important to ensure fairness in terms of what order we schedule pending orchestrations to execute. Unfortunately we can't use a simple sorted dictionary because we have this weird situation where some messages don't have execution IDs, so we need to come up with some logic that figures out how to add messages without execution IDs to batches that do have execution IDs (if any) - i.e. the keys in the "dictionary" may change over time.

Note that these "batches" here are transient. Another thread is constantly pulling batches from this linked list so that it can start executing them. The lock that protects this method (and others) ensures that we don't have to worry about race conditions of batches being modified and pulled for execution at the same time.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -450,6 +445,8 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
 
                     if (targetBatch == null)
                     {
+                        // Since `targetBatch` is null, we know that the orchestration does not have a history associated with it,
+                        // so we add the messages to a pending message batch instead.
```
cgillum, line 449: This comment is technically correct, but doesn't accurately represent what `targetBatch == null` means. Put simply, `targetBatch == null` means we didn't find any existing batch of messages that matches this instance ID / execution ID pair. It's like when you do a "get" on a dictionary for a key that doesn't exist and get a `null` as the return value. In this case, if the batch wasn't found, we create a new batch and add it to the back of the list.

I think it's more accurate and meaningful to say something like "Since `targetBatch` is null (we couldn't find any existing batches of messages to add this message to), we must create a new batch".

cgillum (commented):
Some explanations, which hopefully are helpful.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -346,6 +346,13 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
         }
 
+        /// <summary>
+        /// Adds history messages to a pending orchestration.
+        /// </summary>
+        /// <param name="controlQueue">The orchestration's control-queue.</param>
+        /// <param name="queueMessages">New messages to assign to orchestrators</param>
+        /// <param name="traceActivityId">The "related" ActivityId of this operation.</param>
+        /// <param name="cancellationToken">Cancellation token in case the orchestration is terminated.</param>
         internal void AddMessageToPendingOrchestration(
```
davidmrdavid, line 356: At this point I'm only unsure about the meaning of "Pending" in this case. Do we mean "Pending" as in the orchestration's runtime status or pending in some other sense? I'm asking because I find it strange that a pending orchestration could have an existing session, i.e, be loaded in memory, and not transition to a "Running" state also immediately after.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -429,8 +441,10 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
                         node = node.Previous;
                     }
 
+                    // If there is no batch for this instanceID-executionID pair
                     if (targetBatch == null)
                     {
+                        // Create a batch for this message's instanceID-executionID pair
```
cgillum, line 447: nit: I feel like this comment might be starting to cross the line of helpful vs. redundant. Hopefully it's already clear from the code that we're creating a new batch. Was there maybe another subtle behavior that wasn't clear which we can point out instead?

cgillum (commented):
This latest round looks good to me! Since we know the scenario that wouldn't have resulted in bad behavior prior to your fix (racing between ExecutionStarted and ExternalEvent), should we write a test that verifies that we've successfully fixed the issue?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
davidmrdavid: I think that's fair, I tend to err on the side of over-commenting, sometimes to a fault. I've removed this comment and added just 2 words to the previous one. Hopefully that's less redundant :)

davidmrdavid:
Thanks for the feedback so far, I understand this part of the code so much better now. +1 to adding some minimal testing, I'll scout the testing directory for some examples and put something together. Thanks!

cgillum:
Take a look at the StressTests.cs file I added in my recent de-dup PR: https://github.com/Azure/durabletask/blob/main/Test/DurableTask.AzureStorage.Tests/StressTests.cs. You can probably add another test there which starts (or restarts) a bunch of orchestrations and throws external events at them at the same time, and then verify that we never get stuck in a Pending state.

(Unrelated: as I was looking for that test file, I noticed that it looks like my PR messed up the file system layout on GitHub because of a capitalization difference in "test" vs "Test" - just FYI in case you're wondering why the test code seems organized in a weird way)

davidmrdavid:
A few questions about the testing scenario. So this issue is particularly tied to the singleton pattern, so I don't see much benefit in having concurrently running orchestrations for the test, would you agree? Instead, it seems that we want to have a 2 concurrent loops: one where we continuously try to fire singleton orchestrations if they're not currently running, and another which sends non-stop signals to the fixed orchestrationID. Maybe we do this for 100 generations or so, and if the orchestrationID at the end is in a finished state, then we pass this test.

Does that seem right to you?

cgillum:
Agreed. You don't necessarily need concurrent loops, though. It might be easy enough to do the following:

* Start a singleton orchestration that waits for a single event (wait for it to finish starting)
* Send a large number of external events AND start a new instance of the same orchestration concurrently.
* Repeat this a few times, checking to ensure that the instance never gets stuck in Pending

The above can be done with a single instance. The only concurrency involved is sending external events and a new start message to that single instance at the same time. You should be able to reuse the primitives that exist in the other test that I pointed you to.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -117,5 +117,124 @@ public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
                 ThreadPool.SetMinThreads(minWorkerThreads, minIoThreads);
             }
         }
+
+        internal class Approval : TaskOrchestration<string, TimeSpan, bool, string>
```
davidmrdavid, line 121: this is copy-pasted from a different test, will clean-up once I manage to get the test case passing reliably

davidmrdavid:
Note to self: latest commit failing with exception `Assert.Fail failed. Singleton did not complete within 2 minutes with status Pending  at iteration 4`

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -117,5 +117,124 @@ public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
...
+                    }
+
+                    // (2) Send a large number of external events AND start a new instance of the same orchestration concurrently.
+                    List<Task> externalEvents = new List<Task>();
+                    for (var i = 0; i < numExternalEvents; i++)
+                    {
+                        externalEvents.Add(client.RaiseEventAsync(eventName: "approval", eventData: ""));
+                    }
+
+                    Func<Task> getStartNewSingletonTask = async () =>
```
cgillum, line 205: C# tip: use local functions instead of Func<Task> here. It's much cleaner.

```csharp
async Task StartNewSingleton()
{
    // implementation
}

// invocation
await StartNewSingleton();
```

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -117,5 +117,108 @@ public async Task ConcurrentOrchestrationStarts(bool useSameInstanceId)
...
+        {
+            TaskCompletionSource<bool> waitForApprovalHandle;
+
+            public override async Task<string> RunTask(OrchestrationContext context, TimeSpan timeout)
+            {
+                Task<bool> approvalTask = this.GetWaitForApprovalTask();
+
+
+                await Task.WhenAny(approvalTask);
+                // The timer must be cancelled or fired in order for the orchestration to complete.
```
davidmrdavid, line 131: there's no longer a timer, this is leftover code, will clean-up early next week

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -22,14 +22,18 @@ namespace DurableTask.Core.Logging
     /// Event source logger for DurableTask.Core that uses structured events rather than log messages.
     /// </summary>
     [EventSource(Name = "DurableTask-Core")]
-    class StructuredEventSource : EventSource
+    public class StructuredEventSource : EventSource
```
davidmrdavid, line 25: this is just for debuggability, will remove for the final commit. Currently using this to get a history of events that we can analyze to understand why the orchestration is still stuck

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -30,6 +30,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Serilog.Extensions.Logging.File" Version="1.0.1" />
```
davidmrdavid, line 33: also a temporary addition for observability

test/DurableTask.AzureStorage.Tests/TestHelpers.cs
```diff
@@ -29,6 +38,9 @@ static class TestHelpers
         {
             string storageConnectionString = GetTestStorageAccountConnectionString();
 
+
+            var listener = new CustomEventListener();
+            listener.EnableEvents(StructuredEventSource.Log, EventLevel.LogAlways);
```
davidmrdavid, line 43: temporary, for observability

test/DurableTask.AzureStorage.Tests/TestOrchestrationHost.cs
```diff
@@ -38,7 +39,7 @@ public TestOrchestrationHost(AzureStorageOrchestrationServiceSettings settings)
             this.service.CreateAsync().GetAwaiter().GetResult();
 
             this.settings = settings;
-
+            settings.LoggerFactory = settings.LoggerFactory.AddFile("<mypath>");
```
davidmrdavid, line 42: this is where my history file is being written

davidmrdavid:
In my latest commit, I include some temporary changes that allow me to get some observability into the orchestration history. I need these to understand why the orchestration is still getting stuck with "Pending".

davidmrdavid:
I'm attaching the a history of the stuck orchestration here: 
[history.txt](https://github.com/Azure/durabletask/files/6328322/history.txt)

Around line 1270, I can see the moment where the `ExecutionStarted` which should be started a new history is accidentally getting confused as a duplicate.

Below is the relevant snippet, look for the keyword "duplicate" near the bottom:
```
2021-04-16T16:37:11.3193276-07:00  [INF] Raising 'approval' event with 4 bytes to 'mySingleton' (fdaef063)
2021-04-16T16:37:11.3194512-07:00  [INF] Sending [EventRaised] message to test-control-00 for instance 'mySingleton' (b0b02da4)
2021-04-16T16:37:11.3215609-07:00  [INF] mySingleton: Appended 35 new events to the history table in 234ms (d869bc8b)
2021-04-16T16:37:11.3297799-07:00  [INF] test-control-00: No new messages were found - backing off (d4c35249)
2021-04-16T16:37:11.3445754-07:00  [INF] mySingleton: Updated Instances table and set the runtime status to 'Completed' (ed338f83)
2021-04-16T16:37:11.3446484-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3448669-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3449874-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3450909-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3451991-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3452941-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3453897-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3455271-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3456348-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3457292-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3458282-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3459317-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3460280-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3461220-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3462170-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3463104-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3464062-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3465016-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3466013-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3466936-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3467972-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3469318-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3470787-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3471915-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3473072-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3474509-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3475716-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3477430-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3478933-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3480367-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3481862-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3483500-07:00  [INF] mySingleton: Deleting [EventRaised] message from test-control-00 (3882ffba)
2021-04-16T16:37:11.3498963-07:00  [INF] Fetching tracking state for instance 'mySingleton' (385d5de2)
2021-04-16T16:37:11.3610035-07:00  [INF] Fetched instance status for mySingleton (c90e3d32)
2021-04-16T16:37:11.3613733-07:00  [INF] Scheduling orchestration 'DurableTask.AzureStorage.Tests.StressTests+Approval' with instance ID = 'mySingleton' and 0 bytes of input (39f08328)
2021-04-16T16:37:11.3760966-07:00  [INF] Fetched instance status for mySingleton (c90e3d32)
**2021-04-16T16:37:11.3764922-07:00  [INF] Sending [ExecutionStarted] message to test-control-00 for instance 'mySingleton' (b0b02da4)
2021-04-16T16:37:11.4661882-07:00  [INF] mySingleton: Fetched [EventRaised] message from test-control-00 (delay = 466ms) (16725a75)
2021-04-16T16:37:11.4805023-07:00  [INF] mySingleton: Fetched [ExecutionStarted] message from test-control-00 (delay = 480ms) (16725a75)
2021-04-16T16:37:11.4873555-07:00  [INF] Fetched instance status for mySingleton (c90e3d32)
2021-04-16T16:37:11.4891872-07:00  [WRN] mySingleton: Received a duplicate message [ExecutionStarted] from test-control-00 with ID = 7c01fd08-489f-4842-9a5a-dd9ca9ef8fe5 and dequeue count = 1 (0abeb1ab)
2021-04-16T16:37:11.4895411-07:00  [INF] mySingleton: Deleting [ExecutionStarted] message from test-control-00 (3882ffba)**
2021-04-16T16:37:11.4907548-07:00  [INF] test-control-00: No new messages were found - backing off (d4c35249)
2021-04-16T16:37:11.5012797-07:00  [INF] mySingleton: Updated Instances table and set the runtime status to 'Pending' (ed338f83)
2021-04-16T16:37:11.5014431-07:00  [INF] Raising 'approval' event with 4 bytes to 'mySingleton' (fdaef063)
```

I couldn't figure out the meaning of the ID inside parentheses at the end of each log message, do y'all know?

cgillum:
> I couldn't figure out the meaning of the ID inside parentheses at the end of each log message, do y'all know?

I'm not sure but I'm guessing this is something specific to whatever logging framework you're using to output these logs.

> Around line 1270, I can see the moment where the ExecutionStarted event, which should be started a new history, is accidentally getting confused as a duplicate from an older generation.

This would mean that there is a mismatch between the execution ID of that ExecutedStarted message and the Instances record for that instance. I think we'll need to see the full structured trace events to be able to easily diagnose this. I typically collect these by running another app on my machine that listens to ETW traces and writes them to the console. Let me know if you'd like help with this.

ConnorMcMahon:
There is also a way you can listen to ETW events within the process and emit them as a part of the test logs. We do this for some of our extension tests, and that may be easier to do then running Chris's ETW listening console app.

davidmrdavid:
[eventSourceTrace.txt](https://github.com/Azure/durabletask/files/6339778/eventSourceTrace.txt)

Thanks for the suggestions. I'm attaching a capture of EventSource traces to this thread. Below is a snippet of the relevant segment (line 2387). Look for the keyword "duplicate":

```
DurableTask-AzureStorage | DeletingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, MessageId=48ab1021-d2c3-4779-bf7c-361b11b40d3a, InstanceId=mySingleton, ExecutionId=, PartitionId=test-control-00, SequenceNumber=31, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | WaitingForMoreMessages ||Account=devstoreaccount1, TaskHub=test, PartitionId=test-control-00, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | DeletingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, MessageId=6dd0e52d-c71a-4553-92dc-2d04b6e496b6, InstanceId=mySingleton, ExecutionId=, PartitionId=test-control-00, SequenceNumber=32, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | DeletingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, MessageId=3cbd2946-5014-4d15-9d58-1cc4800116f0, InstanceId=mySingleton, ExecutionId=, PartitionId=test-control-00, SequenceNumber=34, AppName=, ExtensionVersion=1.8.6
DurableTask-Core | FetchingInstanceState ||InstanceId=mySingleton, ExecutionId=, AppName=, ExtensionVersion=2.5.3
DurableTask-AzureStorage | FetchedInstanceStatus ||Account=devstoreaccount1, TaskHub=test, InstanceId=mySingleton, ExecutionId=, LatencyMs=0, AppName=, ExtensionVersion=1.8.6
DurableTask-Core | SchedulingOrchestration ||InstanceId=, ExecutionId=, TargetInstanceId=mySingleton, TargetExecutionId=9f8f2601e2434fcb83b1716ca5780be2, Name=DurableTask.AzureStorage.Tests.StressTests+Approval, TaskEventId=-1, SizeInBytes=0, AppName=, ExtensionVersion=2.5.3
DurableTask-AzureStorage | FetchedInstanceStatus ||Account=devstoreaccount1, TaskHub=test, InstanceId=mySingleton, ExecutionId=, LatencyMs=0, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | SendingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=ExecutionStarted, TaskEventId=-1, InstanceId=, ExecutionId=, SizeInBytes=917, PartitionId=test-control-00, TargetInstanceId=mySingleton, TargetExecutionId=9f8f2601e2434fcb83b1716ca5780be2, SequenceNumber=65, Episode=-1, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | ReceivedMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, InstanceId=mySingleton, ExecutionId=, MessageId=19f2f640-879e-460f-b926-9e79f0d0a33d, Age=116, DequeueCount=1, NextVisibleTime=2021-04-20T00:15:45.0000000, SizeInBytes=600, PartitionId=test-control-00, SequenceNumber=64, Episode=-1, AppName=, ExtensionVersion=1.8.6
DurableTask-Core | Trace ||source=DurableTask-dotnet, instanceId=, executionId=, sessionId=, message=Releasing extended session after 1 batch(es)., info=, eventType=OnProcessWorkItemSession-Release
DurableTask-Core | ProcessWorkItemCompleted ||Dispatcher=TaskOrchestrationDispatcher-69103ada612a4de4b09710af815ae812-0, WorkItemId=mySingleton, AppName=, ExtensionVersion=2.5.3
DurableTask-Core | Info ||source=DurableTask-dotnet, instanceId=, executionId=, sessionId=, message=TaskOrchestrationDispatcher-69103ada612a4de4b09710af815ae812-0: Finished processing workItem mySingleton, info=, eventType=WorkItemDispatcherProcess-End
DurableTask-AzureStorage | ReceivedMessage ||Account=devstoreaccount1, TaskHub=test, EventType=ExecutionStarted, TaskEventId=-1, InstanceId=mySingleton, ExecutionId=9f8f2601e2434fcb83b1716ca5780be2, MessageId=86e4b167-f28c-47e7-8e26-2b0bd059440d, Age=128, DequeueCount=1, NextVisibleTime=2021-04-20T00:15:45.0000000, SizeInBytes=917, PartitionId=test-control-00, SequenceNumber=65, Episode=-1, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | FetchedInstanceStatus ||Account=devstoreaccount1, TaskHub=test, InstanceId=mySingleton, ExecutionId=4d50dffe5840423c9c5f22c195bad06d, LatencyMs=9, AppName=, ExtensionVersion=1.8.6
**DurableTask-AzureStorage | DuplicateMessageDetected ||Account=devstoreaccount1, TaskHub=test, EventType=ExecutionStarted, TaskEventId=-1, MessageId=86e4b167-f28c-47e7-8e26-2b0bd059440d, InstanceId=mySingleton, ExecutionId=9f8f2601e2434fcb83b1716ca5780be2, PartitionId=test-control-00, DequeueCount=1, AppName=, ExtensionVersion=1.8.6**
DurableTask-AzureStorage | DeletingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=ExecutionStarted, TaskEventId=-1, MessageId=86e4b167-f28c-47e7-8e26-2b0bd059440d, InstanceId=mySingleton, ExecutionId=9f8f2601e2434fcb83b1716ca5780be2, PartitionId=test-control-00, SequenceNumber=65, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | WaitingForMoreMessages ||Account=devstoreaccount1, TaskHub=test, PartitionId=test-control-00, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | InstanceStatusUpdate ||Account=devstoreaccount1, TaskHub=test, InstanceId=mySingleton, ExecutionId=9f8f2601e2434fcb83b1716ca5780be2, RuntimeStatus=Pending, Episode=0, LatencyMs=20, AppName=, ExtensionVersion=1.8.6
DurableTask-Core | RaisingEvent ||InstanceId=, ExecutionId=, TargetInstanceId=mySingleton, Name=approval, TaskEventId=-1, SizeInBytes=4, AppName=, ExtensionVersion=2.5.3
DurableTask-AzureStorage | SendingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, InstanceId=, ExecutionId=, SizeInBytes=599, PartitionId=test-control-00, TargetInstanceId=mySingleton, TargetExecutionId=, SequenceNumber=66, Episode=-1, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | FetchedInstanceHistory ||Account=devstoreaccount1, TaskHub=test, InstanceId=mySingleton, ExecutionId=4d50dffe5840423c9c5f22c195bad06d, EventCount=38, Episode=2, RequestCount=1, LatencyMs=23, ETag=W/"datetime'2021-04-20T00%3A10%3A44.85Z'", LastCheckpointTime=4/20/2021 12:10:44 AM, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | ProcessingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, InstanceId=mySingleton, ExecutionId=, MessageId=2c3dbd07-cffc-44bd-81e6-7f2dc296106a, Age=1150, SequenceNumber=33, Episode=-1, IsExtendedSession=False, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | ProcessingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, InstanceId=mySingleton, ExecutionId=, MessageId=cb0c1d11-83db-4e14-9251-dd44bf5e0b7f, Age=1150, SequenceNumber=35, Episode=-1, IsExtendedSession=False, AppName=, ExtensionVersion=1.8.6
DurableTask-AzureStorage | ProcessingMessage ||Account=devstoreaccount1, TaskHub=test, EventType=EventRaised, TaskEventId=-1, InstanceId=mySingleton, ExecutionId=, MessageId=cac54444-8703-49b0-84e4-62daf68d4a04, Age=1150, SequenceNumber=36, Episode=-1, IsExtendedSession=False, AppName=, ExtensionVersion=1.8.6
```

Notice the `ExecutionStarted` event that gets marked as duplicate(MessageId=86e4b167-f28c-47e7-8e26-2b0bd059440d) is meant for ExecutionId = 9f8f2601e2434fcb83b1716ca5780be2.

And yet, the `FetchedInstanceStatus` event that precedes the `DuplicateMessageDetected` event appears to be loading the history for `ExecutionId` = 4d50dffe5840423c9c5f22c195bad06d instead of the target  9f8f2601e2434fcb83b1716ca5780be2.

I'll continue looking into this, but I'm starting to suspect this is not something to be fixed in the `AddMessageToPendingOrchestration` method.

cgillum:
I was able to reproduce the issue locally. It appears to be a flaw in my recently added de-dupe logic. I'll take a closer look and update this PR with a fix.

cgillum:
@davidmrdavid I pushed an update which I think will resolve the problem. I also re-wrote your test to use the same helpers as the other test in the same file. The result is much more compact test logic, so I hope you don't mind. Both stress tests are now passing with my fix in OrchestrationSessionManager.cs

src/DurableTask.Core/Logging/StructuredEventSource.cs
cgillum: BTW, I have a utility that will capture the ETW events so that you can easily just copy them to a file for analysis. No need to actually change any code.

davidmrdavid:
@cgillum, it looks like the `using` declarations preventing the build on the CI, and on my local machine. The concrete error reads: `Severity	Code	Description	Project	File	Line	Suppression State
Error	CS8370	Feature 'using declarations' is not available in C# 7.3. Please use language version 8.0 or greater.	DurableTask.AzureStorage.Tests (net461), DurableTask.AzureStorage.Tests (netcoreapp2.1)	C:\Users\dajusto\source\repos\durabletask\test\DurableTask.AzureStorage.Tests\StressTests.cs	57	Active
`

Do see this error locally as well?

cgillum:
Looks like my enlistment had a local change where I updated the version of C# from 7.3 to 8.0, so it was working for me just fine. I didn't realize this change was not already in `main`. I've pushed that change to this PR, which should hopefully fix the build.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -324,11 +324,13 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
             }
 
-            if (remoteInstance.CreatedTime < msg.TaskMessage.Event.Timestamp)
+            if (remoteInstance.CreatedTime < msg.TaskMessage.Event.Timestamp &&
+                remoteInstance.OrchestrationStatus == OrchestrationStatus.Pending)
             {
-                // The message was inserted after the Instances table was updated, meaning that it's likely.
+                // If the Instances table has a Pending record and the message was inserted after the Instances
+                // table was updated (assuming that the execution IDs are different, which should be established),
+                // then we can know with confidence that this is a redundant message and can be safely discarded.
```
davidmrdavid, line 332: So this seems to be the change that fixed the leftover bug. @cgillum, do you mind explaining how this fixes it? In other words, what used to happen in this if-statement that, with the addition of `&& remoteInstance.OrchestrationStatus == OrchestrationStatus.Pending` no longer occurs? Thanks!

davidmrdavid:
Alright folks, I just tested the latest version of this patch with different levels of concurrency and wasn't able to reproduce the bug. So, as far as I'm concerned, this is ready to merge. I'd appreciate a final LGTM so make sure we're all in the same page. Additionally, @cgillum, I left you a comment above ^ asking about the de-dup patch you've included. I'm interested in understanding why the change you made to that if-statement fixed this behavior, thanks!

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
cgillum: I'll explain in a top-level comment.

cgillum:
> So this seems to be the change that fixed the leftover bug. @cgillum, do you mind explaining how this fixes it? In other words, what used to happen in this if-statement that, with the addition of && remoteInstance.OrchestrationStatus == OrchestrationStatus.Pending no longer occurs? Thanks!

It's important to understand that the fundamental issue at the root of these timing issues is that we cannot atomically write a message to a queue and to a table in a single transaction in Azure Storage when starting a new orchestration instance. We therefore have to have robust logic in place that can deal with all the various problems that can possibly arise with a lack of transactions.

It's hard for me to explain concisely in words, so I'll try to create time-series tables to illustrate how de-dupe is supposed to work, the false positive identified by your test, and the updated logic that fixes the false-positive. Here is an example of a successful de-dupe when singletons are involved:

| Time | Event |
| - | - |
| t1 | Client A schedules a new instance of singleton by enqueueing a new ExecutionStarted message with ExID = 'A' |
| t2 | Client B _also_ schedules a new generation of singleton, but with ExID = 'B' | 
| t3 | Before either client can update the Instances table, the worker receives both ExecutionStarted event and runs through this logic in parallel to see if either message is legit. No row exists yet in the Instances table for this instance, so we know that we just need to wait a bit longer for one of the clients to add it. Defer the messages! |
| t4 | Client A updates the Instances table for singleton with ExID = 'A' and RuntimeStatus = 'Pending' |
| t5 | Client B tries to update the Instances table with ExID = 'B' but fails due to a concurrency violation (eTags FTW) and the caller gets an exception. However, we can't take back the message we already enqueued, so we rely on the worker to figure out how to de-dupe it on its own. |
| t6 | Worker dequeues both ExecutionStarted events a second time and runs through this validation. This time, the message from Client A (ExID = 'A') matches the Instances table (newly added with ExID = 'A'), so this message is allowed through. At the same time, Client B's message (ExID = 'B') is mismatched with the Instances table (ExID = 'A'). This validation code was written with the intent to consider the race condition mentioned previously and was aware of a case where we might be creating a new generation of an existing instance (though that's not part of this scenario). The worker knows that the Instances table was updated with ExID = 'A' _before_ we dequeued the ExID = 'B' message the second time, thus we can safely discard the ExID = 'B' ExecutionStarted message instead of deferring it like we did last time. |

Your test uncovered the following false-positive which involves recreating an existing instance. This scenario was unfortunately not tested in my previous PR:

| Time | Event |
| - | - |
| t1 | Singleton with ExID = 'A' starts and completes, Instances table is updated with RuntimeStatus = 'Completed' |
| t2 | Client B schedules new generation of singleton by enqueueing a new ExecutionStarted message with ExID = 'B' |
| t3 | The worker receives the ExecutionStarted message before Client B can update the instances table. It sees that the execution IDs are different between the message (ExID = 'B') and the Instances table (ExID = 'A'). However, it also sees that the message was received (t3) after the Instances table was updated (t1) and incorrectly decides to discard the message. |

The decision to discard the message at t3 is incorrect because there was only one client trying to start the singleton orchestration! The record in the Instances table from t1 was for an old generation of the singleton. The previous de-dupe logic _tried_ to consider this case, but didn't do so correctly. If the worker would have _also_ checked the runtime status in the Instances table, it would have seen "Completed", which is a signal that the client simply hasn't yet updated the Instances table for the new generation. After adding my fix, the following happens:

| Time | Event |
| - | - |
| t1 | Singleton with ExID = 'A' starts and completes, Instances table is updated with RuntimeStatus = 'Completed' |
| t2 | Client B schedules new generation of singleton by enqueueing a new ExecutionStarted message with ExID = 'B' |
| t3 | The worker receives the ExecutionStarted message before Client B can update the instances table. It sees that the execution IDs are different between the message (ExID = 'B') and the Instances table (ExID = 'A'). It also sees that the message was received (t3) after the Instances table was updated (t1). However, because the Instances table has Status = 'Completed', it knows for sure that the client hasn't yet updated the Instances table and should therefore _defer_ the message to give the client more time to do so. |
| t4 | Client B finally updates the Instances table with ExID = 'B' and RuntimeStatus = 'B' |
| t5 | The worker receives the ExecutionStarted message a second time. This time the message and the Instances table both have ExID = 'B' and the message is allowed to be processed. |

I hope that helps explain the issue and the fix.

cgillum (approved):

PR #532, opened 2021-04-27
Fixes for DurableTask.AzureServiceFabric and DurableTask.Core

shankarsama:
2 changes in this PR:
1. DurableTask.AzureServiceFabric: Disabled instanceId validation to alllow any chars to let the application decide on the chars to be allowed.
2. DurableTask.Core: Fix the issue "Multiple ExecutionCompletedEvent found, potential corruption in state storage"
    Multiple ExecutionCompletedEvents possible when orchestration's termination and completion happens at the same time.

### DIFF
--- a/src/DurableTask.AzureServiceFabric/Extensions.cs
+++ b/src/DurableTask.AzureServiceFabric/Extensions.cs
@@ -45,7 +45,8 @@ namespace DurableTask.AzureServiceFabric
 
         internal static void EnsureValidInstanceId(this string instanceId)
         {
-            // Do not enforce InstanceId validation here. Let the application/user decide.
+            // For now do not enforce InstanceId validation here. Let the application/user decide.
+            // Revist or enable this logic after finding out the chars that are not allowed in the InstanceID
             //if (!instanceId.IsValidInstanceId())
             //{
             //    throw new InvalidInstanceIdException(instanceId);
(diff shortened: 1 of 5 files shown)


src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -480,6 +473,13 @@ public void CompleteOrchestration(string result, string details, OrchestrationSt
             }
             else
             {
+                if (executionTerminated)
+                {
+                    return;
+                }
+
+                executionTerminated = true;
```
cgillum, line 481: nit: the code convention here is to prefix instance variables with `this.`.

cgillum (approved):
LGTM! Just one small nit on code style.

NeetArt (approved):

aresowj:
LGTM.

aresowj (approved):

src/DurableTask.AzureServiceFabric/Extensions.cs
```diff
@@ -45,10 +45,11 @@ public static bool IsValidInstanceId(this string instanceId)
 
         internal static void EnsureValidInstanceId(this string instanceId)
         {
-            if (!instanceId.IsValidInstanceId())
```
lchew7, line 48: 1. I would prefer to remove code, rather than simply commenting it out.
2. Why are we keeping this empty function?  Can't we just remove it altogether?

shankarsama: We need to revisit this logic again. Better to leave it for now. I will leave a comment to expalin this.

PR #536, opened 2021-04-30
Adding provision to perform graceful shutdown before process kill due to max timeout hits

RamjotSingh:
In AzureStorage variant, if maximum number of timeouts are hit per call (5) app gets recycled. This is to bring system back into healthy state. However, this kill can result in issues with requests in flight (as `Environment.FailFast` does not perform graceful shutdown). Adding provision to allow graceful shutdown (e.g. `IHostApplicationLifetime.StopApplication` in AspNetCore).

Graceful shutdown tasks are given 35 seconds to finish (aligning with the 30 second host shutdown that .Net has).

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
@@ -21,6 +21,7 @@ namespace DurableTask.AzureStorage
     using Microsoft.WindowsAzure.Storage.Queue;
     using Microsoft.WindowsAzure.Storage.Table;
     using System.Runtime.Serialization;
+    using System.Threading.Tasks;
 
     /// <summary>
     /// Settings that impact the runtime behavior of the <see cref="AzureStorageOrchestrationService"/>.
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -214,6 +215,13 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public ILoggerFactory LoggerFactory { get; set; } = NoOpLoggerFactory.Instance;
 
+        /// <summary>
+        /// Gets or sets an optional action to be executed before the app is recycled. Reason for shutdown is passed as a string parameter.
+        /// This can be used to perform any pending cleanup tasks or just do a graceful shutdown.
+        /// A wait time of 35 seconds will be given for the task to finish, if the task does not finish in required time, <see cref="Environment.FailFast(string)"/> will be executed.
+        /// </summary>
+        public Func<string, Task> ProcessGracefulShutdownAction { get; set; } = (message) => Task.CompletedTask;
```
cgillum, line 223: I suggest we use a different name for this because I think "ProcessGracefulShutdownAction" gives the impression that we'll run this callback for _any_ graceful shutdown (including normal, routine shutdowns). Perhaps something like `OnUngracefulShutdown` or `OnImminentFailFast` to help the user understand that this is only handling a specific process termination scenario.

Not required, but if you want to make this useful in even more scenarios, consider making this a `Func<string, Task<bool>>` so that the implementer could return a value that cancels the FailFast.

cgillum (commented):
Thanks for submitting this PR! It's simple and I like it. I'm just wondering if we can tweak the public API a bit. See my inline comment.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
RamjotSingh: Done.

src/DurableTask.AzureStorage/TimeoutHandler.cs
```diff
@@ -72,9 +77,32 @@ internal static class TimeoutHandler
...
+                            {
+                                executeFailFast = gracefulShutdownTask.Result;
+                            }
+
+                            if (executeFailFast)
+                            {
+                                TimeoutHandler.ProcessKillAction(message);
+                            }
+
                             return default(T);
```
cgillum, line 106: I appreciate these updates. One final thing - since we may not always FailFast, instead of returning `default(T)`, can we instead throw a TimeoutException?
```suggestion
                            throw new TimeoutException(message);
```

RamjotSingh: Done.

cgillum (approved):
LGTM!

PR #541, opened 2021-05-27
Fixed a message order regression caused by the de-dupe PR (#528)

cgillum:
The key issue being fixed is a problem where execution started messages are executed out of order after https://github.com/Azure/durabletask/pull/528 was merged. None of the existing tests were sensitive to this. However, some of the Durable Functions tests were.

Durable Functions tests to verify:
- [x] WaitForEventAndCallActivity_DroppedEventsTest
- [x] ActorOrchestration_NoWaiting
- [x] BatchedActorOrchestration
- [x] BatchedActorOrchestrationDeleteLastItemAlways
- [x] ParallelBatchedActorOrchestration

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
@@ -216,6 +216,11 @@ namespace DurableTask.AzureStorage
         public ILoggerFactory LoggerFactory { get; set; } = NoOpLoggerFactory.Instance;
 
         /// <summary>
+        /// Gets or sets a value indicating whether to disable the ExecutionStarted de-duplication logic.
+        /// </summary>
+        public bool DisableExecutionStartedDeduplication { get; set; }
+
+        /// <summary>
         /// Gets or sets an optional function to be executed before the app is recycled. Reason for shutdown is passed as a string parameter.
         /// This can be used to perform any pending cleanup tasks or just do a graceful shutdown.
         /// The function returns a <see cref="bool"/>. If 'true' is returned <see cref="Environment.FailFast(string)"/> is executed, if 'false' is returned,
--- a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
@@ -214,23 +214,18 @@ namespace DurableTask.AzureStorage.Messaging
             public int Compare(MessageData x, MessageData y)
             {
                 // Azure Storage is the ultimate authority on the order in which messages were received.
-                if (x.OriginalQueueMessage.InsertionTime < y.OriginalQueueMessage.InsertionTime)
+                // Insertion time only has full second precision, however, so it's not always useful.
+                DateTimeOffset insertionTimeX = x.OriginalQueueMessage.InsertionTime.GetValueOrDefault();
+                DateTimeOffset insertionTimeY = y.OriginalQueueMessage.InsertionTime.GetValueOrDefault();
+                if (insertionTimeX != insertionTimeY)
                 {
-                    return -1;
-                }
-                else if (x.OriginalQueueMessage.InsertionTime > y.OriginalQueueMessage.InsertionTime)
-                {
-                    return 1;
+                    return insertionTimeX.CompareTo(insertionTimeY);
                 }
 
                 // As a tie-breaker, messages will be ordered based on client-side sequence numbers.
-                if (x.SequenceNumber < y.SequenceNumber)
+                if (x.SequenceNumber != y.SequenceNumber)
                 {
-                    return -1;
-                }
-                else if (x.SequenceNumber > y.SequenceNumber)
-                {
-                    return 1;
+                    return x.SequenceNumber.CompareTo(y.SequenceNumber);
                 }
 
                 return 0;
--- a/src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
+++ b/src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
@@ -84,6 +84,12 @@ namespace DurableTask.AzureStorage.Monitoring
             this.highLatencyThreshold = Math.Min(this.maxPollingLatency, 1000);
         }
 
+        /// <summary>
+        /// Gets or sets a value to enable random scale-in (e.g. 10% of recommendations) when queue latencies are low.
+        /// This property should be set to <c>false</c> for unit testing.
+        /// </summary>
+        public bool EnableRandomScaleDownOnLowLatency { get; set; } = true;
+
         internal virtual int PartitionCount => this.currentPartitionCount;
 
         internal List<QueueMetricHistory> ControlQueueLatencies => this.controlQueueLatencies;
@@ -412,7 +418,7 @@ namespace DurableTask.AzureStorage.Monitoring
                 // We also want to avoid scaling in unnecessarily when we've reached optimal scale-out. To balance these
                 // goals, we check for low latencies and vote to scale down 10% of the time when we see this. The thought is
                 // that it's a slow scale-in that will get automatically corrected once latencies start increasing again.
-                bool tryRandomScaleDown = Random.Next(10) == 0;
+                bool tryRandomScaleDown = this.EnableRandomScaleDownOnLowLatency && Random.Next(10) == 0;
                 if (tryRandomScaleDown &&
                     controlQueueLatencyHistory.TrueForAll(IsLowLatency) &&
                     workItemQueueLatencyHistory.TrueForAll(latency => latency < LowLatencyThreshold))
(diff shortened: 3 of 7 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -215,6 +215,11 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public ILoggerFactory LoggerFactory { get; set; } = NoOpLoggerFactory.Instance;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to disable the ExecutionStarted de-duplication logic.
+        /// </summary>
+        public bool DisableExecutionStartedDeduplication { get; set; }
```
cgillum, line 221: @ConnorMcMahon this setting can be set to `true` to disable the new de-dupe code path if we find it's causing significant performance regressions. Naturally, we'll need to expose it in the extension settings in order to allow customers to configure it.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -214,23 +214,18 @@ class MessageOrderingComparer : IComparer<MessageData>
...
                 // As a tie-breaker, messages will be ordered based on client-side sequence numbers.
-                if (x.SequenceNumber < y.SequenceNumber)
-                {
-                    return -1;
-                }
-                else if (x.SequenceNumber > y.SequenceNumber)
+                if (x.SequenceNumber != y.SequenceNumber)
                 {
-                    return 1;
+                    return x.SequenceNumber.CompareTo(y.SequenceNumber);
```
cgillum, line 228: I think we've had this comparison code do the exact opposite of what we want (@davidmrdavid this was the latent bug I was referring to). It turns out that this wasn't the problem that caused test failures, but I wanted to fix it anyway. With this new logic, I don't have to remember what `1` and `-1` actually means when implementing `IComparer<T>`.

src/DurableTask.AzureStorage/Monitoring/DisconnectedPerformanceMonitor.cs
```diff
@@ -84,6 +84,12 @@ public DisconnectedPerformanceMonitor(string storageConnectionString, string tas
             this.highLatencyThreshold = Math.Min(this.maxPollingLatency, 1000);
         }
 
+        /// <summary>
+        /// Gets or sets a value to enable random scale-in (e.g. 10% of recommendations) when queue latencies are low.
+        /// This property should be set to <c>false</c> for unit testing.
+        /// </summary>
+        public bool EnableRandomScaleDownOnLowLatency { get; set; } = true;
```
cgillum, line 91: So it turns out that the earlier iteration of this PR failed because one of the tests failed 3 times in a row due to a 10% random scale-in vote behavior. I think this means I hit a 1/1000 probability of failure. 😅 In order to ensure the scale tests can run deterministically, I decided to add this property to disable the random behavior by default _for unit tests_ (see further down where I actually set it). This should help make our CI a tiny bit more reliable. I made it public in case we or a DTFx user wanted to use this class and disable the random scale-in behavior, which might be a bit controversial.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -185,25 +188,29 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+        /// hasn't been written yet (this happens asynchronously and there is no guaranteed order). Meanwhile, this method will
+        /// return the list of filtered messages.
         /// </summary>
         /// <param name="controlQueue">A reference to the control queue from which these messages were dequeued.</param>
         /// <param name="messages">The full set of messages recently dequeued.</param>
         /// <param name="traceActivityId">The trace activity ID to use when writing traces.</param>
         /// <param name="cancellationToken">Cancellation token.</param>
         /// <returns>Returns the list of non-ExecutionStarted messages. This may be an empty list.</returns>
-        IReadOnlyList<MessageData> FilterOutExecutionStartedForDedupeValidation(
+        async Task<IEnumerable<MessageData>> DedupeExecutionStartedMessagesAsync(
```
cgillum, line 201: Previously this method ran as a background task and scheduled ExecutionStarted events in parallel with other events in the same batch after we did the de-dupe validation. I wrote it this way to minimize any latency impact on our internal control-queue dequeue loop. However, this optimization broke tests that sent an ExecutionStarted event immediately followed by EventRaised events because the EventRaised events were now being scheduled and processed _before_ the ExecutionStarted events due to the async de-dupe validation delay. The EventRaised events were discarded because no such instance existed (the ExecutionStarted event is the one that creates the instance, but it was held up in de-dupe validation). Now, we make _all_ messages in the dequeue batch wait for de-dupe validation to complete before scheduling them for execution. This ensures that the EventRaised events in the same batch don't get discarded because the instance is not (yet) found.

I haven't yet convinced myself this is 100% bullet proof, but I'm about 98% confident that it fully addresses the race condition I mentioned. The remaining 2% of doubt is around whether there's a possible scenario where an ExecutionStarted and an EventRaised event from two different dequeue batches could race with each other. As I write this, I think that shouldn't be an issue because we only have one logical dequeue thread per queue, which means all messages should be received and processed in-order, always.

**IMPORTANT**: This de-dupe logic now runs in the main dequeue loop instead of in a background thread. That means we will dequeue messages from the control queue more slowly when there are ExecutionStarted messages in the queue because of the table storage lookups we do here. This may impact the performance of the _ManySequences_ test where we start tons of instances consistently. If this is problematic, the de-dupe logic can be disabled by exposing the new `DisableExecutionStartedDeduplication` setting and setting it to `false` by default. Hopefully that won't be necessary.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -215,99 +222,95 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
+            IDictionary<string, OrchestrationState> remoteOrchestrationsById =
+                instances.ToDictionary(o => o.OrchestrationInstance.InstanceId);
+
+            foreach (MessageData message in executionStartedMessages)
+            {
+                OrchestrationInstance localInstance = message.TaskMessage.OrchestrationInstance;
+                if (remoteOrchestrationsById.TryGetValue(localInstance.InstanceId, out OrchestrationState remoteInstance) &&
+                    string.Equals(localInstance.ExecutionId, remoteInstance.OrchestrationInstance.ExecutionId, StringComparison.OrdinalIgnoreCase))
+                {
+                    // Happy path: The message matches the table status. Allow it to run.
```
cgillum, line 253: The logic in this `foreach` loop all remains unchanged. The only difference is this line. Previously we added the message to a `messagesToKeep` in the happy path. That list is now gone because we return all the messages that we want to keep from this method. Previously, we separated all the ExecutionStarted messages from the non-ExecutionStarted messages and scheduled them independently. Now this method just returns all messages that we want to keep for execution instead of directly scheduling anything. That enables us to ensure that message order is preserved.

cgillum (commented):
I added some PR comments to provide more detailed context on these changes.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
davidmrdavid: It appears this hasn't been expose yet, I suppose we can wait until after our performance tests to determine if exposing this is necessary

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
davidmrdavid: > However, this optimization broke tests that sent an ExecutionStarted event immediately followed by EventRaised events because the EventRaised events were now being scheduled and processed before the ExecutionStarted events due to the async de-dupe validation delay. The EventRaised events were discarded because no such instance existed (the ExecutionStarted event is the one that creates the instance, but it was held up in de-dupe validation).

@ConnorMcMahon, so what goes wrong in the scenario described above? Is it that the `EventRaised` message is discarded forever?

davidmrdavid (approved):
LGTM. @ConnorMcMahon, I left a question for you as a response to one of Chris' comments, please give it a look if you have a minute. If it's too lengthy of a response, we can discuss it later

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
ConnorMcMahon: Exactly David. That is why our old-school "actor" orchestrations that only receive events were the bugs that manifested, because they always receive an EventRaised message in the same batch as an ExecutionStarted message, so they hit this discard logic consistently.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
ConnorMcMahon: This defaults to false, which is fine. We will run our perf tests, and if we determine it's a big hit, we can add a host.json settings that allows customers to set this to true.

ConnorMcMahon (approved):
LGTM!

PR #544, opened 2021-06-02
DT.AzureStorage: Fix message loss bug in ContinueAsNew scenarios

### LINKED ISSUE #543: Possible message loss when messages from multiple generations are processed concurrently
## Summary
There is a possible message loss issue caused by a race condition in the Azure Storage provider (**DurableTask.AzureStorage**) involving ContinueAsNew and durable timers. The message loss can result in orchestrations getting permanently stuck.

## Technical details
The problem can occur when we receive two or more messages for two or more generations of the same orchestration instance at around the same time (an orchestration instances changes generations every time you call the ContinueAsNew() API). For example, you receive a TimerFired message from an older generation of your orchestration (e.g. a 30-minute timer) around the same time as a TaskCompleted message from the current generation. We're supposed to discard the old TimerFired message and keep the TaskCompleted message. However, it appears there is a bug in our message sorting logic that causes both the TimerFired and TaskCompleted messages to both be deleted.

The problematic code appears to be [here](https://github.com/Azure/durabletask/blob/4dc8aacb88f62ac43fdae7d173e8ae1152afe273/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs#L404-L411).

## Workaround
No known workarounds at this time.

## Mitigation
You can mitigate the stuck orchestration by manually re-creating the stuck instance. Note that you might be required to first terminate the stuck orchestration before it can be recreated.

cgillum:
Fixes #543

We have some logic that incorrectly deletes messages when we see messages from multiple generations at the same time. The logic is both wrong and seems to be unnecessary, so this PR mostly just removes code and adjusts the comments.

A couple things I want to verify before merging:

- [x] Existing test cases continue to pass across both DTFx and Durable Functions
- [x] Invalid timer messages still get discarded appropriately in ContinueAsNew scenarios

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
@@ -372,7 +372,6 @@ namespace DurableTask.AzureStorage
             //  3. Do we need to add messages to a currently executing orchestration?
             lock (this.messageAndSessionLock)
             {
-
                 var existingSessionMessages = new Dictionary<OrchestrationSession, List<MessageData>>();
 
                 foreach (MessageData data in queueMessages)
@@ -382,12 +381,13 @@ namespace DurableTask.AzureStorage
                     string instanceId = data.TaskMessage.OrchestrationInstance.InstanceId;
                     string executionId = data.TaskMessage.OrchestrationInstance.ExecutionId;
 
-                    // Check if the target orchestraion is loaded in memory
-                    if (this.activeOrchestrationSessions.TryGetValue(instanceId, out OrchestrationSession session))
+                    // If the target orchestration is already in memory, we can potentially add the message to the session directly
+                    // rather than adding it to the pending list. This behavior applies primarily when extended sessions are enabled.
+                    // We can't do this for ExecutionStarted messages - those must *always* go to the pending list since they are for
+                    // creating entirely new orchestration instances.
+                    if (data.TaskMessage.Event.EventType != EventType.ExecutionStarted &&
+                        this.activeOrchestrationSessions.TryGetValue(instanceId, out OrchestrationSession session))
                     {
-                        // If the target orchestration is already in memory, we add the message to the session directly,
-                        // rather than adding it to the linked list.
-
                         // A null executionId value means that this is a management operation, like RaiseEvent or Terminate, which
                         // should be delivered to the current session.
                         if (executionId == null || session.Instance.ExecutionId == executionId)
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -400,28 +400,18 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
 
                             pendingMessages.Add(data);
-                        }
-                        else if (session.RuntimeState.ExecutionStartedEvent == null || data.TaskMessage.Event.Timestamp < session.RuntimeState.CreatedTime)
-                        {
-                            // If the current generation of this instance hasn't started or the current 
-                            // generation's `CreatedTime` is newer than the message timestamp indicates that
-                            // this message was created for a previous generation of this instance.
-                            // This is common for canceled timer fired events in ContinueAsNew scenarios.
-                            session.DiscardMessage(data);
```
cgillum, line 410: The logic here on line 404 was completely wrong - particularly the null check. What was happening is that an old timer message would arrive and load a session into memory. Because the timer message was for an older generation, there wouldn't be any runtime state associated with its session (`session.RuntimeState.ExecutionStartedEvent == null`). Eventually, we would've deleted this old timer message as part of another thread. However, after we load this "doomed" session and before we delete the associated timer message, there is a window of time where _another_ message would arrive that was completely legitimate (e.g. a TaskCompleted message) and execute the above lines of code. The logic above would cause us to discard the legit TaskCompleted message simply because some other empty orchestration session was loaded a few milliseconds earlier, which makes no sense. In actuality, the null-check was added only because `session.RuntimeState.CreatedTime` will fail with an assert if `ExecutionStartedEvent` was null (a design decision that predates me). It was never meant to check specifically whether the session is for a non-existent orchestration generation.

One way to fix this issue might have been to change the conditional to the following:

```csharp
else if (session.RuntimeState.ExecutionStartedEvent != null && data.TaskMessage.Event.Timestamp < session.RuntimeState.CreatedTime)
```

That would have actually matched the intended behavior. However, it occurred to me that this logic isn't actually needed at all. If we simply allow the new message to bet added to the linked-list of pending batches below, then we'd keep or delete the message appropriately anyways. It's a little more work because it requires us to attempt to load history for a message that we're pretty sure is invalid, but it simplifies the overall message processing logic, making it easier to understand and reason about.

davidmrdavid (approved):
This looks good to me, and a big +1 to re-using our existing logic, instead of duplicating it in various places.

davidmrdavid (commented):
Actually, is there any chance we could add a test to this? Just to prevent future regressions. I also just realized this PR is still in "draft" mode, sorry for the early review :)

cgillum:
I've taken it out of draft mode now that I've finished manually validating the behavior.

davidmrdavid:
@cgillum, would it be possible to add a unit test for this?

cgillum:
@davidmrdavid it's possible to add a _stress_ test for it (not a true unit test), but I hesitated to do so since it's a hard issue to hit and would add a non-trivial amount of time to our CI test suite. I assumed manual testing plus verifying that we aren't regressing anything would be good enough, especially since I'm really just removing logic rather than adding logic.

davidmrdavid:
That makes sense, thanks for clarifying. Indeed, these race conditions are hard to replicate.

davidmrdavid (approved):
It's got my approval, but let's wait to hear from @ConnorMcMahon before merging. Thank you for this PR!

ConnorMcMahon (approved):
LGTM!

PR #546, opened 2021-06-09
Changed DispatchMiddlewareContext to be public

saguiitay:


cgillum (approved):
Thanks for submitting a PR!

cgillum:
Looks like the build failed because of missing XML comments on the new public constructor. Would you mind fixing this? Removing the constructor altogether should also work.

PR #551, opened 2021-06-10
Updated AppLeaseManager to use new storage classes

amdeel:
Updated AppLeaseManager to use new storage classes. Added storage APIs used in AppLeaseManager to BlobContainer.

An extension of this PR: https://github.com/Azure/durabletask/pull/534


src/DurableTask.AzureStorage/TimeoutHandler.cs
```diff
@@ -56,6 +58,10 @@ internal static class TimeoutHandler
                     Task timeoutTask = Task.Delay(DefaultTimeout, cts.Token);
                     Task<T> operationTask = operation(context, cts.Token);
 
+                    if (stats != null)
+                    {
+                        stats.StorageRequests.Increment();
```
cgillum, line 63: Could we instead add this to AzureStorageClient? That seems like a more natural place to keep track of storage requests.

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -19,79 +19,51 @@ namespace DurableTask.AzureStorage.Partitioning
...
         {
-            try
-            {
-                return await this.taskHubContainer.ExistsAsync();
-            }
-            finally
-            {
-                this.stats.StorageRequests.Increment();
-            }
+            return await taskHubContainer.ExistsAsync();
```
cgillum, line 60: Looks like we don't require `await` in this method. Can you remove `async` from the method signature and remove `await` here? That will slightly improve performance (`await` isn't free) and will improve the debugging experience if there are exceptions.

cgillum (approved):
I really appreciate the code reduction and the simplified abstractions that you've introduced in this PR. Just a few non-blocking nits.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -286,50 +271,22 @@ internal string GetNewLargeMessageBlobName(MessageData message)
...
-                    });
-                
-                stats.StorageRequests.Increment();
-                foreach (IListBlobItem blobListItem in segment.Results)
-                {
-                    var cloudBlockBlob = blobListItem as CloudBlockBlob;
-                    CloudBlockBlob blob = this.cloudBlobContainer.GetBlockBlobReference(cloudBlockBlob?.Name);
-                    blobForDeletionTaskList.Add(blob.DeleteIfExistsAsync());
-                }
+            IEnumerable<Blob> blobList = await this.blobContainer.ListBlobsAsync(sanitizedInstanceId);
```
ConnorMcMahon, line 281: The move from grabbing a segment of the lists and then deleting that segment of blobs concurrently to grabbing all of them at once and then deleting them all concurrently may cause some issues.

I know we are looking at implementing global storage concurrency throttling which would address this concern, but I just wanted to potentially highlight that.

This may not actually be a behavior change though, as I do see that we have set maxResults to `null` in the old code path, so take this with a grain of salt.

src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
```diff
@@ -19,79 +19,51 @@ namespace DurableTask.AzureStorage.Partitioning
...
         readonly TimeSpan leaseInterval;
-        readonly CloudBlobClient storageClient;
-        readonly BlobRequestOptions renewRequestOptions;
-        readonly AzureStorageOrchestrationServiceStats stats;
 
-        CloudBlobContainer taskHubContainer;
-        CloudBlobDirectory leaseDirectory;
-        CloudBlockBlob taskHubInfoBlob;
+        BlobContainer taskHubContainer;
+        private string blobDirectoryName;
```
ConnorMcMahon, line 38: Let's make all of this private or none of them private for consistency.

src/DurableTask.AzureStorage/Storage/Blob.cs
```diff
@@ -0,0 +1,144 @@
+﻿//  ----------------------------------------------------------------------------------
```
ConnorMcMahon, line 1: any idea why these classes are showing up in this PR? It looks like they are in the base branch...

src/DurableTask.AzureStorage/TimeoutHandler.cs
ConnorMcMahon: In theory that would be great, but this has potential retries for read-only operations that timeout, so putting  it in this loop keeps track per attempt.

ConnorMcMahon (commented):
Mostly looking good!

src/DurableTask.AzureStorage/MessageManager.cs
davidmrdavid: Can you please expand on what these potential issues are?

ConnorMcMahon: If we have to delete 1000 large blobs, then trying to do 1000 delete operations concurrently could violate the maximum number of outbound connections allowed on the consumption plan.

src/DurableTask.AzureStorage/TimeoutHandler.cs
davidmrdavid: Perhaps I'm missing something, but even if a storage operation times out, shouldn't we be able to keep track of storage requests **in the StorageClient** so long as the counter is incremented before performing the operation?

ConnorMcMahon: Not really. The storage client passes in a delegate that performs the storage operation in this loop. It is not actually invoking the storage request, this loop is.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -161,14 +149,14 @@ public async Task<MessageData> DeserializeQueueMessageAsync(CloudQueueMessage qu
             return envelope;
         }
 
-        internal Task CompressAndUploadAsBytesAsync(byte[] payloadBuffer, string blobName)
+        public Task CompressAndUploadAsBytesAsync(byte[] payloadBuffer, string blobName)
```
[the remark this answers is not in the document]
amdeel: This class is internal, so public methods are just internal methods. I changed them all to public, instead of having some internal, for consistency.

src/DurableTask.AzureStorage/Storage/Blob.cs
amdeel: I think because I did a squash merge instead of just a merge?

ConnorMcMahon (approved):
LGTM.

PR #560, opened 2021-07-01
Fix null ref in instance completion trace

ConnorMcMahon:
A recent change to log only the history dump instead of the history
itself resulted in a null-ref when running Release code. This is because
in release builds, the Events field on OrchestrationRuntimeStateDump is
null.

This PR fixes that by:

1. Logging the entire dump instead of just Events so we at least get the
   number of Events in the instance completion events trace.
2. Including an empty list of Events and NewEvents on the dump in
   Release mode so we don't run into future NRE.


cgillum (approved):
LGTM!

PR #564, opened 2021-07-03
Support for dynamic concurrency throttles

cgillum:
This PR adds a new `WorkItemThrottler` abstraction that can be used to throttle the fetching of orchestrator and activity executions.

The original motivation for this is to support the dynamic concurrency feature of Azure Functions, but it can be used in non-Azure Functions scenarios as well.

This PR also adds incremental support for nullable reference types, something we're trying to do as we touch various parts of the code.

### DIFF
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -5,9 +5,12 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageId>Microsoft.Azure.DurableTask.Core</PackageId>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <!--NuGet licenseUrl and PackageIconUrl/iconUrl deprecation. -->
     <NoWarn>NU5125;NU5048</NoWarn>
+    <AssemblyVersion>2.6.0</AssemblyVersion>
+    <FileVersion>2.6.0</FileVersion>
+    <Version>2.6.0</Version>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -76,6 +76,15 @@ namespace DurableTask.Core
         }
 
         /// <summary>
+        /// Sets a work-item throttler to use for this <see cref="TaskOrchestrationDispatcher"/>.
+        /// </summary>
+        /// <param name="workItemThrottler">The <see cref="WorkItemThrottler"/> implementation to assign.</param>
+        public void SetWorkItemThrottler(WorkItemThrottler workItemThrottler)
+        {
+            this.dispatcher.WorkItemThrottler = workItemThrottler;
+        }
+
+        /// <summary>
         /// Starts the dispatcher to start getting and processing orchestration events
         /// </summary>
         public async Task StartAsync()
--- a/src/DurableTask.Core/WorkItemDispatcher.cs
+++ b/src/DurableTask.Core/WorkItemDispatcher.cs
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
 namespace DurableTask.Core
 {
     using System;
@@ -46,8 +46,8 @@ namespace DurableTask.Core
         volatile int delayOverrideSecs;
         volatile int activeFetchers;
         bool isStarted;
-        SemaphoreSlim concurrencyLock;
-        CancellationTokenSource shutdownCancellationTokenSource;
+        SemaphoreSlim? concurrencyLock;
+        CancellationTokenSource? shutdownCancellationTokenSource;
 
         /// <summary>
         /// Gets or sets the maximum concurrent work items
@@ -59,6 +59,11 @@ namespace DurableTask.Core
         /// </summary>
         public int DispatcherCount { get; set; } = DefaultDispatcherCount;
 
+        /// <summary>
+        /// Gets or sets a work-item throttler implementation to throttle work-item fetching
+        /// </summary>
+        public WorkItemThrottler? WorkItemThrottler { get; set; }
+
         readonly Func<T, string> workItemIdentifier;
 
         Func<TimeSpan, CancellationToken, Task<T>> FetchWorkItem { get; }
@@ -68,12 +73,12 @@ namespace DurableTask.Core
         /// <summary>
         /// Method to execute for safely releasing a work item
         /// </summary>
-        public Func<T, Task> SafeReleaseWorkItem;
+        public Func<T, Task>? SafeReleaseWorkItem;
 
         /// <summary>
         /// Method to execute for aborting a work item
         /// </summary>
-        public Func<T, Task> AbortWorkItem;
+        public Func<T, Task>? AbortWorkItem;
 
         /// <summary>
         /// Method to get a delay to wait after a fetch exception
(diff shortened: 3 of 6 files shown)


src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -5,9 +5,12 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageId>Microsoft.Azure.DurableTask.Core</PackageId>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
cgillum, line 8: Committed this on accident. Will revert.

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -5,9 +5,12 @@
...
     <!--NuGet licenseUrl and PackageIconUrl/iconUrl deprecation. -->
     <NoWarn>NU5125;NU5048</NoWarn>
+    <AssemblyVersion>2.6.0</AssemblyVersion>
+    <FileVersion>2.6.0</FileVersion>
+    <Version>2.6.0</Version>
```
cgillum, line 13: This was also committed on accident and will be reverted. Or, I might decide to leave it since there's no point in having the assembly and package versions be shared across all projects.

tools/DurableTask.props
```diff
@@ -5,7 +5,7 @@
     <DebugType Condition="'$(Configuration)'=='Release'">pdbonly</DebugType>
     <DebugSymbols>True</DebugSymbols>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
```
cgillum, line 8: This was required to support nullable reference types for non-constrained generic types (e.g. `T?`).

src/DurableTask.Core/WorkItemDispatcher.cs
```diff
@@ -259,7 +269,7 @@ async Task DispatchAsync(WorkItemDispatcherContext context)
                         this.GetFormattedLog(dispatcherId, $"Starting fetch with timeout of {DefaultReceiveTimeout} ({this.concurrentWorkItemCount}/{this.MaxConcurrentWorkItems} max)"));
 
                     Stopwatch timer = Stopwatch.StartNew();
-                    workItem = await this.FetchWorkItem(DefaultReceiveTimeout, this.shutdownCancellationTokenSource.Token);
+                    workItem = await this.FetchWorkItem(DefaultReceiveTimeout, this.shutdownCancellationTokenSource?.Token ?? default);
```
ConnorMcMahon, line 272: I understand the appeal of putting this in DurableTask.Core, but then we don't end up solving one of our bigger problems which is overaggressive prefetching, which happens on a background thread in DurableTask.AzureStorage.

We could potentially rethink our prefetching strategy in DurableTask.AzureStorage to make sure it would take advantage of the throttling done here, but I want to call that out as a limit right now.

tools/DurableTask.props
ConnorMcMahon: Does this have any ramifications on developer experience, since I know starting with C# 8 certain features are framework locked. I'm assuming the compiler will still yell at us appropriately considering our targets?

cgillum: Yes, the build will fail if we try to use any features not allowed by the target framework. In some cases it is just the net461 build that failures. If there's a feature that requires netstandard2.1 or higher (like async dispose) then both targets will fail to compile. This hasn't been a problem so far and I don't expect it will be a problem now.

src/DurableTask.Core/WorkItemDispatcher.cs
cgillum: The advantages of putting this in DT.Core is that it works for all backends, including MSSQL and Netherite. It also solves concurrency issues with activity functions. I consider prefetching to be a serious but more niche case (it's a bigger problem for us because it's harder for users to understand or control). My hope is that we can later expose this abstraction to storage providers so they can go through it for prefetching as well. If you agree that the abstraction as designed right now doesn't fundamentally prevent us from expanding it to cover prefetching later, then I suggest we go ahead and take this and plan to cover prefetching in another PR.

ConnorMcMahon: Yeah, I think the abstraction is general enough that individual backends can use the same WorkItemThrottler without interfering with the core logic.

That being said, I think our concurrency logic for prefetch will potentially want a closer abstraction to ConcurrencyManager to decide how many messages to fetch (see [here](https://github.com/Azure/azure-sdk-for-net/pull/22283/files#diff-5f34ee8ab675519fce43d997901d6cd9b6ddf5b0e2448d9c94017ab71a79d54eR307) for how the queue trigger handles it). Because of that, I do think WorkItemThrottler may not be a powerful enough abstraction for all of our use cases down the line. As long as we can in theory add virtual methods to add that further functionality, I am not opposed to merging in the current state.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -75,6 +75,15 @@ public class TaskOrchestrationDispatcher
             this.concurrentSessionLock = new NonBlockingCountdownLock(maxConcurrentSessions);
         }
 
+        /// <summary>
+        /// Sets a work-item throttler to use for this <see cref="TaskOrchestrationDispatcher"/>.
+        /// </summary>
+        /// <param name="workItemThrottler">The <see cref="WorkItemThrottler"/> implementation to assign.</param>
+        public void SetWorkItemThrottler(WorkItemThrottler workItemThrottler)
```
ConnorMcMahon, line 82: These are on internal only objects. Won't we need some equivalent method on a public facing type like TaskHubWorker to actually make this usable?

cgillum: Apologies if I'm misunderstanding, but these are all public API changes. Here is the code that consumes this in the Durable extension repo (haven't created this PR yet):

```csharp
if (this.concurrencyManager != null)
{
    this.taskHubWorker.TaskOrchestrationDispatcher.SetWorkItemThrottler(
        new DurableFunctionsWorkItemThrottler(
            this.concurrencyManager,
            $"{this.Options.HubName}-orchestrations"));
    this.taskHubWorker.TaskActivityDispatcher.SetWorkItemThrottler(
        new DurableFunctionsWorkItemThrottler(
            this.concurrencyManager,
            $"{this.Options.HubName}-activities"));
}
```

...and my implementation of the `WorkItemThrottler` abstract base class.
```csharp
private sealed class DurableFunctionsWorkItemThrottler : WorkItemThrottler
{
    private readonly ConcurrencyManager concurrencyManager;
    private readonly string throttleId;

    public DurableFunctionsWorkItemThrottler(ConcurrencyManager concurrencyManager, string throttleId)
    {
        this.concurrencyManager = concurrencyManager;
        this.throttleId = throttleId;
    }

    public override async Task WaitIfThrottledAsync()
    {
        if (!this.concurrencyManager.Enabled)
        {
            return;
        }

        ConcurrencyStatus status = this.concurrencyManager.GetStatus(this.throttleId);
        while (status.AvailableInvocationCount <= 0)
        {
            TimeSpan delay = status.NextStatusDelay;
            if (delay <= TimeSpan.Zero)
            {
                // Not sure if the delay will ever be zero or less, but protect ourselves just in case.
                delay = TimeSpan.FromSeconds(1);
            }

            await Task.Delay(delay);

            status = this.concurrencyManager.GetStatus(this.throttleId);
        }
    }
}
```

ConnorMcMahon: Woops, I didn't scroll up far enough. The _constructor_ is internal, but the class is public. Disregard this thread.

src/DurableTask.Core/WorkItemDispatcher.cs
davidmrdavid: I also don't mind adding a pre-fetching abstraction later, so long as a tracking issue is created before (or shortly after) merging :)

cgillum:
FYI @mathewc on this PR, which is the basis for the dynamic concurrency work for Durable Functions.

michaelplavnik:
@cgillum This PR is two years old. Have your team come to a conclusion of what to do about throttling of fetch ?

Issue #565, opened 2021-07-05
Support CancellationTokens for cancelling running operations


RamjotSingh:
Being a public SDK, all externally visible interfaces should support operation cancellation (unless it will send the system into a broken state). We have at times seen it take a long time for blob upload causing us at times to cancel it (because we aren't sure at that point if it is simply stuck or just being insanely slow). Retrying to queue an orchestration at this point at times results into having multiple orchestrations . This would at times be fine, but if the orchestration has been picked up for processing, by the time orchestration is re-queued, operations might already have been executed.

This is just one example. There are other examples like shutting down of service etc where various operations might need to be cancellled.

ConnorMcMahon:
@RamjotSingh, can you elaborate on if the Blob Upload issue is something that your code is doing, or is it something being done by one of the backend implementations (like the Azure Storage implementation)?

If it's being done in the Azure Storage implementation, there is a known issue with the version of the Storage SDK we are using that has deadlocks, regardless of the usage of cancellation tokens. We have attempted to work around this with retries + shutdowns, but the eventual plan is to upgrade to the latest storage SDK. We hope to do this work in the next few months.

If it's being done by your code, then you have total control over the usage of cancellation tokens. Make sure you are only doing IO inside of Activities and not Orchestrations, as Orchestrations hijack the thread pooling and any async/await calls made that are not composed of Tasks returned by the OrchestrationContext can cause the orchestration to hang.

RamjotSingh:
@ConnorMcMahon It is being done by DTF code as part of the upload of large input blobs. So we do not have control on it to cancel it.

ConnorMcMahon:
Good to know. What version of DurableTask.AzureStorage are you using? I'll check to see if we have properly put our deadlock mitigation attempts in place around those calls.

RamjotSingh:
<PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.8.5" />

ConnorMcMahon:
Sure enough, we don't wrap [the calls](https://github.com/Azure/durabletask/blob/urabletask.azurestorage-v1.8.5/src/DurableTask.AzureStorage/MessageManager.cs#L277) in the MessageManager class that handles large messages with our deadlock mitigation code.

[...]

PR #566, opened 2021-07-05
[Azure Storage] Support cancellation of operations during Creation phase

RamjotSingh:
Fixes one small part of the issue brought up in https://github.com/Azure/durabletask/issues/565

Adds support for cancellation of queuing of orchestration for Azure Storage based orchestrations.

### DIFF
--- a/src/DurableTask.Core/IOrchestrationServiceClient.cs
+++ b/src/DurableTask.Core/IOrchestrationServiceClient.cs
@@ -28,18 +28,20 @@ namespace DurableTask.Core
         /// Creates a new orchestration
         /// </summary>
         /// <param name="creationMessage">Orchestration creation message</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
         /// <exception cref="OrchestrationAlreadyExistsException">Will throw an OrchestrationAlreadyExistsException exception If any orchestration with the same instance Id exists in the instance store.</exception>
         /// <returns></returns>
-        Task CreateTaskOrchestrationAsync(TaskMessage creationMessage);
+        Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, CancellationToken cancellationToken = default(CancellationToken));
 
         /// <summary>
         /// Creates a new orchestration and specifies a subset of states which should be de duplicated on in the client side
         /// </summary>
         /// <param name="creationMessage">Orchestration creation message</param>
         /// <param name="dedupeStatuses">States of previous orchestration executions to be considered while de-duping new orchestrations on the client</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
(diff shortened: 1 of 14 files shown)


src/DurableTask.Core/IOrchestrationServiceClient.cs
```diff
@@ -28,18 +28,20 @@ public interface IOrchestrationServiceClient
...
         /// <summary>
         /// Creates a new orchestration and specifies a subset of states which should be de duplicated on in the client side
         /// </summary>
         /// <param name="creationMessage">Orchestration creation message</param>
         /// <param name="dedupeStatuses">States of previous orchestration executions to be considered while de-duping new orchestrations on the client</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
         /// <exception cref="OrchestrationAlreadyExistsException">Will throw an OrchestrationAlreadyExistsException exception If any orchestration with the same instance Id exists in the instance store and it has a status specified in dedupeStatuses.</exception>
         /// <returns></returns>
-        Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[] dedupeStatuses);
+        Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[] dedupeStatuses, CancellationToken cancellationToken = default(CancellationToken));
```
cgillum, line 44: Unfortunately, this is a breaking change in terms of binary compatibility. I think semver requires this to be done as part of introducing a new major version. I'm not sure we're ready to do that right now. @ConnorMcMahon, thoughts?

RamjotSingh: Ah I feared as much :(.

ConnorMcMahon: I'm open to revisiting this down the line, but I agree this is a breaking change that we aren't ready to take on (at least in the next few months).

@RamjotSingh I will reply to your issue filed to see if we can help you address your issues without making a breaking change.

RamjotSingh: If we are just concerned with IL binding creating issues, I suppose we can add the new method as a new one (one requiring cancellation token to be passed always). I don't remember if IL is ok with that or not. One place I see it creating issues would be if people have created their own implementations of ServiceClient.

ConnorMcMahon: In general, I don't know if cancellation tokens even make sense in the context of long-lived orchestrations that can run across multiple VMs.

In general, we support orchestration "termination" for this use case.

jviau: @ConnorMcMahon - I agree that it doesn't make that much sense in terms of cancelling the orchestration. But maybe this makes sense for cancelling the _enqueuing_ of the orchestration, if it is not already started?

PR #568, opened 2021-07-09
Change single multi-instance table query to many parallel queries

ConnorMcMahon:
In an attempt to reduce the latency + storage transaction impact of the
new deduplication of ExecutionStarted events, we tried to query the
status of multiple instance ids in a single query. Unfortunately, for
certain scenarios, this resulted in very large latency requests.

In order to reduce latency, we will now make an individual request for
each instance id (we do make these requests in parallel though).

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -651,87 +651,12 @@ namespace DurableTask.AzureStorage.Tracking
                 return Array.Empty<OrchestrationState>();
             }
 
-            // Example:
-            // "PartitionKey eq '02717dd1-2eda-47a0-9372-62a1bfa47f46' or PartitionKey eq '0d1c021f-63a3-4678-9495-63025491a091' or PartitionKey eq '0fd59a43-1925-456b-99a2-7c13ac66ce2b'"
-            var queryBuilder = new StringBuilder(1024);
-            int count = 0;
-            foreach (string id in instanceIds)
-            {
-                if (count > 0)
-                {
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -651,87 +651,12 @@ public override async Task<IList<OrchestrationState>> GetStateAsync(IEnumerable<
...
-                this.storageAccountName,
-                this.taskHubName,
-                string.Join(",", result.Select(state => state.OrchestrationInstance.InstanceId)),
-                string.Join(",", result.Select(state => state.OrchestrationInstance.ExecutionId)),
-                stopwatch.ElapsedMilliseconds);
-
-            return result;
+            // In theory this could exceed MaxStorageOperationConcurrency, but the hard maximum of parallel requests is tied to control queue
+            // batch size, which is generally roughly the same value as MaxStorageOperationConcurrency. In almost every case, we would expect this
+            // to only be a small handful of parallel requests, so keeping the code simple until the storage refactor adds global throttling.
```
cgillum, line 656: The max is 32, because that's the maximum batch size supported by Azure Queues.

cgillum (approved):
I appreciate how you turned my 80 lines of code into 3 lines of code that will execute much faster. :)

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
ConnorMcMahon: Yeah, and in general our MaximumStorageConcurrency will be 25 or greater, so I figured it was close enough to not warrant extra code complexity.

PR #570, opened 2021-07-12
Storage Refactor: Queues

amdeel:


src/DurableTask.AzureStorage/Messaging/Session.cs
```diff
@@ -60,7 +54,7 @@ public void TraceProcessingMessage(MessageData data, bool isExtendedSession)
             }
 
             TaskMessage taskMessage = data.TaskMessage;
-            CloudQueueMessage queueMessage = data.OriginalQueueMessage;
+            CloudQueueMessage queueMessage = data.OriginalQueueMessage.CloudQueueMessage;
```
cgillum, line 57: Shouldn't this change be to switch from `CloudQueueMessage` to `QueueMessage`?
```suggestion
            QueueMessage queueMessage = data.OriginalQueueMessage;
```

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -310,26 +288,18 @@ public async Task RenewMessageAsync(MessageData message, SessionBase session)
                 await this.storageQueue.UpdateMessageAsync(
                     queueMessage,
                     this.MessageVisibilityTimeout,
-                    MessageUpdateFields.Visibility,
-                    this.QueueRequestOptions,
-                    session.StorageOperationContext);
-
-                this.stats.MessagesUpdated.Increment();
+                    session?.TraceActivityId.ToString());
```
cgillum, line 291: I noticed that you're calling `session?.TraceActivityId.ToString()` in multiple places. Should we change the signature of these events to take a `Guid` as the activity ID so that we don't have to call `.ToString()` for each of these calls?

src/DurableTask.AzureStorage/Storage/Queue.cs
```diff
@@ -0,0 +1,189 @@
...
+    {
+        readonly AzureStorageClient azureStorageClient;
+        readonly CloudQueueClient queueClient;
+        readonly AzureStorageOrchestrationServiceStats stats;
+        readonly CloudQueue cloudQueue;
+        readonly QueueRequestOptions queueRequestOptions;
+
+        public string Name { get; }
+        public Uri Uri { get; }
+        public int? ApproximateMessageCount => this.cloudQueue.ApproximateMessageCount;
```
cgillum, line 33: nit: the common convention for C# code is to put properties after the constructor.

src/DurableTask.AzureStorage/Storage/Queue.cs
```diff
@@ -0,0 +1,189 @@
...
+                (context, cancellationToken) => this.cloudQueue.DeleteIfExistsAsync(this.queueRequestOptions, context, cancellationToken),
+                "Queue Delete");
+        }
+
+        public async Task<IEnumerable<QueueMessage>> GetMessagesAsync(int batchSize, TimeSpan visibilityTimeout, CancellationToken token)
+        {
+            var cloudQueueMessages = await this.azureStorageClient.MakeStorageRequest<IEnumerable<CloudQueueMessage>>(
+                (context, cancellationToken) =>
+                {
+                    using (var finalLinkedCts = CancellationTokenSource.CreateLinkedTokenSource(token, cancellationToken))
```
cgillum, line 138: nit: I wonder if we can name these two CancellationToken variables in a way that better distinguishes them.

cgillum (commented):
Overall, LGTM. Just one change I think we need in Session.cs, plus a few minor nits.

src/DurableTask.AzureStorage/Storage/Queue.cs
ConnorMcMahon: Not sure why this was resolved?

src/DurableTask.AzureStorage/Storage/Queue.cs
ConnorMcMahon: One option would be to rename cancellationToken to "timeoutHandlerToken" or something along those lines.

ConnorMcMahon (commented):
Agree with most of Chris' comments, with a handful more small suggestions, but I think this is almost ready to merge!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -121,13 +119,13 @@ public AzureStorageOrchestrationService(AzureStorageOrchestrationServiceSettings
             this.allControlQueues = new ConcurrentDictionary<string, ControlQueue>();
             for (int i = 0; i < this.settings.PartitionCount; i++)
             {
-                CloudQueue controlStorageQueue = GetControlQueue(this.queueClient, this.settings.TaskHubName, i);
-                ControlQueue controlQueue = new ControlQueue(controlStorageQueue, this.settings, this.stats, this.messageManager);
+                Queue controlStorageQueue = GetControlQueue(this.azureStorageClient, this.settings.TaskHubName, i);
+                ControlQueue controlQueue = new ControlQueue(this.azureStorageClient, controlStorageQueue, this.messageManager);
```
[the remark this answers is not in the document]
amdeel: I'm not opposed to renaming the classes necessarily but I think in this specific case the problem is that AzureStorageOrchestrationService is using the class Queue and initializing a ControlQueue with a Queue. Instead, I think we should just pass the queue name into ControlQueue and ControlQueue should worry about getting a Queue reference. Check out my recent commit where I made this change.

Also, @ConnorMcMahon, it's actually just Blob right now. You're thinking of the AzureStorageClient method GetBlobReference/GetQueueReference which gives you a Blob or Queue.

src/DurableTask.AzureStorage/Storage/Queue.cs
amdeel: I have a preference for renaming the parameter CancellationToken since all of the timeout tokens are named the same. Maybe I could name it callerToken?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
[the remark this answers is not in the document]
ConnorMcMahon: This works for me.

src/DurableTask.AzureStorage/Storage/Queue.cs
ConnorMcMahon: I definitely wouldn't use the PascalCase `CancellationToken` for a parameter, that violates C# naming conventions, and conflicts with the type name `CancellationToken`. 

You could name the GetMessageAsync() token `callerCancellationToken`, and the TimeoutHandler token `timeoutCancellationToken`, but I think if you have two cancellation tokens, they both need some indicator of who presented that token. Clarity is more important than consistency in my opinion.

ConnorMcMahon (approved):
Looks good except for the few lingering issues (a couple of naming suggestions and the [Obsolete] on the SDK type in the Queue class).

PR #574, opened 2021-07-14
Add Storage deadlock mitigation to MessageManager

ConnorMcMahon:
This code path is important to protect, as blobs can often be quite
large which means the requests take longer and are more likely to be
interupted by network issues.

This is going to be replaced shortly by the AzureStorageClient refactor,
but this is easy enough to squeeze into 1.8.8.


cgillum (approved):
LGTM!

amdeel (approved):

PR #576, opened 2021-07-15
TaskActivityDispatcher -> set activity to current before middleware

jviau:
Quick change to set the current activity _before_ running the middleware pipeline. This allows for accessing `Activity.Current` during the middleware. Otherwise it is `null` until the `TaskActivity` is actually executed.

### DIFF
--- a/src/DurableTask.Core/TaskActivityDispatcher.cs
+++ b/src/DurableTask.Core/TaskActivityDispatcher.cs
@@ -97,6 +97,7 @@ namespace DurableTask.Core
             TaskMessage taskMessage = workItem.TaskMessage;
             OrchestrationInstance orchestrationInstance = taskMessage.OrchestrationInstance;
             TaskScheduledEvent scheduledEvent = null;
+            Activity diagnosticActivity = null;
             try
             {
                 if (string.IsNullOrWhiteSpace(orchestrationInstance?.InstanceId))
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -148,11 +149,15 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
...
                 dispatchContext.SetProperty(scheduledEvent);
 
-                await this.dispatchPipeline.RunAsync(dispatchContext, async _ =>
+                // correlation
+                CorrelationTraceClient.Propagate(() =>
                 {
-                    // correlation 
-                    CorrelationTraceClient.Propagate(() => workItem.TraceContextBase?.SetActivityToCurrent());
+                    workItem.TraceContextBase?.SetActivityToCurrent();
+                    diagnosticActivity = workItem.TraceContextBase?.CurrentActivity;
```
cgillum, line 156: Looks like you're not using this local variable for anything?

jviau: Line 207 I am stopping the activity - to make sure it is removed from `Activity.Current` before exiting this function.

cgillum:
@bachuv or @TsuyoshiUshio, would either of you be able to sign off on this change?

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -148,11 +149,15 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
...
+                });
 
+                await this.dispatchPipeline.RunAsync(dispatchContext, async _ =>
+                {
```
bachuv, line 160: @jviau Thank you for contributing! Can you share some scenarios or sample code to explain these changes and how they would be helpful?

jviau: As mentioned in the PR description, this sets `Activity.Current` _before_ the middleware pipeline is executed, so it is accessible during user provided middleware (`TaskHubWorker.AddActivityDispatcherMiddleware`). Having Activity.Current set there is important so any telemetry during your middleware is also correlated to TaskActivity that is about to be executed. Without this change, nothing in the middleware during a TaskActivity can be correlated. However, TaskOrchestration's middleware does not face the same issue - correlation is available there.

TsuyoshiUshio:
Hi @cgillum and @bachuv 

The change looks good. I'd like to run the following test and see if it works.  How can we run the test?
 https://github.com/Azure/durabletask/blob/main/test/DurableTask.AzureStorage.Tests/Correlation/CorrelationScenarioTest.cs

jviau:
@cgillum - do you know why the build failed there? It shows the tests as 'Flaky'?

bachuv (approved):

PR #579, opened 2021-07-19
Fixed an InvalidOperationException if an empty string is provided for an instance ID query

cgillum:
Minor issue I discovered while running a Java app against this backend. I didn't mean to provide an empty instance ID, but I do expect we handle it more gracefully if one does show up.

Also added nullable type information to the updated file as part of slowly modernizing this codebase.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
+++ b/src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
 namespace DurableTask.AzureStorage.Tracking
 {
     using System;
@@ -32,7 +32,7 @@ namespace DurableTask.AzureStorage.Tracking
         /// <summary>
         /// RuntimeStatus
         /// </summary>
-        public IEnumerable<OrchestrationStatus> RuntimeStatus { get; set; }
+        public IEnumerable<OrchestrationStatus>? RuntimeStatus { get; set; }
 
         /// <summary>
         /// CreatedTimeFrom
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -141,12 +145,17 @@ string GetConditions()
                     TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.LessThan, greaterThanPrefix)));
             }
 
-            if (!string.IsNullOrEmpty(this.InstanceId))
+            if (this.InstanceId != null)
```
cgillum, line 148: This is one of the primary bug fixes needed. It's a bug because it's inconsistent with the InstanceId check in `ToTableQuery<T>()` (which calls this method and checks for null but not empty).  Now they both check specifically for null.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -141,12 +145,17 @@ string GetConditions()
...
             {
-                var sanitizedInstanceId = KeySanitation.EscapePartitionKey(this.InstanceId);
+                string sanitizedInstanceId = KeySanitation.EscapePartitionKey(this.InstanceId);
                 conditions.Add(TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, sanitizedInstanceId));
             }
 
+            if (conditions.Count == 0)
+            {
+                return string.Empty;
+            }
```
cgillum, line 157: Previously an empty string instance ID would cause us to have a `conditions` list with zero elements. That resulted in an `InvalidOperationException` when we moved down to the next few lines and tried to run `conditions.Aggregate(...)`. With this fix, we check for an empty conditions list and bail early.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
```
ConnorMcMahon, line 13: Unrelated to this PR, but I think in order to make sure that our gradual efforts to introduce nullability, we should start by adding this flag to all of our data types (especially in DurableTask.Core).

That way as we tackle classes with the business logic on a case by case basis, they have accurate information about what fields are/aren't nullable.

ConnorMcMahon (approved):
LGTM!

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
cgillum: Not a bad idea, though it will of course require a more dedicated effort.

PR #582, opened 2021-07-21
Prepare DurableTask.Core and DurableTask.AzureStorage release

ConnorMcMahon:
The previous fix #560 was not actually released as the fix is contained
in DurableTask.Core, which was not released as a part of
DurableTask.AzureStorage v1.8.8.

The fix in #560 is more severe than we thought as it makes a
TaskHubWorker lose an entire partition when it completes an
orchestration, so we are also releasing a new version of
DurableTask.AzureStorage that depends on DurableTask.Core having this
fix.


cgillum (approved):

jviau:
Any chance we can get https://github.com/Azure/durabletask/pull/576 included here? The tests failed, but if I am reading it correctly, they are flaky?

PR #585, opened 2021-07-29
Fixed bug with BlobContainer permissions

amdeel:
Updated BlobContainerPublicAccessType on CreateIfNotExistsAsync in BlobContainer that was causing an error with storage accounts without public blob access.

Updated AzureStorageClient constructor by removing storageAccountName. AzureStorageClient now sets it's own StorageAccountName.


ConnorMcMahon (approved):
LGTM, glad we caught this!

PR #586, opened 2021-07-29
Remove noisy logs in SafePartitionManager

ConnorMcMahon:
Some of the warnings being emitted by the SafePartitionManager happen
more frequently than initially thought, but are not really a concern. To
avoid taking up too much App Insights space and concerning customers
with unnecessary noise, we are removing these logs.


amdeel (approved):
LGTM!

PR #587, opened 2021-08-02
Fixed NullRef when querying instance status on execution started

amdeel:
This PR addresses a race condition leading to a NullReferenceException when starting an orchestrator.

When the instances table is not yet updated, OrchestrationSessionManager throws a NullRef on InstanceId after attempting to query for existing OrchestrationStates here:
https://github.com/Azure/durabletask/blob/6be1e3a2a23b81fdf681147c9b433610915adfae/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs#L218

This fix changes the result returned from GetStateAsync, called above, to only return non-null OrchestrationStates.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -656,7 +656,7 @@ namespace DurableTask.AzureStorage.Tracking
             // to only be a small handful of parallel requests, so keeping the code simple until the storage refactor adds global throttling.
             var instanceQueries = instanceIds.Select(instance => this.GetStateAsync(instance, allExecutions: true, fetchInput: false));
             IEnumerable<IList<OrchestrationState>> instanceQueriesResults = await Task.WhenAll(instanceQueries);
-            return instanceQueriesResults.SelectMany(result => result).ToList();
+            return instanceQueriesResults.SelectMany(result => result).Where(x => x.OrchestrationInstance.InstanceId != null).ToList();
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


ConnorMcMahon (approved):
LGTM!

Looks like 3rd time is the charm!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -656,7 +656,7 @@ public override async Task<IList<OrchestrationState>> GetStateAsync(IEnumerable<
             // to only be a small handful of parallel requests, so keeping the code simple until the storage refactor adds global throttling.
             var instanceQueries = instanceIds.Select(instance => this.GetStateAsync(instance, allExecutions: true, fetchInput: false));
             IEnumerable<IList<OrchestrationState>> instanceQueriesResults = await Task.WhenAll(instanceQueries);
-            return instanceQueriesResults.SelectMany(result => result).ToList();
+            return instanceQueriesResults.SelectMany(result => result).Where(x => x.OrchestrationInstance.InstanceId != null).ToList();
```
cgillum, line 659: I don't have full context on this PR, but isn't it possible that `x` could be null? If so, do we need to check for that too? Looking at the overloads of `GetStateAsync(string, bool, bool)`, null values for `OrchestrationState` seem possible.
```suggestion
            return instanceQueriesResults.SelectMany(result => result).Where(x => x?.OrchestrationInstance.InstanceId != null).ToList();
```

amdeel: The method that GetStateAsync calls,  FetchInstanceStatusInternalAsync will never return null. It's currently coded in an incorrect way. Currently in code, we return null if the DynamicEntityTable result is null. However, this will never be null even when there is no result returned. This means that FetchInstanceStatusInternalAsync will always return an InstanceStatus with an OrchestrationState. The OrchestrationState will just have null values.

This is something I am going to open an issue to address, as this code should be fixed to avoid any potential NullRefs in the future when using OrchestrationState. However, that will require a bit more code change and testing so this PR is intended to fix the NullRef on line 227 of OrchestrationSessionManager only.

cgillum (approved):
:shipit:

PR #595, opened 2021-08-17
Add more robust determinism checks in orchestration replay logic

### LINKED ISSUE #594: Add more robust non-determinism checks
Current non-determinism checks in TaskOrchestrationExecutor essentially only catch the case where the number of orchestration actions scheduled this replay are less than the number of actions scheduled on previous replays, or a substantial change in the await/async structure of the orchestration.

Improve checks would involve:

- validating that the orchestration action is the same type as previous replays
- validating that activity inputs/names match the activity/name scheduled on previous replays
- validating that timer values have not changed (?) - (not sure if this is worth breaking over)

Currently, non-determinism issues that aren't caught just result in stuck orchestrations, which are incredibly hard to debug.

ConnorMcMahon:
Current checks for nondeterminism are fairly limited, and only catch
major control flow changes. This PR attempts to improve our best-effort
attempts to catch non-determinism by validating that this replay's api
call signatures match the API calls scheduled by the orchestration
originally.

Resolves #594.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -323,9 +323,9 @@ namespace DurableTask.Core
             if (!this.orchestratorActionsMap.ContainsKey(taskId))
             {
                 throw new NonDeterministicOrchestrationException(subOrchestrationCreateEvent.EventId,
-                   $"A previous execution of this orchestration scheduled a suborchestration task with sequence ID {taskId} "
+                   $"A previous execution of this orchestration scheduled a sub-orchestration task with sequence ID {taskId} "
                    + $"and name '{subOrchestrationCreateEvent.Name}' (version '{subOrchestrationCreateEvent.Version}', "
-                   + $"instance id '{subOrchestrationCreateEvent.InstanceId}'), but the current replay execution hasn't (yet?) "
+                   + $"instance ID '{subOrchestrationCreateEvent.InstanceId}'), but the current replay execution hasn't (yet?) "
                    + "scheduled this task. Was a change made to the orchestrator code after this instance had already started running?");
             }
 
@@ -333,9 +333,9 @@ namespace DurableTask.Core
             if (!(orchestrationAction is CreateSubOrchestrationAction currentReplayAction))
             {
                 throw new NonDeterministicOrchestrationException(subOrchestrationCreateEvent.EventId,
-                   $"A previous execution of this orchestration scheduled a suborchestration task with sequence ID {taskId} "
+                   $"A previous execution of this orchestration scheduled a sub-orchestration task with sequence ID {taskId} "
                    + $"and name '{subOrchestrationCreateEvent.Name}' (version '{subOrchestrationCreateEvent.Version}', "
-                   + $"instance id '{subOrchestrationCreateEvent.InstanceId}'), but the current orchestration replay instead "
+                   + $"instance ID '{subOrchestrationCreateEvent.InstanceId}'), but the current orchestration replay instead "
                    + $"scheduled a {orchestrationAction.GetType().Name} task at this sequence number. Was a change made to "
                    + "the orchestrator code after this instance had already started running?");
             }
@@ -343,10 +343,10 @@ namespace DurableTask.Core
             if (!string.Equals(subOrchestrationCreateEvent.Name, currentReplayAction.Name, StringComparison.OrdinalIgnoreCase))
             {
                 throw new NonDeterministicOrchestrationException(subOrchestrationCreateEvent.EventId,
-                   $"A previous execution of this orchestration scheduled a suborchestration task with sequence ID {taskId} "
+                   $"A previous execution of this orchestration scheduled a sub-orchestration task with sequence ID {taskId} "
                    + $"and name '{subOrchestrationCreateEvent.Name}' (version '{subOrchestrationCreateEvent.Version}', "
-                   + $"instance id '{subOrchestrationCreateEvent.InstanceId}'), but the current orchestration replay instead "
-                   + $"scheduled a suborchestration task with name {currentReplayAction.Name} at this sequence number. "
+                   + $"instance ID '{subOrchestrationCreateEvent.InstanceId}'), but the current orchestration replay instead "
+                   + $"scheduled a sub-orchestration task with name {currentReplayAction.Name} at this sequence number. "
                    + "Was a change made to the orchestrator code after this instance had already started running?");
             }
 
@@ -360,7 +360,7 @@ namespace DurableTask.Core
             {
                 throw new NonDeterministicOrchestrationException(eventSentEvent.EventId,
                    $"A previous execution of this orchestration scheduled a send event task with sequence ID {taskId}, "
-                   + $"type '{eventSentEvent.EventType}' name '{eventSentEvent.Name}', instance id '{eventSentEvent.InstanceId}', "
+                   + $"type '{eventSentEvent.EventType}' name '{eventSentEvent.Name}', instance ID '{eventSentEvent.InstanceId}', "
                    + $"but the current replay execution hasn't (yet?) scheduled this task. Was a change made to the orchestrator code "
                    + $"after this instance had already started running?");
             }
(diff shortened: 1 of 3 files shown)


ConnorMcMahon:
I should probably add some tests, and this change definitely warrants some perf validation to make sure there are no major regressions since this is in such a core code path. 

We could potentially make this feature opt-in if it hurts perf too much.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -276,49 +294,97 @@ public void HandleTimerCreatedEvent(TimerCreatedEvent timerCreatedEvent)
...
+                    $"Scheduled event not run this replay: EventSentEvent: {eventSentEvent.EventId} {eventSentEvent.EventType} {eventSentEvent.Name} {eventSentEvent.InstanceId}.");
             }
-            else
+
+            var orchestrationAction = this.orchestratorActionsMap[taskId];
+            if (!(orchestrationAction is SendEventOrchestratorAction currentReplayAction))
+            {
+                throw new NonDeterministicOrchestrationException(eventSentEvent.EventId,
+                    $"Event id {eventSentEvent.EventId} on previous replays was {nameof(EventSentEvent)} but is now {orchestrationAction.GetType().Name}.");
+
```
cgillum, line 366: nit: remove extra whitespace

cgillum (commented):
I'm happy to see this PR which adds more validation checks to orchestration code. I added some suggestions for how to make the exception messages a bit more actionable to users with the hope of reducing the number of GitHub issues that are opened when non-determinism errors are encountered.

I only added suggestions for the TaskScheduledEvent exception messages, but I think we should apply the same basic template for the others as well (sub-orchestrations and sent events).

I'm a little worried about doing comparisons on inputs, especially considering the potential CPU impact. Are you seeing cases where a customer is confused because they changed the input but didn't see a different output? I wonder if we could skip this validation, since it's rather strict and changing inputs doesn't actually break anything.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -256,15 +256,33 @@ public override async Task<T> CreateTimer<T>(DateTime fireAt, T state, Cancellat
         public void HandleTaskScheduledEvent(TaskScheduledEvent scheduledEvent)
         {
             int taskId = scheduledEvent.EventId;
-            if (this.orchestratorActionsMap.ContainsKey(taskId))
+            if (!this.orchestratorActionsMap.ContainsKey(taskId))
             {
-                this.orchestratorActionsMap.Remove(taskId);
+                throw new NonDeterministicOrchestrationException(scheduledEvent.EventId,
+                    $"Scheduled event not run this replay: TaskScheduledEvent {scheduledEvent.EventId} {scheduledEvent.EventType} {scheduledEvent.Name} {scheduledEvent.Version}.");
```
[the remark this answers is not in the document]
ConnorMcMahon: Love it! 

I tried to keep similar to the old messages, but those were always super confusing to customers anyways, so let's fix this while we are touching this here.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -256,15 +256,33 @@ public override async Task<T> CreateTimer<T>(DateTime fireAt, T state, Cancellat
...
+            var orchestrationAction = this.orchestratorActionsMap[taskId];
+            if (!(orchestrationAction is ScheduleTaskOrchestratorAction currentReplayAction))
             {
                 throw new NonDeterministicOrchestrationException(scheduledEvent.EventId,
-                    $"TaskScheduledEvent: {scheduledEvent.EventId} {scheduledEvent.EventType} {scheduledEvent.Name} {scheduledEvent.Version}");
+                    $"Event id {scheduledEvent.EventId} on previous replays was {nameof(TaskScheduledEvent)} but is now {orchestrationAction.GetType().Name}.");
+
             }
+
+            if (scheduledEvent.Name != currentReplayAction.Name)
```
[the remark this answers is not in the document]
ConnorMcMahon: It would probably help prevent false positives, and I can't think of many (any?) situations where a case insensitive comparison would cause us to miss a non-determinism check?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -256,15 +256,33 @@ public override async Task<T> CreateTimer<T>(DateTime fireAt, T state, Cancellat
...
+            {
+                throw new NonDeterministicOrchestrationException(scheduledEvent.EventId,
+                    $"Event id {scheduledEvent.EventId} on previous replays scheduled task with name {scheduledEvent.Name}) but is now {currentReplayAction.Name}.");
+            }
+
+            if (scheduledEvent.Input != currentReplayAction.Input)
```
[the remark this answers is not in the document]
ConnorMcMahon: Good point, especially about input payloads being unbounded making this potentially very expensive.

I already explicitly left-of fireAt from the timer comparison as it doesn't effectively matter, so I think we can make similar justification for activity/suborchestration inputs and event data.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -276,49 +294,97 @@ public void HandleTimerCreatedEvent(TimerCreatedEvent timerCreatedEvent)
...
+                    $"Event id {eventSentEvent.EventId} on previous replays scheduled task with name {eventSentEvent.Name}) but is now {currentReplayAction.EventName}.");
             }
+
+            if (eventSentEvent.Input != currentReplayAction.EventData)
+            {
+                throw new NonDeterministicOrchestrationException(eventSentEvent.EventId,
+                    $"EventSentEvent {eventSentEvent.EventId} has different data this replay than it was scheduled with.");
+            }
+
+            if (eventSentEvent.InstanceId != currentReplayAction.Instance.InstanceId)
```
ConnorMcMahon, line 381: What are our thoughts about validating instance id? In general, it should be fairly cheap to validate, but considering our stances on inputs and eventdata, I think you could make a similar argument for instance id?

cgillum: Yeah, I guess for the sake of consistency, if we're not validating inputs then we probably shouldn't validate the instance ID either since these changes are inconsequential...

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -256,15 +256,35 @@ public override async Task<T> CreateTimer<T>(DateTime fireAt, T state, Cancellat
         public void HandleTaskScheduledEvent(TaskScheduledEvent scheduledEvent)
         {
             int taskId = scheduledEvent.EventId;
-            if (this.orchestratorActionsMap.ContainsKey(taskId))
+            if (!this.orchestratorActionsMap.ContainsKey(taskId))
             {
-                this.orchestratorActionsMap.Remove(taskId);
+                throw new NonDeterministicOrchestrationException(scheduledEvent.EventId,
```
ConnorMcMahon, line 261: This is not super efficient string building, but this should be rare enough that I prefer readability over perf. I am open to being convinced otherwise.

cgillum: I'm not concerned about string building perf for exceptions. Also, the C# compiler is smart enough to optimize-away much of this.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -276,49 +296,98 @@ public void HandleTimerCreatedEvent(TimerCreatedEvent timerCreatedEvent)
...
         {
             int taskId = subOrchestrationCreateEvent.EventId;
-            if (this.orchestratorActionsMap.ContainsKey(taskId))
+            if (!this.orchestratorActionsMap.ContainsKey(taskId))
             {
-                this.orchestratorActionsMap.Remove(taskId);
+                throw new NonDeterministicOrchestrationException(subOrchestrationCreateEvent.EventId,
+                   $"A previous execution of this orchestration scheduled a suborchestration task with sequence ID {taskId} "
+                   + $"and name '{subOrchestrationCreateEvent.Name}' (version '{subOrchestrationCreateEvent.Version}', "
+                   + $"instance id '{subOrchestrationCreateEvent.InstanceId}'), but the current replay execution hasn't (yet?) "
```
cgillum, line 328: nit: replace "id" with "ID" in all of these error messages.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -276,49 +296,98 @@ public void HandleTimerCreatedEvent(TimerCreatedEvent timerCreatedEvent)
...
             {
                 throw new NonDeterministicOrchestrationException(subOrchestrationCreateEvent.EventId,
-                    // ReSharper disable once UseStringInterpolation
-                    string.Format("SubOrchestrationInstanceCreatedEvent: {0} {1} {2} {3} {4}",
-                        subOrchestrationCreateEvent.EventId,
-                        subOrchestrationCreateEvent.EventType,
-                        subOrchestrationCreateEvent.Name,
-                        subOrchestrationCreateEvent.Version,
-                        subOrchestrationCreateEvent.InstanceId));
+                   $"A previous execution of this orchestration scheduled a suborchestration task with sequence ID {taskId} "
```
cgillum, line 336: nit: suborchestration -> sub-orchestration here and below.

cgillum (approved):
A few minor nits on the error messages. Otherwise I think this looks good.

PR #600, opened 2021-09-02
Updated TableQuery filter condition string generation

amdeel:
This PR updates AzureTableTrackingStore to no longer manually create the filter condition for TableQuery, instead using the TableQuery methods to generate it for us.

In GetHistoryEventsAsync, we generated the filter condition containing InstanceId. If customers used an apostrophe in their InstanceId this query would fail as a bad request. Using TableQuery.GenerateFilterCondition handles this case for us and generates a filter condition that does not result in a bad request.

All manual filter condition creation code has been swapped with TableQuery generation methods.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -238,22 +238,20 @@ namespace DurableTask.AzureStorage.Tracking
 
         async Task<HistoryEntitiesResponseInfo> GetHistoryEntitiesResponseInfoAsync(string instanceId, string expectedExecutionId, IList<string> projectionColumns,  CancellationToken cancellationToken = default(CancellationToken))
         {
-            var filterCondition = new StringBuilder(200);
-
-            const char Quote = '\'';
-
-            // e.g. "PartitionKey eq 'c138dd969a1e4a699b0644c7d8279f81'"
             var sanitizedInstanceId = KeySanitation.EscapePartitionKey(instanceId);
-            filterCondition.Append(PartitionKeyProperty).Append(" eq ").Append(Quote).Append(sanitizedInstanceId).Append(Quote);
+            string filterCondition = TableQuery.GenerateFilterCondition(PartitionKeyProperty, QueryComparisons.Equal, sanitizedInstanceId);
             if (!string.IsNullOrEmpty(expectedExecutionId))
             {
                 // Filter down to a specific generation.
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -389,31 +385,38 @@ public override async Task<IList<string>> RewindHistoryAsync(string instanceId,
...
             var recentStartRow = orchestratorStartedEntities.Where(y => y.RowKey == recentStartRowKey).ToList();
             var executionId = recentStartRow[0].Properties["ExecutionId"].StringValue;
             var instanceTimestamp = recentStartRow[0].Timestamp.DateTime;
 
-            var rowsToUpdateFilterCondition = new StringBuilder(200);
+            var rowsToUpdateFilterCondition = partitionFilter;
+            var executionIdFilter = TableQuery.GenerateFilterCondition("ExecutionId", QueryComparisons.Equal, executionId);
+            rowsToUpdateFilterCondition = TableQuery.CombineFilters(rowsToUpdateFilterCondition, TableOperators.And, executionIdFilter);
+            var orchestrationStatusFilter = TableQuery.GenerateFilterCondition("OrchestrationStatus", QueryComparisons.Equal, "Failed");
+            rowsToUpdateFilterCondition = TableQuery.CombineFilters(rowsToUpdateFilterCondition, TableOperators.And, orchestrationStatusFilter);
```
cgillum, line 410: I'm not sure if I'm reading this correctly, but it looks like the query you're building here is a little different from the one that was being generated before. Previously we were building this:

> PartitionKey eq 'X' and ExecutionId eq 'Y' and (OrchestrationStatus eq 'Failed' or EventType eq 'TaskFailed' or EventType eq 'SubOrchestrationInstanceFailed')

Your query seems to be building the following:

> (PartitionKey eq 'X' and ExecutionId eq 'Y' and OrchestrationStatus eq 'Failed') or (EventType eq 'TaskFailed' or EventType eq 'SubOrchestrationInstanceFailed')

Am I reading this correctly?

amdeel: Yeah that's a mistake. This should be fixed now.

cgillum:
My comments are resolved, but it looks like there are some failures in the Rewind tests (which are definitely impacted by these changes).

cgillum (approved):
LGTM! I'm glad you were able to figure out the cause of those failures. I wouldn't have guessed it could be related to string vs. int comparisons.

ConnorMcMahon (approved):
This looks great.

These query conditions definitely seem like the approach we should take, it reminds me of similar approaches to prevent SQL Injections attacks when using SQL-based libraries instead of just building your own queries directly.

Hopefully this won't require too many changes when we switch to track 2 Table SDK!

PR #605, opened 2021-09-07
DT.AzureStorage: Fix a missed exception handling scenario

cgillum:
## Issue summary

Failing to deserialize a control queue message can result in orchestrations getting stuck and lots of duplicate message warnings to be logged. This is because we don't properly abandon the message after we've dequeued it. The result is that the orchestration service internally thinks we're still holding onto the message and refuses to process any new dequeued copies of the message.

## Issue details

We observed a case with users in the M365 org where orchestrations would get stuck for extremely long periods of time. When this happens, only a lease failover or a machine restart would cause the orchestration to continue. When investigating, we noticed two things:

* There were many "Received a duplicate message" warnings for the same message, one every 5 minutes. There was also no "MessageReceived" trace when this happened.
* Around the time this started happening, the following error was observed on this partition:

```
An error occurred while processing messages on xxx-control-08: System.IO.InvalidDataException: The archive entry was compressed using an unsupported compression method.
  at System.IO.Compression.Inflater.Inflate(FlushCode flushCode)
  at System.IO.Compression.Inflater.ReadInflateOutput(Byte* bufPtr, Int32 length, FlushCode flushCode, Int32& bytesRead)
  at System.IO.Compression.Inflater.ReadOutput(Byte* bufPtr, Int32 length, Int32& bytesRead)
  at System.IO.Compression.Inflater.InflateVerified(Byte* bufPtr, Int32 length)
  at System.IO.Compression.DeflateStream.ReadCore(Span`1 buffer)
  at System.IO.Compression.DeflateStream.Read(Byte[] array, Int32 offset, Int32 count)
  at DurableTask.AzureStorage.MessageManager.Decompress(Stream blobStream) in C:\source\durabletask\src\DurableTask.AzureStorage\MessageManager.cs:line 242
  at DurableTask.AzureStorage.MessageManager.DownloadAndDecompressAsBytesAsync(Blob blob) in C:\source\durabletask\src\DurableTask.AzureStorage\MessageManager.cs:line 212
  at DurableTask.AzureStorage.MessageManager.DownloadAndDecompressAsBytesAsync(String blobName) in C:\source\durabletask\src\DurableTask.AzureStorage\MessageManager.cs:line 194
  at DurableTask.AzureStorage.MessageManager.DeserializeQueueMessageAsync(QueueMessage queueMessage, String queueName) in C:\source\durabletask\src\DurableTask.AzureStorage\MessageManager.cs:line 149
  at DurableTask.AzureStorage.Messaging.ControlQueue.<>c__DisplayClass11_0.<<GetMessagesAsync>b__0>d.MoveNext() in C:\source\durabletask\src\DurableTask.AzureStorage\Messaging\ControlQueue.cs:line 126
--- End of stack trace from previous location where exception was thrown ---
  at DurableTask.AzureStorage.Utils.ParallelForEachAsync[TSource](IEnumerable`1 enumerable, Func`2 action) in C:\source\durabletask\src\DurableTask.AzureStorage\Utils.cs:line 50
  at DurableTask.AzureStorage.Messaging.ControlQueue.GetMessagesAsync(CancellationToken cancellationToken) in C:\source\durabletask\src\DurableTask.AzureStorage\Messaging\ControlQueue.cs:line 165
```

We don't know the cause for the above exception, however, after examining the code, I realized that this exception could cause orchestrations to get stuck because we don't correctly handle it. The following sequence of events can happen inside of [ControlQueue.cs](https://github.com/Azure/durabletask/blob/e8b169ac97d2a2a4926412a1dcffbbca5e4c0e78/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs#L102-L126):

1. We dequeue up to 32 messages and process them as a batch ([here](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs#L102-L126)).
2. For every message that's deserialized correctly, we add them to [`this.stats.PendingOrchestratorMessages`](https://github.com/Azure/durabletask/blob/e8b169ac97d2a2a4926412a1dcffbbca5e4c0e78/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs#L109). If that message ID is already in this dictionary, then we log a duplicate-message warning. After we're done processing those messages, they get removed from this dictionary.
3. If any one of those 32 messages fail to deserialize, then the entire try-block fails, we log an error, delay, and try to dequeue messages again ([source](https://github.com/Azure/durabletask/blob/e8b169ac97d2a2a4926412a1dcffbbca5e4c0e78/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs#L142-L158)). However, we never remove the successful entries from `this.stats.PendingOrchestratorMessages`!
4. 5-minutes later when we dequeue those same 32 messages again, all of them (except for the one that failed to deserialize, ironically) get blocked by the [`this.stats.PendingOrchestratorMessages.TryAdd`](https://github.com/Azure/durabletask/blob/e8b169ac97d2a2a4926412a1dcffbbca5e4c0e78/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs#L109) duplicate message check since those message IDs were never removed from the dictionary. These messages are now forever stuck until the app process gets recycled.

## Fix summary

The fix is to properly handle message download/dequeue exceptions to ensure that we clean up our internal data structures (PendingOrchestratorMessages) if a failure occurs. This PR required some refactoring of the abandon message path because we didn't have any code that allowed us to abandon a raw Azure Storage queue message. We only had code to abandon the fully-deserialized version of a message.

This PR also increments the DT.AzureStorage package version to v1.9.1.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
@@ -101,9 +101,32 @@ namespace DurableTask.AzureStorage.Messaging
                         var batchMessages = new ConcurrentBag<MessageData>();
                         await batch.ParallelForEachAsync(async delegate (QueueMessage queueMessage)
                         {
-                            MessageData messageData = await this.messageManager.DeserializeQueueMessageAsync(
-                                queueMessage,
-                                this.storageQueue.Name);
+                            MessageData messageData;
+                            try
+                            {
+                                messageData = await this.messageManager.DeserializeQueueMessageAsync(
+                                    queueMessage,
+                                    this.storageQueue.Name);
+                            }
+                            catch (Exception e)
+                            {
+                                // We have limited information about the details of the message
+                                // since we failed to deserialize it.
+                                this.settings.Logger.MessageFailure(
+                                    this.storageAccountName,
+                                    this.settings.TaskHubName,
+                                    queueMessage.Id /* MessageId */,
+                                    string.Empty /* InstanceId */,
+                                    string.Empty /* ExecutionId */,
+                                    this.storageQueue.Name,
+                                    string.Empty /* EventType */,
+                                    0 /* TaskEventId */,
+                                    e.ToString());
+
+                                // Abandon the message so we can try it again later.
+                                await this.AbandonMessageAsync(queueMessage);
+                                return;
+                            }
 
                             // Check to see whether we've already dequeued this message.
                             if (!this.stats.PendingOrchestratorMessages.TryAdd(queueMessage.Id, 1))
@@ -162,6 +185,18 @@ namespace DurableTask.AzureStorage.Messaging
                 this.IsReleased = true;
                 return EmptyMessageList;
             }
+        }
+
+        // This overload is intended for cases where we aren't able to deserialize an instance of MessageData.
+        public Task AbandonMessageAsync(QueueMessage queueMessage)
+        {
+            this.stats.PendingOrchestratorMessages.TryRemove(queueMessage.Id, out _);
+            return base.AbandonMessageAsync(
+                queueMessage,
+                taskMessage: null,
+                instance: null,
+                traceActivityId: null,
+                sequenceNumber: -1);
         }
 
         public override Task AbandonMessageAsync(MessageData message, SessionBase? session = null)
--- a/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
@@ -214,11 +214,32 @@ namespace DurableTask.AzureStorage.Messaging
             return initialVisibilityDelay;
         }
 
-        public virtual async Task AbandonMessageAsync(MessageData message, SessionBase? session = null)
+        public virtual Task AbandonMessageAsync(MessageData message, SessionBase? session = null)
         {
             QueueMessage queueMessage = message.OriginalQueueMessage;
             TaskMessage taskMessage = message.TaskMessage;
             OrchestrationInstance instance = taskMessage.OrchestrationInstance;
+            long sequenceNumber = message.SequenceNumber;
+
+            return this.AbandonMessageAsync(
+                queueMessage,
+                taskMessage,
+                instance,
+                session?.TraceActivityId,
+                sequenceNumber);
+        }
+
+        protected async Task AbandonMessageAsync(
+            QueueMessage queueMessage,
+            TaskMessage? taskMessage,
+            OrchestrationInstance? instance,
+            Guid? traceActivityId,
+            long sequenceNumber)
+        {
+            string instanceId = instance?.InstanceId ?? string.Empty;
+            string executionId = instance?.ExecutionId ?? string.Empty;
+            string eventType = taskMessage?.Event.EventType.ToString() ?? string.Empty;
+            int taskEventId = taskMessage != null ? Utils.GetTaskEventId(taskMessage.Event) : -1;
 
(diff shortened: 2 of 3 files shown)


src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -101,9 +101,32 @@ public async Task<IReadOnlyList<MessageData>> GetMessagesAsync(CancellationToken
...
-                                queueMessage,
-                                this.storageQueue.Name);
+                            MessageData messageData;
+                            try
+                            {
+                                messageData = await this.messageManager.DeserializeQueueMessageAsync(
+                                    queueMessage,
+                                    this.storageQueue.Name);
+                            }
+                            catch (Exception e)
```
cgillum, line 111: This try/catch block is the most important part of the change. Without it, any individual failure in `this.messageManager.DeserializeQueueMessageAsync` could cause the entire `ParallelForEachAsync` task to fail, which wouldn't be a big deal except that we weren't updating `this.stats.PendingOrchestratorMessages` as part of our exception handling (the `AbandonMessageAsync` method updates this for us).

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -214,11 +214,32 @@ async Task<MessageData> AddMessageAsync(TaskMessage taskMessage, OrchestrationIn
...
+            QueueMessage queueMessage,
+            TaskMessage? taskMessage,
+            OrchestrationInstance? instance,
+            Guid? traceActivityId,
+            long sequenceNumber)
+        {
+            string instanceId = instance?.InstanceId ?? string.Empty;
+            string executionId = instance?.ExecutionId ?? string.Empty;
+            string eventType = taskMessage?.Event.EventType.ToString() ?? string.Empty;
+            int taskEventId = taskMessage != null ? Utils.GetTaskEventId(taskMessage.Event) : -1;
```
cgillum, line 242: This refactoring really highlights the value of nullable reference types in C# 8. Previously we didn't expect any of this data to be null, but now we do. Nullable reference types allowed me to refactor this code in a way that I'm fairly confident I didn't miss a null-check somewhere.

cgillum (commented):
A couple notes on the code changes:

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
amdeel: Catching this exception is definitely an important improvement, abandoning the message makes sense to me as well. I'm wondering what you mean by updating stats.PendingOrchestratorMessages being a big deal itself. I might not be understanding the significance of this but it also looks like we don't even add this message to PendingOrchestratorMessages until after this line of code (meaning if serialization succeeds). Could you explain what you meant by that?

cgillum: Sure, yeah, I kinda skimmed over this. I've updated the PR description with more details. Let me know if there's anything else I can add to make the issue more clear. To summarize here, you're right that we don't add the failed message to this dictionary. The problem is actually with all the messages that _didn't_ fail to deserialize. Those get added to the dictionary, but the error for a bad messages causes us to stop processing the good messages and then we fail to reprocess them again later because we never cleaned up this dictionary.

amdeel: I see, so you're just talking about the side effect of not having this catch on all of the other messages in the batch. Thanks for the clarification.

amdeel (approved):

PR #607, opened 2021-09-11
Fix large inputs for pending orchestrations on Azure Storage

ConnorMcMahon:
Pending orchestrations for large messages stored in blobs in the Azure
Storage implementation stored the blob name in the Input field in the
status table. This does not comply with the expectations that the
Input/Output fields should have either the raw value or an absolute URL
to the blob containing the raw values.

This PR addresses replaces the blob name with the blob URL.


cgillum (approved):
LGTM!

PR #608, opened 2021-09-13
Update DurableTask.props to work with different sln file locations

cgillum:
This is to fix a build signing issue in the alternate release pipeline (azure-pipelines-release.yml)

### DIFF
--- a/tools/DurableTask.props
+++ b/tools/DurableTask.props
@@ -23,9 +23,9 @@
   </ItemGroup>
 
   <!-- Sign assemblies if the snk file is present -->
-  <PropertyGroup Condition="Exists('$(SolutionDir)\tools\sign.snk') And $(Configuration) == 'Release'">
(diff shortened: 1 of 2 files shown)


tools/DurableTask.props
```diff
@@ -23,9 +23,9 @@
   </ItemGroup>
 
   <!-- Sign assemblies if the snk file is present -->
-  <PropertyGroup Condition="Exists('$(SolutionDir)\tools\sign.snk') And $(Configuration) == 'Release'">
+  <PropertyGroup Condition="Exists('$(MSBuildProjectDirectory)\..\..\tools\sign.snk') And $(Configuration) == 'Release'">
```
cgillum, line 26: The new build pipeline uses a different *.sln file which is outside of the root directory. That was causing this condition to evaluate to false when building the alternate solution file, which meant that the output binaries wouldn't be strong-name signed. This change makes it so we look for the snk file relative to the csproj that needs it, rather than relative to the .sln file.

ConnorMcMahon (approved):
LGTM!

PR #609, opened 2021-09-15
Update access modifiers

ChenglongLiu:
Update two access modifiers to `public`, which is helpful when users want to inherit `ReflectionBasedTaskActivity`

### DIFF
--- a/src/DurableTask.Core/Common/Utils.cs
+++ b/src/DurableTask.Core/Common/Utils.cs
@@ -81,7 +81,12 @@ namespace DurableTask.Core.Common
             return input;
         }
 
-        internal static JArray ConvertToJArray(string input)
+        /// <summary>
+        /// Convert input string to JArray
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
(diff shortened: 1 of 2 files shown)


src/DurableTask.Core/Common/Utils.cs
```diff
@@ -81,7 +81,12 @@ public static string Truncate(this string input, int maxLength)
             return input;
         }
 
-        internal static JArray ConvertToJArray(string input)
+        /// <summary>
+        /// Convert input string to JArray
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
+        public static JArray ConvertToJArray(string input)
```
cgillum, line 89: I'm curious: why do you need this method to be public instead of just using `JArray.Parse(string input)`?

ChenglongLiu: In commit #442 `JArray.Parse' was changed to this function, I'm not sure they are equal or not. If yes, just use `JArray.Parse(string input)' is a choice

PR #611, opened 2021-09-22
Added catch block in dequeue loop, improved logging

amdeel:
Added a catch block to the dequeue loop to prevent unhandled exceptions from stopping the dequeue loop.

Added CloseReason to logs when removing control queues.

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
@@ -21,6 +21,7 @@ namespace DurableTask.AzureStorage
     using System.Threading.Tasks;
     using DurableTask.AzureStorage.Messaging;
     using DurableTask.AzureStorage.Monitoring;
+    using DurableTask.AzureStorage.Partitioning;
     using DurableTask.AzureStorage.Storage;
     using DurableTask.AzureStorage.Tracking;
     using DurableTask.Core;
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -516,7 +516,7 @@ async Task<bool> RenewLeaseAsync(T lease)
                 lease.PartitionId,
                 renewed,
                 lease.Token,
-                errorMessage);
+                $"Renewal result for {this.leaseType}: {renewed}");
```
cgillum, line 519: Are we potentially missing out on useful information if we stop logging `errorMessage` when `renewed == true`?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -147,16 +148,23 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
                     }
                 }
+                catch (Exception e)
+                {
+                    this.settings.Logger.PartitionManagerWarning(
+                        this.storageAccountName,
+                        this.settings.TaskHubName,
+                        this.settings.WorkerId,
+                        partitionId,
+                        $"Exception in the dequeue loop for control queue {controlQueue.Name}. Exception: {e}");
```
cgillum, line 158: I wonder if we should add a small delay to avoid the possibility of getting into a tight error loop, which would consume a lot of CPU and generate a lot of logs. Maybe just add `await Thread.Sleep(TimeSpan.FromSeconds(1));` after the log message is written?

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -147,16 +148,23 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
                         this.AddMessageToPendingOrchestration(controlQueue, filteredMessages, traceActivityId, cancellationToken);
                     }
                 }
+                catch (Exception e)
```
cgillum, line 151: I think we might need another catch block to handle cancellation tokens firing. Something like this:
```csharp
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
{
    // shutting down
    break;
}

cgillum (approved):

ConnorMcMahon (approved):
LGTM, this should result in making these situations a lot easier to diagnose (and hopefully fix some of them altogether).

PR #616, opened 2021-09-30
JsonSerializerSettings made static

Scooletz:
This PR makes a few `JsonSerializerSettings` static. This was partially a result of profiling `AzureTableOrchestrationHistoryEventEntity` behavior when reading/writing entities. `JsonSerializerSettings` isn't the smallest class, allocating it less won't hurt 😉


cgillum (approved):
Looks good to me! Thanks for these optimizations!

PR #625, opened 2021-10-14
Add CloudStorageAccount support in AzureTableInstanceStore

shwetabhartimsft:
Modify AzureTableInstanceStore & AzureTableClient

### DIFF
--- a/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
+++ b/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
@@ -48,13 +48,18 @@ namespace DurableTask.ServiceBus.Tracking
         volatile CloudTable jumpStartTable;
 
         public AzureTableClient(string hubName, string tableConnectionString)
-            :this(hubName, CloudStorageAccount.Parse(tableConnectionString))
+            : this(hubName, CloudStorageAccount.Parse(tableConnectionString))
         {
 
         }
 
         public AzureTableClient(string hubName, CloudStorageAccount cloudStorageAccount)
         {
+            if (string.IsNullOrWhiteSpace(hubName))
+            {
+                throw new ArgumentException("Invalid hub name", nameof(hubName));
+            }
+
             if (cloudStorageAccount == null)
             {
                 throw new ArgumentException("Invalid Cloud Storage Account", nameof(cloudStorageAccount));
--- a/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
+++ b/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
@@ -28,7 +28,6 @@ namespace DurableTask.ServiceBus.Tracking
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.RetryPolicies;
     using Microsoft.WindowsAzure.Storage.Table;
-    using Microsoft.WindowsAzure.Storage.Table.Protocol;
 
     internal class AzureTableClient
     {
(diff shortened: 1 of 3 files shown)


src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -49,27 +48,32 @@ internal class AzureTableClient
...
-                MaxRetries);
-            this.tableClient.DefaultRequestOptions.MaximumExecutionTime = MaximumExecutionTime;
-
+            this.tableClient = CreateAzureTableClient(cloudStorageAccount);
             this.hubName = hubName;
             this.historyTable = this.tableClient.GetTableReference(TableName);
             this.jumpStartTable = this.tableClient.GetTableReference(JumpStartTableName);
         }
 
+        public static CloudTableClient CreateAzureTableClient(CloudStorageAccount cloudStorageAccount)
```
zheg, line 69: need public?

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -49,27 +48,32 @@ internal class AzureTableClient
...
-            }
 
-            if (string.IsNullOrWhiteSpace(hubName))
+        }
+
+        public AzureTableClient(string hubName, CloudStorageAccount cloudStorageAccount)
+        {
+            if (cloudStorageAccount == null)
             {
-                throw new ArgumentException("Invalid hub name", nameof(hubName));
```
cgillum, line 60: It looks like you removed the null-check for `hubName`?

src/DurableTask.ServiceBus/Tracking/AzureTableInstanceStore.cs
```diff
@@ -45,12 +46,45 @@ public class AzureTableInstanceStore : IOrchestrationServiceInstanceStore
...
+        /// <summary>
+        /// Creates a new AzureTableInstanceStore using the supplied hub name and cloud storage account
+        /// </summary>
+        /// <param name="hubName">The hub name for this instance store</param>
+        /// <param name="cloudStorageAccount">Cloud Storage Account</param>
+        public AzureTableInstanceStore(string hubName, CloudStorageAccount cloudStorageAccount)
+        {
+            if (cloudStorageAccount == null)
+            {
+                throw new ArgumentException("Invalid CLoud Storage Account", nameof(cloudStorageAccount));
```
cgillum, line 74: nit: CLoud --> Cloud

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -49,27 +48,32 @@ internal class AzureTableClient
         volatile CloudTable jumpStartTable;
 
         public AzureTableClient(string hubName, string tableConnectionString)
+            :this(hubName, CloudStorageAccount.Parse(tableConnectionString))
```
cgillum, line 51: nit: add a space between `:` and `this`.

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
shwetabhartimsft: The idea is that only AzureTableInstanceStore ctor calls AzureTableClient ctor, and I am having a null check in AzureTableInstanceStore ctor.  AzureTableClient ctor is not called anywhere else. If there is a possibility that AzureTableClient  ctor can be called from anywhere else, then I can put a check for hub name. Please let me know. Thanks

src/DurableTask.ServiceBus/Tracking/AzureTableInstanceStore.cs
shwetabhartimsft: My bad, thanks for pointing.

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
cgillum: I think it would be good to keep the arg-null check. It protects against future changes and also makes the code more self-documenting (i.e., the reader can know for sure whether or not null values are acceptable when looking at a particular class).

shwetabhartimsft: Makes sense. Made the change.

cgillum (approved):
Looks good to me!

PR #627, opened 2021-10-15
Fix exception when Task returns void

bachuv:
This PR changes the behavior to return the default value of a type instead of a `NullReferenceException` when a user calls `CallActivityAsync()` with a typed call (e.g. `await context.CallActivityAsync<bool>("MyFunctions_RunTaskAsync", input)`) and expects a Task, but the function returns void.

Example from GitHub issue:
```
[FunctionName("MyFunctions_RunTaskAsync")]
public async Task RunTaskAsync([ActivityTrigger] string input)
{
    ...
}
```
```
await context.CallActivityAsync<bool>("MyFunctions_RunTaskAsync", input);
```
```
Function 'MyFunctions (Orchestrator)' failed with an error.Reason: System.NullReferenceException: Object reference not set to an instance of an object.
at DurableTask.Core.TaskOrchestrationContext.ScheduleTaskToWorker[TResult](String name, String version, String taskList, Object[] parameters) in C:\source\durabletask\src\DurableTask.Core\TaskOrchestrationContext.cs:line 80
at DurableTask.Core.TaskOrchestrationContext.ScheduleTask[TResult](String name, String version, Object[] parameters) in C:\source\durabletask\src\DurableTask.Core\TaskOrchestrationContext.cs:line 70
at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableOrchestrationContext.CallDurableTaskFunctionAsync[TResult](String functionName, FunctionType functionType, Boolean oneWay, String instanceId, String operation, RetryOptions retryOptions, Object input, Nullable`1 scheduledTimeUtc) in d:\a\r1\a\azure-functions-durable-extension\src\WebJobs.Extensions.DurableTask\ContextImplementations\DurableOrchestrationContext.cs:line 607
at MyNamespace.MyFunctions.RunOrchestrator(IDurableOrchestrationContext context)
```

With this fix, the function would return the default value for the type (e.g. 0 for int).

Resolves https://github.com/Azure/azure-functions-durable-extension/issues/1245

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -77,6 +77,11 @@ namespace DurableTask.Core
         {
             object result = await ScheduleTaskInternal(name, version, taskList, typeof(TResult), parameters);
 
+            if (result == null && typeof(TResult).IsPrimitive)
+            {
+                throw new InvalidOperationException("Task result is null and cannot be cast to TResult. Please use the non-generic overload method.");
+            }
+
             return (TResult)result;
         }
 
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -79,7 +79,7 @@ namespace DurableTask.Core
 
             if (result == null && typeof(TResult).IsPrimitive)
             {
-                throw new InvalidOperationException("Task result is null and cannot be cast to TResult. Please use the non-generic overload method.");
+                return default(TResult);
             }
 
             return (TResult)result;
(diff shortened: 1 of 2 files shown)


src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -77,6 +77,11 @@ internal void ClearPendingActions()
         {
             object result = await ScheduleTaskInternal(name, version, taskList, typeof(TResult), parameters);
 
+            if (result == null && typeof(TResult).IsPrimitive)
+            {
+                throw new InvalidOperationException("Task result is null and cannot be cast to TResult. Please use the non-generic overload method.");
```
cgillum, line 82: Another option we could consider is this:
```suggestion
                return default(TResult);
```

This is what we did to handle a very similar situation in another part of the code: https://github.com/Azure/durabletask/blob/f2dc50f392ac03a5389d60f582184819310ffb8a/src/DurableTask.Core/Serializing/DataConverter.cs#L52-L61

cgillum (approved):
LGTM2!

PR #631, opened 2021-10-22
StructuredLogEvent uses regular Dictionary now

Scooletz:
This PR changes the value type of `SharedPropertiesCache` from `ConcurrentDictionary` to a regular `Dictionary`. 

The reasoning for this is that once the dictionary is created it's only read. Additionally, `.Values` and `.Keys` might be used by the logger and for `ConcurrentDictionary` these are full blocking properties. With dictionary, they are not


cgillum (approved):
Looks good to me! Thanks for the contribution! Since this can be a high-traffic code path, I think it's very possible there could be some nice gains from this improvement.

PR #640, opened 2021-11-12
Updated AppLeaseManager hash algorithm to be FIPS compliant

### LINKED ISSUE #637: Latest DT.AzureStorage versions are not FIPS compliant
This [code](https://github.com/Azure/durabletask/blob/c56f066e6f9ef9a8455f584d7305de2cbad8824b/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs#L74-L78) in DurableTask.AzureStorage uses MD5 APIs to generate a hash. However, the use of MD5 APIs violates FIPS compliance rules and prevents certain users from being able to use the framework.

```
Description: The application requested process termination through System.Environment.FailFast(string message).
Message: RunAsync failed due to an unhandled exception causing the host process to crash: Autofac.Core.DependencyResolutionException: An exception was thrown while executing a resolve operation. See the InnerException for details. ---> Exception has been thrown by the target of an invocation. (See inner exception for details.) ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.
   at System.Security.Cryptography.MD5CryptoServiceProvider..ctor()
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Security.Cryptography.CryptoConfig.CreateFromName(String name, Object[] args)
   at System.Security.Cryptography.MD5.Create()
   at DurableTask.AzureStorage.Partitioning.AppLeaseManager..ctor(AzureStorageClient azureStorageClient, IPartitionManager partitionManager, String appLeaseContainerName, String appLeaseInfoBlobName, AppLeaseOptions options) in /_/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs:line 72
   at DurableTask.AzureStorage.AzureStorageOrchestrationService..ctor(AzureStorageOrchestrationServiceSettings settings, IOrchestrationServiceInstanceStore customInstanceStore) in /_/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs:line 176
```

We'll need to look into using some other API to generate this hash.

It's been suggested that we can test this by changing the following registry key and a Windows OS to reproduce the failure:

> Setting following in windows registry before running your test should be enough, I guess.
> `HKLM\System\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy\Enabled` and set `Enabled` to `1`

This might require a reboot, but not sure.

FYI @amdeel

amdeel:
AppLeaseManager is currently using MD5 to create a hash of the appName used for appLeaseId, which is not FIPS compliant. This PR replaces that hash algorithm with an existing algorithm we use to group instanceIds to partitions.

resolves #637

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
@@ -15,8 +15,6 @@ namespace DurableTask.AzureStorage.Partitioning
 {
     using System;
     using System.Net;
-    using System.Security.Cryptography;
-    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using DurableTask.AzureStorage.Storage;
@@ -71,7 +69,9 @@ namespace DurableTask.AzureStorage.Partitioning
             this.appLeaseContainer = this.azureStorageClient.GetBlobContainerReference(this.appLeaseContainerName);
             this.appLeaseInfoBlob = this.appLeaseContainer.GetBlobReference(this.appLeaseInfoBlobName);
 
-            this.appLeaseId = Fnv1aHashHelper.ComputeHash(this.appName).ToString();
+            var appNameHashInBytes = BitConverter.GetBytes(Fnv1aHashHelper.ComputeHash(this.appName));
+            Array.Resize(ref appNameHashInBytes, 16);
+            this.appLeaseId = new Guid(appNameHashInBytes).ToString();
 
             this.isLeaseOwner = false;
             this.shutdownCompletedEvent = new AsyncManualResetEvent();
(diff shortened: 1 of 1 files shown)


cgillum:
The code change itself looks good (I'm glad we're able to reuse existing code). However, what will happen when existing apps upgrade to this new version of the storage provider? If the app lease ID changes after an upgrade, will that cause any issues? For example, if a customer has a primary and backup region already established, and if they upgrade one of those regions initially, will that cause both primary and backup apps to think they own the lease since they're (at least temporarily) using different app lease IDs?

amdeel:
@cgillum I'll have to be really clear about this in the release notes.

When this change goes into effect and the app is restarted, neither the primary nor the backup will be able to immediately acquire the lease while the old leaseId still is considered the app owner. So, if the customer does nothing, there will be a delay while the old lease times out and another app can acquire the lease.

I added the ForceChangeAppLeaseAsync API a while back which will allow customers to decide which app the primary is and allow for an immediate switch. This means that when the customer updates their DurableTask.AzureStorage version, they will have to use this API on their desired primary app. Customers who use the AppLease feature might be doing this already.

However, it is a bit different for customers who might not intentionally be using this feature. Because UseAppLease defaults to true currently, customers could see a delay after updating if they don't use the ForceChangeAppLease API.

In the release notes, I will be clear about using that API to prevent the delay or suggesting setting UseAppLease to false temporarily.

cgillum (approved):

cgillum (approved):

PR #641, opened 2021-11-18
Add delegate function support construct for creator

ChenglongLiu:
Support add TaskOrchestrations like this
```
            this.AddTaskOrchestrations(
               new NameValueObjectCreator<TestCreatorOrchestration>("TestCreatorOrchestration", "V1", () =>
               {
                   var orc = new TestCreatorOrchestration();
                   // do something 
                   return orc ;
               }),
           );
```

### DIFF
--- a/src/DurableTask.Core/DefaultObjectCreator.cs
+++ b/src/DurableTask.Core/DefaultObjectCreator.cs
@@ -45,6 +45,16 @@ namespace DurableTask.Core
         }
 
         /// <summary>
+        /// Creates a new DefaultObjectCreator using delegate function
+        /// </summary>
+        /// <param name="func">Object instances to infer the type from</param>
+        public DefaultObjectCreator(Func<T> func)
+        {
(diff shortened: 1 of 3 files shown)


src/DurableTask.Core/DefaultObjectCreator.cs
```diff
@@ -44,6 +44,16 @@ public DefaultObjectCreator(T instance)
             Initialize(instance);
         }
 
+        /// <summary>
+        /// Creates a new DefaultObjectCreator using delegate function
+        /// </summary>
+        /// <param name="func">Object instances to infer the type from</param>
+        public DefaultObjectCreator(Func<T> func)
+        {
+            this.instance = func();
```
cgillum, line 53: I'm not sure I understand the benefit of this new overload. Without this, couldn't the caller just call `func()` themselves and pass in the result to the `DefaultObjectCreator(T instance)` constructor overload?

PR #643, opened 2021-11-21
Add Storage Service Url Support

wsugarman:
## Description
This changes adds the ability for users to specify the URL for various Azure Storage Account services manually, as supported by [Azure Functions here](https://sugarmandicomtest.blob.core.windows.net/). Specifically:

- Update the `StorageAccountDetails` class with the following changes:
  - Add new properties `BlobServiceUri`, `QueueServiceUri`, and `TableServiceUri` (whose names match the above Azure Functions documentation)
  - Update the method `ToCloudStorageAccount` to check for these properties when creating a new instance of `CloudStorageAccount`
- Add new utility class `StorageAccount` to help generate default service URLs if only some are specified

## Resolves
Helps resolve issue https://github.com/Azure/azure-functions-durable-extension/issues/1793 by providing feature parity with the core Azure Functions framework


src/DurableTask.AzureStorage/StorageService.cs
```diff
@@ -0,0 +1,36 @@
...
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage
+{
+    /// <summary>
+    /// Specifies a data service within the Azure Storage platform.
+    /// </summary>
+    public enum StorageService
+    {
+        /// <summary>
```
wsugarman, line 21: I used the descriptions [here](https://docs.microsoft.com/en-us/azure/storage/common/storage-introduction?toc=/azure/storage/blobs/toc.json#core-storage-services) as a basis.

src/DurableTask.AzureStorage/StorageAccount.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            string suffix = location switch
+            {
+                StorageLocation.Primary => string.Empty,
+                StorageLocation.Secondary => DefaultSecondarySuffix,
+                _ => throw new ArgumentOutOfRangeException(nameof(location), $"Unknown storage location '{location}'."),
+            };
+
+            return new Uri(
+                $"{DefaultScheme}://{accountName + suffix}.{service.ToString("G").ToLowerInvariant()}.{DefaultEndpointSuffix}",
```
wsugarman, line 60: We could also make the scheme configurable, but it's probably best not to encourage `http` over `https`

src/DurableTask.AzureStorage/StorageAccountDetails.cs
```diff
@@ -45,14 +46,39 @@ public sealed class StorageAccountDetails
...
         {
             if (!string.IsNullOrEmpty(this.ConnectionString))
             {
                 return CloudStorageAccount.Parse(this.ConnectionString);
             }
+            else if (this.BlobServiceUri != null || this.QueueServiceUri != null || this.TableServiceUri != null)
+            {
+                return new CloudStorageAccount(
+                    this.StorageCredentials,
+                    this.BlobServiceUri ?? StorageAccount.GetDefaultServiceUri(this.AccountName, StorageService.Blob),
```
wsugarman, line 77: Instead of passing only primary URLs for default, we _could_ provide default `StorageUri` objects instead with both a primary and second endpoint. However, I think this would be strange to give the default URLs both primary and secondary endpoints, as users can only provide primary endpoint overrides.

This behavior is currently equivalent to that for the [Azure Functions SDK](https://github.com/Azure/azure-webjobs-sdk/blob/830b86a1f1f03a7405315b0cc0b94a310d1df5ee/src/Microsoft.Azure.WebJobs.Host.Storage/BlobServiceClientProvider.cs#L93).

src/DurableTask.AzureStorage/StorageService.cs
```diff
@@ -0,0 +1,36 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage
+{
+    /// <summary>
+    /// Specifies a data service within the Azure Storage platform.
+    /// </summary>
+    public enum StorageService
```
cgillum, line 19: How about StorageServiceType?

src/DurableTask.AzureStorage/StorageAccount.cs
cgillum: Do we need to worry about local emulator support for either Azurite or the Storage Emulator (which I believe uses `http`)?

src/DurableTask.AzureStorage/StorageAccount.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace DurableTask.AzureStorage
+{
+    /// <summary>
+    /// Provides a set of <see langword="static"/> methods concerning Azure Storage Accounts.
+    /// </summary>
+    public static class StorageAccount
+    {
+        private const string DefaultEndpointSuffix = "core.windows.net";
+        private const string DefaultScheme = "https";
+        private const string DefaultSecondarySuffix = "-secondary";
```
cgillum, line 26: Maybe add a comment with a link to where the `-secondary` suffix is documented?

cgillum (commented):
Looks great! Just a few thoughts/questions.

src/DurableTask.AzureStorage/StorageAccount.cs
wsugarman: Added

wsugarman:
@cgillum - Regarding your comment about the storage emulator, I think you're right. We should support http, but our approach should probably be equivalent to that of the `CloudStorageAccount` class we're using. Looking at that, I think I may want to move these utility methods I've created to the durable functions lib.

The `CloudStorageAccont` allows users to create the object from a connection string, which is probably the easiest way to leverage the emulator:
```csharp
CloudStorageAccount.Parse("UseDevelopmentStorage=true")
```

Of course, the user could also manually pass the corresponding URLs, account key, etc. But there is no "backfilling" of URLs for different service URLs. If a user decides to pass endpoint URLs themselves, no additional URLs are generated.

I initially added these utility methods and backfilling to match the current behavior in the WebJobs SDK. I am unsure if this is an intended feature, but since the URLs appear to be evaluated independently, a user _could_ specify just the blob service URL, but when a component needs to interact with the queue service a default URL is generated (if account name is present). If we want to preserve this behavior for durable functions, I could move these default URLs up to the durable functions extension and leave the behavior for durable task closer to `CloudStorageAccount`. Or I could emulate the behavior in both places.

src/DurableTask.AzureStorage/StorageAccount.cs
wsugarman: @cgillum - Regarding your comment about the storage emulator, I think you're right. We should support http, but our approach should probably be equivalent to that of the `CloudStorageAccount` class we're using. Looking at that, I think I may want to move these utility methods I've created to the durable functions lib.

The `CloudStorageAccont` allows users to create the object from a connection string, which is probably the easiest way to leverage the emulator:
```csharp
CloudStorageAccount.Parse("UseDevelopmentStorage=true")
```

Of course, the user could also manually pass the corresponding URLs, account key, etc. But there is no "backfilling" of URLs for different service URLs. If a user decides to pass endpoint URLs themselves, no additional URLs are generated.

I initially added these utility methods and backfilling to match the current behavior in the WebJobs SDK. I am unsure if this is an intended feature, but since the URLs appear to be evaluated independently, a user _could_ specify just the blob service URL, but when a component needs to interact with the queue service a default URL is generated (if account name is present). If we want to preserve this behavior for durable functions, I could move these default URLs up to the durable functions extension and leave the behavior for durable task closer to `CloudStorageAccount`. Or I could emulate the behavior in both places.

cgillum: @wsugarman I'm simply asking whether your PR allows us to continue having storage emulator support. I wasn't able to determine the answer from your response. Apologies if I'm just missing some context.

wsugarman: Sorry -- my answer was a bit all over the place. To answer your question: yes. Users can still use the storage emulator with the following code:
```csharp
// Connection String
new StorageAccountDetails { ConnectionString = "UseDevelopmentStorage=true" }

// If you enjoy lots of typing
new StorageAccountDetails
{
    StorageCredentials = new StorageCredentials(
        "devstoreaccount1",
        "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="),
    BlobServiceUri = new Uri("http://127.0.0.1:10000/devstoreaccount1", UriKind.Absolute),
    QueueServiceUri = new Uri("http://127.0.0.1:10001/devstoreaccount1", UriKind.Absolute),
    TableServiceUri = new Uri("http://127.0.0.1:10002/devstoreaccount1", UriKind.Absolute),
}
```

wsugarman: The longer point I was discussing was whether or not this library should attempt to backfill the service URIs if only a subset are specified. And I've come to the conclusion: no (at least for the Durable Task library). If a user created a `CloudStorageAccount` object without the blob storage URI and then attempted to access blob storage, they would be met with an error like `"No blob endpoint configured"`. And so it seems reasonable to be just as strict here (we just throw the error earlier).

However, I think this "backfilling" will be necessary for the durable functions extension as it seems to be a (perhaps unintentional) feature for the existing Azure Functions identity-based connection implementation.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -5,7 +5,7 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.9.4</FileVersion>
+    <FileVersion>1.10.0</FileVersion>
```
wsugarman, line 8: I bumped the minor version because I added new properties. Let me know if you'd like a different version.

src/DurableTask.AzureStorage/StorageAccount.cs
cgillum: ok, thanks. This answers my question.

cgillum (approved):
Looks good to me!

PR #646, opened 2021-11-29
Improve FetchedInstanceStatus logs

cgillum:
I was debugging a customer issue and noticed that it would be really nice if the _FetchedInstanceStatus_ log event had more information. This PR adds the following columns:

* **ExecutionId**: We already had this column, but it was always empty. This PR populates it with whatever execution ID we actually found in table storage.
* **RuntimeStatus**: This column is new and includes the status of the found orchestration instance or "NotFound" if none was found. This is helpful for debugging issues related to dedupe or for simply verifying what the status of an instance really is.

This is just a quality-of-life improvement for debugging.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -424,9 +424,9 @@ namespace DurableTask.AzureStorage.Tracking
         {
             return this.FetchInstanceStatusInternalAsync(instanceId, executionId: null, fetchInput: false);
         }
-
+#nullable enable
         /// <inheritdoc />
-        async Task<InstanceStatus> FetchInstanceStatusInternalAsync(string instanceId, string executionId, bool fetchInput)
+        async Task<InstanceStatus?> FetchInstanceStatusInternalAsync(string instanceId, string executionId, bool fetchInput)
         {
             if (instanceId == null)
             {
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -424,9 +424,9 @@ public override Task<InstanceStatus> FetchInstanceStatusAsync(string instanceId)
         {
             return this.FetchInstanceStatusInternalAsync(instanceId, executionId: null, fetchInput: false);
         }
-
+#nullable enable
         /// <inheritdoc />
-        async Task<InstanceStatus> FetchInstanceStatusInternalAsync(string instanceId, string executionId, bool fetchInput)
+        async Task<InstanceStatus?> FetchInstanceStatusInternalAsync(string instanceId, string executionId, bool fetchInput)
```
davidmrdavid, line 429: Since we're enabling nullable references for this method, could we use that to also rule-out `instanceId` being null? I'm asking since there's a check in line 431 for whether `instanceId` is null

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -441,24 +441,30 @@ async Task<InstanceStatus> FetchInstanceStatusInternalAsync(string instanceId, s
...
+            {
+                orchestrationState = await this.ConvertFromAsync(tableEntity);
+            }
+
             this.settings.Logger.FetchedInstanceStatus(
                 this.storageAccountName,
                 this.taskHubName,
                 instanceId,
-                executionId ?? string.Empty,
+                orchestrationState?.OrchestrationInstance.ExecutionId ?? executionId ?? string.Empty,
```
davidmrdavid, line 456: Just for my own learning, what's the difference between `orchestrationState.OrchestrationInstance.ExecutionId` and `executionId`?

Also, while I realize this is slightly outside the scope of this PR, I would appreciate if this difference could be made explicit in the code either via a comment or by renaming the variables slightly :)

davidmrdavid (approved):
LGTM but I've also left some non-blocking questions and suggestions

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: Since I'm limiting the scope of nullable reference types for just this method, we're not able to use it to check whether any callers of this method are potentially passing a null value for `instanceId`. We'd need to expand the scope of nullable reference types to include all callers of this method, which I didn't want to do for this small PR.

That said, your comment reminded me that we should put a nullable annotation on the `executionId` parameter since callers usually pass `null` here.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
amdeel: The executionId off of orchestrationState is the result returned from the table query. While in theory it should be the same as executionId, the parameter passed in, its not actually used in the table query so orchestrationState.ExecutionId would be the actual value. I think this is fine personally. You could make an argument for it being redundant, but I really don't think someone reading this code would need a comment.

cgillum: Now that you mention it, we don't actually use `executionId` at all in this method! I'll just remove it. I think the original idea was for it to be a filter parameter, but it's clearly not being used for this. Since I'm removing it, I might also increase the scope of the nullable reference types too.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: Update on this comment: I'm actually removing `executionId` since it's not being used. Regarding `instanceId`, we never expect it to be null so I don't want to add a nullable annotation to it. However, we still need to guard against a null value since public APIs are feeding into this method and we can't guarantee that those external callers are not passing in null values.

Issue #596, opened 2021-08-23
TaskHubWorker to implement IHostedService


cgillum:
This would make it easier for .NET developers to host `TaskHubWorker` instances in [.NET background worker](https://docs.microsoft.com/aspnet/core/fundamentals/host/hosted-services#ihostedservice-interface) projects. It also more closely aligns DTFx with modern .NET design patterns.

In terms of implementation, this shouldn't be very difficult. It's mostly a matter of adding `StartAsync` and `StopAsync` overloads that take a `CancellationToken` parameter. As part of this, we could consider passing down these cancellation tokens to inner StartAsync and StopAsync calls.

PR #647, opened 2021-11-30
Implement IHostedService interface for TaskHubworker

ChenglongLiu:
A draft PR of adding IHostedService  implement in TaskHubworker #596

### DIFF
--- a/src/DurableTask.Core/TaskHubWorker.cs
+++ b/src/DurableTask.Core/TaskHubWorker.cs
@@ -20,13 +20,14 @@ namespace DurableTask.Core
     using System.Threading.Tasks;
     using DurableTask.Core.Logging;
     using DurableTask.Core.Middleware;
+    using Microsoft.Extensions.Hosting;
     using Microsoft.Extensions.Logging;
 
     /// <summary>
     ///     Allows users to load the TaskOrchestration and TaskActivity classes and start
     ///     dispatching to these. Also allows CRUD operations on the Task Hub itself.
     /// </summary>
-    public sealed class TaskHubWorker : IDisposable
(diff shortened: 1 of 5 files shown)


src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -170,8 +182,8 @@ public async Task<TaskHubWorker> StartAsync()
                     this.logHelper);
 
                 await this.orchestrationService.StartAsync();
```
ChenglongLiu, line 184: the `orchestrationService` also should implement `IHostedService` to support cancel token

cgillum (commented):
These changes seem fine to me and I'm glad to see the public API get a bit more modernized. Some of the duplication around the StopAsync implementations seem unfortunate, but I understand that it's probably hard to reuse some of the existing logic.

ChenglongLiu:
> These changes seem fine to me and I'm glad to see the public API get a bit more modernized. Some of the duplication around the StopAsync implementations seem unfortunate, but I understand that it's probably hard to reuse some of the existing logic.

ChenglongLiu:
> /azp run

A small update of the while loop. @cgillum your last `/azp run` didn't run successfully. Are there any restrictions on branch names?

```
Azure Pipelines could not run because the pipeline triggers exclude this branch/path.
```

jviau:
I would encourage to add this and other integrations into `Microsoft.Extension` ecosystem in a separate package(s). Something like `DurableTask.Extensions.*`. As these should be optional enhancements and DTFx should not force these dependencies on consumers unless they want it.

cecilphillip:
@ChenglongLiu  Did this end up somewhere else?

PR #648, opened 2021-12-01
Distributed Tracing v2

cgillum:
## Description

This PR adds distributed tracing support that is compatible with all storage providers: Azure Storage, Microsoft SQL, and Netherite. The implementation is a storage provider-agnostic and can therefore be used with any storage provider that knows how to fully serialize DTFx history events.

### What about the existing Distributed Tracing (Correlation) support?
This PR is intended to replace the [previous Distributed Tracing implementation](https://github.com/Azure/durabletask/tree/main/samples/Correlation.Samples#readme), which had a few important shortcomings:
* Only worked with the Azure Storage orchestration service provider.
* Required a potentially large amount of storage. In some cases, the required storage was [unbounded](https://github.com/Azure/azure-functions-durable-extension/issues/1999#issuecomment-970931979).

This new implementation addresses these shortcomings and adds new capabilities, as described below.

image

### Is this compatible with Azure Functions?

Yes, with these changes, orchestrations and activities have the correct trace context set automatically and logs that are sent to Application Insights will be correctly correlated. No additional changes are required for basic correlation. 

## PR Progress
This PR is being developed incrementally. Below is the list of tasks, which will be checked off as they are completed:
- [x] Support for ExecutionStarted (orchestrator) and TaskScheduled (activity) events
- [x] Support for EventRaised (external events)
- [x] Support for Timer events
- [x] Support for Sub-Orchestration events
- [x] Reuse activities to show an accurate representation of orchestration executions
- [x] Support separate spans for scheduling and executing functions (client and server spans)
- [x] Added descriptive tags and Activity.ActivityKind values
- [x] Integration tests that cover more orchestration scenarios (reties, continue-as-new, etc.)
- [x] Bump DT.Core to v2.7.0 since this represents new surface area
- [x] Added a DurableTask.ApplicationInsights package to convert Activitys to App Insights specific telemetry so the Functions Host can read the information
- [x] Add a new sample project that works with Application Insights
- ~~[ ] Remove previous implementation~~ (We will remove the preview implementation after GA in a separate release)


davidmrdavid:
I gave this an initial look and it seems fairly clear to me! At the risk of expanding the scope of this PR, but given the fact that this is intending to replace our existing implementation, it'd be nice if this new design also made it easier to test and mock correlation tracing edge cases.

For example, in the current design, a lot of the behavior is dependent of `CorrelationTraceClient` which, as a static class, is  tricky to _mock_ and to alter its behavior for unit testing purposes. I wonder, for example, if the CorrelationTraceClient and helpers alike could be injected via DI, as that'd make them easier to replace the implementation on a test.

Just some ideas!

cgillum:
I can look into using an object that can be passed around in place of static TraceHelper methods. The hard part about this is that `Activity` objects cannot be created with particular values, so I worry that mocking might not be very useful even if we do the work to enable it. We could try to create an abstraction around `Activity` that can be mocked, but then I worry that we might be dramatically increasing complexity for questionable benefit. I'll keep an eye out for some potential low-hanging-fruit improvements around this as I work through the remaining implementation.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -298,8 +299,15 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
             OrchestrationRuntimeState originalOrchestrationRuntimeState = runtimeState;
 
-            OrchestrationState instanceState = null;
+            // Distributed tracing support: each orchestration execution is a trace activity
+            // that derives from an established parent trace context. It is expected that some
+            // listener will receive these events and publish them to a distributed trace logger.
+            ExecutionStartedEvent startEvent =
+                runtimeState.ExecutionStartedEvent ??
+                workItem.NewMessages.Select(msg => msg.Event).OfType<ExecutionStartedEvent>().FirstOrDefault();
+            using Activity traceActivity = TraceHelper.StartTraceActivityForExecution(startEvent);
```
jviau, line 308: The behavior of activities for orchestrations, especially during replays, was one of my main hurdles when consuming the existing DTFx tracing. Primarily that the activity was only set on the first execution of the orchestration.

Can you describe how it will behave in this implementation? My questions are:
1) Will the activity be set for each replay of the orchestration?
1) If it is set for each replay, will it have the same spanId?

Through some middleware I was able to force the Activity to exist for each replay of the orchestration, and I used reflection to set the SpanId of the activity to force it to be consistent on each activity. This is very helpful for logs emitted during the orchestration to pick up the correct parent Activity.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -548,6 +562,11 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                 instanceState.Status = runtimeState.Status;
             }
 
+            // Add the runtime status to the trace context so that listeners can know
+            // whether the orchestration is still running. Also, stop the activity here so
+            // that we don't include the time required to persist the side effects.
+            traceActivity?.SetTag("dt.runtimestatus", runtimeState.OrchestrationStatus.ToString());
```
jviau, line 568: If you upgrade to DiagnosticSource 6.0, I believe `Activity` has some new properties around `Status` and `StatusDescription`. The OTEL Sdk also has some extension methods for setting status their own way.

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
```diff
@@ -0,0 +1,53 @@
...
+#nullable enable
+namespace DurableTask.Core.Tracing
+{
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// W3C-compliant distributed trace context.
+    /// Spec: https://www.w3.org/TR/trace-context/.
+    /// </summary>
+    public class DistributedTraceContext
```
jviau, line 22: `Baggage` would also be very helpful to include here. This would users propagate some custom metadata to include in logs. Keep in mind there are _two_ sources baggage to consider: `Activity.Baggage` and then otel has their own `Baggage` class.

jviau (commented):
Very happy to see OTEL support here! An additional question/concern:

What are the plans for tracking exceptions? otel does have a `Activity.SetException` extension method. Would that be appropriate to use? Consider also including the exception itself on the Activity - maybe via `SetCustomObject`? This way during the stop event, people can pull it off and enrich the Activity from the exception. I wonder if OTEL has a pattern for this already?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -26,6 +28,91 @@ public class TraceHelper
     {
         const string Source = "DurableTask";
 
+        // TODO: Add tracing for external event send, which could be used to initialize an entity
+        static readonly ActivitySource ActivityTraceSource = new ActivitySource(Source);
```
cijothomas, line 32: consider Fully qualified name instead of "DurableTask"

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -26,6 +28,91 @@ public class TraceHelper
...
+
+        internal static Activity? CreateActivityForNewOrchestration(ExecutionStartedEvent startEvent)
+        {
+            Activity? newActivity = ActivityTraceSource.StartActivity(
+                name: startEvent.Name,
+                kind: ActivityKind.Internal,
+                parentContext: Activity.Current?.Context ?? default,
+                tags: new KeyValuePair<string, object?>[]
+                {
+                    new("dt.type", "client"),
```
cijothomas, line 42: Could this be better captured in ActivityKind.Client?

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DistributedTraceContext"/> class.
+        /// </summary>
+        /// <param name="traceParent">The W3C traceparent ID.</param>
+        /// <param name="traceState">The optional W3C tracestate data.</param>
+        public DistributedTraceContext(string traceParent, string? traceState = null)
```
cijothomas, line 29: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/ActivityContext.cs

Do you need to expose a new class?

cijothomas:
> I can look into using an object that can be passed around in place of static TraceHelper methods. The hard part about this is that `Activity` objects cannot be created with particular values, so I worry that mocking might not be very useful even if we do the work to enable it. We could try to create an abstraction around `Activity` that can be mocked, but then I worry that we might be dramatically increasing complexity for questionable benefit. I'll keep an eye out for some potential low-hanging-fruit improvements around this as I work through the remaining implementation.

You could consider using `InMemoryExporter<Activity>` from OpenTelemetry for testing. Or setup own ActivityListener as well.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: FYI @bachuv I think we should explore reusing the same Activity/SpanID for each orchestration replay. It will require custom reflection (I think Jacob shared some pointers to code on how he does this), but it will also create a nicer visual where users can easily see both the start and the end of an orchestration's execution.

jviau:
@cgillum we should propose a new `Part B` section to [CommonSchema v4.0](https://github.com/microsoft/common-schema) as well as a semantic convention for [open telemetry itself](https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions/trace).

This would involve deciding what properties in these traces are important are giving them well defined names and meanings.

For starters, we can select a namespace for these properties. I propose `dtfx`. Then we can choose some properties we want:

| Name | Description | Example |
| ------- | ------------ | --------- |
| `dtfx.type` | The type of durable task | Enum. Only one of `orchestration` or `activity` |
| `dtx.name` | The name of the orchestration or activity. | `GreetingOrchestration` |
| `dtfx.version` | The version of the orchestration or activity | `1.0`, can be null |

We may also want something to indicate success. Do we want just a `dtfx.success` with `true`/`false`? Or do we want something more granular like `dtfx.result_code` for open ended strings - or even use HTTP status codes there (we are using HTTP status codes for our orchestrations, as it helps quickly identify known failures. Or differentiate client vs server error.) OTel and CommonSchema already have exception semantics, so those can also help out here.

jviau:
@cgillum , @bachuv can we come up with a spec on what the complete expectations of activities / traces DTFx will emit? This should cover the names of all activities, what they represent, what tags, links, events they will potentially have. Having this will be very helpful in defining what the expected outcome of this work is.

To support why I think this valuable, here is something that may not be considered in this work yet: activity links between DTFx events. What I mean by this is: sending an event to an orchestration is a `Producer` span. I think that producer activity should show up as an `ActivityLink` in the target orchestration - we can then add a tag to that link for the name of the event sent.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -26,6 +29,171 @@ public class TraceHelper
...
+            }
+
+            string activityName = startEvent.Name;
+            ActivityKind activityKind = ActivityKind.Internal;
+            KeyValuePair<string, object?>[] activityTags = new KeyValuePair<string, object?>[]
+            {
+                new("dtfx.type", "orchestrator"),
+                new("dtfx.instance_id", startEvent.OrchestrationInstance.InstanceId),
+                new("dtfx.execution_id", startEvent.OrchestrationInstance.ExecutionId),
+            };
```
RohitRanjanMS, line 91: Consider setting tags after StartActivity. If the activity returned is null, we can avoid any allocation.

RohitRanjanMS: Also, consider honoring IsAllDataRequested flag.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: Resolving this comment as we currently can't upgrade. Tracking this in #698

src/DurableTask.Core/Tracing/TraceHelper.cs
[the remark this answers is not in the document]
bachuv: Updated to "dtfx"

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -26,6 +28,91 @@ public class TraceHelper
...
+                return null;
+            }
+
+            return ActivityTraceSource.StartActivity(
+                name: startEvent.Name,
+                kind: ActivityKind.Internal,
+                parentContext: activityContext,
+                tags: new KeyValuePair<string, object?>[]
+                {
+                    new("dt.type", "orchestrator"),
```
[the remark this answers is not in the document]
bachuv: I believe these values were chosen to represent terms usually used in the Durable Functions context.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: This Activity can get created internally in an application and not externally.

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
bachuv: Can you explain why we need to expose a new class?

jviau: I think Chris is saying we can just re-use `ActivityContext` and not expose our own class at all.

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau: That is not really what `Internal` means for activities. Internal means this activity conceptually is completely internal to this application, it was created as a result of some incoming action, and it also is not primarily responsible for sending out some external action.

In DTFx, very few activities should actually be `Internal`, as they tend to all have some external component to them - primarily around the orchestration service. They are either being created because work was popped off the orchestration service (`server`), or they are sending new work to the orchestration service (`client`, or `producer` in the case of starting a new orchestration via `TaskHubClient`).

lsl-2020:
Is this PR abandoned? Any updates on this topic?

jviau:
@lsl-2020 we are actively working on OpenTelemetry support through other PRs into this branch. We are also working on having a detailed spec available to the community before we lock anything in place.

lsl-2020:
> @lsl-2020 we are actively working on OpenTelemetry support through other PRs into this branch. We are also working on having a detailed spec available to the community before we lock anything in place.

Hi @jviau , thanks for your kind reply. Actually I am working on a PR to enable DT in ASP and Azure Functions with durable extensions for my team in Microsoft. Could you please provide some links to the work you referred or connect me to the corresponding team? Much appreciated!

jviau:
I would be an appropriate contact for that - you can reach out to me internally and we can discuss it.

This work will cover distributed tracing in Durable Functions as well. We are also hoping to address OpenTelemetry support in Azure Functions as a whole. In the meantime, there is only so much you can do on your own for DT in durable functions and it is very involved work. I am happy to share some stuff I have done for my previous team with DurableTask (not functions) though.

src/DurableTask.Core/Tracing/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+            s_setSpanId = typeof(Activity).GetField("_spanId", flags).CreateSetter<Activity, string>();
+            s_setId = typeof(Activity).GetField("_id", flags).CreateSetter<Activity, string>();
+            s_setStatus = CreateSetStatus();
+        }
+
+        public static void SetId(this Activity activity, string id)
+            => s_setId(activity, id);
+        public static void SetSpanId(this Activity activity, string spanId)
+            => s_setSpanId(activity, spanId);
+        public static void SetStatus(this Activity activity, ActivityStatusCode status, string description)
```
bachuv, line 53: Just adding a note that we used reflection to write the SetStatus() method since we currently cannot upgrade to the latest version of System.Diagnostics.DiagnosticSource because of the issue described in #698. The current version of System.Diagnostics.DiagnosticSource is 5.0.1.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -26,6 +29,454 @@ public class TraceHelper
...
+        /// </returns>
+        internal static Activity? StartActivityForNewOrchestration(ExecutionStartedEvent startEvent)
+        {
+            Activity? newActivity = ActivityTraceSource.StartActivity(
+                name: CreateSpanName("create_orchestration", startEvent.Name, startEvent.Version),
+                kind: ActivityKind.Producer);
+
+            if (newActivity != null)
+            {
+                newActivity.SetTag(Schema.Task.Type, "orchestration");
```
davidmrdavid, line 49: I notice we have these constant strings that appear often like `"orchestration"` and `"create_orchestration"` that will probably become keys in the traces that users may want to filter on. To avoid typos and keep them consistent, should we consider setting them as constants somewhere and referencing them instead of inlining them?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -26,6 +29,454 @@ public class TraceHelper
...
+        {
+            Activity? newActivity = ActivityTraceSource.StartActivity(
+                name: CreateSpanName("orchestration_event", eventRaisedEvent.Name, null),
+                kind: ActivityKind.Producer,
+                parentContext: Activity.Current?.Context ?? default);
+
+            if (newActivity == null)
+            {
+                return null;
+            }
```
davidmrdavid, line 388: I notice we are utilizing `null` as return values a lot, particularly in this file, which worries me due to the potential of nullref-exceptions. Is there something about how Distributed Tracing works that makes these null-cases common?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -201,6 +201,7 @@ async Task OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem)
                     () =>
                     {                
                         // Check if it is extended session.
+                        // TODO: Remove this code - it looks incorrect and dangerous
                         isExtendedSession = this.concurrentSessionLock.Acquire();
```
davidmrdavid, line 205: Was this TODO resolved yet? Is it related to Dist. Tracing or just something you spotted while going through the code?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2227,6 +2232,288 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
...
+
+                    var client = await host.StartOrchestrationAsync(typeof(Orchestrations.SayHelloWithActivity), "World");
+                    var status = await client.WaitForCompletionAsync(StandardTimeout);
+
+                    await host.StopAsync();
+                }
+            }
+
+            // (1) Explanation about indexes:
+            // The orchestration Activity's start at Invocation[1] and each action logs
```
davidmrdavid, line 2264: Here, when we say "Activity" we mean a Dist. Tracing Activity and not a DF Activity, correct? I'm having a bit of a hard time context switching between these two concepts when looking at the code. Most of the time, I'm able to catch when we're talking about a different Activity concept, but it requires careful attention.  I think it might get trickier once this is merged and it's not so clear when the Dist. Tracing code starts and the rest of the DTFx continues.

I wonder if we could establish a convention to be able to differentiate between the two, at least in comments. Do other people think this is something to improve readability-wise or just me? If it's just me, I won't push on this!

davidmrdavid (commented):
Left some comments, but mostly questions to try and grok the implementation so I'm prepared for future maintainability. Overall looks good, thanks for this work.

samples/DistributedTraceSample/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                input: null);
+            await client.WaitForOrchestrationAsync(orchestrationInstance, TimeSpan.FromMinutes(5));
+        }
+
+        static async Task<(IOrchestrationService, IOrchestrationServiceClient)> GetOrchestrationServiceAndClient()
+        {
+            var settings = new AzureStorageOrchestrationServiceSettings
+            {
+                TaskHubName = "OpenTelemetrySampleTaskHub",
+                StorageConnectionString = "UseDevelopmentStorage=true",
```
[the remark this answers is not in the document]
bachuv: Yes, the customer would have to replace this if they were deploying an app using DTFx. Is there a way to configure storage accounts for apps using DTFx similar to how we have "local.settings.json" for Durable Functions?

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -423,7 +423,7 @@ public static bool IsFatal(Exception exception)
             }
 
             TraceHelper.Trace(TraceEventType.Error, "ExecuteWithRetry-RetriesExhausted", "Exhausted all retries for operation " + operation);
-            TraceHelper.TraceExceptionSession(TraceEventType.Error, "ExecuteWithRetryRetriesExhausted", sessionId, lastException).Throw();
+            TraceHelper.TraceExceptionSession(TraceEventType.Error, "ExecuteWithRetryRetriesExhausted", sessionId, lastException!).Throw();
```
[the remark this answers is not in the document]
bachuv: Yes

src/DurableTask.Core/History/EventRaisedEvent.cs
```diff
@@ -48,5 +50,11 @@ public EventRaisedEvent(int eventId, string input)
         /// </summary>
         [DataMember]
         public string Input { get; set; }
+
+        /// <summary>
+        /// The W3C trace context associated with this event.
```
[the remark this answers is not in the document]
bachuv: OpenTelemetry supports W3C by default.

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -18,6 +18,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.6" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <PackageReference Include="WindowsAzure.Storage" version="9.3.3" />
+    <PackageReference Include="Azure.Monitor.OpenTelemetry.Exporter" Version="1.0.0-beta.3" />
```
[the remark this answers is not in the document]
bachuv: Yes, the latest version is 1.0.0-beta.9.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: Yes, I can add them to the constants file.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: Yes, if there is no listener (i.e. the user doesn't subscribe to listening from the "DurableTask" source) then `Activity`s are not created/started. For example, `ActivityTraceSource.StartActivity()` will return null.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
bachuv: Yes, I started using "Task activity" when describing a DF activity and "Activity" when describing a Distributed Tracing activity.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I believe this is from @cgillum in a very early commit. @cgillum, is this something that has been fixed? I can create a GitHub issue for it if not.

cgillum: Let's leave this for now. We can remove it when we remove the old distributed tracing implementation.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
jviau: It is confusing with such an overloaded term as `Activity`. We could refer to it as `Span` - which is what OTel calls it.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -607,16 +608,31 @@ public Task<OrchestrationInstance> CreateOrchestrationInstanceAsync(string name,
...
+            }
 
             if (eventData != null)
             {
                 string serializedEventData = this.defaultConverter.Serialize(eventData);
                 var eventRaisedEvent = new EventRaisedEvent(-1, serializedEventData) { Name = eventName };
+                if (newActivity != null)
+                {
+                    // TODO: Set the parent trace context?
+                }
```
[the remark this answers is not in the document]
bachuv: The parent is actually set in `TraceHelper.StartActivityForNewOrchestration(startedEvent)` when the Activity is created. I removed this comment. If a parent isn't specified when creating an activity (like below), then `Activity.Current` is set as the parent.

```
    name: CreateSpanName("create_orchestration", startEvent.Name, startEvent.Version),
    kind: ActivityKind.Producer);
```

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,140 @@
...
+                Sample = (ref ActivityCreationOptions<ActivityContext> _) => ActivitySamplingResult.AllData,
+                SampleUsingParentId = (ref ActivityCreationOptions<string> _) => ActivitySamplingResult.AllData
+            };
+
+            ActivitySource.AddActivityListener(this.listener);
+        }
+
+        void OnEndActivity(Activity activity)
+        {
+            if (!activity.IsAllDataRequested)
```
jviau, line 54: I think we need to look into if this is the right behavior. if `IsAllDataRequested` is false, does that mean do not emit the span at all, or does that mean to just not populate the span with all of the information? And thus we still emit a span, but it has minimal tags/details.

I _think_ it is the later. `IsAllDataRequested == false` essentially amounts to "do not enrich the activity with all the extra tags." This is fine for preview, but we should probably fix it before GA.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -130,7 +130,13 @@ public async Task<string> FetchLargeMessageIfNecessary(string message)
 
         internal static bool TryGetLargeMessageReference(string messagePayload, out Uri blobUrl)
         {
-            return Uri.TryCreate(messagePayload, UriKind.Absolute, out blobUrl);
+            if (Uri.IsWellFormedUriString(messagePayload, UriKind.Absolute))
```
jviau, line 133: this change seems unrelated to telemetry, do you have any background on it?

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Diagnostics;
+using System.Threading;
+using Newtonsoft.Json;
+
+namespace DurableTask.Core.Tracing
+{
+    /// <summary>
+    /// Manage Activity for orchestration execution.
+    /// </summary>
+    internal class DistributedTraceActivity
```
jviau, line 24: Is this type used anywhere?

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
jviau: Let's make a note of this as follow up work.

jviau (commented):
This looks good for preview. However, half of this is going to be in a package (`DurableTask.Core`) which is not "preview", and you can't just say that functionality or types within a non-preview package are preview (that violates SemVer). So, we should figure out how we want to preview this which follows correct SemVer.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
bachuv: This is a good point! The definition in the [Activity.IsAllDataRequested Property doc](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.isalldatarequested?view=net-7.0) is "gets or sets a value that indicates whether this activity should be populated with all the propagation information, as well as all the other properties, such as links, tags, and events." This matches what you described above. I'll add it to the list of changes.

src/DurableTask.AzureStorage/MessageManager.cs
bachuv: Yes, this was actually added in https://github.com/Azure/durabletask/pull/853# because exceptions weren't getting formatted correctly. Information about this change is in [this comment](https://github.com/Azure/durabletask/pull/853#discussion_r1087081311). I moved these changes to #867 and they've already been merged so I think I need to merge `main` to this PR to remove this code.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
bachuv: `DistributedTraceActivity.Current` is used to keep track of the orchestration execution.

bachuv:
> This looks good for preview. However, half of this is going to be in a package (`DurableTask.Core`) which is not "preview", and you can't just say that functionality or types within a non-preview package are preview (that violates SemVer). So, we should figure out how we want to preview this which follows correct SemVer.

Thanks for pointing this out! Tagging @lilyjma -- do you have any recommendations?

jviau (approved):

Issue #650, opened 2021-12-01
Orchestrations stuck in Pending when using mismatched versions of DurableTask.AzureStorage


cgillum:
### Issue
When using DT.AzureStorage v1.8.4 or earlier to start orchestrations that get picked up by a worker running DT.AzureStorage v1.8.5 or later, the orchestrations will get permanently stuck in a "Pending" state.

### Root Cause
This is a regression caused by the orchestration de-duplication start PR: https://github.com/Azure/durabletask/pull/528. The de-dupe detection assumes that newly started orchestrations will have an ExecutionId field in the Instances table, which is true for v1.8.5+ clients. However, if this field doesn't exist, the de-dupe logic incorrectly flags the start message as a duplicate and discards it.

### Workaround
When upgrading from DT.AzureStorage v1.8.4 or earlier to a newer version of DT.AzureStorage, you must either ensure all clients and workers are upgraded together or ensure that all clients are upgraded BEFORE any workers get upgraded.

PR #651, opened 2021-12-01
Fix compatibility issue between older clients and newer workers on DT.AzureStorage

cgillum:
Fixes https://github.com/Azure/durabletask/issues/650.

The fix is to essentially short-circuit out of de-dupe detection if the Instances table has a record with a null value for ExecutionId. This will be true only for orchestrations that are started with clients on v1.8.4 or earlier (ExecutionId wasn't added until after the orchestration checkpointed once). For newer clients that do add ExecutionId when scheduling an orchestration, de-dupe detection should continue to work as normal - no change in behavior.

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
@@ -238,8 +238,8 @@ namespace DurableTask.AzureStorage
                 if (remoteOrchestrationsById.TryGetValue(localInstance.InstanceId, out OrchestrationState remoteInstance) &&
                     (remoteInstance.OrchestrationInstance.ExecutionId == null || string.Equals(localInstance.ExecutionId, remoteInstance.OrchestrationInstance.ExecutionId, StringComparison.OrdinalIgnoreCase)))
                 {
-                    // Happy path: The message matches the table status -OR- the table doesn't have an execution ID field (older clients, pre-v1.8.5),
-                    // in which case we have no way of knowing whether it's a duplicate. Allow it to run.
+                    // Happy path: The message matches the table status. Alternatively, if the table doesn't have an ExecutionId field (older clients, pre-v1.8.5),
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -236,9 +236,10 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
             {
                 OrchestrationInstance localInstance = message.TaskMessage.OrchestrationInstance;
                 if (remoteOrchestrationsById.TryGetValue(localInstance.InstanceId, out OrchestrationState remoteInstance) &&
-                    string.Equals(localInstance.ExecutionId, remoteInstance.OrchestrationInstance.ExecutionId, StringComparison.OrdinalIgnoreCase))
+                    (remoteInstance.OrchestrationInstance.ExecutionId == null || string.Equals(localInstance.ExecutionId, remoteInstance.OrchestrationInstance.ExecutionId, StringComparison.OrdinalIgnoreCase)))
                 {
-                    // Happy path: The message matches the table status. Allow it to run.
+                    // Happy path: The message matches the table status -OR- the table doesn't have an execution ID field (older clients, pre-v1.8.5),
+                    // in which case we have no way of knowing whether it's a duplicate. Allow it to run.
```
davidmrdavid, line 242: nit: In this comment, I'm unsure if "in which case" applies only to the older-clients-case or if it applies to the entirety of cases under this branch.

As a non-blocking suggestion, I would like to see this made clear. If it applies to both cases, then to say "in both cases" and if it applies only to the latter, then to say: "in the latter (older clients) case, there's no way of knowing ..."

davidmrdavid (approved):
LGTM with non-blocking nit.

cgillum:
Got an internal customer to independently verify this fix as well.

PR #652, opened 2021-12-04
Fix retry interceptor sleeps

pavisalavisa:
The final sleep inside the retry loop is redundant. 

The sleep period should take place between two retries. 

For example: If the max number of attempts is three, the total number of sleep periods should be two, as illustrated below.
image

This effect is amplified when retry intervals are relatively big (e.g., 30 mins), leading to undesirable side effects in the client code.

I created a couple of test cases reproducing the defect and verifying that the applied fix is working as intended.

### DIFF
--- a/Test/DurableTask.Core.Tests/RetryInterceptorTests.cs
+++ b/Test/DurableTask.Core.Tests/RetryInterceptorTests.cs
@@ -1,0 +1,100 @@
+﻿// ---------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// ---------------------------------------------------------------
+
+namespace DurableTask.Core.Tests
+{
+    using System;
+    using System.IO;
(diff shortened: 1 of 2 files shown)


cgillum (approved):
Thanks for the PR and associated unit tests! I'm surprised something like this hasn't been noticed earlier. I'll run this through our CI to ensure no unintended regressions.

cgillum:
It just occurred to me that this change could impact existing orchestration instances that already have the extra TimerFired event in their history. If those apps upgrade to this version, then I'm worried their orchestrations could fail with [one of these NonDeterministicOrchestrationException's](https://github.com/Azure/durabletask/blob/859311cfc2223eda9404448c7ef9587518988d13/src/DurableTask.Core/TaskOrchestrationContext.cs#L304-L320). In order to make this change backwards compatible, we may need to keep the extra timer in place, but perhaps with a delay of zero.

PR #653, opened 2021-12-07
change OrchestrationInstance and IsReplaying to virtual make it mockable

ChenglongLiu:
Add unit test for orchestrations is a strong requirement. But the `OrchestrationInstance` and `IsReplaying` do not support mock, also the `internal protected set;` access modifiers prevented us update them.

```
[TestMethod]
        public async Task RunTest()
        {
            var mockContext = new Mock<OrchestrationContext>();
            mockContext.Setup(c => c.OrchestrationInstance).Returns(
                () => new OrchestrationInstance()
                {
                    InstanceId = Guid.NewGuid().ToString()
                });
            mockContext.Setup(c => c.CreateClient<ITask>()).Returns(
                () => new MockTask());
            //other mock

            //Test orchestration
            TestOrchestration orchestration = new TestOrchestration();
            var result = await orchestration.RunTask(mockContext.Object, "input");
            Assert.AreEqual("Succeeded", result);
        }
```

### DIFF
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -45,7 +45,7 @@ namespace DurableTask.Core
         /// <summary>
         /// Instance of the currently executing orchestration
         /// </summary>
-        public OrchestrationInstance OrchestrationInstance { get; internal protected set; }
+        public virtual OrchestrationInstance OrchestrationInstance { get; internal protected set; }
 
         /// <summary>
         /// Replay-safe current UTC datetime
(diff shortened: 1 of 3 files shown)


jviau:
Adding `virtual` to a property is considered a binary breaking change.

https://docs.microsoft.com/en-us/dotnet/core/compatibility/

> ❌ DISALLOWED: Adding the virtual keyword to a member
>
> While this often is not a breaking change because the C# compiler tends to emit callvirt Intermediate Language (IL) instructions to call non-virtual methods (callvirt performs a null check, while a normal call doesn't), this behavior is not invariable for several reasons:
>
> C# is not the only language that .NET targets.
>
> The C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and is probably not null (such as a method accessed through the ?. null propagation operator).
>
> Making a method virtual means that the consumer code would often end up calling it non-virtually.

cgillum:
Closing out this old PR, which I assume is abandoned.

PR #655, opened 2021-12-10
Add SBOM task to release pipeline

davidmrdavid:
Includes SBOM-generation task to our pipeline, which documents our dependencies for cybersecurity compliance, It also uses the new 1ES-approved VM, another compliance requirement.

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -1,8 +1,10 @@
 trigger: none
 pr: none
 
-pool:
-  vmImage: 'windows-latest'
+ pool:
+  name: '1ES-Hosted-DurableTaskFramework'
+  demands:
+    - ImageOverride -equals MMS2022TLS
 
 steps:
 # Start by restoring all the dependencies. This needs to be its own task
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -60,12 +60,6 @@ steps:
           }
       ]
 
-- task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
-  displayName: 'SBOM Generation Task'
-  inputs:
-    BuildDropPath: '$(build.artifactStagingDirectory)'
-    Verbosity: 'Information'
-
 # Packaging needs to be a separate step from build.
 # This will automatically pick up the signed DLLs.
 - task: DotNetCoreCLI@2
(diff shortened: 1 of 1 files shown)


azure-pipelines-release.yml
```diff
@@ -58,6 +58,12 @@ steps:
           }
       ]
 
+- task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
+  displayName: 'SBOM Generation Task'
+  inputs:
+    BuildDropPath: '$(Build.ArtifactStagingDirectory)'
```
bachuv, line 64: Should `$(Build.ArtifactStagingDirectory)` be `$(build.artifactStagingDirectory)` to match other places in this yml file or is it fine?

davidmrdavid: @bachuv, I don't see `$(build.artifactStagingDirectory` being used in other places. What line(s) are you referring to?

bachuv: - https://github.com/Azure/durabletask/blob/a2aa3f0430d3a2ec7a667dc1ca88ae6b1aeb3e04/azure-pipelines-release.yml#L76
- https://github.com/Azure/durabletask/blob/a2aa3f0430d3a2ec7a667dc1ca88ae6b1aeb3e04/azure-pipelines-release.yml#L85
- https://github.com/Azure/durabletask/blob/a2aa3f0430d3a2ec7a667dc1ca88ae6b1aeb3e04/azure-pipelines-release.yml#L107

davidmrdavid: Thanks! My latest commit utilizes this capitalization scheme

azure-pipelines-release.yml
```diff
@@ -58,6 +60,12 @@ steps:
           }
       ]
 
+- task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
```
bachuv, line 63: Is this `task` value standard for Manifest Generator tasks? When I try populating this task in DevOps, I get `AzureArtifacts.drop-validator-task.manifest-generator-task.ManifestGeneratorTask@0`

davidmrdavid: Yeah I think in DevOps there's somehow two ManifestGenerator tasks, one of them resolves to this path, and the other resolves to the path you shared. Ultimately they both seem to do the same thing, so I can't really judge which one is more appropiate.

azure-pipelines-release.yml
```diff
@@ -58,6 +60,12 @@ steps:
...
+  displayName: 'SBOM Generation Task'
```
bachuv, line 64: Can you explain how we know that this is the correct place to add the SBOM Generation Task? I was looking at another PR: https://github.com/Azure/azure-functions-eventgrid-extension/pull/98/files and it's the last step before publishing the pipeline artifact.

davidmrdavid: I thought we had to generate the SBOM before running `pack`, but it seems other PRs disagree. Let me play with the location of this step. If I can still generate the SBOM when running this later in the pipeline, then I'm ok with moving this further down.

davidmrdavid: My latest commit moves this further down

amdeel (approved):
LGTM!

PR #656, opened 2021-12-18
Initial implementation of out-of-proc support

cgillum:
This PR introduces changes to the DT.Core middleware implementation to enable injecting a custom orchestration processor. **See the new unit tests for a primitive example of how this works.** The intent is to leverage this for out-of-process execution of orchestrations in Durable Functions and elsewhere.

Part of the change requires a stronger decoupling of the orchestration executor from the dispatcher layer. In particular, two important changes were made:

- Removed usage of the `IsPlayed` property. This value was set in the executor, which is not a change that could be persisted across process boundaries. For that reason, I've decoupled event history into PastEvents and NewEvents as a way to distinguished "played" and "not-yet-played" history events.
- Similarly, custom status is now being saved in the dispatcher layer instead of in the executor layer.

In order to implement .NET-based out-of-process handlers, several internal types were made public.

### DIFF
--- a/src/DurableTask.Core/History/HistoryEvent.cs
+++ b/src/DurableTask.Core/History/HistoryEvent.cs
@@ -75,7 +75,6 @@ namespace DurableTask.Core.History
         /// Gets the IsPlayed status
         /// </summary>
         [DataMember]
-        [Obsolete("The IsPlayed property is no longer being set")]
         public bool IsPlayed { get; set; }
 
         /// <summary>
--- a/src/DurableTask.Core/History/HistoryEvent.cs
+++ b/src/DurableTask.Core/History/HistoryEvent.cs
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
 namespace DurableTask.Core.History
 {
     using System;
@@ -25,7 +25,7 @@ namespace DurableTask.Core.History
     [KnownType(nameof(KnownTypes))]
     public abstract class HistoryEvent : IExtensibleDataObject
     {
-        private static IReadOnlyCollection<Type> knownTypes;
+        private static IReadOnlyCollection<Type>? knownTypes;
 
         /// <summary>
         /// List of all event classes, for use by the DataContractSerializer
@@ -52,7 +52,6 @@ namespace DurableTask.Core.History
         /// </summary>
         internal HistoryEvent()
         {
-            IsPlayed = false;
             Timestamp = DateTime.UtcNow;
         }
 
@@ -63,7 +62,6 @@ namespace DurableTask.Core.History
         protected HistoryEvent(int eventId)
         {
             EventId = eventId;
-            IsPlayed = false;
             Timestamp = DateTime.UtcNow;
         }
 
@@ -77,6 +75,7 @@ namespace DurableTask.Core.History
         /// Gets the IsPlayed status
         /// </summary>
         [DataMember]
+        [Obsolete("The IsPlayed property is no longer being set")]
         public bool IsPlayed { get; set; }
 
         /// <summary>
@@ -94,7 +93,6 @@ namespace DurableTask.Core.History
         /// <summary>
         /// Implementation for <see cref="IExtensibleDataObject.ExtensionData"/>.
         /// </summary>
-        public ExtensionDataObject ExtensionData { get; set; }
-
+        public ExtensionDataObject? ExtensionData { get; set; }
     }
 }
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -77,8 +77,8 @@ namespace DurableTask.Core
         /// <param name="events">List of events for this runtime state</param>
         public OrchestrationRuntimeState(IList<HistoryEvent>? events)
         {
-            Events = new List<HistoryEvent>(events?.Count ?? 32);
-            PastEvents = new List<HistoryEvent>(events?.Count ?? 32);
+            Events = events != null ? new List<HistoryEvent>(events.Count) : new List<HistoryEvent>();
+            PastEvents = events != null ? new List<HistoryEvent>(events.Count) : new List<HistoryEvent>();
             NewEvents = new List<HistoryEvent>();
             completedEventIds = new HashSet<int>();
 
@@ -310,17 +310,21 @@ namespace DurableTask.Core
 
             if (evt is TaskScheduledEvent taskScheduledEvent)
             {
-                returnedEvent = new TaskScheduledEvent(
-                    eventId: taskScheduledEvent.EventId,
-                    name: taskScheduledEvent?.Name ?? "(unnamed)",
-                    version: taskScheduledEvent?.Version,
-                    input: "[..snipped..]");
+                returnedEvent = new TaskScheduledEvent(taskScheduledEvent.EventId)
+                {
+                    Timestamp = taskScheduledEvent.Timestamp,
+                    IsPlayed = taskScheduledEvent.IsPlayed,
+                    Name = taskScheduledEvent.Name,
+                    Version = taskScheduledEvent.Version,
+                    Input = "[..snipped..]",
+                };
             }
             else if (evt is TaskCompletedEvent taskCompletedEvent)
             {
                 returnedEvent = new TaskCompletedEvent(taskCompletedEvent.EventId, taskCompletedEvent.TaskScheduledId, "[..snipped..]")
                 {
                     Timestamp = taskCompletedEvent.Timestamp,
+                    IsPlayed = taskCompletedEvent.IsPlayed,
                 };
             }
             else if (evt is SubOrchestrationInstanceCreatedEvent subOrchestrationInstanceCreatedEvent)
@@ -328,6 +332,7 @@ namespace DurableTask.Core
                 returnedEvent = new SubOrchestrationInstanceCreatedEvent(subOrchestrationInstanceCreatedEvent.EventId)
                 {
                     Timestamp = subOrchestrationInstanceCreatedEvent.Timestamp,
+                    IsPlayed = subOrchestrationInstanceCreatedEvent.IsPlayed,
                     Name = subOrchestrationInstanceCreatedEvent.Name,
                     Version = subOrchestrationInstanceCreatedEvent.Version,
                     Input = "[..snipped..]",
@@ -339,6 +344,7 @@ namespace DurableTask.Core
                     subOrchestrationInstanceCompletedEvent.TaskScheduledId, "[..snipped..]")
                 {
                     Timestamp = subOrchestrationInstanceCompletedEvent.Timestamp,
+                    IsPlayed = subOrchestrationInstanceCompletedEvent.IsPlayed,
                 };
             }
             else if (evt is TaskFailedEvent taskFailedEvent)
@@ -347,6 +353,7 @@ namespace DurableTask.Core
                     taskFailedEvent.TaskScheduledId, taskFailedEvent.Reason, "[..snipped..]")
                 {
                     Timestamp = taskFailedEvent.Timestamp,
+                    IsPlayed = taskFailedEvent.IsPlayed,
                 };
             }
             else if (evt is SubOrchestrationInstanceFailedEvent subOrchestrationInstanceFailedEvent)
@@ -355,6 +362,7 @@ namespace DurableTask.Core
                     subOrchestrationInstanceFailedEvent.TaskScheduledId, subOrchestrationInstanceFailedEvent.Reason, "[..snipped..]")
                 {
                     Timestamp = subOrchestrationInstanceFailedEvent.Timestamp,
+                    IsPlayed = subOrchestrationInstanceFailedEvent.IsPlayed,
                 };
             }
             else if (evt is ExecutionStartedEvent executionStartedEvent)
@@ -362,6 +370,7 @@ namespace DurableTask.Core
                 returnedEvent = new ExecutionStartedEvent(executionStartedEvent.EventId, "[..snipped..]")
                 {
                     Timestamp = executionStartedEvent.Timestamp,
+                    IsPlayed = executionStartedEvent.IsPlayed,
                 };
             }
             else if (evt is ExecutionCompletedEvent executionCompletedEvent)
@@ -370,6 +379,7 @@ namespace DurableTask.Core
                     executionCompletedEvent.OrchestrationStatus)
                 {
                     Timestamp = executionCompletedEvent.Timestamp,
+                    IsPlayed = executionCompletedEvent.IsPlayed,
                 };
             }
             else if (evt is ExecutionTerminatedEvent executionTerminatedEvent)
@@ -377,6 +387,7 @@ namespace DurableTask.Core
                 returnedEvent = new ExecutionTerminatedEvent(executionTerminatedEvent.EventId, "[..snipped..]")
                 {
                     Timestamp = executionTerminatedEvent.Timestamp,
+                    IsPlayed = executionTerminatedEvent.IsPlayed,
                 };
             }
             // ContinueAsNewEvent is covered by the ExecutionCompletedEvent block
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -10,7 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
 namespace DurableTask.Core
 {
     using System;
@@ -36,6 +36,11 @@ namespace DurableTask.Core
         /// </summary>
         public IList<HistoryEvent> NewEvents { get; }
 
+        /// <summary>
+        /// A subset of <see cref="this.Events"/> that contains only events that have been previously played and should not be serialized
+        /// </summary>
+        public IList<HistoryEvent> PastEvents { get; }
+
         readonly ISet<int> completedEventIds;
 
         /// <summary>
@@ -46,7 +51,7 @@ namespace DurableTask.Core
         /// <summary>
         /// Gets the execution completed event
         /// </summary>
-        public ExecutionCompletedEvent ExecutionCompletedEvent { get; set; }
+        public ExecutionCompletedEvent? ExecutionCompletedEvent { get; set; }
 
         /// <summary>
         /// Size of the serialized state (uncompressed)
@@ -56,7 +61,7 @@ namespace DurableTask.Core
         /// <summary>
         /// The string status of the runtime state
         /// </summary>
-        public string Status;
+        public string? Status;
 
         /// <summary>
         /// Creates a new instance of the OrchestrationRuntimeState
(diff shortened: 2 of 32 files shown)


src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -69,28 +93,41 @@ IEnumerable<OrchestratorAction> ExecuteCore(IEnumerable<HistoryEvent> eventHisto
...
 
+                    // Replay the old history to rebuild the local state of the orchestration.
+                    // TODO: Log a verbose message indicating that the replay has started (include event count?)
+                    this.context.IsReplaying = true;
+                    ProcessEvents(pastEvents);
+
+                    // Play the newly arrived events to determine the next action to take.
+                    // TODO: Log a verbose message indicating that new events are being processed (include event count?)
+                    this.context.IsReplaying = false;
+                    ProcessEvents(newEvents);
```
cgillum, line 119: The main reason I made the change to distinguish "past" events from "new" events is because I needed a way to preserve the semantic of `HistoryEvent.IsPlayed` for out-of-proc to work (since any changes we make to history events here aren't visible in the "host" process). I could've potentially solved this by just blindly setting all history events to `IsPlayed = true` in TaskOrchestrationDispatcher, but it felt like a strange thing to do. When thinking through this problem, it seemed to me that we didn't need the `IsPlayed` property at all. It's only used to assign the `TaskOrchestrationContext.IsReplaying` property, and this property can already be assigned in a deterministic way without needing to save any IsPlayed data into the storage layer. For these reasons, I decided to just deprecate the IsPlayed property in history and updated this code to simply distinguish past events from new events.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -116,11 +153,14 @@ IEnumerable<OrchestratorAction> ExecuteCore(IEnumerable<HistoryEvent> eventHisto
...
-                return this.context.OrchestratorActions;
+                return new OrchestratorExecutionResult
+                {
+                    Actions = this.context.OrchestratorActions,
+                    CustomStatus = this.taskOrchestration.GetStatus(),
+                };
             }
             finally
             {
-                this.orchestrationRuntimeState.Status = this.taskOrchestration.GetStatus();
```
cgillum, line 123: Similar to `IsPlayed`, assigning the `Status` property of the local runtime state object is useless if this code is running out-of-process. For that reason, I instead moved the custom status saving into the `OrchestratorExecutionResult`, which allows custom status to be correctly saved whether we're in-proc or out-of-proc.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -195,6 +195,51 @@ public async Task GetPaginatedStatuses()
             }
         }
 
+        [TestMethod]
+        public async Task GetInstanceIdsByPrefix()
```
cgillum, line 199: I created this test a long time ago as part of an unrelated investigation. I decided to include it as part of this PR for convenience. It's not actually related to any of the other changes in this PR.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1248,7 +1293,7 @@ public async Task HandledActivityException(bool enableExtendedSessions)
 
                 // Empty string input should result in ArgumentNullException in the orchestration code.
                 var client = await host.StartOrchestrationAsync(typeof(Orchestrations.TryCatchLoop), 5);
-                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(10));
+                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(15));
```
cgillum, line 1296: This test was being flakey in the CI because sometimes the orchestration was taking more than 10 seconds to complete. Increasing the timeout made it more reliable.

test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj
```diff
@@ -2,7 +2,7 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
```
cgillum, line 5: This fixes all the warnings about .NET Core 2.1 being a deprecated framework. It's not critical for this PR, however.

tools/DurableTask.props
```diff
@@ -5,7 +5,7 @@
     <DebugType Condition="'$(Configuration)'=='Release'">pdbonly</DebugType>
     <DebugSymbols>True</DebugSymbols>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
```
cgillum, line 8: All DTFx projects use C# 9.0 by default now. This was necessary to get certain features working correctly with nullable reference types.

cgillum (commented):
Added some inline comments to help explain why certain changes were made.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -20,20 +20,28 @@ namespace DurableTask.Core
...
-    using DurableTask.Core.Command;
     using DurableTask.Core.Common;
     using DurableTask.Core.Exceptions;
     using DurableTask.Core.History;
 
-    internal class TaskOrchestrationExecutor
+    /// <summary>
+    /// Utility for executing task orchestrators.
+    /// </summary>
+    public class TaskOrchestrationExecutor
```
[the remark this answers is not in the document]
davidmrdavid: I'll make sure we're lucky then ;)

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -20,20 +20,28 @@ namespace DurableTask.Core
...
         readonly TaskScheduler decisionScheduler;
         readonly OrchestrationRuntimeState orchestrationRuntimeState;
         readonly TaskOrchestration taskOrchestration;
         readonly bool skipCarryOverEvents;
-        Task<string> result;
-
+        Task<string>? result;
+
+        /// <summary>
+        /// Initializes a new instances of the <see cref="TaskOrchestrationExecutor"/> class.
```
davidmrdavid, line 40: nit: minor typo

```suggestion
        /// Initializes a new instance of the <see cref="TaskOrchestrationExecutor"/> class.
```

src/DurableTask.Core/TaskOrchestrationExecutor.cs
davidmrdavid: In the current Python/JS OOProc model, the `IsPlayed` History property is used to assign the orchestration context's `IsReplaying` flag. I suspect this change would caused a regression there, or won't it?

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -156,36 +156,76 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
...
-                        this.logHelper.TaskActivityFailure(orchestrationInstance, scheduledEvent.Name, (TaskFailedEvent)eventToRespond, e);
-                    }
+                        if (taskActivity == null)
+                        {
+                            // This likely indicates a deployment error of some kind. Because these unhandled exceptions are
+                            // automatically retried, resolving this may require redeploying the app code so that the activity exists again.
+                            // CONSIDER: Should this be changed into a permanent error that fails the orchestration? Perhaps
+                            //           the app owner doesn't care to preserve existing instances when doing code deployments?
+                            throw new TypeMissingException($"TaskActivity {scheduledEvent.Name} version {scheduledEvent.Version} was not found");
+                        }
```
davidmrdavid, line 171: To me, it makes sense to permanently error-out from instances with this kind of error. Otherwise, I worry about how this may clutter the History and/or lead to some inconsistent state

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -156,36 +156,76 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
...
 
-                    if (eventToRespond is TaskCompletedEvent completedEvent)
-                    {
-                        this.logHelper.TaskActivityCompleted(orchestrationInstance, scheduledEvent.Name, completedEvent);
-                    }
-                });
+                        // TODO : pass workflow instance data
+                        var context = new TaskContext(taskMessage.OrchestrationInstance);
```
davidmrdavid, line 174: Hasn't this `TODO` already been fulfilled with the line immediately below?

src/DurableTask.Core/History/HistoryEvent.cs
```diff
@@ -77,6 +75,7 @@ protected HistoryEvent(int eventId)
         /// Gets the IsPlayed status
         /// </summary>
         [DataMember]
+        [Obsolete("The IsPlayed property is no longer being set")]
```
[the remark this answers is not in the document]
davidmrdavid: I believe this would affect replay in JS/Python OOProc, because those use the History's `IsPlayed` field to set the orchestration context's `isReplaying` flag.

Btw, I also note this in a later comment.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -123,15 +122,20 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
...
+                    this.logHelper.TaskActivityDispatcherError(workItem, message);
+                    throw TraceHelper.TraceException(
+                        TraceEventType.Error,
+                        "TaskActivityDispatcher-MissingActivityName",
+                        new InvalidOperationException(message));
                 }
 
+                this.logHelper.TaskActivityStarting(orchestrationInstance, scheduledEvent);
+                TaskActivity? taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version);
+
```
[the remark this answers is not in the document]
davidmrdavid: Unfortunately I don't fully follow the significance of this change. I could use some context.

What are examples of activities supported in the middleware layer, and why wouldn't they be registered as `TaskActivity` classes?

davidmrdavid (commented):
Exciting changes! Thanks for this PR. I found that, after reading the new tests, the changes in this PR became a lot clearer to me.

I left a few questions, and I'm also a bit concerned about the `isPlayed` change; I understand that failing to update it would break the existing JS/Python implementations.

cgillum (commented):
Thanks @davidmrdavid for the review comments? Let me revisit IsPlayed - I think you're probably right that I'm breaking existing DF OOProc scenarios.

See my replies to your other comments below.

cgillum:
@davidmrdavid I reverted the changes specifically related to `HistoryEvent.IsPlayed` and confirmed that DF JS is now getting the correct values for `context.df.isReplaying`. Thanks for promptly pointing this out. Other PR feedback should also be addressed.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -515,23 +519,15 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
-
                             if (workItem.LockedUntilUtc < DateTime.UtcNow.AddMinutes(1))
                             {
                                 await this.orchestrationService.RenewTaskOrchestrationWorkItemLockAsync(workItem);
                             }
+
+                            // REVIEW: This is a behavior change from v2.6.x, where ContinueAsNew went through the Cursor/Resume path.
+                            //         It's not clear if that was ever necessary, or if there are potentially breaking changes if we move
+                            //         away from Cursor/Resume for ContinueAsNew.
+                            workItem.Cursor = null;
```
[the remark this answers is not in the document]
sebastianburckhardt: Although it is not currently in use, it is conceivable we still want to bring back the "slow continue-as-new" path again at some point.

sebastianburckhardt (changes requested):
Looks good to me in general other than some minor things I commented on.

My confidence about fully understanding the ramifications of the suggested changes is a bit low though.

There is a lot of complexity in the DF shims today, in particular for entities. Many of the things that are changed in this PR were the reason for some complicated workarounds (e.g. the precise way in which middleware gets called, exceptions are handled or converted, and how timeoutes are handled, azure/azure-functions-durable-extension#1917, quite  a nightmare). Those complicated brittle workarounds are at a high risk to break with this PR so I think we need to have a careful look at  them before committing this PR to the main branch. 

In the long run we should be able to simplify things everywhere as we rework the layers but we may break a lot of things in the process unless we coordinate the changes (and the testing) across all the repos. Perhaps we need to have a coordinated "alternate" branch on all the repos?

I also want to test whether this works with Netherite.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -123,15 +122,20 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
                                                   taskMessage.Event.EventType));
                 }
 
-                // call and get return message
                 scheduledEvent = (TaskScheduledEvent)taskMessage.Event;
-                this.logHelper.TaskActivityStarting(orchestrationInstance, scheduledEvent);
-                TaskActivity taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version);
-                if (taskActivity == null)
+                if (scheduledEvent.Name == null)
```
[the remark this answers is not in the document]
davidmrdavid: My understanding is that a `string?` type is either a `string` or `null`. So I had assumed that changing the type to `string` would force the type-checker to ensure no code-paths allow for a `TaskScheduledEvent.Name` to be `null`. 

Am I misunderstanding something?

src/DurableTask.Core/TaskActivityDispatcher.cs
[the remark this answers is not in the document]
davidmrdavid: Thanks for the clarification. I understand and agree with your comments on the downsides of the current design.

However, I'm personally still confused about this comment: _"With this change, we defer the exception throwing to after all **middleware** has had a chance to run so that we can support activities implemented in the **middleware** layer and not necessarily as registered TaskActivity classes (in theory, **middleware** could also inject a TaskActivity object now, but that's not a scenario we particularly care about)."_

I think the confusing bit here is that that I don't have a firm grasp of what the middleware was doing before, and what it's doing now. I do have _some_ working understanding, but it's finicky. I'd love to have a chat about this, as I'm unsure I'll get much clarity on this via async comments alone :)

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -101,7 +107,7 @@ public DateTime CreatedTime
             get
             {
                 Debug.Assert(ExecutionStartedEvent != null);
-                return ExecutionStartedEvent.Timestamp;
+                return ExecutionStartedEvent?.Timestamp ?? default;
```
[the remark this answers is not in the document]
cgillum: I think what I'll do is replace the assert statements with exceptions so that we have a proper contract that has stronger null guarantees.

cgillum (commented):
I'll add another iteration with my proposed changes, but here's my feedback responses for now.

src/DurableTask.Core/TaskActivityDispatcher.cs
davidmrdavid: Good to know this won't clutter the history. I don't have strong feelings about this behavior, but the repeated exceptions (forever or until expiration) still seems like undesirable behavior to me. Just giving my 2 cents, I realize this behavior is outside the scope of this PR :)

src/DurableTask.Core/TaskOrchestrationExecutor.cs
davidmrdavid: Yeah I'm also fully on board with removing IsPlayed and other logically unecessary fields in the near future. If / when we finalize the design for the OOProc Task layer, it would be good to revisit these things.

src/DurableTask.Core/TaskActivityDispatcher.cs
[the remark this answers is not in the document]
cgillum: Let me answer your original question differently: we can't change from `string?` to `string` because there are existing public constructors that allow you to create a `TaskScheduledEvent` without initializing the `Name` property. Removing those constructors would be a breaking change, which I want to avoid. In addition to this, the C# compiler is unable to make any guarantees about nullability when objects of a given type are created/manipulated via reflection. The Newtonsoft.Json serializer is an example of a component that can create an initialize types using private reflection. For that reason, in order to be defensive against bad JSON payloads, I'm intentionally using `string?` instead of `string` even though theoretically I might not need to (though in reality, I actually do need `string?` due to my earlier point).

src/DurableTask.Core/TaskActivityDispatcher.cs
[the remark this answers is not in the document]
cgillum: Hopefully this will be more clear when I publish my Durable Functions extension PR.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -36,6 +36,11 @@ public class OrchestrationRuntimeState
         /// </summary>
         public IList<HistoryEvent> NewEvents { get; }
 
+        /// <summary>
+        /// A subset of <see cref="Events"/> that contains only events that have been previously played and should not be serialized
+        /// </summary>
+        public IList<HistoryEvent> PastEvents { get; }
```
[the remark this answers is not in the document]
cgillum: I take it back - it looks like there are several places in the code where `NewEvents` gets cleared but all the existing events are supposed to be kept in memory. `Events` is therefore _not always_ a combination of these two lists and some redundancy is required. It would probably take a bunch of refactoring to remove this redundancy, but I don't want to risk it as part of this PR.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -70,9 +75,10 @@ public OrchestrationRuntimeState()
         /// Creates a new instance of the OrchestrationRuntimeState with the supplied events
         /// </summary>
         /// <param name="events">List of events for this runtime state</param>
-        public OrchestrationRuntimeState(IList<HistoryEvent> events)
+        public OrchestrationRuntimeState(IList<HistoryEvent>? events)
         {
-            Events = new List<HistoryEvent>();
+            Events = new List<HistoryEvent>(events?.Count ?? 32);
+            PastEvents = new List<HistoryEvent>(events?.Count ?? 32);
```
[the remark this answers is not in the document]
cgillum: I changed the expressions to remove the hardcoded numbers.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -113,49 +114,28 @@ public DateTime CreatedTime
...
-            }
-        }
+        public string Version => GetExecutionStartedEventOrThrow().Version;
 
         /// <summary>
         /// Gets the tags from the ExecutionStartedEvent
         /// </summary>
-        // This gets called by json.net for deserialization, we can't assert if there is no ExecutionStartedEvent
-        public IDictionary<string, string> Tags => ExecutionStartedEvent?.Tags;
+        // This gets called by json.net for deserialization, we can't throw if there is no ExecutionStartedEvent
```
davidmrdavid, line 137: nit: I'd appreciate a tiny edit to this comment where we specify that, even though we're not throwing in here, we do throw an exception later in the code if `ExecutionStartedEvent` is null.

davidmrdavid (approved):
LGTM, especially now that `isPlayed` is preserved. Left a tiny nit.
I'd love to get @sebastianburckhardt's approval before merge as well.

src/DurableTask.Core/OrchestrationRuntimeState.cs
cgillum: I think I'll just get rid of this comment. I'm not even sure it's accurate (the original comment predates me).

sebastianburckhardt (approved):
The changes look good to me.

I also did a quick test with Netherite and it seems to work without any issues.

PR #657, opened 2021-12-30
Updated Out of Order message checks to include call entity results

amdeel:
Currently, if an EventRaised is returned from a CallEntityAsync call before the orchestrator writes itself in history the message will be discarded and the orchestrator will be stuck.

This change updates our Out of Order message checks to allow EventRaised events to be considered Out of Order messages when the message came from an Entity.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
+++ b/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
@@ -136,7 +136,8 @@ namespace DurableTask.AzureStorage.Messaging
                 message.TaskMessage.Event.EventType != EventType.TaskFailed &&
                 message.TaskMessage.Event.EventType != EventType.SubOrchestrationInstanceCompleted &&
                 message.TaskMessage.Event.EventType != EventType.SubOrchestrationInstanceFailed &&
-                message.TaskMessage.Event.EventType != EventType.TimerFired)
+                message.TaskMessage.Event.EventType != EventType.TimerFired &&
+                !(message.TaskMessage.Event.EventType == EventType.EventRaised && message.Sender.InstanceId.StartsWith("@")))
             {
                 // The above message types are the only ones that can potentially be considered out-of-order.
                 return false;


src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -136,7 +136,8 @@ public bool IsOutOfOrderMessage(MessageData message)
                 message.TaskMessage.Event.EventType != EventType.TaskFailed &&
                 message.TaskMessage.Event.EventType != EventType.SubOrchestrationInstanceCompleted &&
                 message.TaskMessage.Event.EventType != EventType.SubOrchestrationInstanceFailed &&
-                message.TaskMessage.Event.EventType != EventType.TimerFired)
+                message.TaskMessage.Event.EventType != EventType.TimerFired &&
+                !(message.TaskMessage.Event.EventType == EventType.EventRaised && message.Sender.InstanceId.StartsWith("@")))
```
amdeel, line 140: I need to make this check a configurable option still.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -175,6 +176,16 @@ public bool IsOutOfOrderMessage(MessageData message)
                 }
             }
 
+            if (message.TaskMessage.Event.EventType == EventType.EventRaised)
+            {
+                // This EventRaised message is a response to an EventSent message. 
+                HistoryEvent mostRecentTaskEvent = this.RuntimeState.Events.LastOrDefault(e => e.EventType == EventType.EventSent);
```
cgillum, line 182: I don't think this is necessarily a response to an EventSent message. The current EventRaised event could have come from a completely unrelated entity. If you're specifically looking for request/response pattern of an orchestrator calling an entity, then I think you need to at least check to see whether the EventSent message and the EventRaised message are part of the same operation. I'll defer to @sebastianburckhardt on how we can do this.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -175,6 +176,16 @@ public bool IsOutOfOrderMessage(MessageData message)
...
+                if (mostRecentTaskEvent != null)
+                {
+                    return false;
+                }
+            }
+
```
sebastianburckhardt, line 188: This check is a bit approximate; it detects an out-of-order message from an entity only if it is the first such message received. This is probably good enough to start, it is already an improvement.

I will think about how a more precise check could be done in the future. Most likely this would require some changes to the DF extension side, otherwise there may not be enough information available in DTFx.

sebastianburckhardt (changes requested):

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
sebastianburckhardt: An event sent from an entity to an orchestration is always going to be a response of some kind, which must have been preceded by a prior EventSent, so the check is safe (though not fully precise).

sebastianburckhardt:
To do precise correlation, the thing to use is the request identifier (a GUID).
The request identifier can be found as follows:

- `EventRaisedEvent.Name` contains the request identifier as a string, i.e. `guid.ToString()`
- `EventSentEvent.Input` is a JSON object (corresponding to `RequestMessage` in DF extension) which as a property `id`. 

So we can solve this problem by doing a bit of parsing:
```csharp
            Guid? FindRequestId(string input)
            {
                try
                {
                    JsonTextReader reader = new JsonTextReader(new StringReader(json));
                    reader.Read(); // JsonToken.StartObject
                    while (reader.Read() && reader.TokenType == JsonToken.PropertyName)
                    {
                        switch (reader.Value)
                        {
                            case "$type":
                            case "op":
                            case "signal":
                            case "input":                             
                                reader.Read(); // skip these, they may appear before the id
                                continue;
                            case "id":
                                return Guid.Parse(reader.ReadAsString());
                            default:
                                break;
                        }
                    }
                }
                catch
                {
                }
                return null;
            }
```

I don't love this as a permanent solution since it creates an invisible dependency (the format of the RequestMessage) but it may be our best shot at fixing this quickly and without introducing compatibility issues. We can revise this later as part of our planned overhaul of the DT/DF layering architecture.

sebastianburckhardt:
Oh, and by the way: thanks for finding this bug and submitting a PR! 

It looks to me like this may indeed have caused some pain for our customers (orchestrations that use entities hang without apparent reason).

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
davidmrdavid: Why is it important that the receiving entity not be an entity?

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
davidmrdavid: We do this correlation when implementing replay in OOProc. You could technically inspect the `Input` field of the matching EventSent and see if it contains within a JSON-serialized `ID` corresponding to the entity instanceID.

This logic is implemented in Python here:  https://github.com/Azure/azure-functions-durable-python/blob/4b4d7c6ba9cf45cf071f69eb49923b2e8416c5fd/azure/durable_functions/models/TaskOrchestrationExecutor.py#L113-L124

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -178,18 +181,51 @@ public bool IsOutOfOrderMessage(MessageData message)
...
             {
-                // This EventRaised message is a response to an EventSent message. 
-                HistoryEvent mostRecentTaskEvent = this.RuntimeState.Events.LastOrDefault(e => e.EventType == EventType.EventSent);
-                if (mostRecentTaskEvent != null)
+                // This EventRaised message is a response to an EventSent message.
+                var requestId = ((EventRaisedEvent)message.TaskMessage.Event).Name;
+                if (requestId != null)
                 {
-                    return false;
+                    HistoryEvent mostRecentTaskEvent = this.RuntimeState.Events.LastOrDefault(e => e.EventType == EventType.EventSent && FindRequestId(((EventSentEvent)e).Input)?.ToString() == requestId);
```
sebastianburckhardt, line 188: Not sure why this is LastOrDefault. I don't think it is possible that there is more than one matching entry. So FirstOrDefault does the job also and executes a bit faster.

sebastianburckhardt (approved):
Just a few nits: the compile is broken because `Util` became ambiguous, and I would prefer `FirstOrDefault` instead of `LastOrDefault`

Otherwise this looks good to me.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
sebastianburckhardt: Entities already contain their own message sorting logic. So we should always deliver all messages to an entity, otherwise that logic breaks.

src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs
```diff
@@ -175,10 +178,59 @@ public bool IsOutOfOrderMessage(MessageData message)
...
+                // This EventRaised message is a response to an EventSent message.
+                var raisedEventRequestId = ((EventRaisedEvent)message.TaskMessage.Event).Name;
+                if (raisedEventRequestId != null)
+                {
+                    foreach (HistoryEvent historyEvent in this.RuntimeState.Events)
+                    {
+                        if (historyEvent.EventType == EventType.EventSent)
+                        {
+                            var sentEventRequestId = FindRequestId(((EventSentEvent)historyEvent).Input);
+                            if (sentEventRequestId == null || sentEventRequestId.ToString() == raisedEventRequestId)
```
sebastianburckhardt, line 192: I think is better to remove the first part of this condition (sentEventRequestId == null). The problem is that even in DF, it is still possible that an orchestration sends some arbitrary event (which cannot be parsed) before later sending a message to an entity.

It does mean that if a non-DF user is using DT with autostarted orchestrations they may have some messages classified as out of order even if they are not out of order. But this is extremely unlikely (because there are no such users) and I think also not actually harmful (since out-of-order messages will be redelivered later).

PR #661, opened 2022-01-06
Fix retry interceptor regression

cgillum:
PR https://github.com/Azure/durabletask/pull/652 introduced a breaking change in history replay by removing a timer from the history when a retry policy executes all retries. The fix was "correct", but it was a breaking change for existing orchestration instances. This PR re-introduces the removed timer to fix the breaking change but sets the final timer delay to zero seconds to mitigate the impact of scheduling an extra timer.

This PR also updates the C# version to 9.0 in order to allow nullable generic type references.


davidmrdavid (approved):
LGTM! Thank you for the prompt patch :)

PR #666, opened 2022-01-19
Updated Queue TTL to infinite

### LINKED ISSUE #660: Message added to Azure Storage Control queue is created with TTL of 7 days by default
As `null` is passed for the param `timeToLive`:

https://github.com/Azure/durabletask/blob/859311cfc2223eda9404448c7ef9587518988d13/src/DurableTask.AzureStorage/Storage/Queue.cs#L46-L59

In `Microsoft.Azure.Storage.Queue` this results in the `initialVisibilityDelayInSeconds` var never being set:
https://github.com/Azure/azure-storage-net/blob/master/Lib/ClassLibraryCommon/Queue/CloudQueue.cs#L2472-L2524

```csharp
        private RESTCommand<NullType> AddMessageImpl(CloudQueueMessage message, TimeSpan? timeToLive, TimeSpan? initialVisibilityDelay, QueueRequestOptions options)
        {
            long? timeToLiveInSeconds = null;
            int? initialVisibilityDelayInSeconds = null;


            if (timeToLive.HasValue)
            {
                if (timeToLive.Value.Equals(TimeSpan.FromSeconds(-1)) || timeToLive.Value > TimeSpan.Zero)
                {
                    timeToLiveInSeconds = (long)timeToLive.Value.TotalSeconds;
                }
                else
                {
                    throw new ArgumentOutOfRangeException("timeToLive", string.Format(CultureInfo.InvariantCulture, SR.ArgumentOutOfRangeError, timeToLive));
                }
            }
            ...
```
Then the internal REST call is made, but due to the `timeToLiveInSeconds` param being null - it's never added to the body of the request. 
https://github.com/Azure/azure-storage-net/blob/master/Lib/ClassLibraryCommon/Queue/Protocol/QueueHttpRequestMessageFactory.cs#L187-L203
```csharp
        public static StorageRequestMessage AddMessage(Uri uri, int? timeout, long? timeToLiveInSeconds, int? visibilityTimeoutInSeconds, HttpContent content, OperationContext operationContext, ICanonicalizer canonicalizer, StorageCredentials credentials)
        {
            UriQueryBuilder builder = new UriQueryBuilder();


            if (timeToLiveInSeconds != null)
            {
                builder.Add(Constants.QueryConstants.MessageTimeToLive, timeToLiveInSeconds.Value.ToString());
            }


            if (visibilityTimeoutInSeconds != null)
            {
                builder.Add(Constants.QueryConstants.VisibilityTimeout, visibilityTimeoutInSeconds.Value.ToString());
            }


            StorageRequestMessage request = HttpRequestMessageFactory.CreateRequestMessage(HttpMethod.Post, uri, timeout, builder, content, operationContext, canonicalizer, credentials);
            return request;
        }
```
Therefore, resulting in the storage REST API defaulting to 7 days, as per the documentation:
https://docs.microsoft.com/en-us/rest/api/storageservices/put-message

messagettl=<int-seconds> | Optional. Specifies the time-to-live interval for the message, in seconds. Prior to version 2017-07-29, the maximum time-to-live allowed is 7 days. For version 2017-07-29 or later, the maximum time-to-live can be any positive number, as well as -1 indicating that the message does not expire. **If this parameter is omitted, the default time-to-live is 7 days.**

Are there any implications passing `TimeSpan.FromSeconds(-1)` as the `timeToLive` instead of `null`? The unit tests seem to still pass...:
https://github.com/Azure/durabletask/blob/859311cfc2223eda9404448c7ef9587518988d13/src/DurableTask.AzureStorage/Storage/Queue.cs#L46-L59

We were caught out by this on return from the holidays, whereby our dev environment was shutdown (durable functions on AKS). All our eternal orchestrations (1000s) that where awaiting `IDurableOrchestrationContext.CreateTimer` became corrupted due to the underlying message being deleted as it had lapsed 7 days. 

We had to terminate all the orchestrations and re-create them as they were stuck with a `Running` status.

amdeel:
Updated queue time to live to infinite to avoid messages from being deleted after 7 days resulting in stuck orchestrations.

resolves #660

### DIFF
--- a/src/DurableTask.AzureStorage/Storage/Queue.cs
+++ b/src/DurableTask.AzureStorage/Storage/Queue.cs
@@ -45,10 +45,16 @@ namespace DurableTask.AzureStorage.Storage
 
         public async Task AddMessageAsync(QueueMessage queueMessage, TimeSpan? visibilityDelay, Guid? clientRequestId = null)
         {
+            // Infinite time to live
+            TimeSpan? timeToLive = TimeSpan.FromSeconds(-1);
+#if NET461
+            // When using net461 SDK version WindowsAzure.Storage 7.2.1 does not allow infinite time to live. Passing in null will default the time to live to 7 days.
+            timeToLive = null;
+#endif
             await this.azureStorageClient.MakeStorageRequest(
                 (context, cancellationToken) => this.cloudQueue.AddMessageAsync(
                     queueMessage.CloudQueueMessage,
-                    TimeSpan.FromSeconds(-1) /* disable TTL */,
+                    timeToLive,
                     visibilityDelay,
                     null,
                     context),
--- a/src/DurableTask.AzureStorage/Storage/Queue.cs
+++ b/src/DurableTask.AzureStorage/Storage/Queue.cs
@@ -48,7 +48,7 @@ namespace DurableTask.AzureStorage.Storage
             await this.azureStorageClient.MakeStorageRequest(
                 (context, cancellationToken) => this.cloudQueue.AddMessageAsync(
                     queueMessage.CloudQueueMessage,
-                    TimeSpan.FromSeconds(-1) /* timeToLive */,
+                    TimeSpan.FromSeconds(-1) /* disable TTL */,
                     visibilityDelay,
                     null,
                     context),
--- a/src/DurableTask.AzureStorage/Storage/Queue.cs
+++ b/src/DurableTask.AzureStorage/Storage/Queue.cs
@@ -48,7 +48,7 @@ namespace DurableTask.AzureStorage.Storage
             await this.azureStorageClient.MakeStorageRequest(
                 (context, cancellationToken) => this.cloudQueue.AddMessageAsync(
                     queueMessage.CloudQueueMessage,
-                    null /* timeToLive */,
+                    TimeSpan.FromSeconds(-1) /* timeToLive */,
                     visibilityDelay,
                     null,
                     context),


src/DurableTask.AzureStorage/Storage/Queue.cs
```diff
@@ -48,7 +48,7 @@ public async Task AddMessageAsync(QueueMessage queueMessage, TimeSpan? visibilit
             await this.azureStorageClient.MakeStorageRequest(
                 (context, cancellationToken) => this.cloudQueue.AddMessageAsync(
                     queueMessage.CloudQueueMessage,
-                    null /* timeToLive */,
+                    TimeSpan.FromSeconds(-1) /* timeToLive */,
```
cgillum, line 51: Can you update the comment to explain why we're setting this to `TimeSpan.FromSeconds(-1)`?

Something like:
```suggestion
                    TimeSpan.FromSeconds(-1) /* disable TTL */,
```

cgillum (approved):
LGTM! It'd be great if we can update the comment to make it clear what we're doing. Also, it would be good to test this in a Functions v1 environment (if you haven't already) since we use an even older storage SDK version there.

cgillum (approved):

src/DurableTask.AzureStorage/Storage/Queue.cs
```diff
@@ -45,10 +45,16 @@ public Queue(AzureStorageClient azureStorageClient, CloudQueueClient queueClient
 
         public async Task AddMessageAsync(QueueMessage queueMessage, TimeSpan? visibilityDelay, Guid? clientRequestId = null)
         {
+            // Infinite time to live
+            TimeSpan? timeToLive = TimeSpan.FromSeconds(-1);
+#if NET461
+            // When using net461 SDK version WindowsAzure.Storage 7.2.1 does not allow infinite time to live. Passing in null will default the time to live to 7 days.
```
M0ns1gn0r, line 51: Would `TimeSpan.MaxValue` work in that case? It would be much better than the 7 days.

PR #671, opened 2022-01-24
Allow Implicit (and Multiple) Azure Storage Accounts for Services

wsugarman:
The Azure Storage account name is used for logging messages, but it may not be specified when Azure Storage Account Service URIs are supplied directly. However, while deriving the account name implicitly brings its own challenges, it also surfaces an additional question: what happens when the service accounts are different?

In this PR, I've made the following changes:
- Remove `StorageAccountName` from `AzureStorageClient` and replace it with `BlobAccountName`, `QueueAccountName`, and `TableAccountName`
- When the account name isn't specified explicitly, the account name is instead derived from each service URI based on the storage SDK's logic
- Use these new properties in lieu of the general `StorageAccountName` property based on which service is in use
  - For general log messages, when the diagnostics/log message refer to multiple services, we'll use `QueueAccountName`

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -5,7 +5,7 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.10.2</FileVersion>
+    <FileVersion>1.10.1</FileVersion>
     <AssemblyVersion>$(FileVersion)</AssemblyVersion>
     <Version>$(FileVersion)</Version>
     <IncludeSymbols>true</IncludeSymbols>
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -5,7 +5,7 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.10.1</FileVersion>
+    <FileVersion>1.10.2</FileVersion>
     <AssemblyVersion>$(FileVersion)</AssemblyVersion>
     <Version>$(FileVersion)</Version>
     <IncludeSymbols>true</IncludeSymbols>
--- a/src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
+++ b/src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
@@ -107,13 +107,13 @@ namespace DurableTask.AzureStorage.Storage
         }
 
         public Task<T> MakeBlobStorageRequest<T>(Func<OperationContext, CancellationToken, Task<T>> storageRequest, string operationName, string? clientRequestId = null) =>
-            this.MakeStorageRequest(storageRequest, BlobAccountName, operationName, clientRequestId);
+            this.MakeStorageRequest<T>(storageRequest, BlobAccountName, operationName, clientRequestId);
 
         public Task<T> MakeQueueStorageRequest<T>(Func<OperationContext, CancellationToken, Task<T>> storageRequest, string operationName, string? clientRequestId = null) =>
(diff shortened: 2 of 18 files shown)


src/DurableTask.AzureStorage/Storage/StorageUriExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+                        return segments[0];
+                    break;
+                case UriHostNameType.IPv4:
+                case UriHostNameType.IPv6:
+                    // In other scenarios, like for Azurite, the service URI looks like <protocol>://<host>:<port>/<account>
+                    return serviceUri.GetComponents(UriComponents.Path, UriFormat.Unescaped);
+            }
+
+            // Otherwise, simply return the entire host + path
+            return serviceUri.PathAndQuery == "/"
```
wsugarman, line 43: I am unsure what we should return for non-standard URIs

cgillum (approved):
I'm fine with these changes. Having a single value for the storage account name was more important back when we weren't keeping track of the app name. I also think it will be pretty uncommon for customers to use multiple storage accounts, so I'm not particularly concerned about the differences in logging behavior there.

Tagging @amdeel for awareness of these changes, in case she has any additional feedback.

src/DurableTask.AzureStorage/Storage/StorageUriExtensions.cs
wsugarman: Now the logic is based on the Azure SDK

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -120,15 +142,16 @@ public async Task<T> MakeStorageRequest<T>(Func<OperationContext, CancellationTo
...
+            this.MakeStorageRequest((context, cancellationToken) => WrapFunctionWithReturnType(storageRequest, context, cancellationToken), accountName, operationName, clientRequestId);
 
         private static async Task<object?> WrapFunctionWithReturnType(Func<OperationContext, CancellationToken, Task> storageRequest, OperationContext context, CancellationToken cancellationToken)
         {
             await storageRequest(context, cancellationToken);
             return null;
         }
+
+        private static string GetAccountName(StorageCredentials credentials, AzureStorageOrchestrationServiceSettings settings, StorageUri serviceUri, string service) =>
+            credentials.AccountName ?? settings.StorageAccountDetails?.AccountName ?? serviceUri.GetAccountName(service) ?? string.Empty;
```
wsugarman, line 155: Is empty string a reasonable default if we cannot figure out what the account name is?

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -5,7 +5,7 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <FileVersion>1.10.1</FileVersion>
+    <FileVersion>1.10.2</FileVersion>
```
wsugarman, line 8: This seems like a patch fix as the API surface area is unchanged (I think). It's more like a bug fix

cgillum: Actually, you can revert this since we haven't released 1.10.1 yet.

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
cgillum: I would prefer something more explicit, like "(unknown)".

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
wsugarman: Ah my bad!

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
wsugarman: That makes sense! Added the default value

saguiitay:
> I also think it will be pretty uncommon for customers to use multiple storage accounts,

@cgillum using multiple storage accounts is needed when you scale out. You might start with a certain SA (which is shared be multiple workloads), and then "graduate" to a dedicated SA....

cgillum:
@saguiitay that's fair - I guess what I meant to say is that I think it will be uncommon for customers to split a single task hub across multiple storage accounts (e.g. queues in storage account A, blobs in storage account B, etc.). It would be more common to migrate between storage accounts by having two separate task hubs.

PR #673, opened 2022-02-01
Use the SimpleBufferManager when downloading Lease blobs

saguiitay:
This PR changes the `BlobLeaseManager` to use the `SimpleBufferManager` when reading the lease blobs. This is done since the call to `DownloadTextAsync` results in allocation of a new `MemoryStream`, which allocates a new `byte[]`. For high performance services, such allocations have a high impact on performance and GC times.


saguiitay:
@cgillum can this PR be completed?

cgillum (commented):
Merged! Thanks for the contribution!

PR #678, opened 2022-02-15
Reset NumTimeoutsHit periodically

### LINKED ISSUE #676: NumTimeoutsHit is never reset
The `NumTimeoutsHit` field in `TimeoutHandler` is never reset. This means that if you happen to have a sporadic timeout every now-and-then, the count keeps increasing, resulting in the process being killed.
While this might work for short-lived processes, this is impacting long-running processes, which might have sporadic storage timeouts, even if those are spread over a very long period.

A workaround could be adding a timestamp to "remember" when the last timeout happened, and reset it if enough time happened, or using some form of trickling-bucket algorithm.

saguiitay:
This PR resolves #676.

This PR resets the `NumTimeoutsHit` count in `TimeoutHandler`, to avoid long-running processes from being impacted by sporadic timeouts.

### DIFF
--- a/src/DurableTask.AzureStorage/TimeoutHandler.cs
+++ b/src/DurableTask.AzureStorage/TimeoutHandler.cs
@@ -29,8 +29,11 @@ namespace DurableTask.AzureStorage
         private const int MaxNumberOfTimeoutsBeforeRecycle = 5;
 
         private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(2);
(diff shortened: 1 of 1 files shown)


saguiitay:
@amdeel could you please take a look? This is a critical issue to ADX

saguiitay:
@cgillum, @amdeel could you please re-run the pipeline

cgillum (approved):
Re-run of the CI is green. Thanks for this contribution! It will be included automatically in our next planned release of DurableTask.AzureStorage.

PR #679, opened 2022-02-16
Adding OpenTelemetry tests

bachuv:
This PR includes end to end tests that check traces after running orchestrations. The tests include an orchestration that calls an activity and also includes orchestrations that handle external events.

### DIFF
--- a/test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -32,7 +32,12 @@ namespace DurableTask.AzureStorage.Tests
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Blob;
     using Microsoft.WindowsAzure.Storage.Table;
+    using Moq;
     using Newtonsoft.Json.Linq;
+#if NETCOREAPP3_1
+    using OpenTelemetry;
+    using OpenTelemetry.Trace;
+#endif
 
     [TestClass]
     public class AzureStorageScenarioTests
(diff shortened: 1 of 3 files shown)


test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -32,7 +32,12 @@ namespace DurableTask.AzureStorage.Tests
     using Microsoft.WindowsAzure.Storage;
     using Microsoft.WindowsAzure.Storage.Blob;
     using Microsoft.WindowsAzure.Storage.Table;
+    using Moq;
     using Newtonsoft.Json.Linq;
+#if NETCOREAPP3_1
```
cgillum, line 37: Rather than checking for NETCOREAPP3_1, I suggest checking for `#if !NET461` since we're much more likely to change from .NET Core 3.1 to something newer (like .NET 6) whereas we're very unlikely to change the .NET 4.6.1 dependency.

Not just here, but in other places as well.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2078,6 +2083,78 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
...
+                    var status = await client.WaitForCompletionAsync(StandardTimeout);
+
+                    await host.StopAsync();
+                }
+            }
+
+            // Explanation about indexes:
+            // The orchestration Activity's start at Invocation[1] and each action logs
+            // two activities - (Processor.OnStart(Activity) and Processor.OnEnd(Activity)
+            // so we start looking at invocations from index 2 and look at every other one from there
```
cgillum, line 2117: In addition to this explanation, it might be useful to include the expected output of the console exporter to the comments. Otherwise, it's hard for me to know what exact traces are expected.

cgillum (commented):
Great start! I think it would be good to include a test with external events as well, since that's another place where we emit traces.

cgillum (approved):

PR #681, opened 2022-02-18
Support for error propagation

cgillum:
The current error handling behavior of DurableTask.Core is problematic for a few reasons:

1. Exception information is lost if there are problems serializing or deserializing an unhandled exception
2. Serialized exception data in history is a string blob, making it difficult for tools to learn more about the failures
3. Out-of-process language workers may not support serializable exceptions at all, making error handling extra difficult

This PR alleviates these issues by introducing a `ErrorPropagationMode` setting on the task hub worker. The default behavior is (`ErrorPropagationMode.SerializeExceptions`) is the existing behavior, so there's no behavioral change introduced in this PR by default. However, when setting `ErrorPropagationMode` to `UseFailureDetails`, the `TaskFailedException` that gets raised when a activity or sub-orchestration failures will contain a new `FailureDetails` property with more structured information, such as the name of the exception (it's type), the exception message, and the full `exception.ToString()` value. In this mode, error propagation is consistent regardless of the language worker being used and regardless of whether a .NET exception is serializable.

The desire for this new feature is primarily to better support error propagation for out-of-process language workers.

Future extensions of `FailureDetails` could include property bags (key/value pairs) for additional information. However, that's considered out-of-scope for this PR.

This PR also updates the version of DurableTask.Core to **v2.8.0**.

### DIFF
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -58,7 +58,7 @@ namespace DurableTask.Core
         public bool IsReplaying { get; internal protected set; }
 
         /// <summary>
-        /// Gets or sets a value indicating how to propogate unhandled exception metadata.
+        /// Gets or sets a value indicating how to propagate unhandled exception metadata.
         /// </summary>
         internal ErrorPropagationMode ErrorPropagationMode { get; set; }
 
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -58,6 +58,11 @@ namespace DurableTask.Core
         public bool IsReplaying { get; internal protected set; }
 
         /// <summary>
+        /// Gets or sets a value indicating how to propogate unhandled exception metadata.
+        /// </summary>
+        internal ErrorPropagationMode ErrorPropagationMode { get; set; }
+
+        /// <summary>
         ///     Create a proxy client class to schedule remote TaskActivities via a strongly typed interface.
         /// </summary>
         /// <typeparam name="T">The interface for the proxy client</typeparam>
--- a/src/DurableTask.Core/TaskContext.cs
+++ b/src/DurableTask.Core/TaskContext.cs
@@ -33,7 +33,7 @@ namespace DurableTask.Core
         public OrchestrationInstance OrchestrationInstance { get; private set; }
 
         /// <summary>
-        /// Gets or sets a value indicating how to propogate unhandled exception metadata.
+        /// Gets or sets a value indicating how to propagate unhandled exception metadata.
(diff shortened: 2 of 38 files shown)


src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -57,6 +57,11 @@ public abstract class OrchestrationContext
         /// </summary>
         public bool IsReplaying { get; internal protected set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating how to propogate unhandled exception metadata.
```
davidmrdavid, line 61: tiny typo fix :)

```suggestion
        /// Gets or sets a value indicating how to propagate unhandled exception metadata.
```

src/DurableTask.Core/TaskContext.cs
```diff
@@ -31,5 +31,10 @@ public TaskContext(OrchestrationInstance orchestrationInstance)
         /// Gets the OrchestrationInstance for this task context
         /// </summary>
         public OrchestrationInstance OrchestrationInstance { get; private set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating how to propogate unhandled exception metadata.
```
davidmrdavid, line 36: tiny typo :)

```suggestion
        /// Gets or sets a value indicating how to propagate unhandled exception metadata.
```

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -418,10 +422,22 @@ public void HandleTaskFailedEvent(TaskFailedEvent failedEvent)
             if (this.openTasks.ContainsKey(taskId))
             {
                 OpenTaskInfo info = this.openTasks[taskId];
-                Exception cause = Utils.RetrieveCause(failedEvent.Details, this.ErrorDataConverter);
 
-                var taskFailedException = new TaskFailedException(failedEvent.EventId, taskId, info.Name, info.Version,
-                    failedEvent.Reason, cause);
+                // Legacy behavior is to try to serialize and deserialize exceptions from activities.
+                // However, this behavior is not reliable and can be turned off in favor of alternate settings.
+                Exception cause = this.ErrorPropagationMode == ErrorPropagationMode.SerializeExceptions ?
```
davidmrdavid, line 428: Would "cause" be included in `FailureDetails` if the alternative ErrorPropagationMode is used?
If so, it would be good to document this.

davidmrdavid (approved):
LGTM, just left a tiny suggestion on a typo I found.

That said, I'm still a bit unsure about how this blocks our Java work. Aren't Java exceptions serializable by default? I realize this structured approach is better for data inspection, but I'd like to triple-check if this was required, at a technical level, for Java SDK support.

PR #683, opened 2022-02-25
Added log when a decrease in partition count has been detected

amdeel:
When customers decrease their partition count without switching to a new task hub or removing their unused storage resources old blobs associated with partitions will cause partition load balancing to be inefficient.

This PR adds a partition error log message when this case is detected.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/BlobLeaseManager.cs
@@ -88,6 +88,18 @@ namespace DurableTask.AzureStorage.Partitioning
 
             blobLeases.AddRange(downloadTasks.Select(t => t.Result));
 
+            if (blobLeases.Count > this.settings.PartitionCount)
+            {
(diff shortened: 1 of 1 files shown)


cgillum:
If the partition count is decreased, do we still poll _all_ the queues (including the ones which we technically shouldn't be)? I'm not familiar with the expected behavior in this scenario. I thought we'd start polling only a subset of the queues (which could result in bigger problems) but that's different from the behavior you're describing and my own knowledge may be outdated (or wrong).

amdeel:
@cgillum AzureStorageOrchestrationService only creates ControlQueues based on the partition count. So these control queues will get all of the messages while we will still balance partitions based on workers according to the old partition count.

cgillum:
Okay, so clients will correctly load balance to the smaller number of queues but workers will still poll the larger number of queues? I was trying to understand if orchestrations could get stuck after reducing the partition count, but from what you’re saying, it sounds like that’s not the case. (If it was, I was thinking we’d want to add that information to the error message)

cgillum (approved):

PR #689, opened 2022-03-10
Adding an option to log structured events via LoggerFactory in Service Fabric provider

NeetArt:
## Summary
* DurableTask.Core library supports emitting structured logs if an `ILoggerFactory` is provided during `TaskHubWorker` or `TaskHubClient` creation. This pull requests adds an option to provide the logger as part of `FabricOrchestrationProviderSettings` which is then passed into the initialization of `TaskHubWorker` or `TaskHubClient` objects.
* Updated the verbosity level for `LogStoreCount` trace. This trace is more like a counter which is emitted every minute for all reliable collections, reporting the number of objects in the collection. This can turn out to be very noisy, and drowning out other legitimate traces.
* I have not yet converted the ServiceFabric traces into structured logs. That will come in as part of a future PR


NeetArt:
@shankarsama Can you please review this change?

shankarsama (approved):

PR #692, opened 2022-03-16
Reuse orchestration Activity to show an accurate representation of orchestration executions

bachuv:
This PR adds an update for OpenTelemetry support to reuse activities for each orchestration execution instead of creating a new one for each replay. This change improves the visuals in telemetry exporters because it doesn't show each replay. The images below show Gantt charts in Zipkin before and after making these changes.

Before:
image

After:
image

Summary of changes: For each new orchestration, we create an `Activity` and set that `Activity` as `DistributedTraceActivity.Current`. We can then reference that `Activity` during each orchestration replay to help with correlation. After the orchestration execution is done, we set `DistributedTraceActivity.Current` to null to reset the `Activity`.


cgillum:
@bachuv glad to see this PR! The "after" looks _much_ better than the "before"!

One thing I don't quite understand...what's the difference between the first two bars in the "after" image? I was sort of expecting there'd be one bar for the orchestration and one for each of the activities, for a total of four bars. What's the fifth one?

bachuv:
> One thing I don't quite understand...what's the difference between the first two bars in the "after" image? I was sort of expecting there'd be one bar for the orchestration and one for each of the activities, for a total of four bars. What's the fifth one?

The first bar is the client (the tags/metadata show "dt.type" = "client"). I can see how that can be confusing because the name isn't accurate. It shows the orchestration name.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -801,6 +795,10 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                 }
             }
 
+            DistributedTraceActivity.Current?.SetTag("dt.runtimestatus", runtimeState.OrchestrationStatus.ToString());
+            DistributedTraceActivity.Current?.Stop();
+            DistributedTraceActivity.Current = null;
```
cgillum, line 800: This code doesn't seem thread safe. `DistributedTraceActivity.Current` is an ordinary static variable, so if multiple orchestration threads are running in parallel, they may attempt to make conflicting modifications to the same activity and make the results unpredictable.

Should we be using an `AsyncLocal<T>` for `DistributedTraceActivity.Current` to avoid this problem?

cgillum (changes requested):

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -81,6 +86,10 @@ public class TraceHelper
                     new("dt.instanceid", startEvent.OrchestrationInstance.InstanceId),
```
jviau, line 86: As mentioned in the other PR, I recommend using `dtfx` and not `dt` as the prefix. `dt` may be confused with "Distributed Tracing". "dt" is already used in some telemetry schemas for that.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I added `AsyncLocal<T>` to `DistributedTraceActivity.Current`, but I'm running into an issue. I explained the changes I made below. Any guidance or help would be appreciated!

Changes:
1. Added `AsyncLocal<T>` to `DistributedTraceActivity`
```
public class DistributedTraceActivity
{
  static AsyncLocal<Activity> current = new AsyncLocal<Activity>();
  ...
  public static Activity Current
  {
      get { return current.Value; }
      set { current.Value = value; }
  }
  ...
}
```

2. With the previous change, we now need to introduce a way to "restore" the original orchestrator `Activity` during each replay since `DistributedTraceActivity.Current` is null during replays. To address that, I added a `SerializedActivity` string property to `ExecutionStartedEvent` to store the serialized orchestrator `Activity` that is originally created. 
 
This is set in `TraceHelper.StartTraceActivityForExecution()` after the initial orchestrator`Activity` is created.

```
public class ExecutionStartedEvent : HistoryEvent
{
  ...
  [DataMember]
  public string SerializedActivity { get; set; }
  ...
}
```

3. Once the `SerializedActivity` is created, we need a way to "restore" or deserialize the `SerializedActivity` during a replay to assign it back to `DistributedTraceActivity.Current`. The issue with this is that it throws the error below when deserializing the `Activity`. After researching, it seems like there's a read only enumerable in `Activity` that prevents it from getting deserialized. One workaround for this would be to reconstruct the Activity by manually reading the JSON, but this might lead to lost information between copies. 

```
JsonConvert.DeserializeObject<Activity>(json)

'JsonConvert.DeserializeObject<Activity>(json)' threw an exception of type 'System.NotSupportedException'
Message: "Collection was of a fixed size."
```

jviau: @bachuv I don't think we need to serialize the whole activity here and I would advise against it for performance reasons. I think we only need two things: `TraceContext` and `TraceParent`. Everything else, links, tags, events, etc should all be re-added to the activity in the same fashion they are added the first execution. What I am getting at is we can rely on the predictable replayability of DTFx Orchestrations to re-populate the `Activity` object the same way each replay.

This is also better behavior as now the presence of a tag, event, link, etc will not be duplicated and will be predictable in both existing and _point in time_ it is added to the activity (relative to the code execution position during a replay).

bachuv: @jviau Thank you for the explanation! That makes sense. It will also be simple and reliable. I started implementing that solution (serializing only `TraceParent`, `TraceState`, and `SpanId`) and I'm having trouble using reflection to set the `SpanId` in the "restored" `Activity`.  I tried using the following code to set the `SpanId`:
```
PropertyInfo propertyInfo = typeof(Activity).GetProperty("SpanId", BindingFlags.Public | BindingFlags.Instance);
propertyInfo.SetValue(restoredActivity, ActivitySpanId.CreateFromString(spanIdString.ToCharArray()))
```

and got this error
>System.ArgumentException: 'Property set method not found.'

This seems expected because `SpanId` doesn't have a setter ([reference](https://github.com/dotnet/runtime/blob/0e35b8d33851efde52e4e7aaadac239f3c2e97d5/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L743)). 

I'm still looking for workarounds to set the `SpanId`. Do you recommend trying anything else?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -801,6 +795,10 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                 }
             }
 
+            DistributedTraceActivity.Current?.SetTag("dt.runtimestatus", runtimeState.OrchestrationStatus.ToString());
+            DistributedTraceActivity.Current?.Stop();
```
jviau, line 799: When is this code hit? Is it after every replay, or only on completion? We want to make sure `.Stop()` is **only** called on completion (success or fail) of an orchestration and activity. It should not be called for replays.

For all replays, we need to manually set `Activity.Current` back to `Activity.Current.Parent`.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,109 @@
...
+using Newtonsoft.Json;
+
+namespace DurableTask.Core
+{
+    /// <summary>
+    /// Manage Activity for orchestration execution.
+    /// </summary>
+    public class DistributedTraceActivity
+    {
+        static AsyncLocal<Activity> current = new AsyncLocal<Activity>();
```
jviau, line 27: nit: add `private` access modifier.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        [JsonIgnore]
+        static JsonSerializerSettings CustomJsonSerializerSettings { get; }
+
+        /// <summary>
+        /// Share the Activity across an orchestration execution.
+        /// </summary>
+        public static Activity Current
+        {
+            get { return current.Value; }
```
jviau, line 47: Not sure the coding style of DTFx, but typically you do one of the following: 1) use `this.` modifier or 2) no `this.` modifier, but all instance fields are prefixed with `_`.

So either: `this.current` or `_current`.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,109 @@
...
+            SerializedActivity restored = JsonConvert.DeserializeObject<SerializedActivity>(json, CustomJsonSerializerSettings);
+
+            return restored;
+        }
+    }
+
+    /// <summary>
+    /// Manage serialized Activity
+    /// </summary>
+    public class SerializedActivity
```
jviau, line 87: Does this need to be `public`, can you make it a private nested class?

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        /// <summary>
+        /// DistributedTraceContext which stores the traceparent and tracestate
+        /// </summary>
+        public DistributedTraceContext DistributedTraceContext { get; set; }
+
+        /// <summary>
```
jviau, line 94: If `DistributedTraceContext` has the traceparent and tracestate, then we only need `SpanId` and `StartTime` here. `Id` can be built from `TraceParent` and `SpanId`.

src/DurableTask.Core/Tracing/FieldInfoExtensionMethods.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace DurableTask.Core.Tracing
+{
+    /// <summary>
+    /// Extensions for <see cref="FieldInfo"/>.
+    /// </summary>
+    public static class FieldInfoExtensionMethods
```
jviau, line 23: Make this `internal` please so it is not part of public API surface.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -71,16 +81,54 @@ public class TraceHelper
...
+                }
+
+                Activity? activity = ActivityTraceSource.StartActivity(
+                    name: startEvent.Name,
+                    kind: ActivityKind.Internal,
+                    parentContext: activityContext,
+                    startTime: restoredSerializedActivity.StartTime,
+                    tags: new KeyValuePair<string, object?>[]
+                    {
+                    new("dtfx.type", "orchestrator"),
```
jviau, line 116: nit: indentation.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace DurableTask.Core
+{
+    /// <summary>
+    /// Manage Activity for orchestration execution.
+    /// </summary>
+    public class DistributedTraceActivity
+    {
+        static AsyncLocal<Activity> current = new AsyncLocal<Activity>();
+
+        static DistributedTraceActivity()
```
jviau, line 29: not need for static ctor. You can move this to the static property:

``` CSharp
private static JsonSerializerSettings CustomJsonSerializerSettings { get; } = new() { ... }
```

But if it is private and used only here we can make it just a `readonly` field instead

jviau (changes requested):
Some comments for now, I will continue reviewing this next weel.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -39,9 +49,9 @@ public class TraceHelper
                 parentContext: Activity.Current?.Context ?? default,
                 tags: new KeyValuePair<string, object?>[]
                 {
-                    new("dt.type", "client"),
-                    new("dt.instanceid", startEvent.OrchestrationInstance.InstanceId),
-                    new("dt.executionid", startEvent.OrchestrationInstance.ExecutionId),
+                    new("dtfx.type", "client"),
```
jviau, line 52: It would be good to also set `ActivityKind` to an appropriate value based on the activity at had:

TaskHubClient -> `Producer` (this produces a new message on the DTFx queue)
TaskHubWorker processing an orchestration or activity -> `Server` (this is the 'server' running work)
TaskHubWorker sending an outbound orchestration or activity -> `Client` (this is acting as a 'client' waiting for external work to process)

`dtfx.type` should then be used to indicate what is the _task type_, as in it is only even `Activity` or `Orchestration` values. So it can be read as "this **ActivityKind.Server** span is processing a dtfx.type of **Orchestration**"

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -39,9 +49,9 @@ public class TraceHelper
...
+                    new("dtfx.instanceid", startEvent.OrchestrationInstance.InstanceId),
```
jviau, line 53: nit: I think OTel conventions have snake casing for single sections. So `dtfx.instance_id`, `dtfx.executiion_id`

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -39,9 +49,9 @@ public class TraceHelper
                 parentContext: Activity.Current?.Context ?? default,
```
jviau, line 49: I am not sure if you need to specify `parentContext` here - it may just infer it from `Activity.Current` for you?

jviau (changes requested):

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: It's hit when the execution is completed in `ProcessWorkflowCompletedTaskDecision()`.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
bachuv: `current` is static so it doesn't look like it applies in this situation. I renamed `current` to `CurrentActivity` though.

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
bachuv: `SerializedActivity` is used in `TraceHelper.StartTraceActivityForExecution()` after calling `DistributedTraceActivity.Restore()`.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: The [OTel documentation](https://opentelemetry.io/docs/instrumentation/net/manual/#add-tags-to-an-activity) that I found about tags doesn't show this. Is there another place where they explain this?

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
jviau: Oh, can it be `internal` then?

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
jviau: Sorry didn't notice that. Yeah, for status it is typically pascal case.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
jviau: Is `Activity.Current` set for every orchestration replay?

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau: ah yeah might not be formally documented - I even see a case hyphenation there. I was basing it off of examples in otel, such as `http.status_code`

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
bachuv: Yes, I changed it to `internal`.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: No, I don't think it is. Would it make sense to set it in TraceHelper.cs when initializing and restoring the Activity in [StartTraceActivityForExecution()](https://github.com/Azure/durabletask/blob/05ce0dbd57e27b409127568d803fc68bbd2da24e/src/DurableTask.Core/Tracing/TraceHelper.cs#L72)?

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: I talked to an engineer from the Azure Functions Monitoring team and they are following the conventions listed [here](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/faas/). To keep it consistent, I'll also use the same conventions which means changing dtfx.instanceid to dtfx.instance_id.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
jviau: Yes - it is critical that `Activity.Current` is set correctly for every replay. This helps with correlation, as more than just DurableTask will be looking at that activity for different purposes. A primary example is logging - any logs emitted from an Orchestration execution would want `Activity.Current` set so they can pull the traceId and spanId off of it for correlation.

bachuv: That makes sense. Do you have an example of what this would look like? I want to make sure I'm testing the scenario that you are describing.

jviau: Yeah the general idea would be:

1. Before being sent to the orchestration dispatcher pipeline, DTFx.Core would prepare and set `Activity.Current`. Tags, start time, spanId, traceId, all of that would be set the same on every replay (with the IDs being set via reflection).
    - On the _first_ play, you wouldnt need to do the reflection for IDs as you want to let it generate that for you.
1. On a replay finishing you would need to branch based on if the orchestration is terminal or not (succeeded or failed).
    - Is terminal, dispose the activity so the "Stop" flow is triggered.
    - Is not terminal, do **not** dispose the activity or call `.Stop()`. Manually set `Activity.Parent` back to what it was _before_ running this orchestration (`Activity.Current.Parent`).
        - We will need to document that instrumentation libraries or listeners of this activity must NOT hold onto it in memory and expect a `.Stop()` call for this activity in all cases.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: It looks like `parentContext` is needed based on looking at the different StartActivity [overloads](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource.startactivity?view=net-6.0#overloads). It probably doesn't hurt to set it either.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: Thanks for the explanation! I was originally asking for an example of the logging scenario that you were explaining:

> A primary example is logging - any logs emitted from an Orchestration execution would want Activity.Current set so they can pull the traceId and spanId off of it for correlation.

Sorry I didn't clarify that.

jviau: Ah okay. So this is in regards to just using the `ILogger` from `Microsoft.Extensions.Logging`. Many logger implementations will check `Activity.Current` on log and pull a couple extra dimensions off of it. Take the application insights `ILogger` implementation, it will send the log into its telemetry pipeline, which has a step to check `Activity.Current` and populate `operation_Id` and `parent_Id` for logs. With this I can now associate all logs back to not just what orchestration emitted the log (category name), but also back to the trace context of that orchestration.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -71,16 +79,54 @@ public class TraceHelper
...
+                 parentContext: activityContext,
+                 tags: new KeyValuePair<string, object?>[]
+                 {
+                    new("dtfx.type", "orchestrator"),
+                    new("dtfx.instance_id", startEvent.OrchestrationInstance.InstanceId),
+                    new("dtfx.execution_id", startEvent.OrchestrationInstance.ExecutionId),
+                 });
+
+                startEvent.SerializedActivity = DistributedTraceActivity.SerializeActivity(startEvent.ParentTraceContext, DistributedTraceActivity.Current);                
+            }
```
cgillum, line 96: nit: the indentation for this block needs to be cleaned up.

src/DurableTask.Core/History/ExecutionStartedEvent.cs
```diff
@@ -90,6 +90,12 @@ internal ExecutionStartedEvent()
         [DataMember]
         public string Correlation { get; set; }
 
+        /// <summary>
+        /// Gets or sets the serialized end-to-end trace activity.
+        /// </summary>
+        [DataMember]
+        public string SerializedActivity { get; set; }
```
cgillum, line 97: It's not very clear to me what the difference between `SerializedActivity` is and the `ParentTraceContext` property that was added earlier. Can you expand on the summary comment to help make the distinction more clear?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: > 2. On a replay finishing you would need to branch based on if the orchestration is terminal or not (succeeded or failed).
>     - Is terminal, dispose the activity so the "Stop" flow is triggered.
>     - Is not terminal, do **not** dispose the activity or call `.Stop()`. Manually set `Activity.Parent` back to what it was _before_ running this orchestration (`Activity.Current.Parent`).
>         - We will need to document that instrumentation libraries or listeners of this activity must NOT hold onto it in memory and expect a `.Stop()` call for this activity in all cases.

@jviau I believe this scenario is already covered with the changes in `ProcessWorkflowCompletedTaskDecision()`. Can you clarify the "terminal/is not terminal" scenario that you are describing with an example?

jviau: Terminal == orchestration is complete, whether it is success, fail, or termination/cancelled. It will not be ran again and the activity should be tracked now.
Not-Terminal == the opposite of above - orchestration has more work / will be replayed later.

src/DurableTask.Core/History/ExecutionStartedEvent.cs
```diff
@@ -90,6 +90,15 @@ internal ExecutionStartedEvent()
...
         public string Correlation { get; set; }
 
+        /// <summary>
+        /// SerializedActivity includes the following serialized
+        /// information: ParentTraceContext, SpanId, and the Activity's StartTime.
+        /// It's set when the Activity is first created. This information is saved
+        /// and used to restore the Activity during replays.
+        /// </summary>
+        [DataMember]
+        public string SerializedActivity { get; set; }
```
cgillum, line 100: I have a couple concerns with this design:

1. I'm not comfortable having a JSON-serialized string field in the event history because this creates a tighter dependency on Newtonsoft.Json. It's also inefficient because we're double serializing data.
2. I'm also not comfortable with having two different properties that overlap with each other, partly because it's confusing and partly because it's inefficient: SerializedActivity internally includes a DistributedTraceContext object, and yet we have another DistributedTraceContext below.

I would be happier if we could somehow combine both the parent context and the current activity context into a single `DistributedTraceContext` field and remove all serialized string properties. These history events are already serialized using whatever serializer the backend storage provider chooses, so there shouldn't be any need to introduce nested serialization.

cgillum (changes requested):

src/DurableTask.Core/History/ExecutionStartedEvent.cs
bachuv: This makes sense now looking at it again. The design should definitely change to store all relevant information in one place.

To restore the activity, we actually only need the original Activity's `SpanId` and `StartTime` so we don't have to store the entire Activity. This would mean that we need to save the following:
1. Parent trace context
2. Current activity's SpanId
3. Current activity's start time

There are 2 design implementations I could see as options:
1. Create a new class, `DistributedTraceInfo` (or another name) which stores those 3 fields. `ExecutionStartedEvent` would include a `DistributedTraceInfo` field and `TraceHelper.StartTraceActivityForExecution()` would access and modify these fields when creating and restoring the current activity.
2. `ExecutionStartedEvent` would store the 3 fields mentioned above without a wrapper class like in Option 1 and `TraceHelper.StartTraceActivityForExecution()` would access and modify these fields directly on the `ExecutionStartedEvent`. 

Option 1 seems like it would be organized and clear to the reader about how those 3 fields are related. @cgillum What are your thoughts about this? Is there another design option that you recommend?

cgillum: @bachuv instead of introducing a new class, could we just add `SpanId` and `ActivityStartTime` fields to the existing `DistributedTraceContext` property?

I like the idea of have just one "wrapper" class because it simplifies the backend-specific serialization changes we have to make. :)

bachuv: Yes, I just pushed those changes. Let me know if I should make any changes!

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -71,16 +79,54 @@ public class TraceHelper
...
+                    name: startEvent.Name,
+                    kind: ActivityKind.Internal,
+                    parentContext: activityContext,
+                    startTime: (DateTimeOffset)startEvent.ParentTraceContext.ActivityStartTime,
+                    tags: new KeyValuePair<string, object?>[]
+                    {
+                        new("dtfx.type", "orchestrator"),
+                        new("dtfx.instance_id", startEvent.OrchestrationInstance.InstanceId),
+                        new("dtfx.execution_id", startEvent.OrchestrationInstance.ExecutionId),
+                    });
```
cgillum, line 117: It looks like we're creating the same object in both the `if` and `else` cases. If so, can we refactor this a bit to reduce duplication?

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace DurableTask.Core.Tracing
+{
+    /// <summary>
+    /// Manage Activity for orchestration execution.
+    /// </summary>
+    internal class DistributedTraceActivity
+    {
+        private static AsyncLocal<Activity> CurrentActivity = new AsyncLocal<Activity>();
+
+        private static readonly JsonSerializerSettings CustomJsonSerializerSettings = new JsonSerializerSettings()
```
cgillum, line 28: I didn't see this referenced anywhere. Can it be removed?

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: Thanks for bringing this up! I created Issue #718 to track this and added it to the [task list](https://github.com/Azure/durabletask/pull/648#issue-1067859057) for the main DurableTask OTel PR.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: The `Activity` that's created in the `else` case includes setting a read-only `startTime` field as a parameter. I refactored both cases to reduce some of the duplication though.

cgillum (approved):

src/DurableTask.Core/Tracing/DistributedTraceActivity.cs
```diff
@@ -0,0 +1,37 @@
...
+using Newtonsoft.Json;
+
+namespace DurableTask.Core.Tracing
+{
+    /// <summary>
+    /// Manage Activity for orchestration execution.
+    /// </summary>
+    internal class DistributedTraceActivity
+    {
+        private static AsyncLocal<Activity> CurrentActivity = new AsyncLocal<Activity>();
```
jviau, line 26: nit: consider adding `readonly` modifier.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -28,9 +29,16 @@ public class TraceHelper
     {
         const string Source = "DurableTask";
 
-        // TODO: Add tracing for external event send, which could be used to initialize an entity
         static readonly ActivitySource ActivityTraceSource = new ActivitySource(Source);
 
+        private static readonly Action<Activity, string> s_spanIdSet;
```
jviau, line 34: I think you also need to set `_id` as well.

jviau (changes requested):
Just the setting of `_id` ontop of `_spanId`. I _think_ at least, I did this in my implementation. I imagine I had a reason for it?

This is looking awesome though!

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -71,16 +79,53 @@ public class TraceHelper
...
+                if (string.IsNullOrEmpty(spanId) || startEvent.ParentTraceContext.ActivityStartTime == null)
                 {
-                    new("dt.type", "orchestrator"),
-                    new("dt.instanceid", startEvent.OrchestrationInstance.InstanceId),
-                    new("dt.executionid", startEvent.OrchestrationInstance.ExecutionId),
-                });
+                    return null;
+                }
+
+                Activity? activity = ActivityTraceSource.StartActivity(
```
jviau, line 111: Can we use the activity that was already started on line 44?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -71,16 +79,53 @@ public class TraceHelper
...
+            KeyValuePair<string, object?>[] activityTags = new KeyValuePair<string, object?>[]
+            {
+                new("dtfx.type", "orchestrator"),
+                new("dtfx.instance_id", startEvent.OrchestrationInstance.InstanceId),
+                new("dtfx.execution_id", startEvent.OrchestrationInstance.ExecutionId),
+            };
+
+            if (string.IsNullOrEmpty(startEvent.ParentTraceContext.SpanId))
+            {
+                DistributedTraceActivity.Current = ActivityTraceSource.StartActivity(
```
jviau, line 93: Same here - this function already starts an activity. We should be re-using that and not starting multiple.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: The activity seems to get restored using just the `SpanId`. Can you explain why we would need both? It looks like `Activity.Id` is nullable whereas `Activity.SpanId` isn't nullable.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: The `Activity` started on line 44 is for the `client` function and this one is for the `orchestrator`. Can you clarify why we would use that `Activity`?

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau: `_id` is lazily constructed in `Activity`. So if something does access `Id` before you set `_spanId`, then `_id` will be snapped into place with the old (now invalid) `_spanId`.

You may have someone listening to the `OnStart` event via an `ActivityListener` and they could read `Id` which will snap it in place. So right now it may appear to work, but a consumer of this may easily break it.

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau: Ah yeah disregard what I said. I didn't expand the collapsed code sections in the PR view enough.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: Thank you for the explanation! That makes sense. Sorry for the delayed response to this PR feedback. I just pushed new changes based on your comments. Let me know if you have any other comments and I'll make updates! Thanks!

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -71,16 +81,42 @@ public class TraceHelper
...
-                });
+                startTime: startTime,
+                tags: activityTags);
+
+            if (activity == null)
+            {
+                return null;
+            }
+
+            if (startEvent.ParentTraceContext.Id != null && startEvent.ParentTraceContext.SpanId != null)
```
bachuv, line 106: @jviau I changed this check to `!= null` instead of `!string.IsNullOrEmpty()` because `startEvent.ParentTraceContext.Id` and `startEvent.ParentTraceContext.SpanId` are nullable so `string.IsNullOrEmpty()` wasn't enough for a check and it showed up as an error. Are there scenarios where those values would fall into the `OrEmpty` case from the `string` method? If so, would it be worth adding both checks?

jviau: Ah yeah you are using strict nullables. You can solve this by adding a `!` (you tell the compiler "I know this is not null here") after your usage of it on line 108 and 109. So:

``` CSharp
if (string.IsNullOrEmpty(startEvent.ParentTraceContext.Id) && string.IsNullOrEmpty(startEvent.ParentTraceContext.SpanId))
{
    s_idSet(activity, startEvent.ParentTraceContext.Id!);
    s_spanIdSet(activity, startEvent.ParentTraceContext.SpanId!);
}
```

jviau (approved):
This looks awesome!

PR #696, opened 2022-03-18
Add Non-Generic Overload for AddTaskActivitiesFromInterface

aueloka:
This change does the following:

* Adds a new non-generic `AddTaskActivitiesFromInterface` overload that lets the caller pass in a type instead.

### DIFF
--- a/src/DurableTask.Core/TaskHubWorker.cs
+++ b/src/DurableTask.Core/TaskHubWorker.cs
@@ -314,11 +314,24 @@ namespace DurableTask.Core
         /// </summary>
         /// <typeparam name="T">Interface</typeparam>
         /// <param name="activities">Object that implements this interface</param>
+        public TaskHubWorker AddTaskActivitiesFromInterface<T>(T activities)
+        {
+            return this.AddTaskActivitiesFromInterface(activities, false);
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -314,9 +314,13 @@ public TaskHubWorker AddTaskActivities(params ObjectCreator<TaskActivity>[] task
         /// </summary>
         /// <typeparam name="T">Interface</typeparam>
         /// <param name="activities">Object that implements this interface</param>
-        public TaskHubWorker AddTaskActivitiesFromInterface<T>(T activities)
```
cgillum, line 317: This is technically a breaking change from a binary compatibility perspective. Can we leave this method signature in?

cgillum (changes requested):
Thanks for submitting this PR! Requesting one small change:

src/DurableTask.Core/TaskHubWorker.cs
aueloka: Sounds good. I'll revert that.

cgillum (approved):
Looks good to me!

aueloka:
Thanks

aueloka:
@cgillum , I may need help merging this since I don't have write access to this repository.

PR #699, opened 2022-03-31
Include LeaseType in lease-stealing flow

saguiitay:
Include LeaseType in lease-stealing flow. This should make it easier to troubleshoot lease-stealing flows.

### DIFF
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -757,6 +757,7 @@ namespace DurableTask.AzureStorage
             string TaskHub,
             string WorkerName,
             string FromWorkerName,
+            string LeaseType,
             string PartitionId,
             string AppName,
             string ExtensionVersion)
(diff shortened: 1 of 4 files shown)


saguiitay:
@cgillum

cgillum:
@amdeel can you take a look?

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -778,6 +780,7 @@ public void GeneralError(string Account, string TaskHub, string Details, string
             string TaskHub,
             string WorkerName,
             string FromWorkerName,
+            string LeaseType,
```
cgillum, line 783: Since we're adding a new field, we need to update the `Version` attribute value above to `3`.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -757,6 +757,7 @@ public void GeneralError(string Account, string TaskHub, string Details, string
...
```
cgillum, line 760: Since we're adding a new field, we need to update the `Version` attribute value above to `3`.

cgillum (changes requested):

cgillum (approved):
LGTM!

PR #704, opened 2022-04-13
Purge instance history support

shreyas-gopalakrishna:
- Adds purge instance history support

Reference: https://github.com/microsoft/durabletask-java/issues/10

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1772,7 +1772,7 @@ namespace DurableTask.AzureStorage
         /// </summary>
         /// <param name="instanceId">Instance ID of the orchestration.</param>
         /// <returns>Class containing number of storage requests sent, along with instances and rows deleted/purged</returns>
-        public Task<Core.PurgeHistoryResult> PurgeInstanceHistoryAsync(string instanceId)
+        public Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string instanceId)
         {
             return this.trackingStore.PurgeInstanceHistoryAsync(instanceId);
         }
@@ -1784,7 +1784,7 @@ namespace DurableTask.AzureStorage
         /// <param name="createdTimeTo">CreatedTime of orchestrations. Purges history less than this value.</param>
         /// <param name="runtimeStatus">RuntimeStatus of orchestrations. You can specify several status.</param>
         /// <returns>Class containing number of storage requests sent, along with instances and rows deleted/purged</returns>
-        public Task<Core.PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus)
+        public Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus)
         {
             return this.trackingStore.PurgeInstanceHistoryAsync(createdTimeFrom, createdTimeTo, runtimeStatus);
         }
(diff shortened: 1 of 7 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1783,11 +1784,33 @@ public Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string instanceId)
         /// <param name="createdTimeTo">CreatedTime of orchestrations. Purges history less than this value.</param>
         /// <param name="runtimeStatus">RuntimeStatus of orchestrations. You can specify several status.</param>
         /// <returns>Class containing number of storage requests sent, along with instances and rows deleted/purged</returns>
-        public Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus)
+        public Task<Core.PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus)
```
shreyas-gopalakrishna, line 1787: Should we retain this to AzureStorage/PurgeHistoryResult and convert it to Core/PurgeHistoryResult in the `PurgeInstanceStateAsync` method?

src/DurableTask.Core/IOrchestrationServicePurgeClient.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public interface IOrchestrationServicePurgeClient
+    {
+        // Purge instance history operations
+
+        /// <summary>
+        /// Purge history for an orchestration with a specified instance id.
+        /// </summary>
+        /// <param name="instanceId">Instance ID of the orchestration.</param>
+        /// <returns><see cref="PurgeHistoryResult"/> object containing number of storage requests sent, along with instances and rows deleted/purged.</returns>
```
cgillum, line 31: This description assumes the Azure Storage backend, which isn't always going to be the one used. Let's simplify it.
```suggestion
        /// <returns><see cref="PurgeHistoryResult"/> object containing more information about the purged instance.</returns>
```

src/DurableTask.Core/PurgeHistoryResult.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Number of requests sent to Storage during this execution of purge history
+        /// </summary>
+        public int StorageRequests { get; }
+
+        /// <summary>
+        /// Number of instances deleted during this execution of purge history
+        /// </summary>
+        public int InstancesDeleted { get; }
```
cgillum, line 42: For DurableTask.Core, this should be the only property we use. The other two may not always be relevant. We can add more properties later that may be universally interesting. For example, the number of bytes freed.

src/DurableTask.Core/IOrchestrationServicePurgeClient.cs
```diff
@@ -0,0 +1,43 @@
...
+        Task<PurgeHistoryResult> PurgeInstanceStateAsync(string instanceId);
+
+        /// <summary>
+        /// Purge history for orchestrations that match the specified parameters.
+        /// </summary>
+        /// <param name="createdTimeFrom">CreatedTime of orchestrations. Purges history grater than this value.</param>
+        /// <param name="createdTimeTo">CreatedTime of orchestrations. Purges history less than this value.</param>
+        /// <param name="runtimeStatus">RuntimeStatus of orchestrations. You can specify several status.</param>
+        /// <returns><see cref="PurgeHistoryResult"/> object containing number of storage requests sent, along with instances and rows deleted/purged.</returns>
+        Task<PurgeHistoryResult> PurgeInstanceStateAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus);
```
cgillum, line 41: Rather than having discrete parameters, I think we should have a single object with multiple properties as the one parameter. That way we can more easily add new criteria without making a breaking interface change.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: We need to retain the existing methods. Otherwise, we're introducing a breaking change. which we need to avoid for now. This best way to make this work is to have `AzureStorageOrchestrationService` implement the `IOrchestrationServicePurgeClient` interface explicitly. The syntax will look something like this:

```csharp
Task<Core.PurgeHistoryResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync(string instanceId)
{
    return this.PurgeInstanceHistoryAsync(instanceId);
}

async Task<Core.PurgeHistoryResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus)
{
    PurgeHistoryResult storagePurgeResult = await this.PurgeInstanceHistoryAsync(createdTimeFrom, createdTimeTo, runtimeStatus);
    return storagePurgeResult.ToCorePurgeResult(); // <-- some new method that converts between the two types
}
```

src/DurableTask.AzureStorage/PurgeHistoryResult.cs
```diff
@@ -13,9 +13,11 @@
 
 namespace DurableTask.AzureStorage
 {
+    using System;
     /// <summary>
     /// Class to hold statistics about this execution of purge history
     /// </summary>
+    [Obsolete("Not used any more. Will use DurableTask.Core.PurgeHistoryResult going forward.", true)]
```
cgillum, line 20: Let's not mark this as obsolete yet, since introducing warnings could be a breaking change for some. Also, the new Core.PurgeHistoryResult isn't a perfect replacement, so we need to think through the deprecation strategy a bit more (but this can be done later).

cgillum (changes requested):
Thanks for this PR! I think we need to make this change in a way that's non-breaking. See my comments below.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1788,6 +1789,28 @@ public Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTime createdTimeFr
...
+        {
+            PurgeHistoryResult storagePurgeHistoryResult = await this.PurgeInstanceHistoryAsync(instanceId);
+            return storagePurgeHistoryResult.ToCorePurgeHistoryResult();
+        }
+
+        /// <summary>
+        /// Purge history for orchestrations that match the specified parameters.
+        /// </summary>
+        /// <param name="purgeInstanceCondition"><see cref="PurgeInstanceCondition"/>Conditions that should match to purge orchestration history.</param>
+        /// <returns><see cref="PurgeResult"/> object containing more information about the purged instance.</returns>
```
cgillum, line 1807: Rather than duplicating the description of this method, just add `/// <inheritdoc />`. That will cause it to refer to the documentation comments on the interface declaration.

src/DurableTask.Core/IOrchestrationServicePurgeClient.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Purge history for an orchestration with a specified instance id.
+        /// </summary>
+        /// <param name="instanceId">Instance ID of the orchestration.</param>
+        /// <returns><see cref="PurgeResult"/> object containing more information about the purged instance.</returns>
+        Task<PurgeResult> PurgeInstanceStateAsync(string instanceId);
+
+        /// <summary>
+        /// Purge history for orchestrations that match the specified parameters.
+        /// </summary>
+        /// <param name="purgeInstanceCondition"><see cref="PurgeInstanceCondition"/>Conditions that should match to purge orchestration history.</param>
```
cgillum, line 35: The formatting of this comment doesn't look right. The description is also a little ambiguous. Here's a suggestion:

```suggestion
        /// <param name="purgeInstanceCondition">A <see cref="PurgeInstanceCondition"/> object that determines which orchestration instance to purge.</param>
```

src/DurableTask.Core/PurgeInstanceCondition.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace DurableTask.Core
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Class to hold conditions that should match to purge orchestration history
+    /// </summary>
+    public class PurgeInstanceCondition
```
cgillum, line 22: I gave similar feedback in Kaibo's PR, but let's not use the "Condition" suffix. It doesn't sound natural. Something like `PurgeInstanceCriteria` or `PurgeInstanceFilter` would be better.

cgillum (approved):

PR #705, opened 2022-04-14
Include assembly type for the exception on FailureDetails

SamGuoMsft:
Add exception assembly type info also in FailureDetails

### DIFF
--- a/src/DurableTask.Core/FailureDetails.cs
+++ b/src/DurableTask.Core/FailureDetails.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.Core
 {
     using System;
+    using System.Reflection;
     using System.Runtime.Serialization;
     using DurableTask.Core.Exceptions;
     using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/FailureDetails.cs
```diff
@@ -29,14 +30,16 @@ public class FailureDetails : IEquatable<FailureDetails>
         /// </summary>
         /// <param name="errorType">The name of the error, which is expected to the the namespace-qualified name of the exception type.</param>
         /// <param name="errorMessage">The message associated with the error, which is expected to be the exception's <see cref="Exception.Message"/> property.</param>
+        /// <param name="errorAssemblyName">The assembly name of the error.</param>
         /// <param name="stackTrace">The exception stack trace.</param>
         /// <param name="innerFailure">The inner cause of the failure.</param>
         /// <param name="isNonRetriable">Whether the failure is non-retriable.</param>
         [JsonConstructor]
-        public FailureDetails(string errorType, string errorMessage, string? stackTrace, FailureDetails? innerFailure, bool isNonRetriable)
+        public FailureDetails(string errorType, string errorMessage, string errorAssemblyName, string? stackTrace, FailureDetails? innerFailure, bool isNonRetriable)
```
cgillum, line 38: Since we've already shipped this public API, we can't make breaking changes to it. Please add a new constructor which includes the "assembly" name.

cgillum (commented):
Thanks for submitting this PR! A couple things:

1. Since this is a shipped, public API, any changes we make need to be non-breaking.
2. The concept of "assemblies" is .NET-specific but the goal of this API is to be platform agnostic.

I'm wondering if we can either use a more generic name, like "package" or "source" instead of "assembly". Alternatively, we could try to put the fully-qualified type name in the `ErrorType` field, which wouldn't require any public surface API changes.

Thoughts?

PR #708, opened 2022-04-19
Added Generation to ExecutionStartedEvent for Dedupe Checks

amdeel:
This PR addresses issues when running dedupe orchestration started checks when reusing instanceIds. Currently if the instances table hasn't been updated by the time the dedupe checks run, the execution stated event can be discarded if the instanceId was reused. This is because the remoteInstance will be the previous instance execution, instead of the current instance execution. This PR addresses this issue by adding the column Generation to the Instances table, which represents the sequence number of each instance to ensure the correct execution is being represented in remoteInstance.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1561,8 +1561,6 @@ namespace DurableTask.AzureStorage
                 {
                     executionStartedEvent.Generation = 1;
                 }
-
-                creationMessage.Event = executionStartedEvent;
             }
 
             ControlQueue controlQueue = await this.GetControlQueueAsync(creationMessage.OrchestrationInstance.InstanceId);
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1578,8 +1578,6 @@ namespace DurableTask.AzureStorage
                 inputStatusOverride = this.messageManager.GetBlobUrl(internalMessage.CompressedBlobName);
             }
 
-            Thread.Sleep(5000);
-
             await this.trackingStore.SetNewExecutionAsync(
                 executionStartedEvent,
                 existingInstance?.ETag,
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1551,6 +1551,20 @@ namespace DurableTask.AzureStorage
                 return;
             }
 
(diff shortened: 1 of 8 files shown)


src/DurableTask.Core/OrchestrationState.cs
```diff
@@ -107,6 +107,12 @@ public class OrchestrationState : IExtensibleDataObject
         [DataMember]
         public string Version;
 
+        /// <summary>
+        /// The orchestration gneration. Reused instanceIds will increment this value.
```
cgillum, line 111:
```suggestion
        /// The orchestration generation. Reused instanceIds will increment this value.
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1551,6 +1551,20 @@ public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, Orch
...
+                if (existingInstance != null)
+                {
+                    executionStartedEvent.Generation = existingInstance.State.Generation + 1;
+                }
+                else
+                {
+                    executionStartedEvent.Generation = 1;
+                }
+
+                creationMessage.Event = executionStartedEvent;
```
cgillum, line 1565: This is redundant, right? `creationMessage.Event` already references `executionStartedEvent`, as we can see on line 1531.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -235,13 +235,14 @@ async Task DequeueLoop(string partitionId, ControlQueue controlQueue, Cancellati
...
                 OrchestrationInstance localInstance = message.TaskMessage.OrchestrationInstance;
+                var expectedGeneration = ((ExecutionStartedEvent)message.TaskMessage.Event).Generation;
                 if (remoteOrchestrationsById.TryGetValue(localInstance.InstanceId, out OrchestrationState remoteInstance) &&
                     (remoteInstance.OrchestrationInstance.ExecutionId == null || string.Equals(localInstance.ExecutionId, remoteInstance.OrchestrationInstance.ExecutionId, StringComparison.OrdinalIgnoreCase)))
                 {
                     // Happy path: The message matches the table status. Alternatively, if the table doesn't have an ExecutionId field (older clients, pre-v1.8.5),
                     // then we have no way of knowing if it's a duplicate. Either way, allow it to run.
                 }
-                else if (this.IsScheduledAfterInstanceUpdate(message, remoteInstance))
+                else if (expectedGeneration == remoteInstance?.Generation && this.IsScheduledAfterInstanceUpdate(message, remoteInstance))
```
cgillum, line 245: Would it be useful to do `<=` instead of `==`?

cgillum (approved):
Looks good. Just a couple very minor optional things.

PR #715, opened 2022-04-26
Use `using` for disposable objects

saguiitay:
Use `using` when working with IDisposable objects. This allows better memory management by the GC.


cgillum (approved):
I never noticed this - thanks for this PR!

PR #716, opened 2022-04-26
Avoid many instances of JsonDataConverter

saguiitay:
Since JsonDataConverter is an immutable class, there's no point creating new instances it of it every time Activity/Orchestration/Context/Client objects are created.
Users are still able to provide their own implementation as before.

### DIFF
--- a/src/DurableTask.Core/ReflectionBasedTaskActivity.cs
+++ b/src/DurableTask.Core/ReflectionBasedTaskActivity.cs
@@ -26,6 +26,8 @@ namespace DurableTask.Core
     /// </summary>
     public class ReflectionBasedTaskActivity : TaskActivity
     {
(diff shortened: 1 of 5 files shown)


davidmrdavid (commented):
Thanks for this PR, it seems like a good optimization to me. It's got my tentative LGTM, but I'll defer to @cgillum or @amdeel, who look after this repo more often, to give the final word.

cgillum (approved):
This looks good to me, though it would be even better if you were to add a static `JsonDataConverter.Default` property and replace all the individual `defaultDataConverter` objects with this one static singleton.

saguiitay:
@cgillum done

cgillum (approved):
LGTM!

PR #717, opened 2022-04-26
Fixes for new partition management

sebastianburckhardt:
This fixes a bug in the partition management logic for ownership leases: all available ownership leases (for partitions that are expired and for which intent leases are owned) must be acquired, regardless of how many nodes there are and how many partitions have been already acquired. The balancing is already taken care of by the intent leases.

Also, this PR adds an exclusion so lease renewals are not subjected to request throttling; because request throttling on lease renewals can be counterproductive in heavily loaded situations.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
@@ -171,115 +171,102 @@ namespace DurableTask.AzureStorage.Partitioning
             }
         }
 
-        Task LeaseRenewer()
+        async Task LeaseRenewer()
         {
-            // use a dedicated thread to make the lease renewal less susceptible to thread pool starvation
-            var taskCompletionSource = new TaskCompletionSource<bool>();
-            var leaseRenewalThread = new Thread(LeaseRenewalLoop);
-            leaseRenewalThread.Name = "LeaseRenewalThread";
-            leaseRenewalThread.Start();
-            return taskCompletionSource.Task;
+            this.settings.Logger.PartitionManagerInfo(
+                this.accountName,
+                this.taskHub,
+                this.workerName,
+                string.Empty /* partitionId */,
+                $"Starting background renewal of {this.leaseType} leases with interval: {this.options.RenewInterval}.");
 
-            void LeaseRenewalLoop()
+            while (this.isStarted == 1 || !shutdownComplete)
             {
-                this.settings.Logger.PartitionManagerInfo(
-                    this.accountName,
-                    this.taskHub,
-                    this.workerName,
-                    string.Empty /* partitionId */,
-                    $"Starting background renewal of {this.leaseType} leases with interval: {this.options.RenewInterval}.");
-
-                var stopwatch = new Stopwatch();
-                var leaseTasks = new List<Task>();
-
-                while (this.isStarted == 1 || !this.shutdownComplete)
+                try
                 {
-                    try
+                    var nonRenewedLeases = new ConcurrentBag<T>();
+                    var renewTasks = new List<Task>();
+
+                    // Renew leases for all currently owned partitions in parallel
+                    foreach (T lease in this.currentlyOwnedShards.Values)
                     {
-                        this.settings.Logger.PartitionManagerInfo(
-                           this.accountName,
-                           this.taskHub,
-                           this.workerName,
-                           string.Empty /* partitionId */,
-                           $"Starting next round of {this.leaseType} lease renewals after {stopwatch.Elapsed}.");
-
-                        stopwatch.Restart();
-
-                        foreach (T lease in this.currentlyOwnedShards.Values.ToList())
+                        if (this.shouldRenewLeaseDelegate(lease.PartitionId))
                         {
-                            leaseTasks.Add(RenewOwnedLease(lease));
-                        }
-                        foreach (T shutdownLease in this.keepRenewingDuringClose.Values.ToList())
-                        {
-                            leaseTasks.Add(RenewShutdownLease(shutdownLease));
-                        }
-
-                        Task.WhenAll(leaseTasks).Wait();
-                        leaseTasks.Clear();
-
-                        var delayBy = this.options.RenewInterval - stopwatch.Elapsed;
-                        if (delayBy > TimeSpan.Zero)
-                        {
-                            this.leaseRenewerCancellationTokenSource.Token.WaitHandle.WaitOne(delayBy);
-                        }
-
-                        async Task RenewOwnedLease(T lease)
-                        {
-                            if (this.shouldRenewLeaseDelegate(lease.PartitionId))
+                            renewTasks.Add(this.RenewLeaseAsync(lease).ContinueWith(renewResult =>
                             {
-                                if (!await this.RenewLeaseAsync(lease))
+                                if (!renewResult.Result)
                                 {
-                                    await this.RemoveLeaseAsync(lease, false);
+                                    // Keep track of all failed attempts to renew so we can trigger shutdown for these partitions
+                                    nonRenewedLeases.Add(lease);
                                 }
-                            }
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -115,18 +115,21 @@ public Table GetTableReference(string tableName)
         public Task<T> MakeTableStorageRequest<T>(Func<OperationContext, CancellationToken, Task<T>> storageRequest, string operationName, string? clientRequestId = null) =>
             this.MakeStorageRequest<T>(storageRequest, TableAccountName, operationName, clientRequestId);
 
-        public Task MakeBlobStorageRequest(Func<OperationContext, CancellationToken, Task> storageRequest, string operationName, string? clientRequestId = null) =>
-            this.MakeStorageRequest(storageRequest, BlobAccountName, operationName, clientRequestId);
+        public Task MakeBlobStorageRequest(Func<OperationContext, CancellationToken, Task> storageRequest, string operationName, string? clientRequestId = null, bool throttle = true) =>
```
cgillum, line 118: Non-blocking suggestion: I think it's more common in .NET API design to name a variable like this as `force` with a default value of `false`.

cgillum (approved):
Thanks for contributing this fix!

saguiitay:
Can this be completed and released, so that I can test it?

sebastianburckhardt:
Ideally, you could help us test these changes before they go in the next release, it would give us confidence we are not breaking anything. Is this possible? @cgillum, what do you think?

I also have two more modifications to the partition manager I wanted to add and test.

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -138,12 +141,15 @@ private async Task<T> MakeStorageRequest<T>(Func<OperationContext, CancellationT
...
+                if (!force)
+                {
+                    requestThrottleSemaphore.Release();
+                }
             }
         }
 
-        private Task MakeStorageRequest(Func<OperationContext, CancellationToken, Task> storageRequest, string accountName, string operationName, string? clientRequestId = null) =>
-            this.MakeStorageRequest<object?>((context, cancellationToken) => WrapFunctionWithReturnType(storageRequest, context, cancellationToken), accountName, operationName, clientRequestId);
+        private Task MakeStorageRequest(Func<OperationContext, CancellationToken, Task> storageRequest, string accountName, string operationName, string? clientRequestId = null, bool throttle = true) =>
```
cgillum, line 151: Can we rename this parameter too for consistency?

cgillum:
I would definitely prefer having some validation before doing a release. @saguiitay are you able to test a private version of this deployed to myget.org?

sebastianburckhardt:
I have added a commit that changes the renew loop to make its timing more consistent:
- now uses a dedicated thread instead of an async task to avoid observed issues with high variability of `await Time.Delay(x)`
- now uses a stopwatch to time successive iterations of the renew loop
- has a trace statement so we can observe timing anomalies, such as excessive delays between iterations
- uses async local tasks and a single task collection, instead of Task.ContinueWith and multiple collections, to make the code easier to read

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -171,102 +171,115 @@ public async Task TryReleasePartitionAsync(string partitionId, string leaseToken
...
+                        {
+                            leaseTasks.Add(RenewOwnedLease(lease));
+                        }
+                        foreach (T shutdownLease in this.keepRenewingDuringClose.Values.ToList())
                         {
-                            renewTasks.Add(this.RenewLeaseAsync(lease).ContinueWith(renewResult =>
+                            leaseTasks.Add(RenewShutdownLease(shutdownLease));
+                        }
+
+                        Task.WhenAll(leaseTasks).Wait();
```
cgillum, line 217: If there's an exception, the `.Wait()` method throws an `AggregateException` instead of throwing the original exception. I worry about whether this might cause unexpected regressions or make debugging more challenging. If you want to preserve the previous exception behavior, you should instead use `.GetAwaiter().GetResult()`.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -171,102 +171,115 @@ public async Task TryReleasePartitionAsync(string partitionId, string leaseToken
...
+                        {
+                            this.leaseRenewerCancellationTokenSource.Token.WaitHandle.WaitOne(delayBy);
+                        }
+
+                        async Task RenewOwnedLease(T lease)
+                        {
+                            if (this.shouldRenewLeaseDelegate(lease.PartitionId))
                             {
-                                if (!renewResult.Result)
+                                if (!await this.RenewLeaseAsync(lease))
```
cgillum, line 230: In your comments you specifically mentioned improving the reliability of resuming from delays by using a dedicated thread. However, we will temporarily jump off the dedicated thread and use worker pool threads here and everywhere else that we fall back to async/await. Is that going to be acceptable in terms of the goals of this latest reliability iteration?

cgillum (commented):
Just a couple comments on this latest iteration.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
sebastianburckhardt: Good catch. Indeed, we want to preserve the previous exceptions.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
sebastianburckhardt: I understand that we may still depend on the worker thread pool here, which is not ideal. But I was reluctant to make the code more complicated given that I am not completely sure in the first place that thread pool starvation is really the reason of why Task.Delay has such a high variance.

cgillum: That's fine. I don't think we need to block on this and I agree that it would increase complexity so long as you still feel comfortable with the value of this approach in general.

sebastianburckhardt: This may not be the final iteration of tuning the lease renewal loop. But we have to start somewhere and at the moment the code looks reasonably simple to me.

saguiitay:
I would suggest doing the refactoring in a separate PR. This looks like an unncessary risk. Please consider first fixing the bug, and then improving the code (for readability/stability/performance).
That being said, I'll see if I get run it a few times to test...

sebastianburckhardt:
Given the unfavorable risk/benefit ratio I have reverted the last commit (refactoring of the renewal loop) as you suggested, @saguiitay.

I will pursue this in a separate PR at a later point of time.

saguiitay:
@sebastianburckhardt I wasn't able to fully test the change, but I did some testing. I manually deployed the DLLs with the change to a 3-nodes cluster:

* I then ran MANY orchestrations on it, which all worked as expected
* I then stopped one of the nodes, "forcing" the 2 remaining nodes to take leases for the leases. 
* I then ran MANY orchestrations on it, which all worked as expected
* I then started the stopped node back to life - "forcing" it to steal the leases from other nodes
* I then ran MANY orchestrations on it, which all worked as expected

So it does looks like things are working, but it's not a 100% test coverage.

@cgillum if you can release an updated version, I run it through our pipelines, which usually encounter the issue after 4-5 runs (usually take less than 2 days to appear)

cgillum:
Thanks @saguiitay for helping with this validation! This gives me pretty good confidence that we can make a release to nuget.org with this fix. I'll try to do this in the next couple days.

PR #723, opened 2022-05-04
Improve logging for lease management

cgillum:
I was doing some manual testing of the lease management code to try and diagnose some issues we've been facing recently and found a few places where we could use some improvements. Here are the changes in this PR:

* Add "lease type" fields to all the lease management logs for easier analysis
* Fixes an `ObjectDisposedException` when restarting the task hub worker (reported by an internal DTFx user).
* A small number of minor code cleanup changes

I found some more fundamental lease management issues as well, but I'll try to tackle those in a separate PR.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -228,6 +228,11 @@ namespace DurableTask.AzureStorage
                 throw new ArgumentOutOfRangeException(nameof(settings), "The number of partitions must be a positive integer and no greater than 16.");
             }
 
+            if (string.IsNullOrEmpty(settings.TaskHubName))
+            {
+                throw new ArgumentNullException(nameof(settings), $"A {nameof(settings.TaskHubName)} value must be configured in the settings.");
+            }
+
             // TODO: More validation.
         }
 
--- a/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
@@ -25,6 +25,8 @@ namespace DurableTask.AzureStorage.Partitioning
     /// </summary>
     sealed class AppLeaseManager
     {
+        const string LeaseType = "app";
+
         readonly AzureStorageClient azureStorageClient;
         readonly IPartitionManager partitionManager;
         readonly AzureStorageOrchestrationServiceSettings settings;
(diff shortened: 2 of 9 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -228,6 +228,11 @@ static void ValidateSettings(AzureStorageOrchestrationServiceSettings settings)
                 throw new ArgumentOutOfRangeException(nameof(settings), "The number of partitions must be a positive integer and no greater than 16.");
             }
 
+            if (string.IsNullOrEmpty(settings.TaskHubName))
```
sebastianburckhardt, line 231: Seems thematically not related to this PR. Is there a good reason to include it?

Is there some risk of breaking old versions? I think I remember Connor going through lots of iterations about how/when to set default taskhub name for different versions.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
```diff
@@ -168,6 +170,7 @@ public async Task StopAsync()
 
             this.starterTokenSource.Cancel();
             this.starterTokenSource.Dispose();
+            this.starterTokenSource = null;
```
sebastianburckhardt, line 173: Are we absolutely sure that we don't get StopAsync() called twice? In that case the preceding lines could throw a NullReferenceException.

sebastianburckhardt (commented):
some minor questions, other than that this looks good to me.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: It's not related - just something I ran into when setting up my manual test. I don't anticipate this breaking anything because we always set this property from Durable Functions already. Any app that wasn't setting this explicitly already would have failed with a `NullReferenceException` during startup.

While not required in this PR, it's one of those tiny improvements that we'd otherwise never do. I try to be careful about these in general.

src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs
cgillum: Good point. Let me add a null check to this.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: I'm fairly confident as well that the TaskHub name is already set by default in DF :) . I know this because the default taskhub name was recently the source of some challenges in another GH issue

davidmrdavid (approved):
LGTM, although I'm not sure where the object-disposed exception was addressed.

cgillum:
@davidmrdavid thanks! The ObjectDisposedException was in AppLeaseManager, where we try to reuse an old `starterTokenSource`, which had been previously disposed. My change was to assign it to `null` as part of shutdown so that we'd reinitialize it on the subsequent startup.

PR #725, opened 2022-05-05
Fail when orchestrator does not complete despite having no "open" tasks

davidmrdavid:
Possibly related to: https://github.com/Azure/azure-functions-durable-extension/issues/2133, https://github.com/Azure/azure-functions-durable-extension/issues/2135

Recently, we've received a few Durable Functions reports of orchestrators getting mysteriously "stuck" despite completing all their scheduled Tasks. The telemetry generated by these incidents match those of an "illegal await" at the end of the user code execution: when an orchestrator awaits a non-DurableTask API before returning.

At this time, we have not found a root cause for these incidents, but we can still improve our core logic to transition the orchestrator to a `Failed` state when we detect an orchestrator will get stuck. This PR adds logic to do just that.

I did not add a test as I'm not familiar enough with this part of the stack to know exactly how to test with ease *and* because this is a small edge-case change. I'm happy to add one if we think this small change warrants it, but I'll need pointers.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationExecutor.cs
+++ b/src/DurableTask.Core/TaskOrchestrationExecutor.cs
@@ -151,9 +151,13 @@ namespace DurableTask.Core
                                 this.context.CompleteOrchestration(this.result.Result);
                             }
                         }
-
-                        // TODO: It is an error if result is not completed when all OpenTasks are done.
-                        // Throw an exception in that case.
+                        else
+                        {
+                            // It is an error if result is not completed when all OpenTasks are done.
+                            // Throw an exception in that case.
+                            var exception = new Exception("Possible constraint violation: No 'open' tasks remain but the orchestrator is incomplete. Please verify that only DurableTask APIs are awaited.");
+                            this.context.FailOrchestration(exception);
+                        }
                     }
                 }
                 catch (NonDeterministicOrchestrationException exception)


src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -151,9 +151,13 @@ void ProcessEvents(IEnumerable<HistoryEvent> events)
...
                         }
-
-                        // TODO: It is an error if result is not completed when all OpenTasks are done.
-                        // Throw an exception in that case.
+                        else
+                        {
+                            // It is an error if result is not completed when all OpenTasks are done.
+                            // Throw an exception in that case.
+                            var exception = new Exception("Possible constraint violation: No 'open' tasks remain but the orchestrator is incomplete. Please verify that only DurableTask APIs are awaited.");
+                            this.context.FailOrchestration(exception);
```
cgillum, line 159: In spite of what the comment says, I'm pretty confident that this is not actually an "error". For example, if an orchestration is waiting for some external event before it completes, you will reach this else block. Such a case is perfectly valid and failing the orchestration would be obviously wrong.

We were able to address this in the Java out-of-process SDK by keeping track of any external events that the orchestrator declared it was waiting for. However, unlike Durable Functions, DTFx Core doesn't actually have a concept of "external event subscriptions", thus we have no way of knowing whether an orchestration is legitimately waiting for some external event before it allows itself to complete.

cgillum (changes requested):

src/DurableTask.Core/TaskOrchestrationExecutor.cs
davidmrdavid: Good catch, thank you. **Hmm, this complicates things.** After looking at the DTFx and DF code, it doesn't seem we currently have a good way of giving DTFx visiblity into the WaitForExternalEvent API without implementing the API itself in the DTFx context object.

As part of our discussion to implemented a "Task Layer" API: how easy do you think it would be to move the implementation of external events into DTFx? It seems to me that it should be fairly straightforward.

I ask, **in large part,** because the related incidents mentioned above would really benefit from having this PR's proposed functionality. So, if giving DTFx an implementation of External Events is not too hard, then I would be willing to do that as a precursor to this PR's functionality.

cgillum: It's not hard to add an DF-like external events API to DTFx. But I don't see how that solves the problem since most existing DTFx apps probably won't use it. Any change we make here has to be non-breaking for existing apps. I'm not convinced that we've actually explored all options for detecting illegal async usage in DF - I'd like to see proposals that would be less risky than this one.

davidmrdavid: Just for clarity, since I'm unsure if this was communicated well - my reason to propose adding external events to DTFx is primarily so DTFx can correctly track open tasks when DF uses external events, which in turn enables us to error out on illegal async.

That aside, yeah I would also prefer an alternative design. I've looked for solutions at the DF-level, but so far I've hit a wall. I'll look to explore some more, but I'll probably be requesting assistance :-).

I'll look to close this PR and to track this work item in a separate issue within the DF repo.

davidmrdavid: Just FYI - I have a WIP change in DF that at least logs a warning on illegal awaits - https://github.com/Azure/azure-functions-durable-extension/pull/2168

jviau:
As @cgillum mentioned, this may be extremely hard to get correct. But this is a problem worth tackling - I have been bitten by it a couple times now. I was thinking maybe some DTFx analyzers would be a good approach? We could warn / error on incorrect async/await usage within DTFx, and then add some attributes or something for users to explicitly mark some APIs as safe to await (maybe they wrap the DTFx task with some post-work).

davidmrdavid:
I was under the impression the analyzers would already catch this, but I don't know that for sure; I'll leave that to @amdeel to confirm. 

And yes, I agree on the importance of this. I currently have some internal reports of this happening to folks who, from the looks of it, aren't explicitly awaiting any non-DF APIs (maybe unidentified part of the DF code invocation pipeline might be to blame?) so this is becoming more and more important on my radar.

jviau:
Ah I see there is an analyzer package for WebJobs / DurableFunctions. Does this package work for base DTFx?

davidmrdavid:
Ah, my mistake for misreading your comment as relating to DF instead of DTFx. @amdeel, does the DF analyzer also work with DTFx?

jviau:
I was thinking about the external input scenario: what if we do introduce an explicit scenario similar to this PR, but make the following changes:

1. Introduce our own external input task, which is basically just a custom `TaskCompletionSource`. The idea here is that it is _some_ task that DTFx would be aware of and consider it an "open" task. Would need to figure out the specifics for how that is done.
2. This would be a non-breaking **opt-in** behavior. Maybe a default opt-in value for the worker as a whole, and then individual orchestrations could also opt in or out (so users can migrate to this behavior over time).

davidmrdavid:
For clarity: @jviau, is your reason to introduce external-input support in DTFx is so that (1) Durable Functions can utilize this DTFx feature during external events, which then leads to (2) the ability to fail an orchestrator during open tasks as originally intended in this PR, correct?

jviau:
Yes exactly. It would be an optional more strict way to perform external event waiting, where the consumers need to explicitly declare they are waiting on external events (this API needs to be designed obviously). Then we will at least know _not_ to fail the orchestration when there are pending external events. However, we will have no way of ensuring the correctness of those events.

davidmrdavid:
Personally, I like that idea. 

At a glance, it does seem a bit roundabout to have to introduce a new API in DTFx in order to enable this, but we have other internal proposals which would also require bringing DF features down to the DTFx layer.

@jviau: would you mind opening an issue describing this request? That would help us prioritize it, and it would help keep the discussion in this PR (open tasks and failing orchestrator) separate from the design of a DTFx-level external-event API

davidmrdavid:
Hey @jviau - I'm interested in following up on this work item. How difficult do you think it would be to implement your suggestion here: https://github.com/Azure/durabletask/pull/725#issuecomment-1136273482 ? If it's low hanging fruit, then I think this would provide a lot of value in preventing stuck orchestrations.

jviau:
@davidmrdavid it overall isn't much work from a coding perspective (at least I think so). But there is a lot of discussion to be had on if this is a good approach to start with. There might be better approaches to consider, like how durable functions doesn't use the on-event callback, but instead you subscribe to events via the context object. Is that better? Or even if it isn't better, do we even want DTFx Core to go down a fairly different path?

torosent:

Issue #722, opened 2022-05-03
Split AzureServiceFabric into multiple projects


dsempi:
DurableTask.AzureServiceFabric contains multiple logical units in a single project:

- The core implementation of IOrchestrationServiceClient and IOrchestrationService on top of Service Fabric reliable collections.
- The Service Fabric stateful service exposing the orchestration service over Web API
- The remote client implementing IOrchestrationServiceClient communicating with the Web API

This brings some difficulties, the main ones are:

1. The current implementation detail of the Service Fabric stateful service requires .NET Framework, which prohibits using the core part or the remote client in .NET 5/6 or .NET Core environments.
2. The consumer of the remote client gets a lot of dependencies which are useless are from the remote client prospective.

Proposed split:

- DurableTask.AzureServiceFabric.Core - contains the core implementation
- DurableTask.AzureServiceFabric.Service - contains Service Service stateful service
- DurableTask.AzureServiceFabric.Client - contains the remote client
- DurableTask.AzureServiceFabric - empty project for the backward compatibility referencing all above mentioned projects (NuGets)

cgillum:
@shankarsama I believe this one belongs to your or someone on your team?

dsempi:
If agreed, I can prepare a pull request with the proposed changes ...

davidmrdavid:
PRs are always appreciated, @dsempi. If you have an implementation, I think we can help get the right reviewers (probably someone in @shankarsama's team) to take a look at it :)

dsempi:
Here it is #726

lilyjma:
@shankarsama -- could you take a look at the PR please.

PR #726, opened 2022-05-05
Split DurableTask.AzureServiceFabric

dsempi:
Split DurableTask.AzureServiceFabric into 3 projects:
- DurableTask.AzureServiceFabric.Core: The core implementation of IOrchestrationServiceClient and IOrchestrationService on top of Service Fabric reliable collections.
- DurableTask.AzureServiceFabric.Service: The Service Fabric stateful service exposing the orchestration service over Web API.
- DurableTask.AzureServiceFabric.Client: The remote client implementing IOrchestrationServiceClient communicating with the Web API.

Leave DurableTask.AzureServiceFabric for backward compatibility (aggregates above mentioned packages).


dsempi:
Hi @davidmrdavid, can we please get someone else for the review as @shankarsama seems not to be active anymore :-(? Thanks!

davidmrdavid:
Hi @dsempi, let me try to reach out internally, I believe @shankarsama should be available but I don't work too directly with him.

FYI @cgillum

shankarsama:
@dsempi What's the motivation for this change? Please provide detailed explanation with use cases or examples where this split is beneficial.

dsempi:
Hi @shankarsama, I summarized the reasons in the  Issue #722. The current architecture goes against the Minimization security principle (to minimize the attack surface). My main motivation was to get core implementation without the dependencies on ASP.NET as I want to use durable framework within a Service Fabric service but without exposing it via Web APIs. Since I don't want to use this functionality, I also don't want to include unnecessary dependencies. Similarly, if someone wants to communicate with the DurableTask Service Fabric service, he has to include a lot of unnecessary dependencies.

dsempi:
@davidmrdavid, @shankarsama, it has been laying here for month ... I'd understand, if you had objections against the implementation, but it's not nice pretending it does not exists.

davidmrdavid:
Hi @dsempi. Apologies for the delay. I can't personally review this PR as I'm not an owner of the ServiceFabric component of DTFx. Let me try to reach @shankarsama about this PR

shankarsama:
@dsempi 
Sorry for the delay.
We can make the Owin listener (or Http endpoint) optional so that the TaskHubWorker can be started without the proxy listener. There is no need to split it. Its the Service Fabric provider, we need to ship it with Service Fabric stateful service implementation to reduce the developer effort. So core implementation comes with Service Fabric's stateful service. Only extra addition is the proxy. We can make it optional

Check the example here, we can make changes or add new API to skip ProxyListener during startup: https://github.com/Azure/durabletask/blob/main/test/TestFabricApplication/TestApplication.StatefulService/Program.cs#:~:text=var%20listener%20%3D%20new%20TaskHubProxyListener(testProvider.GetFabricOrchestrationProviderSettings()%2C%20testProvider.RegisterOrchestrations)%3B

Another thing is, you are adding new provider guids for EventSource, this will be a breaking change as the developers already started their trace listeners with the old guids. This going to be a difficult production rollout.

Let me know what you think.

dsempi:
Hi @shankarsama,
thanks for looking into it. Making Owin listener optional parameter would not solve the main issues:
1. Usage of the NuGet package is bound to .NET Framework as Owin is supported in .NET Framework only.
2. The package requires dependencies (AspNetCore, Owin) which are not needed from the perspective of storing DurableTask state in the Service Fabric reliable store.

The first issue could be workarounded, if you approved #806 which uses Kestrel instead of Owin. The second problem still remains. The only purpose of the TaskHubStatefulService is to enable the DurableTask controller in the Service Fabric environment. Without the listener it would not have any purpose. If we take it further, the controller could actually be used with any DurableTask store. This clearly shows the things are not separated correctly.

If EventSource ID was a problem, we could define it only in Core and make it public (rather nasty hack) or rewrite Core and Services to use ILogger instead and register LoggerProvider for the EventSource.

Thanks. What do you think?

Issue #392, opened 2020-05-14
Easy way to set DataConverter for entire orchestration?


NickMoores:
Hello!

I was using a struct as an input and return value throughout one of my orchestrations, and I spotted problems with the value transferring between the orchestrations and the tasks. I realised the struct probably wasn't playing ball with JSON.NET, so found a constructor on TaskHubClient that accepted a `DataConverter` that I could pass a `JsonDataConverter` to with custom serialization settings, and custom `JsonConverter`s - hurrah!

I was surprised that this only affected the conversion of the initial inputs to the orchestration and not the conversion between the orchestration, suborchestrations, and tasks.

I found another `DataConverter` on the `OrchestrationContext`, and another on the `TaskOrchestration` itself, and one on the `AsyncTaskActivity` class - that's quite a few, and I'm sure there are benefits to them all using their own `DataConverters`, but is there an easy way to specify a default `DataConverter` that gets used across an entire orchestration instance (and maybe all orchestration instances)? I might very well be missing an easy way to do this, I confess!

Thanks!

ConnorMcMahon:
@amdeel, this may be an opportunity for us to make this code path a bit simpler, as I know this is something we struggled with when making our customer serialization for Durable Functions work in each place that it was supposed to.

PR #728, opened 2022-05-06
 DataConverters injection

rozhnovl:
A custom implementation of DataConverters can be injected into framework to handle serialization/deserialization of Orchestrations and Activities parameters
issue ref: #392

### DIFF
--- a/src/DurableTask.Core/TaskActivity.cs
+++ b/src/DurableTask.Core/TaskActivity.cs
@@ -25,8 +25,18 @@ namespace DurableTask.Core
     ///     User activity should almost always derive from either TypedTaskActivity
     ///     &lt;TInput, TResult&gt; or TaskActivity&lt;TInput, TResult&gt;
     /// </summary>
-    public abstract class TaskActivity
+    public abstract class TaskActivity : IDataConverterConfigurable
     {
+        /// <summary>
+        /// The DataConverter to use for input and output serialization/deserialization
+        /// </summary>
+        public DataConverter DataConverter { get; internal set; } = new JsonDataConverter();
+
+        DataConverter IDataConverterConfigurable.DataConverter
+        {
+            set => DataConverter = value;
+        }
+
         /// <summary>
         /// Abstract method for executing a task activity synchronously
         /// </summary>
@@ -54,28 +64,6 @@ namespace DurableTask.Core
     /// <typeparam name="TResult">Output type of the activity</typeparam>
     public abstract class AsyncTaskActivity<TInput, TResult> : TaskActivity
     {
-        /// <summary>
-        /// Creates a new AsyncTaskActivity with the default DataConverter
-        /// </summary>
-        protected AsyncTaskActivity()
-        {
-            DataConverter = new JsonDataConverter();
-        }
-
-        /// <summary>
-        /// Creates a new AsyncTaskActivity with the supplied DataConverter
-        /// </summary>
-        /// <param name="dataConverter"></param>
-        protected AsyncTaskActivity(DataConverter dataConverter)
-        {
-            DataConverter = dataConverter ?? new JsonDataConverter();
-        }
-
-        /// <summary>
-        /// The DataConverter to use for input and output serialization/deserialization
-        /// </summary>
-        public DataConverter DataConverter { get; protected set; }
-
         /// <summary>
         /// Synchronous execute method, blocked for AsyncTaskActivity
         /// </summary>
--- a/src/DurableTask.Core/TaskActivityDispatcher.cs
+++ b/src/DurableTask.Core/TaskActivityDispatcher.cs
@@ -22,6 +22,7 @@ namespace DurableTask.Core
     using DurableTask.Core.History;
     using DurableTask.Core.Logging;
     using DurableTask.Core.Middleware;
+    using DurableTask.Core.Serializing;
     using DurableTask.Core.Tracing;
 
     /// <summary>
@@ -35,19 +36,22 @@ namespace DurableTask.Core
         readonly DispatchMiddlewarePipeline dispatchPipeline;
         readonly LogHelper logHelper;
         readonly ErrorPropagationMode errorPropagationMode;
+        readonly DataConverter dataConverter;
 
         internal TaskActivityDispatcher(
             IOrchestrationService orchestrationService,
             INameVersionObjectManager<TaskActivity> objectManager,
             DispatchMiddlewarePipeline dispatchPipeline,
             LogHelper logHelper,
-            ErrorPropagationMode errorPropagationMode)
+            ErrorPropagationMode errorPropagationMode,
+            DataConverter dataConverter)
         {
             this.orchestrationService = orchestrationService ?? throw new ArgumentNullException(nameof(orchestrationService));
             this.objectManager = objectManager ?? throw new ArgumentNullException(nameof(objectManager));
             this.dispatchPipeline = dispatchPipeline ?? throw new ArgumentNullException(nameof(dispatchPipeline));
             this.logHelper = logHelper;
             this.errorPropagationMode = errorPropagationMode;
+            this.dataConverter = dataConverter;
 
             this.dispatcher = new WorkItemDispatcher<TaskActivityWorkItem>(
                 "TaskActivityDispatcher",
@@ -137,7 +141,7 @@ namespace DurableTask.Core
                 }
 
                 this.logHelper.TaskActivityStarting(orchestrationInstance, scheduledEvent);
-                TaskActivity? taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version);
+                TaskActivity? taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version, this.dataConverter);
 
                 if (workItem.LockedUntilUtc < DateTime.MaxValue)
                 {
--- a/src/DurableTask.Core/TaskHubClient.cs
+++ b/src/DurableTask.Core/TaskHubClient.cs
@@ -51,7 +51,7 @@ namespace DurableTask.Core
         /// </summary>
         /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
         /// <param name="dataConverter">The <see cref="JsonDataConverter"/> to use for message serialization.</param>
-        public TaskHubClient(IOrchestrationServiceClient serviceClient, JsonDataConverter dataConverter)
+        public TaskHubClient(IOrchestrationServiceClient serviceClient, DataConverter dataConverter)
             : this(serviceClient, dataConverter, null)
         {
         }
--- a/src/DurableTask.Core/TaskHubWorker.cs
+++ b/src/DurableTask.Core/TaskHubWorker.cs
@@ -21,6 +21,7 @@ namespace DurableTask.Core
     using DurableTask.Core.Exceptions;
     using DurableTask.Core.Logging;
     using DurableTask.Core.Middleware;
+    using DurableTask.Core.Serializing;
     using Microsoft.Extensions.Logging;
 
     /// <summary>
@@ -31,6 +32,7 @@ namespace DurableTask.Core
     {
         readonly INameVersionObjectManager<TaskActivity> activityManager;
         readonly INameVersionObjectManager<TaskOrchestration> orchestrationManager;
+        readonly DataConverter dataConverter;
 
         readonly DispatchMiddlewarePipeline orchestrationDispatchPipeline = new DispatchMiddlewarePipeline();
         readonly DispatchMiddlewarePipeline activityDispatchPipeline = new DispatchMiddlewarePipeline();
@@ -67,12 +69,14 @@ namespace DurableTask.Core
         /// </summary>
         /// <param name="orchestrationService">Reference the orchestration service implementation</param>
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging</param>
-        public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null)
+        /// <param name="dataConverter">Data converter used for task serialization. By default - Newtonsoft.Json based implementation is used.</param>
+        public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null, DataConverter dataConverter = null)
             : this(
                   orchestrationService,
                   new NameVersionObjectManager<TaskOrchestration>(),
                   new NameVersionObjectManager<TaskActivity>(),
-                  loggerFactory)
+                  loggerFactory,
+                  dataConverter)
         {
         }
 
@@ -94,7 +98,6 @@ namespace DurableTask.Core
         {
         }
 
-
         /// <summary>
         ///     Create a new <see cref="TaskHubWorker"/> with given <see cref="IOrchestrationService"/> and name version managers
         /// </summary>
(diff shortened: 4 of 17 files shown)


jviau:
It is already possible to customize the data converter today via middleware. If this does not cover your needs, can you help me understand what gaps it has? If you want examples on how to do this, let me know.

Just to be clear: I am not against improving the data converter experience, but I think we may want to minimize changes for now and have a bigger rethink on serialization as a whole. Right now DTFx makes an assumption of JSON, and specifically Newtonsoft.Json. I think DTFx should abstract away the serialization implementation and just deal with streams, letting the user of DTFx determine their serialization approach (JSON, message pack, binary, whatever they want). DTFx can of course offer some packages to implement to most common forms of serialization.

rozhnovl:
If such an option already exists - sure, I'd be glad to just use it. But from the source code - I didn't quite get it, and in some places - it's pretty hardcoded without possibility to override (OrchestrationContext). If you have some example how it can be implemented in our code alternatively - can you share it please
I can probably give some info about our case, why we do need such capability in our service.
The main issue we have - we're utilizing some third-party models in our code (Synapse SDK) to perform background tasks (deploying Azure resources, including Synapse pipelines). This Synapse SDK relies heavily on System.Text.Json serialization and has some hand-written data converters to control how data is serialized/deserialized. These models are not serializeable by Newtonsoft out of box correctly. What's worse - these models are coming to our service API so we cannot control initial serialization. We could write some custom Newtonsoft serializers, but that's quite a lot of work.
Just passing these models as TaskActivity arguments doesn't work for us, so we had to make our tasks accept string parameters and doing serialization/deserialization manually. I believe that's not the right way to do so.
Some of serialization/deserialization now can be overriden for sure, but there also comes a question that there is a plenty of places it needs to be set. In this PR I wanted to make sure there is only one (two as we have TaskHubClient and Worker roles) entry point requiring configuration.

If there is some global refactoring coming regarding it - can you tell some ETA?

jviau:
> If there is some global refactoring coming regarding it - can you tell some ETA?

Nothing planned that I know of. This is just my suggestion for DTFx. I am not a maintainer so one of the maintainers can always disagree with me and accept your PR.

To do this with existing functionality, it takes a combination of a couple base classes and interfaces to access the protected setters on `TaskActivity` and `TaskOrchestration` and then middleware to set the data converter for all your orchestrations.

``` CSharp
public interface IActivityBase
{
    DataConverter DataConverter { get; set; }
}

public abstract class ActivityBase<TInput, TResult> : TaskActivity<TInput, TResult>
{
    DataConverter IActivityBase.DataConverter // exposes the data converter for the middleware
    {
        get => DataConverter;
        set => DataConverter = value;
    }
}

public interface IOrchestrationBase
{
    DataConverter DataConverter { get; set; }
}

public abstract class OrchestrationBase<TInput, TResult> : TaskOrchestration<TInput, TResult>
{
    DataConverter IOrchestrationBase.DataConverter // exposes the data converter for the middleware
    {
        get => DataConverter;
        set => DataConverter = value;
    }

    public override Task<TResult> RunTask(OrchestrationContext context, TInput input)
    {
        // This step will take our data converter and set it back on the orchestration context provided, so all serialization off of it works as we desire.
        if (DataConverter is JsonDataConverter converter)
        {
            // OrchestrationContext has this typed as JsonDataConverter for some reason, but you can probably ignore Newtonsoft.Json details in it and
            // override it all yourself. Or you can write a shim JsonDataConverter which takes your DataConverter in and just forwards the calls to it.
            context.MessageDataConverter = converter;
            context.ErrorDataConverter = converter;
        }

        return RunAsync(context, input);
    }

    protected abstract Task<TResult> RunAsync(OrchestrationContext context, TInput input);
}

// set up the middleware:
TaskHubWorker worker = ...;
worker.AddOrchestrationDispatcherMiddleware((context, next) =>
{
    if (context.GetProperty<TaskOrchestration>() is IOrchestrationBase baseOrchestration)
    {
        baseOrchestration.DataConverter = // set your converter here
    }

    return next();
});

worker.AddActivityDispatcherMiddleware((context, next) =>
{
    if (context.GetProperty<TaskActivity>() is IActivityBase baseActivity)
    {
        baseActivity.DataConverter = // set your converter here
    }

    return next();
});
```

rozhnovl:
@shankarsama, @cgillum do you mind giving some feedback on that PR?
We'd like to clean up some temporary workarounds in our service, which are caused by the issue this PR is trying to cover.
If you don't like the implementation and think it might be done in a more appropriate way - let's discuss how it should be implemented and I don't mind doing it another way.

cgillum:
The main problem that I see with this PR is that it's a breaking change - specifically that changes to existing public interfaces.  Given that there are workarounds as @jviau suggested, I'd prefer not to take this kind of change to the base framework. I'd be willing to consider a version of this if it can be done in a non-breaking, opt-in way.

rozhnovl:
That's kinda the idea I pursued, the default behavior remains the same, if no custom DataConverter is passed - it initializes with the same Newtonsoft.Json implementation as before.
Also I believe I didn't change any "publically used" classes contracts, only ones which are used internally in the library. If I missed something and altered some public contract - please point to it and I'll fix it in a compatible manner.

jviau:
There are several public API breaking changes in this PR. If the class is `public` and you added a parameter to an existing method or constructor, that is a breaking change. Even with default params, you can cause `MissingMethodException`s. Adjusting public interfaces also is largely a breaking change (adding methods, adding params to methods). This does not matter if it is "publicly used" or not. If the type is public, its public and under breaking change restrictions.

Additionally, I strongly advise against setting DataConverter for `IOrchestrationService`. This will not affect your orchestration/activity serialization - that is already a plain `string` when it gets to them. This risks breaking that component if they have any custom serialization (which you remove by overriding their serializer), or the serializer you provide is not compatible with their infrastructure. 

I have started working on an [extensions lib for DurableTask](https://github.com/jviau/durabletask-hosting/tree/main/src/DurableTask.Extensions). I have not published it yet, but you can reference it for how I set the data converter. This can be seen in the middleware.

src/DurableTask.Core/TaskActivity.cs
```diff
@@ -25,8 +25,18 @@ namespace DurableTask.Core
     ///     User activity should almost always derive from either TypedTaskActivity
     ///     &lt;TInput, TResult&gt; or TaskActivity&lt;TInput, TResult&gt;
     /// </summary>
-    public abstract class TaskActivity
+    public abstract class TaskActivity : IDataConverterConfigurable
     {
+        /// <summary>
+        /// The DataConverter to use for input and output serialization/deserialization
+        /// </summary>
+        public DataConverter DataConverter { get; internal set; } = JsonDataConverter.Default;
```
jviau, line 33: I would prefer we don't add this here and instead add this contract to the types which already had `DataConverter` on it.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -51,7 +51,7 @@ public TaskHubClient(IOrchestrationServiceClient serviceClient)
         /// </summary>
         /// <param name="serviceClient">Object implementing the <see cref="IOrchestrationServiceClient"/> interface </param>
         /// <param name="dataConverter">The <see cref="JsonDataConverter"/> to use for message serialization.</param>
-        public TaskHubClient(IOrchestrationServiceClient serviceClient, JsonDataConverter dataConverter)
+        public TaskHubClient(IOrchestrationServiceClient serviceClient, DataConverter dataConverter)
```
jviau, line 54: This is a breaking change as well. You can add a separate constructor:

``` CSharp
public TaskHubClient(IOrchestrationServiceClient serviceClient, JsonDataConverter dataConverter)
    : this(serviceClient, (DataConverter)dataConverter) // downcast to call the new ctor.
{
}

public TaskHubClient(IOrchestrationServiceClient serviceClient, DataConverter dataConverter)
{
    // add this one separately.
}
```

src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -67,12 +69,14 @@ public TaskHubWorker(IOrchestrationService orchestrationService)
         /// </summary>
         /// <param name="orchestrationService">Reference the orchestration service implementation</param>
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging</param>
-        public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null)
+        /// <param name="dataConverter">Data converter used for task serialization. By default - Newtonsoft.Json based implementation is used.</param>
+        public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null, DataConverter dataConverter = null)
```
jviau, line 73: Breaking change - will need to add a ctor independently.

``` CSharp
public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null)
{
    // leave unchanged
}

public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory, DataConverter dataConverter)
{
    // add separately. Make sure to NOT include default params here.
}
```

src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -94,23 +98,25 @@ public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory
...
         /// <param name="activityObjectManager">The <see cref="INameVersionObjectManager{TaskActivity}"/> for activities</param>
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging</param>
+        /// <param name="dataConverter">Data converter used for task serialization. By default - Newtonsoft.Json based implementation is used.</param>
         public TaskHubWorker(
             IOrchestrationService orchestrationService,
             INameVersionObjectManager<TaskOrchestration> orchestrationObjectManager,
             INameVersionObjectManager<TaskActivity> activityObjectManager,
-            ILoggerFactory loggerFactory = null)
+            ILoggerFactory loggerFactory = null,
+            DataConverter dataConverter = null)
```
jviau, line 114: Same as above, this needs to be added separately. Also will have to ensure default params are **NOT** used. You can update the existing ctors to call the new one you add.

src/DurableTask.Core/TaskOrchestration.cs
```diff
@@ -25,8 +25,18 @@ namespace DurableTask.Core
     ///     TaskOrchestration&lt;TResult, TInput&gt; or 
     ///     TaskOrchestration&lt;TResult, TInput, TEvent, TStatus&gt;
     /// </summary>
-    public abstract class TaskOrchestration
+    public abstract class TaskOrchestration : IDataConverterConfigurable
```
jviau, line 28: Same as activity - lets only add this to types that already had a data converter.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -34,6 +34,7 @@ namespace DurableTask.Core
     /// </summary>
     public class TaskOrchestrationDispatcher
     {
+        readonly DataConverter DataConverter;
```
jviau, line 37: nit: field should be lower case
```suggestion
        readonly DataConverter dataConverter;
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -563,7 +566,7 @@ async Task<OrchestrationExecutionCursor> ExecuteOrchestrationAsync(Orchestration
             // Get the TaskOrchestration implementation. If it's not found, it either means that the developer never
             // registered it (which is an error, and we'll throw for this further down) or it could be that some custom
             // middleware (e.g. out-of-process execution middleware) is intended to implement the orchestration logic.
-            TaskOrchestration? taskOrchestration = this.objectManager.GetObject(runtimeState.Name, runtimeState.Version!);
+            TaskOrchestration? taskOrchestration = this.objectManager.GetObject(runtimeState.Name, runtimeState.Version!, this.DataConverter);
```
jviau, line 569: We can avoid adding to object manager contract by setting the property ourselves here:

```suggestion
            TaskOrchestration? taskOrchestration = this.objectManager.GetObject(runtimeState.Name, runtimeState.Version!);
            if (taskOrchestration is IDataConverterConfigurable configurable)
            {
                configurable.DataConverter = this.dataConverter;
            }
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -563,7 +566,7 @@ async Task<OrchestrationExecutionCursor> ExecuteOrchestrationAsync(Orchestration
...
 
             var dispatchContext = new DispatchMiddlewareContext();
             dispatchContext.SetProperty(runtimeState.OrchestrationInstance);
```
jviau, line 572: I think it would be a good idea to also set the data converter in this context:

``` CSharp
dispatchContext.SetProperty(this.dataConverter);
```

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -42,17 +43,20 @@ public class TaskOrchestrationExecutor
         /// <param name="orchestrationRuntimeState"></param>
         /// <param name="taskOrchestration"></param>
         /// <param name="eventBehaviourForContinueAsNew"></param>
+        /// <param name="dataConverter">Data converter used for task serialization.</param>
         /// <param name="errorPropagationMode"></param>
         public TaskOrchestrationExecutor(
             OrchestrationRuntimeState orchestrationRuntimeState,
             TaskOrchestration taskOrchestration,
             BehaviorOnContinueAsNew eventBehaviourForContinueAsNew,
+            DataConverter dataConverter,
```
jviau, line 52: This is a breaking change. You will need to add it as a separate ctor.

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -87,6 +87,7 @@ public class ServiceBusOrchestrationService : IOrchestrationService, IOrchestrat
         /// </summary>
         public readonly ServiceBusOrchestrationServiceStats ServiceStats;
 
+        private readonly DataConverter dataConverter;
```
jviau, line 90: As with service fabric, lets not update anything outside of `DurableTask.Core`.

src/DurableTask.ServiceBus/Tracking/AzureTableInstanceStore.cs
```diff
@@ -35,14 +35,16 @@ public class AzureTableInstanceStore : IOrchestrationServiceInstanceStore
...
         readonly AzureTableClient tableClient;
 
         /// <summary>
         /// Creates a new AzureTableInstanceStore using the supplied hub name and table connection string
         /// </summary>
         /// <param name="hubName">The hub name for this instance store</param>
         /// <param name="tableConnectionString">Azure table connection string</param>
-        public AzureTableInstanceStore(string hubName, string tableConnectionString)
+        /// <param name="dataConverter">Data converter used for task serialization.</param>
+        public AzureTableInstanceStore(string hubName, string tableConnectionString, DataConverter dataConverter = null)
```
jviau, line 47: Same here, lets not update anything outside of `DurableTask.Core`.

src/DurableTask.SqlServer/Tracking/SqlServerInstanceStore.cs
```diff
@@ -30,17 +30,20 @@ namespace DurableTask.SqlServer.Tracking
...
+        private readonly DataConverter dataConverter;
         private readonly SqlServerInstanceStoreSettings settings;
 
         /// <summary>
         /// Creates a new SqlServerInstanceStore using the supplied settings
         /// </summary>
         /// <param name="settings">Configuration values for the Instnace Store</param>
-        public SqlServerInstanceStore(SqlServerInstanceStoreSettings settings)
+        /// <param name="dataConverter">Data converter used for task serialization.</param>
+        public SqlServerInstanceStore(SqlServerInstanceStoreSettings settings, DataConverter dataConverter = null)
```
jviau, line 41: Here as well - lets not update this. Especially since it is not used.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -137,7 +141,7 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
                 }
 
                 this.logHelper.TaskActivityStarting(orchestrationInstance, scheduledEvent);
-                TaskActivity? taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version);
+                TaskActivity? taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version, this.dataConverter);
```
jviau, line 144: We can avoid having to change object manager contract by setting the data converter here ourselves:
```suggestion
                TaskActivity? taskActivity = this.objectManager.GetObject(scheduledEvent.Name, scheduledEvent.Version);
                if (taskActivity is IDataConverterConfigurable configurable)
                {
                    configurable.DataConverter = this.dataConverter;
                }
```

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -35,19 +36,22 @@ public sealed class TaskActivityDispatcher
...
+            DataConverter dataConverter)
         {
             this.orchestrationService = orchestrationService ?? throw new ArgumentNullException(nameof(orchestrationService));
             this.objectManager = objectManager ?? throw new ArgumentNullException(nameof(objectManager));
             this.dispatchPipeline = dispatchPipeline ?? throw new ArgumentNullException(nameof(dispatchPipeline));
             this.logHelper = logHelper;
             this.errorPropagationMode = errorPropagationMode;
+            this.dataConverter = dataConverter;
 
             this.dispatcher = new WorkItemDispatcher<TaskActivityWorkItem>(
```
jviau, line 56: Also, please set the `DataConverter` on the `DispatchMiddlewareContext`. Line 155:

``` CSharp
dispatchContext.SetProperty(this.dataConverter);
```

jviau (changes requested):

PR #736, opened 2022-05-23
Add a backend-configurable flag to disable problematic OrchestrationRuntimeState behavior

sebastianburckhardt:
Currently, the TaskOrchestrationDispatcher, when completing a work item, does a tricky thing with the `OrchestrationRuntimeState` field of the work item:

1. replaces the new runtime state with the original runtime state object 
2. calls `CompleteTaskOrchestrationWorkItemAsync`
3. assigns it the new runtime state

This is a problem with the Netherite backend since the latter may immediately resubmit the workitem to continue the orchestration, causing a race condition with number 3 that leads to state corruption.

This PR adds a flag so that backends can turn off this behavior. 

_None of the other backends are affected since the flag is on by default._

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -534,7 +534,11 @@ namespace DurableTask.Core
                 } while (continuedAsNew);
             }
 
-            workItem.OrchestrationRuntimeState = originalOrchestrationRuntimeState;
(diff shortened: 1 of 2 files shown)


cgillum:
@sebastianburckhardt can you clarify what is the "new runtime state" you are referring to?

cgillum (approved):
Interesting. The existing behavior seems completely counterintuitive to me. I can't think of any reason why we'd want to reference the old runtime state like this. The behavior you're enabling seems like the right one to me, but I'm clearly missing some context.

PR #738, opened 2022-05-28
Improve versioning of DT.Core and DT.AzureStorage

cgillum:
This PR does a few primary things:

* Ensures the assembly version is always X.Y.0.0, making it easier to upgrade assembles for bug fixes. This is also more closely aligned with best practices based on learnings from the ASP.NET team.
* Ensures the file version is X.Y.Z.N, where N is a number from the CI or 0 locally, making it easier to distinguish locally built assemblies from CI-built assemblies (and to link them back to a particular CI run).

This is in response to the latest release where we had to upgrade both DT.Core and DT.AzureStorage even though we only made bugfix changes to DT.Core.

I also incremented the minor versions as part of this PR since the assembly versions would be going backwards with this change if we did another patch release.

### DIFF
--- a/DurableTask.sln
+++ b/DurableTask.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29409.12
+# Visual Studio Version 17
+VisualStudioVersion = 17.2.32516.85
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DurableTask.ServiceBus", "src\DurableTask.ServiceBus\DurableTask.ServiceBus.csproj", "{6F5D2EAD-726D-4FE5-A575-AEB96D1CCE37}"
 EndProject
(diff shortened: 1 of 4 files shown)


davidmrdavid:
> Ensures the assembly version is **always** X.Y.0.0, making it easier to upgrade assembles for bug fixes. This is also more closely aligned with best practices based on learnings from the ASP.NET team.

I don't quite follow this bit. Does this mean we would never set the patch version? Can you please clarify this? Thanks!

cgillum:
For the [assembly version](https://docs.microsoft.com/dotnet/standard/assembly/versioning), yes.  For the nuget and file versions, however, we'll continue to set the patch version. This allows us to make patch version updates to DT.Core without needing to also update DT.AzureStorage.

FWIW, we only set the *major* version in the WebJobs extension, as shown [here](https://github.com/Azure/azure-functions-durable-extension/blob/7e8d4895f9efbd3aca18a18c1f4ac1fcb92ce359/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj#L7-L12).

```xml
  <MinorVersion>7</MinorVersion>
  <PatchVersion>2</PatchVersion>
  <Version>$(MajorVersion).$(MinorVersion).$(PatchVersion)</Version>
  <FileVersion>$(MajorVersion).$(MinorVersion).$(PatchVersion)</FileVersion>
  <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
```

This follows best practice advise we were given by former members of the ASP.NET team, who had to deal with related assembly versioning nightmares back in the day, when we started the DF project. This was never applied to the DTFx project since it was being managed by another team at the time. In their case, they recommended only incrementing the major version, as shown above. I'm not quite going that far in this PR, but I wanted to at least exclude the patch version number from the assembly version.

davidmrdavid (approved):
LGTM

cgillum:
Got a green run using these (staged) changes on the Durable Functions extension repo: https://github.com/Azure/azure-functions-durable-extension/pull/2199. Merging...

PR #740, opened 2022-06-01
add FetchParentInstancesOnly to OrchestrationQuery

hsnsalhi:
Hello,
As mentioned in my PR for [durabletask-mssql](https://github.com/microsoft/durabletask-mssql) at this link https://github.com/microsoft/durabletask-mssql/actions/runs/2422744112 
This PR adds the property FetchParentInstancesOnly to the OrchestrationQuery class.
This property will be used in every provider (mssql, AzureStorage...) to enable filtering instances by ParentInstanceId (in order to fetch only parent instances if needed).


cgillum:
It's unfortunately not sufficient for us to accept a PR to just a new property on the `OrchestrationQuery` class. We'll also need to include an implementation for the DurableTask.AzureStorage if we're to accept this, plus integration tests that demonstrate that this works end-to-end.

PR #744, opened 2022-06-15
Support Generic Interface Methods with Interface-Based Activities

elokaac:
Currently, it is not possible to invoke a generic method on an interface created by `context.CreateClient<T>()` or `context.CreateRetryableClient<T>()`.

This change adds this support by:

* Modifying `ScheduleProxy.cs` to retrieve the generic type arguments from the `InvokeMemberBinder` using reflection, and then adding these generic parameters to the arguments provided to the `ReflectionBasedTaskActivity`. Since `System.Type` cannot be easily serialized, this is done by instead passing the type metadata (Assembly + Fully Qualified Name), such that the `Type` object can be reconstructed.

* Modifying `ReflectionBasedTaskActivity` to retrieve the generic type arguments and use those to invoke the implemented activity object.

The unit test added in this changeset shows how this change is expected to be utilized.

Tested scenarios:

* Method is generic and has an array return type where the array element is generic.

```c#
public interface IMyInterface
{
    Task<T[]> GetT<T>();
}
```

* Method is generic but has no return type

```c#
public interface IMyInterface
{
    Task PutT<T>();
}
```

* Method is generic and returns a nested generic type:

```c#
public interface IMyInterface
{
    Task<Dictionary<T, List<U>> PutT<T, U>();
}
```

* Method is generic and has generic input types:

```c#
public interface IMyInterface
{
    Task Insert<T, U>(T arg1, U[] arg2);
}
```

### DIFF
--- a/src/DurableTask.Core/RetryProxy.cs
+++ b/src/DurableTask.Core/RetryProxy.cs
@@ -13,7 +13,6 @@
 
 namespace DurableTask.Core
 {
-    using DurableTask.Core.Common;
     using System;
     using System.Diagnostics;
     using System.Linq;
--- a/src/DurableTask.Core/ScheduleProxy.cs
+++ b/src/DurableTask.Core/ScheduleProxy.cs
@@ -19,6 +19,7 @@ namespace DurableTask.Core
     using System.Reflection;
     using System.Threading.Tasks;
     using Castle.DynamicProxy;
+    using DurableTask.Core.Common;
 
     internal class ScheduleProxy : IInterceptor
     {
@@ -56,14 +57,12 @@ namespace DurableTask.Core
             }
 
             Type[] genericArgumentValues = invocation.GenericArguments ?? Array.Empty<Type>();
-            List<object?> arguments = new(invocation.Arguments);
+            List<object> arguments = new(invocation.Arguments);
 
             foreach (var typeArg in genericArgumentValues)
             {
-                arguments.Add(new TypeMetadata(typeArg.Assembly.FullName!, typeArg.FullName!));
+                arguments.Add(new Utils.TypeMetadata { AssemblyName = typeArg.Assembly.FullName!, FullyQualifiedTypeName = typeArg.FullName });
             }
-
-            object[] args = arguments.ToArray()!;
 
             string normalizedMethodName = NameVersionHelper.GetDefaultName(invocation.Method, this.useFullyQualifiedMethodNames);
 
@@ -72,7 +71,7 @@ namespace DurableTask.Core
                 invocation.ReturnValue = this.context.ScheduleTask<object>(
                     normalizedMethodName,
                     NameVersionHelper.GetDefaultVersion(invocation.Method),
-                    args);
+                    arguments.ToArray());
                 return;
             }
 
@@ -89,7 +88,7 @@ namespace DurableTask.Core
             {
                 normalizedMethodName,
                 NameVersionHelper.GetDefaultVersion(invocation.Method),
-                invocation.Arguments.ToArray(),
+                arguments.ToArray(),
             });
 
             return;
--- a/test/DurableTask.ServiceBus.Tests/SerializationTests.cs
+++ b/test/DurableTask.ServiceBus.Tests/SerializationTests.cs
@@ -873,12 +873,12 @@ namespace DurableTask.ServiceBus.Tests
                 Assert.IsNotNull(a);
                 Assert.AreEqual(2, a[1].Count);
 
-                GenericInterfaceOrchestrationInput b = await client.GetWhenNoParams<GenericInterfaceOrchestrationInput>();
-                Assert.IsNull(b);
-
                 GenericInterfaceOrchestrationInput[] c = await retryableClient.GetWhenMultipleGenericTypes<double, GenericInterfaceOrchestrationInput>(input.Property, input);
                 Assert.IsNotNull(c);
                 Assert.AreEqual(1, c.Length);
+
+                GenericInterfaceOrchestrationInput b = await client.GetWhenNoParams<GenericInterfaceOrchestrationInput>();
+                Assert.IsNull(b);
 
                 return string.Empty;
             }


aueloka:
@cgillum , is there anyone that can help review this?

cgillum:
@aueloka I'm taking a look now.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -512,5 +516,111 @@ public static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int task
...
+        /// <param name="genericArguments">The generic method arguments.</param>
+        internal static Type GetGenericReturnType(MethodInfo methodInfo, Type[] genericArguments)
+        {
+            if (!methodInfo.ReturnType.IsGenericType)
+            {
+                throw new Exception("Return type is not a generic type. Type Name: " + methodInfo.ReturnType.FullName);
+            }
+
+            var genericArgument = methodInfo.ReturnType.GetGenericArguments().SingleOrDefault() ??
+                throw new Exception("Generic Parameters are not equal to 1. Type Name: " + methodInfo.ReturnType.FullName);
```
cgillum, line 533: I'm not sure what this exception message is trying to communicate to the user. Can we put it in more actionable terms? Also, can we use a more specific exception type? If I understand correctly, I wonder if the following would be clearer:
```suggestion
                throw new NotSupportedException($"The method {methodInfo.Name} cannot be used because its return type '{methodInfo.ReturnType.FullName}' has more than one generic parameter.");
```

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -512,5 +516,111 @@ public static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int task
...
+            if (methodInfo.IsGenericMethod && !string.Equals(binderType.Name, "CSharpInvokeMemberBinder"))
+            {
+                throw new InvalidOperationException("Generic method invoked but method binder is not C#");
+            }
+
+            Type[] genericTypeArguments = binderType.GetProperty("TypeArguments")?.GetValue(binder) as Type[] ?? Array.Empty<Type>();
+
+            if (genericTypeArguments.Length != methodInfo.GetGenericArguments().Length)
+            {
+                throw new InvalidOperationException("Generic method mismatch");
```
cgillum, line 558: I don't understand what "Generic method mismatch" means. Can we use a more explicit exception message? Ideally it provides enough information that the developer knows what to do to try and fix the problem.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -512,5 +516,111 @@ public static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int task
...
+            }
+
+            if (typeToConvert.IsGenericType)
+            {
+                var genericArgs = typeToConvert.GetGenericArguments();
+                List<Type> genericTypeValues = new();
+
+                foreach (var genericArg in genericArgs)
+                {
+                    var index = Array.IndexOf(genericParameters, genericArg);
```
cgillum, line 595: Prefer explicit types over `var` if the right-hand expression doesn't already include the explicit type name.
```suggestion
                    int index = Array.IndexOf(genericParameters, genericArg);
```

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -512,5 +516,111 @@ public static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int task
...
+                    // Return the value of the generic argument.
+                    genericTypeValues.Add(ConvertFromGenericType(genericParameters, genericArguments, genericArguments[index]));
+                }
+
+                return typeToConvert.GetGenericTypeDefinition().MakeGenericType(genericTypeValues.ToArray());
+            }
+
+            if (typeToConvert.IsGenericParameter)
+            {
+                var index = Array.IndexOf(genericParameters, typeToConvert);
```
cgillum, line 606:
```suggestion
                int index = Array.IndexOf(genericParameters, typeToConvert);
```

cgillum (commented):
A few initial comments.

src/DurableTask.Core/Common/Utils.cs
aueloka: Agreed. This part was moved from existing code. Updating.

src/DurableTask.Core/Common/Utils.cs
aueloka: Got it. I've updated the message in the latest iteration.

src/DurableTask.Core/RetryProxy.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace DurableTask.Core
 {
+    using DurableTask.Core.Common;
```
cgillum, line 16: nit: in this project, `System` usings go before non-System usings.

src/DurableTask.Core/ScheduleProxy.cs
```diff
@@ -55,14 +55,24 @@ public void Intercept(IInvocation invocation)
...
 
+            Type[] genericArgumentValues = invocation.GenericArguments ?? Array.Empty<Type>();
+            List<object?> arguments = new(invocation.Arguments);
+
+            foreach (var typeArg in genericArgumentValues)
+            {
+                arguments.Add(new TypeMetadata(typeArg.Assembly.FullName!, typeArg.FullName!));
+            }
+
+            object[] args = arguments.ToArray()!;
```
cgillum, line 66: nit: to avoid an allocation and reduce the scope of this variable, can it be moved into the `if` block after line 71?

test/DurableTask.ServiceBus.Tests/SerializationTests.cs
```diff
@@ -807,5 +808,81 @@ public class MoreParamsOrchestrationInput
...
+                IGenericMethodInterface retryableClient = context.CreateRetryableClient<IGenericMethodInterface>(new RetryOptions(TimeSpan.FromMilliseconds(1), 1));
+
+                Dictionary<int, List<string>> a = await client.GetWhenTIsInput<string>(input.Property.ToString(), new[] { "test" }, new List<string>());
+                Assert.IsNotNull(a);
+                Assert.AreEqual(2, a[1].Count);
+
+                GenericInterfaceOrchestrationInput b = await client.GetWhenNoParams<GenericInterfaceOrchestrationInput>();
+                Assert.IsNull(b);
+
+                GenericInterfaceOrchestrationInput[] c = await retryableClient.GetWhenMultipleGenericTypes<double, GenericInterfaceOrchestrationInput>(input.Property, input);
```
cgillum, line 879: One of the previous issues we had was that the generated retriable client was non-deterministic. To help ensure that this version is fully deterministic, can you switch the order of the calls to move `retryableClient.GetWhenMultipleGenericTypes<...>(...)` to run *before* `client.GetWhenNoParams<GenericInterfaceOrchestrationInput>();`?  This is preferred because a non-deterministic action will often go undetected if it's the last action that happens in the orchestration.

cgillum (commented):
Just a few very small comments

test/DurableTask.ServiceBus.Tests/SerializationTests.cs
aueloka: Done

src/DurableTask.Core/RetryProxy.cs
aueloka: Turns out the dependency was no longer needed. Dropped it now

cgillum (approved):
LGTM! Thanks again for this contribution! We'll include it in our next release of DurableTask.Core, which will be v2.11.0.

PR #747, opened 2022-06-17
Make batch instance purging more efficient

cgillum:
Resolves https://github.com/Azure/azure-functions-durable-extension/issues/2211 (though perhaps not perfectly).

This PR makes batch purging more efficient in the following ways:

1. Scan the Instances table in multiple batches rather than scanning the whole thing in one go. This will result in decreased memory usage and prevent purge operations from making zero progress in timeout scenarios.
1. Remove outputs from the result of the initial scan to reduce network bandwidth and memory usage
1. Execute purge commands in parallel rather than serially, reducing the total amount of time expected to purge the history.

I added a small amount of code cleanup as well as made some fixes related to storage operation counting that I found while working through this code.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -623,55 +623,87 @@ namespace DurableTask.AzureStorage.Tracking
             return orchestrationStates;
         }
 
-        async Task<PurgeHistoryResult> DeleteHistoryAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus)
+        async Task<PurgeHistoryResult> DeleteHistoryAsync(
+            DateTime createdTimeFrom,
+            DateTime? createdTimeTo,
+            IEnumerable<OrchestrationStatus> runtimeStatus)
         {
-            TableQuery<OrchestrationInstanceStatus> query = OrchestrationInstanceStatusQueryCondition.Parse(createdTimeFrom, createdTimeTo, runtimeStatus)
-                .ToTableQuery<OrchestrationInstanceStatus>();
+            TableQuery<OrchestrationInstanceStatus> query = OrchestrationInstanceStatusQueryCondition.Parse(
+                createdTimeFrom,
+                createdTimeTo,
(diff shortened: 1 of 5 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -623,55 +623,87 @@ async Task<IList<OrchestrationState>> QueryStateAsync(TableQuery<OrchestrationIn
...
         {
-            TableQuery<OrchestrationInstanceStatus> query = OrchestrationInstanceStatusQueryCondition.Parse(createdTimeFrom, createdTimeTo, runtimeStatus)
-                .ToTableQuery<OrchestrationInstanceStatus>();
+            TableQuery<OrchestrationInstanceStatus> query = OrchestrationInstanceStatusQueryCondition.Parse(
+                createdTimeFrom,
+                createdTimeTo,
+                runtimeStatus).ToTableQuery<OrchestrationInstanceStatus>();
+
+            // Limit to batches of 100 to avoid excessive memory usage and table storage scanning
+            query.TakeCount = 2;
```
amdeel, line 637: It looks like you forgot to change this back to 100.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -168,13 +183,18 @@ string GetConditions()
...
+            DateTime? createdTimeTo,
+            IEnumerable<OrchestrationStatus> runtimeStatus)
         {
             var condition = new OrchestrationInstanceStatusQueryCondition
             {
                 CreatedTimeFrom = createdTimeFrom,
                 CreatedTimeTo = createdTimeTo ?? default(DateTime),
-                RuntimeStatus = runtimeStatus
+                RuntimeStatus = runtimeStatus,
+                FetchInput = false,
```
amdeel, line 196: Should we be worried about this changing the default behavior?

amdeel (commented):
This looks good! Just a few comments.

cgillum:
Thanks for the review, @amdeel. I've updated the PR with the fixes you called out. I also proactively incremented the package version number in the latest commit.

amdeel (approved):
LGTM!

Issue #545, opened 2021-06-04
Activity calls from CreateRetryableClient are non-deterministic


cgillum:
## Issue
It appears that proxy objects generated by `OrchestrationContext.CreateRetriableClient` result in clients that invoke activities in a way that is non-deterministic. In particular, there appears to be threading problems with these clients that often result in stuck orchestrations or orchestrations that fail with errors like the following:

```
DurableTask.Core.Exceptions.NonDeterministicOrchestrationException: Non-Deterministic workflow detected: TaskScheduledEvent: 4 TaskScheduled Method5 
   at DurableTask.Core.TaskOrchestrationContext.HandleTaskScheduledEvent(TaskScheduledEvent scheduledEvent) in C:\GitHub\durabletask\src\DurableTask.Core\TaskOrchestrationContext.cs:line 268
   at DurableTask.Core.TaskOrchestrationExecutor.ProcessEvent(HistoryEvent historyEvent) in C:\GitHub\durabletask\src\DurableTask.Core\TaskOrchestrationExecutor.cs:line 141
   at DurableTask.Core.TaskOrchestrationExecutor.ExecuteCore(IEnumerable`1 eventHistory) in C:\GitHub\durabletask\src\DurableTask.Core\TaskOrchestrationExecutor.cs:line 82
   at DurableTask.Core.TaskOrchestrationContext.HandleTaskScheduledEvent(TaskScheduledEvent scheduledEvent) in C:\GitHub\durabletask\src\DurableTask.Core\TaskOrchestrationContext.cs:line 268
```

Previous maintainers of this repro have mentioned that this API is buggy.

The issue may be related to the use of [Dynamity](https://www.nuget.org/packages/Dynamitey) for dynamically invoking activity functions.

## Workarounds
* Use `OrchestrationContext.ScheduleWithRetry` - unfortunately this is not a type-safe way to call activity tasks.
* Use `OrchestrationContext.CreateClient` - unfortunately this doesn't support automatic retries.

saguiitay:
Any update on this? Can you provide a snippet for using `OrchestrationContext.ScheduleWithRetry` (what name/version should I be using?)

cgillum:
[...]

PR #750, opened 2022-06-20
#545 Replace Dynamity with Castle.Core for CreateRetryableClient

aueloka:
This change addresses #545 by replacing the Dynamity Library with [Castle.Core DynamicProxy](https://github.com/castleproject/Core/blob/master/docs/dynamicproxy.md).

### DIFF
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -29,13 +29,11 @@
   </PropertyGroup>
   
   <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="ImpromptuInterface" Version="6.2.2" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
     <PackageReference Include="Newtonsoft.Json" Version="7.0.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="ImpromptuInterface" Version="7.0.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -44,6 +44,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.1" />
     <PackageReference Include="System.Reactive.Core" Version="4.4.1" />
     <PackageReference Include="System.Reactive.Compatibility" Version="4.4.1" />
+    <PackageReference Include="Castle.Core" Version="5.0.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(Configuration)'=='Release'">
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -43,6 +43,7 @@
     <PackageReference Include="System.Reactive.Core" Version="4.4.1" />
     <PackageReference Include="System.Reactive.Compatibility" Version="4.4.1" />
     <PackageReference Include="Castle.Core" Version="5.0.0" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(Configuration)'=='Release'">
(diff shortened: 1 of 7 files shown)


src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -44,6 +44,7 @@
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.1" />
     <PackageReference Include="System.Reactive.Core" Version="4.4.1" />
     <PackageReference Include="System.Reactive.Compatibility" Version="4.4.1" />
+    <PackageReference Include="Castle.Core" Version="5.0.0" />
```
cgillum, line 47: With this change, are we able to get rid of the `ImpromptuInterface` dependencies further up?

aueloka: > With this change, are we able to get rid of the `ImpromptuInterface` dependencies further up?

Not yet. ReflectionBasedTaskActivity.cs has some dependency on it [here](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/ReflectionBasedTaskActivity.cs#:~:text=DataConverter.Serialize(await%20((dynamic)invocationTask))%3B)

```
src\DurableTask.Core\ReflectionBasedTaskActivity.cs(130,44): error CS0656: Missing compiler required member 'Microsoft.CSharp.RuntimeBinder.CS
harpArgumentInfo.Create'
```
My PR on generics removes this by interacting with `Task<T>` directly without using `dynamic`.

aueloka: Actually, we can remove it if I update line 130 of `ReflectionBasedTaskActivity.cs` to this:

```c#
await invocationTask;

var returnType = this.MethodInfo.ReturnType.GetGenericArguments().Single(); //Task<T>
var resultProperty = typeof(Task<>).MakeGenericType(returnType).GetProperty("Result");
serializedReturn = this.DataConverter.Serialize(resultProperty.GetValue(invocationTask));
```

If this is fine, I'll push a commit with this change included.

aueloka: I've gone ahead and made this change and also dropped the refeences to `ImpromptuInterface`.

aueloka:
> @aueloka would you mind taking a look at the build errors reported by the CI? It looks like they're coming from some of the projects that depend on DurableTask.Core.

Sure. Is there a way I can run the pipeline myself to verify?

aueloka:
> /azp run

/azp run

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -44,6 +42,8 @@
...
+    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
```
cgillum, line 46: Why is this new `Microsoft.CSharp` dependency necessary?

aueloka: There's a part of `ServiceBusUtils.cs` that attempts to cast to a dynamic type: https://github.com/Azure/durabletask/blob/main/src/DurableTask.ServiceBus/Common/ServiceBusUtils.cs#:~:text=(dynamic)new%20TaskMessage

This causes the compiler error `src\DurableTask.ServiceBus\Common\ServiceBusUtils.cs(283,24): error CS0656: Missing compiler required member 'Microsoft.CSharp.RuntimeBinder.B 
inder.Convert'`. This [answer](https://stackoverflow.com/questions/49637389/missing-compiler-required-member-microsoft-csharp-runtimebinder-binder-convert) points out it's due to a missing Microsoft.CSharp dll, which Dynamity was previously providing.

aueloka: image

cgillum: That makes sense, but why do we add this as a dependency to DurableTask.Core? The code you referenced lives in DurableTask.ServiceBus. Could we instead just add this dependency to the projects that strictly need it? I'm just trying to be really careful about any changes we make to dependencies because they can have lots of unintended consequences. :)

aueloka: Sounds good. I added here since this is the project being updated. I've moved the reference now.

cgillum (approved):
Thanks - this looks great!

cgillum:
Hi @aueloka. We recently discovered an issue with our CI that was causing tests to not run correctly. We've fixed the CI problem but discovered a few tests that are consistently failing.

image

Based on the changes that have been made since the CI stopped working, my guess is that this PR is the root cause of the failures. Can you please investigate to see if you can confirm the problem and/or a fix?  This is unfortunately urgent for us because it's blocking a release. If we can't get this resolved soon, we'll need to revert this CreateRetryableClient PR (and any related PRs) so that we can unblock the next release of DurableTask.Core. Thanks for your understanding!

In a brief debugging session I did for one of the failing tests, it appears that the retry proxy logic has been broken, causing us to not correctly count the number of retries.

PR #756, opened 2022-06-24
Support Add Task Activities from Abstract or Non-Sealed Class

aueloka:
This change extends `TaskHubWorker.AddTaskActivitiesFromInterface` by supporting adding activities from `abstract` or non-sealed classes.

### DIFF
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -88,12 +88,18 @@ namespace DurableTask.Core
         /// <returns></returns>
         public virtual T CreateClient<T>(bool useFullyQualifiedMethodNames) where T : class
         {
-            if (!typeof(T).IsInterface)
+            IInterceptor scheduleProxy = new ScheduleProxy(this, useFullyQualifiedMethodNames);
+
+            if (typeof(T).IsClass)
             {
-                throw new InvalidOperationException("Pass in an interface.");
+                if (typeof(T).IsSealed)
+                {
+                    throw new InvalidOperationException("Class cannot be sealed.");
+                }
+
+                return ProxyGenerator.CreateClassProxy<T>(scheduleProxy);
             }
 
-            IInterceptor scheduleProxy = new ScheduleProxy(this, useFullyQualifiedMethodNames);
             return ProxyGenerator.CreateInterfaceProxyWithoutTarget<T>(scheduleProxy);
         }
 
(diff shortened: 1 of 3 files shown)


cgillum (commented):
I worry a little bit about this getting messy since methods on classes aren't always going to be properly designed to handle this kind of "remoting". It seems like an area where developers could run into various issues accidentally.

Would you be willing to consider a design where classes must opt certain methods into being used as interfaces, perhaps using an attribute? For example:

```csharp
public abstract class ClassClient
{
    [TaskMethod]
    public abstract Task<T[]> Run<T>();
}
```

aueloka:
> I worry a little bit about this getting messy since methods on classes aren't always going to be properly designed to handle this kind of "remoting". It seems like an area where developers could run into various issues accidentally.
> 
> Would you be willing to consider a design where classes must opt certain methods into being used as interfaces, perhaps using an attribute? For example:
> 
> ```cs
> public abstract class ClassClient
> {
>     [TaskMethod]
>     public abstract Task<T[]> Run<T>();
> }
> ```

Yeah fair enough. Making it explicit with an attribute makes sense here. I'll do that.

aueloka:
@cgillum , just curious, when you say "methods on classes aren't always going to be properly designed to handle this kind of remoting, what do you mean?

What differentiates class methods from interface methods in this context? Asides from the possibility of classes having private, protected or static methods?

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -88,12 +88,23 @@ public abstract class OrchestrationContext
         /// <returns></returns>
         public virtual T CreateClient<T>(bool useFullyQualifiedMethodNames) where T : class
         {
-            if (!typeof(T).IsInterface)
+            if (!typeof(T).IsInterface || !typeof(T).IsClass)
```
aueloka, line 91:
```suggestion
            if (!typeof(T).IsInterface && !typeof(T).IsClass)
```

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -132,14 +143,24 @@ public abstract class OrchestrationContext
         /// <returns>Dynamic proxy that can be used to schedule the remote tasks</returns>
         public virtual T CreateRetryableClient<T>(RetryOptions retryOptions, bool useFullyQualifiedMethodNames) where T : class
         {
-            if (!typeof(T).IsInterface)
+            if (!typeof(T).IsInterface || !typeof(T).IsClass)
```
aueloka, line 146:
```suggestion
            if (!typeof(T).IsInterface && !typeof(T).IsClass)
```

aueloka:
> @cgillum , just curious, when you say "_methods on classes aren't always going to be properly designed to handle this kind of remoting_", what do you mean?
> 
> What differentiates class methods from interface methods in this context? Asides from the possibility of classes having private, protected or static methods?



> @cgillum , just curious, when you say "_methods on classes aren't always going to be properly designed to handle this kind of remoting_", what do you mean?
> 
> What differentiates class methods from interface methods in this context? Asides from the possibility of classes having private, protected or static methods?

@cgillum , just checking in on this if you don't mind sharing your thoughts.

aueloka:
> LGTM!

Thanks @cgillum. Can you help me merge it? Or how can I get access to be able to merge pull requests I create?

PR #758, opened 2022-06-29
add logic to support reverse order of query results.

kaibocai:
This PR add feature to reverse the query results. https://github.com/microsoft/durabletask-java/issues/73


cgillum:
The thing that's missing from this PR is the code that actually ensures that we return the data in the reversed order. However, as we discussed, this may not actually be supported by Azure Storage. If it's not supported, then we'll probably need to abandon this work item.

PR #759, opened 2022-06-29
Update ActivityKind values

### LINKED ISSUE #718: Update ActivityKind values
Originally taken from #692 ([comment](https://github.com/Azure/durabletask/pull/692#discussion_r836667217)).

> It would be good to also set ActivityKind to an appropriate value based on the activity at had:
> 
> TaskHubClient -> Producer (this produces a new message on the DTFx queue)
> TaskHubWorker processing an orchestration or activity -> Server (this is the 'server' running work)
> TaskHubWorker sending an outbound orchestration or activity -> Client (this is acting as a 'client' waiting for external work to process)
> 
> dtfx.type should then be used to indicate what is the task type, as in it is only even Activity or Orchestration values. So it can be read as "this ActivityKind.Server span is processing a dtfx.type of Orchestration"

bachuv:
This PR updates the external event `Activity`'s `ActivityKind` value. [This link](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitykind?view=net-6.0) lists the `ActivityKind` enum values.

image

Resolves #718

### DIFF
--- a/src/DurableTask.Core/Tracing/TraceHelper.cs
+++ b/src/DurableTask.Core/Tracing/TraceHelper.cs
@@ -153,7 +153,7 @@ namespace DurableTask.Core.Tracing
         {
             Activity? newActivity = ActivityTraceSource.StartActivity(
                 name: eventRaised.Name,
-                kind: ActivityKind.Internal,
+                kind: ActivityKind.Server,
                 parentContext: Activity.Current?.Context ?? default,
                 tags: new KeyValuePair<string, object?>[]
                 {
@@ -184,7 +184,7 @@ namespace DurableTask.Core.Tracing
 
             return ActivityTraceSource.StartActivity(
                 name: eventRaisedEvent.Name,
-                kind: ActivityKind.Producer,
+                kind: ActivityKind.Server,
                 parentContext: activityContext,
                 tags: new KeyValuePair<string, object?>[]
                 {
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -153,7 +153,7 @@ static TraceHelper()
         {
             Activity? newActivity = ActivityTraceSource.StartActivity(
                 name: eventRaised.Name,
-                kind: ActivityKind.Internal,
```
cgillum, line 156: This trace event gets created from TaskHubClient. Is there any particular reason you decided to use Server instead of Client?

jviau: +1 - I think this should be either `Client` or `Producer`

bachuv: Thanks for the feedback! After revisiting it, I updated the value to `Producer` since it's an asynchronous operation.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -184,7 +184,7 @@ static TraceHelper()
 
             return ActivityTraceSource.StartActivity(
                 name: eventRaisedEvent.Name,
-                kind: ActivityKind.Producer,
+                kind: ActivityKind.Server,
```
jviau, line 187: This is for when an external event is sent to an existing orchestration?

I am not exactly familiar with the flow, but I just want to make sure we consider of `Consumer` is appropriate here? Or is `Server` the correct choice?

RohitRanjanMS (approved):

Issue #516, opened 2021-03-04
Upgrade to a later version of the Storage SDK for DurableTask.AzureStorage


ConnorMcMahon:
We currently use v9.3.1 of the Azure Storage SDK, which is very out of date and is no longer maintained.

Upgrading to later versions of the SDK should allow us to get bug-fixes for issues like the deadlock we get that has caused many issues for us historically, as well as get us new features and support for key vault integration, as well as hopefully performance improvements.

Ideally, due to the structure of the Azure Storage team updating the namespaces of their types across major versions, we may be able to accomplish this as a non-breaking change, at least for Durable Functions.

ConnorMcMahon:
SDK versions to target:

Tables - Azure.Data.Tables v12 (still in beta)
Blobs - Azure.Storage.Blobs v12 
Queues - Azure.Storage.Queues v12

kzu:
Kinda mind-boggling that this is still using a 3 year old, deprecated API that's in production *right now*
![image](https://user-images.githubusercontent.com/169707/125869270-d59e821a-830d-4a0f-b73d-a6721a5fc345.png)


So much for the benefits of (mostly) evergreen SaaS via serverless Azure functions... 🤦 .

dim-37:
Because of this crazy issue with `WindowsAzure.Storage` package where it cannot parse `UseDevelopmentStorage=True` where true is written in capital I lost all the working day!!! 

Why the hell do you still use it???

xperiandri:
Really. Is it so hard to rewrite a few classes?!

cgillum:
@xperiandri @dim-37 Swapping NuGet packages is obviously not hard. Navigating conflicting dependencies with other projects (e.g., Azure Functions through Durable Functions) is the thing that's hard - otherwise we'd have done this years ago. The good news is that most of the blocking dependency issues have been resolved and we plan to do this work soon.

EricStG:
Our Azure storage accounts are recommending we upgrade packages, but since it's by the durable extensions it's not something we can do easily.

![image](https://user-images.githubusercontent.com/6012934/144629213-48b58f12-ecd0-4ec9-b90d-0dc5991bcb73.png)

a2741890:
Any update? 🤦

tom-configo:
Ditto, any update? We're being severely clobbered by  [#1403](https://github.com/Azure/azure-functions-durable-extension)

xperiandri:
[...]

PR #763, opened 2022-07-12
Migrate to the New Azure Storage SDKs

wsugarman:
Fixes https://github.com/Azure/durabletask/issues/516

This Pull Request is unfortunately rather long as it required replacing `WindowsAzure.Storage` with the three new libraries for the Azure Blob, Queue, and Table Storage service client SDKs. These SDKs continue to wrap the REST APIs for each of the respective services, but they rely upon the new primitives leveraged across all of the Azure SDKs for .NET. The changes can be summarized (so far) as follows:

- Remove the dependency on `WindowsAzure.Storage` across the libraries and replace it with:
    - `Azure.Data.Tables`
    - `Azure.Storage.Blobs`
    - `Azure.Storage.Queues`
- Replace all of the data models with their replacements including:
    - `string` -> `ETag?`
    - `DynamicTableEntity` -> `TableEntity`
    - `CloudQueueMessage` and Durable Task's `QueueMessage` with the built-in `QueueMessage`
    - `CloudTable` -> `TableClient`
    - `CloudQueue` -> `QueueClient`
    - `CloudBlob` -> `BlobClient`
- Replace the removed `TableQuery<T>` with raw OData queries
    - We could replace them with `Expression<T>` trees as well, but for performance it probably makes sense to use the OData if it's not too difficult to read and validate
- Update the `TableEntityConverter` to use cached dynamic serialization and deserialization delegates using `Expression` trees
- Replace `StorageAccountDetails` with `StorageAccountClientProvider` the encapsulates factories for blob, queue, and table service clients
- Replace the `TimeoutHandler` and some of the `AzureStorageClient` with a series of `HttpPipelinePolicy` objects that run throughout the processing of an HTTP request; we rely on the clients own built-in retry as opposed to building our own
- Add `CancellationToken` parameters and update method signatures to use `IAsyncEnumerable<T>`/`AsyncPageable<T>` throughout the internal clients; consider adding to the public APIs (for now I added them for APIs that return `IAsyncEnumerable<T>`)


cgillum (commented):
Thanks for this PR, Will! I think this would also be a good opportunity for us to bump the version of this package to v2.0.0, since we're changing a major dependency and since we're removing a lot of legacy public APIs.

jviau:
Love to see this change. It is pretty dense though as you said. I do have a question: for consumers that already use these new clients and specifically the dependency injection extensions from Azure SDK, have you included a pattern so that we can just re-use all that configuration and pass an already configured `TableServiceClient`/`BlobServiceClient`/`QueueServiceClient` to DurableTask.AzureStorage?

wsugarman:
> Love to see this change. It is pretty dense though as you said. I do have a question: for consumers that already use these new clients and specifically the dependency injection extensions from Azure SDK, have you included a pattern so that we can just re-use all that configuration and pass an already configured `TableServiceClient`/`BlobServiceClient`/`QueueServiceClient` to DurableTask.AzureStorage?
> 
> Reading a bit of the PR, I do have concerns for this as is. Specifically around the dependency on the custom pipeline policies being injected to the all the clients. I understand how you design needs those, but can we explore if there is a good way to allow consumer to use existing clients? I think this is important as the Azure SDK team has already provided a great way to configure and build clients via `IConfiguration` and dependency injection. It would be nice to see this change embrace that instead of side stepping it. But of course this PR is not where I expect DTFx to bring in DI and IConfiguration - that is a whole other chunk of work. What I am thinking is:
> 
> 1. Are the pipeline policies the only way these tasks can be accomplished?
> 2. If they are required, can we expose some helpers/APIs consumers can use to apply them to their existing clients?

Yes, I have been thinking about this, as we'll ultimately read the client configurations for the Durable Functions extensions from the app settings (similar to how we do it today for Managed Identity and how the Functions Host handles the new client SDKs). I am not yet satisfied with my current draft's solution, so I'll be noodling on that a bit more. I agree that we want to enable users to be able to easily configure their client options and potentially re-use their service clients. There are extensions provided by the Azure extensions SDK to resolve the options and the client from the configuration automatically.

But should we use the policies? I think it's the most desired way of performing these sorts of actions, as we're relying on the native retry capabilities of the SDKs (it may become confusing to allow users to configure retries separately from the client options). However, it is definitely not easy to give users the flexibility of providing their own clients and our own policies. Perhaps DI could indeed help here with something like named options...?

src/DurableTask.AzureStorage/AzureStorageProvider.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        /// <summary>
+        /// Creates an <see cref="AzureStorageProvider{TClient, TClientOptions}"/> for the Azure Queue Storage service.
+        /// </summary>
+        /// <param name="accountName">An Azure Storage account name.</param>
+        /// <param name="tokenCredential">An optional token credential for accessing the service.</param>
+        /// <param name="options">An optional set of client options.</param>
+        /// <returns>An Azure Queue Storage service client whose connection is based on the given <paramref name="accountName"/>.</returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="accountName"/> is <see langword="null"/> or consists entirely of white space characters.
```
cgillum, line 95: nit: I think [whitespace](https://en.wikipedia.org/wiki/Whitespace_character) is one word?

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -32,20 +32,24 @@
   </PropertyGroup>
```
cgillum, line 32: Just as a reminder, let's update `<MajorVersion>`, `<MinorVersion>`, and `<PatchVersion>` above to make this package into a v2.0.0 release.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -32,20 +32,24 @@
...
   
   <ItemGroup>
+    <PackageReference Include="Azure.Core" Version="1.24.0" />
+    <PackageReference Include="Azure.Data.Tables" Version="12.6.0" />
```
cgillum, line 36: I see that there's already a 12.6.1 version available. Should we make the version numbers here to be {major}.{minor}.* just so that we automatically pick up the patch fixes?

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -32,20 +32,24 @@
...
+    <PackageReference Include="Azure.Storage.Blobs" Version="12.12.0" />
+    <PackageReference Include="Azure.Storage.Queues" Version="12.10.0" />
```
cgillum, line 38: We'll need to be careful about our version selection here since we need to remain compatible with other storage related packages, like [Microsoft.Azure.WebJobs.Extensions.Storage](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.Storage#dependencies-body-tab). In particular, we'll need to see which versions exist for which extension bundles so that we can know which Azure Functions extension bundles we can safely include this change in. FYI @davidmrdavid

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -32,20 +32,24 @@
...
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.*" PrivateAssets="All" />
+    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
```
cgillum, line 41: I'm not sure if we'll be able to safely add dependencies to 6.x framework package versions since many of them remove support for .NET Core 2.x (and we still have Azure Functions customers that depend on this older framework). We'll need to do some follow-up research on this.

src/DurableTask.AzureStorage/Http/MonitoringHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,43 @@
...
+        readonly AzureStorageOrchestrationServiceStats stats;
+
+        public MonitoringHttpPipelinePolicy(AzureStorageOrchestrationServiceStats stats)
+        {
+            this.stats = stats ?? throw new ArgumentNullException(nameof(stats));
+        }
+
+        public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+        {
+            stats.StorageRequests.Increment();
```
cgillum, line 33: I love this pipeline stuff you've added! It seems to nicely cover much of the functionality we had before!

cgillum (commented):
First round of comments before I head out on vacation. I'm really excited about this PR! I'm hoping that @davidmrdavid and @amdeel can help push it forward in the meantime!

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
davidmrdavid: Yes, this is something to be careful with. A good source of reference to quickly validate compatibility with Extension Bundles is to look at the release notes [here](https://github.com/Azure/azure-functions-extension-bundles/releases), which contain the versions of each NuGet package included in a bundles release.

I'm happy to help test more in-depth once other parts of the PR feedback are addressed. For now, I'd say we can keep this in the backburner

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -14,150 +14,102 @@
...
-            }
-            finally
-            {
-                if (!force)
-                {
-                    requestThrottleSemaphore.Release();
-                }
-            }
-        }
+            options.AddPolicy(exceptionPolicy, HttpPipelinePosition.PerCall);
```
jviau, line 108: I like the pipeline policies, but I do wonder if the exception one should be removed. Or at the very least, ensure it is the outer-most policy. This is for two reasons: 1) incase any user defined policies throw exceptions we want to transform and 2) we don't want to change exception type on uses in their own policy. (They should be able to re-use policies that work off `RequestFailedExceptions` and not our own type).

src/DurableTask.AzureStorage/Storage/BlobContainer.cs
```diff
@@ -16,132 +16,81 @@ namespace DurableTask.AzureStorage.Storage
...
             }
 
-            return await this.azureStorageClient.MakeBlobStorageRequest<bool>(
-                (context, cancellationToken) => this.cloudBlobContainer.DeleteIfExistsAsync(accessCondition, null, context, cancellationToken),
-                "Delete Container");
+            return await this.blobContainerClient.DeleteIfExistsAsync(conditions, cancellationToken);
         }
 
-        public async Task<IEnumerable<Blob>> ListBlobsAsync(string? blobDirectory = null)
+        public IAsyncEnumerable<Page<Blob>> ListBlobsAsync(string? prefix = null, CancellationToken cancellationToken = default)
```
jviau, line 71: Can we return `AsyncPageable<Blob>` here directly? Can copy this file over for a helper to do this:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Shared/PageResponseEnumerator.cs

src/DurableTask.AzureStorage/Storage/TableQueryResponse.cs
```diff
@@ -0,0 +1,57 @@
...
+        public static implicit operator AsyncPageable<T>(TableQueryResponse<T> response) => response._query;
+
+        public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+        {
+            return this._query.GetAsyncEnumerator(cancellationToken);
+        }
+
+        public async Task<TableQueryResults<T>> GetResultsAsync(string? continuationToken = null, int? pageSizeHint = null, CancellationToken cancellationToken = default)
+        {
+            var sw = new Stopwatch();
```
jviau, line 42: `var sw = Stopwatch.StartNew();`

src/DurableTask.Core/IOrchestrationServiceClient.cs
```diff
@@ -76,12 +76,13 @@ public interface IOrchestrationServiceClient
...
+        /// Asynchronously enumerates over the list of orchestration states from the instance storage for the most current execution (generation) of the specified instance.
         /// </summary>
         /// <param name="instanceId">Instance id</param>
         /// <param name="allExecutions">True if method should fetch all executions of the instance, false if the method should only fetch the most recent execution</param>
-        /// <returns>List of OrchestrationState objects that represents the list of orchestrations in the instance store</returns>
-        Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions);
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+        /// <returns>An asynchronous enumerable over the list of orchestrations in the instance store</returns>
+        IAsyncEnumerable<OrchestrationState> GetOrchestrationStateAsync(string instanceId, bool allExecutions, CancellationToken cancellationToken = default);
```
jviau, line 85: Is this breaking change necessary in this PR? Would it be possible to keep this all scoped to only AzureStorage implementation? This will let us rev the major version for just that one package.

We can worry about deeper API changes like this later.

jviau (commented):
This is looking good. However, is it possible to scope down the changes to just DurableTask.AzureStorage? This will make it easier to review. I am not sure we are ready to introduce breaking changes to the core layer yet. The DurableTask.ServiceBus migration to the new clients can happen in a separate PR as well.

jviau:
> > Love to see this change. It is pretty dense though as you said. I do have a question: for consumers that already use these new clients and specifically the dependency injection extensions from Azure SDK, have you included a pattern so that we can just re-use all that configuration and pass an already configured `TableServiceClient`/`BlobServiceClient`/`QueueServiceClient` to DurableTask.AzureStorage?
> > Reading a bit of the PR, I do have concerns for this as is. Specifically around the dependency on the custom pipeline policies being injected to the all the clients. I understand how you design needs those, but can we explore if there is a good way to allow consumer to use existing clients? I think this is important as the Azure SDK team has already provided a great way to configure and build clients via `IConfiguration` and dependency injection. It would be nice to see this change embrace that instead of side stepping it. But of course this PR is not where I expect DTFx to bring in DI and IConfiguration - that is a whole other chunk of work. What I am thinking is:
> > 
> > 1. Are the pipeline policies the only way these tasks can be accomplished?
> > 2. If they are required, can we expose some helpers/APIs consumers can use to apply them to their existing clients?
> 
> Yes, I have been thinking about this, as we'll ultimately read the client configurations for the Durable Functions extensions from the app settings (similar to how we do it today for Managed Identity and how the Functions Host handles the new client SDKs). I am not yet satisfied with my current draft's solution, so I'll be noodling on that a bit more. I agree that we want to enable users to be able to easily configure their client options and potentially re-use their service clients. There are extensions provided by the Azure extensions SDK to resolve the options and the client from the configuration automatically.
> 
> But should we use the policies? I think it's the most desired way of performing these sorts of actions, as we're relying on the native retry capabilities of the SDKs (it may become confusing to allow users to configure retries separately from the client options). However, it is definitely not easy to give users the flexibility of providing their own clients and our own policies. Perhaps DI could indeed help here with something like named options...?

I think something that may help would be to have the bare minimum API be to provide the built client(s) to the DurableTask.AzureStorage layer. We can then have different levels of APIs to help users get up and running with the correct configuration.

1. User is bringing their own client. We expose extension methods that we require them to call when configuring their client which adds the appropriate policies.
2. User wants DTFx to build the client. We have the builder/provider patterns you are working on in your PR.

wsugarman:
Ok, we're finally passing the Azure Storage tests again. Let me make my way through the comments!

src/DurableTask.AzureStorage/AzureStorageProvider.cs
wsugarman: The method [string.IsNullOrWhiteSpace](https://learn.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?view=net-7.0) uses two words, but I see the [wikipedia page](https://en.wikipedia.org/wiki/Whitespace_character) uses one. A cursory Google search shows a hyphenated approach with "white-space." I'll let you choose which one we use lol!

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
wsugarman: We can definitely do that to make it easier. However, I don't see other package references using the wildcard. Is it ok we do it for these?

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
wsugarman: While it looks like a framework package, it is actually outside of the BCL so it shouldn't conflict with earlier frameworks on its own. However, its dependencies like [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) is an out-of-band (OOB) package meant to shim `IAsyncEnumerable<T>` into .NET Standard (and standard-compatible) frameworks. I don't think there should be an issue using the lates .NET 6-equivalent APIs in a .NET Core 3.1 app, but I'll check.

src/DurableTask.ServiceBus/Common/AsyncEnumerableExtensions.cs
```diff
@@ -11,23 +11,23 @@
...
-        /// <summary>
-        /// The continuation token to obtain more results
-        /// </summary>
-        public string ContinuationToken;
-
-        /// <summary>
-        /// The results for the partial query
-        /// </summary>
-        public IEnumerable<OrchestrationState> Results;
+        public static async IAsyncEnumerable<TResult> SelectMany<TSource, TResult>(this IAsyncEnumerable<TSource> source, Func<TSource, IEnumerable<TResult>> selector)
```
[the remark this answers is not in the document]
wsugarman: I didn't see anything at the time, but I also no longer need it haha

src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
```diff
@@ -153,20 +156,18 @@ public async Task ForceTerminateTaskOrchestrationAsync(string instanceId, string
             }
         }
 
-        public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions)
+        public async AsyncStateEnumerable GetOrchestrationStateAsync(string instanceId, bool allExecutions, [EnumeratorCancellation] CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
wsugarman: While I think your points make sense in isolation (you're right that the "Async" moniker is not entirely accurate, and the multiple ways of specifying tokens can be confusing to end users), it seems that the Azure SDKs have settled on a pattern of calling these methods "Async" and allowing users to specify tokens directly.

For example, `TableClient` provides [QueryAsync<T>](https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.tableclient.queryasync?view=azure-dotnet#azure-data-tables-tableclient-queryasync-1(system-linq-expressions-expression((system-func((-0-system-boolean))))-system-nullable((system-int32))-system-collections-generic-ienumerable((system-string))-system-threading-cancellationtoken)) for fetching paginated entities, while `SecretClient` allows users to query properties via [GetPropertiesOfSecretsAsync](https://learn.microsoft.com/en-us/dotnet/api/azure.security.keyvault.secrets.secretclient.getpropertiesofsecretsasync?view=azure-dotnet). It appears the naming helps differentiate them from their "synchronous" counterparts that return `Pageable<T>`/`IEnumerable<T>` instead of `AsyncPageable<T>`/`IAsyncEnumerable<T>`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1655,37 +1658,31 @@ public async Task<OrchestrationState> GetOrchestrationStateAsync(string instance
...
-        public async Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool allExecutions, bool fetchInput = true)
+        /// <param name="fetchInput">If <see langword="true"/>, fetch and return the input for the orchestration instance.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+        /// <returns>An asynchronous enumeration of the <see cref="OrchestrationState"/> objects that represent the list of orchestrations.</returns>
+        public async IAsyncEnumerable<OrchestrationState> GetOrchestrationStateAsync(string instanceId, bool allExecutions, bool fetchInput, [EnumeratorCancellation] CancellationToken cancellationToken = default)
         {
             // Client operations will auto-create the task hub if it doesn't already exist.
             await this.EnsureTaskHubAsync();
-            return await this.trackingStore.GetStateAsync(instanceId, allExecutions, fetchInput);
+            await foreach (OrchestrationState state in this.trackingStore.GetStateAsync(instanceId, allExecutions, fetchInput))
```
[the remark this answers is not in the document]
wsugarman: Ah good catch!

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
wsugarman: The error one didn't quite worked as I hoped, so it has been removed and changed into an extension method.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -56,48 +78,21 @@ public class AzureStorageOrchestrationServiceSettings
...
         /// </summary>
-        public string TrackingStoreNamePrefix { get; set; } = "DurableTask";
+        public string TaskHubName { get; set; } = "Default";
 
         /// <summary>
-        /// Gets or sets the name of the task hub. This value is used to group related storage resources.
+        /// Gets or sets the optional prefix used for the various tracking table names.
         /// </summary>
-        public string TaskHubName { get; set; } = "Default";
+        /// <remarks>If left unspecified, tables reuse the <see cref="TaskHubName"/> as their prefix.</remarks>
```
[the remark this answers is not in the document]
wsugarman: Yeah, perhaps it's not ideal. I think the reason I did this was because we used to have 1 single "default" connection for Azure Storage, where the tracking table (i.e. Azure table storage) connection could be overridden. Now there are 3 connections: one for each kind of Azure Storage service. Perhaps instead we can use the base URL (i.e. the account) being different as a signal to use `TrackingStoreNamePrefix` over `TaskHubName`?

src/DurableTask.AzureStorage/Storage/BlobContainer.cs
wsugarman: It's a bit of a pain because we would need to convert each `BlobItem` to a `Blob` which would mean re-creating the `Page<T>`. For ease, since we only use this method twice, I forwarded the `BlobItem` objects and made the caller transform them

src/DurableTask.Core/IOrchestrationServiceClient.cs
wsugarman: Given that all of the APIs in the new SDK operate on `IAsyncEnumerable<T>` and `AsyncPageable<T>`, I thought to include it here. The alternative would probably be a bunch of "shims" like for any method with continuation tokens
```csharp
var firstPage = _myClient.GetStuffAsync.AsPages(continuationToken: myToken).FirstOrDefault();
return /* the existing abstraction wrapping the token */;
```
However, we can probably leave the token-less APIs like the referenced. I was just in the zone lol

wsugarman:
> > > Love to see this change. It is pretty dense though as you said. I do have a question: for consumers that already use these new clients and specifically the dependency injection extensions from Azure SDK, have you included a pattern so that we can just re-use all that configuration and pass an already configured `TableServiceClient`/`BlobServiceClient`/`QueueServiceClient` to DurableTask.AzureStorage?
> > > Reading a bit of the PR, I do have concerns for this as is. Specifically around the dependency on the custom pipeline policies being injected to the all the clients. I understand how you design needs those, but can we explore if there is a good way to allow consumer to use existing clients? I think this is important as the Azure SDK team has already provided a great way to configure and build clients via `IConfiguration` and dependency injection. It would be nice to see this change embrace that instead of side stepping it. But of course this PR is not where I expect DTFx to bring in DI and IConfiguration - that is a whole other chunk of work. What I am thinking is:
> > > 
> > > 1. Are the pipeline policies the only way these tasks can be accomplished?
> > > 2. If they are required, can we expose some helpers/APIs consumers can use to apply them to their existing clients?
> > 
> > 
> > Yes, I have been thinking about this, as we'll ultimately read the client configurations for the Durable Functions extensions from the app settings (similar to how we do it today for Managed Identity and how the Functions Host handles the new client SDKs). I am not yet satisfied with my current draft's solution, so I'll be noodling on that a bit more. I agree that we want to enable users to be able to easily configure their client options and potentially re-use their service clients. There are extensions provided by the Azure extensions SDK to resolve the options and the client from the configuration automatically.
> > But should we use the policies? I think it's the most desired way of performing these sorts of actions, as we're relying on the native retry capabilities of the SDKs (it may become confusing to allow users to configure retries separately from the client options). However, it is definitely not easy to give users the flexibility of providing their own clients and our own policies. Perhaps DI could indeed help here with something like named options...?
> 
> I think something that may help would be to have the bare minimum API be to provide the built client(s) to the DurableTask.AzureStorage layer. We can then have different levels of APIs to help users get up and running with the correct configuration.
> 
> 1. User is bringing their own client. We expose extension methods that we require them to call when configuring their client which adds the appropriate policies.
> 2. User wants DTFx to build the client. We have the builder/provider patterns you are working on in your PR.

I think this makes sense. Although for (1), users today have no means of providing their own clients; they can only provide connection metadata. And when we look at Azure Functions, that too relies entirely on maintaining its own separate clients that are created via a configuration. So, I am unsure if it makes sense to expose such an API for Durable Task where users can bring their own already initialized clients.

At the very least, we can allow users to create the new `StorageAccountClientProvider` instances like they would create `StorageAccountDetails` today, and inside of the Durable Function Extension we can provide a configuration-based approach that aligns with the rest of the Azure Functions host.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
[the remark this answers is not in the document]
wsugarman: I've refactored this to bring back the original pattern

src/DurableTask.AzureStorage/Tracking/TableEntityConverter.cs
```diff
@@ -17,93 +17,210 @@ namespace DurableTask.AzureStorage.Tracking
...
     /// <summary>
-    /// Utility class for converting [DataContract] objects into DynamicTableEntity and back.
+    /// Utility class for converting objects that use <see cref="DataContractAttribute"/> into TableEntity and back.
     /// This class makes heavy use of reflection to build the entity converters.
     /// </summary>
     /// <remarks>
-    /// This class is safe for concurrent usage by multiple threads.
+    /// This class is thread-safe.
     /// </remarks>
     class TableEntityConverter
```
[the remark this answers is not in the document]
wsugarman: I wrote a simple benchmark that I can share if you'd like that compares the performance of serializing and deserializing an `ExecutionStartedEvent` object repeatedly:

```
| Method |     Mean |     Error |    StdDev |      Min |       Q1 |       Q3 |      Max | Ratio | RatioSD |   Gen0 |   Gen1 | Allocated | Alloc Ratio |
|------- |---------:|----------:|----------:|---------:|---------:|---------:|---------:|------:|--------:|-------:|-------:|----------:|------------:|
|    Old | 5.483 us | 0.1080 us | 0.2325 us | 5.084 us | 5.314 us | 5.623 us | 6.099 us |  1.00 |    0.00 | 1.0147 | 0.0076 |   6.23 KB |        1.00 |
|    New | 3.818 us | 0.0762 us | 0.0935 us | 3.661 us | 3.752 us | 3.885 us | 4.000 us |  0.68 |    0.04 | 0.8812 | 0.0038 |   5.41 KB |        0.87 |
```

wsugarman:
I hope to remove any changes to the Service Bus backend by completing this other PR first: https://github.com/Azure/durabletask/pull/808

src/DurableTask.AzureStorage/Partitioning/BlobPartitionLease.cs
```diff
@@ -13,24 +13,25 @@
 
 namespace DurableTask.AzureStorage.Partitioning
 {
+    using System.Threading;
     using System.Threading.Tasks;
     using DurableTask.AzureStorage.Storage;
     using Newtonsoft.Json;
 
-    class BlobLease : Lease
+    class BlobPartitionLease : Lease
```
wsugarman, line 21: The name `BlobLease` is used by the blob library, so I thought I'd change these type names to avoid the ambiguity and make their purpose a bit clearer

src/DurableTask.AzureStorage/SimpleBufferManager.cs
```diff
@@ -16,13 +16,12 @@ namespace DurableTask.AzureStorage
     using System;
     using System.Collections.Concurrent;
     using System.Threading;
-    using Microsoft.WindowsAzure.Storage;
 
     /// <summary>
     /// Simple buffer manager intended for use with Azure Storage SDK and compression code.
     /// It is not intended to be robust enough for external use.
     /// </summary>
-    class SimpleBufferManager : IBufferManager
```
wsugarman, line 25: This interface no longer exists

src/DurableTask.AzureStorage/Partitioning/Lease.cs
```diff
@@ -53,7 +56,10 @@ public Lease(Lease source)
 
         /// <summary>Determines whether the lease is expired.</summary>
         /// <returns>true if the lease is expired; otherwise, false.</returns>
-        public virtual bool IsExpired => false;
+        public virtual Task<bool> IsExpiredAsync(CancellationToken cancellationToken = default)
+        {
+            return Task.FromResult(false);
```
jviau, line 61: Tiny nit: tasks have non-negligible GC impact (depending how often this is called). I like to cache constant tasks like this in a readonly static.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -82,7 +81,7 @@ private static bool DefaultLeaseDecisionDelegate(string leaseId)
         public async Task InitializeAsync()
```
jviau, line 81: Is there a cancellation token to forward here? Doing such a good job adding them everywhere else 😄

src/DurableTask.AzureStorage/Storage/AsyncPageableAsyncProjection.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace DurableTask.AzureStorage.Storage
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Azure;
+
+    sealed class AsyncPageableAsyncProjection<TSource, TResult> : AsyncPageable<TResult>
```
jviau, line 23: Is this type needed/used? Curious what this does that `System.Linq.Async` can't?

src/DurableTask.AzureStorage/Storage/AsyncPageableProjection.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace DurableTask.AzureStorage.Storage
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Azure;
+
+    sealed class AsyncPageableProjection<TSource, TResult> : AsyncPageable<TResult>
```
jviau, line 23: Same here - can this be replaced with usage of `System.Linq.Async`?

src/DurableTask.AzureStorage/Storage/TableClientExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    using System;
+    using System.Linq.Expressions;
+    using System.Reflection;
+    using Azure.Data.Tables;
+
+    static class TableClientExtensions
+    {
+        static readonly Func<TableClient, Uri?> getEndpointFunc = CreateGetEndpointFunc();
+
+        public static Uri? GetEndpoint(this TableClient client)
```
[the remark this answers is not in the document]
jviau: Instead of returning `null` in `CreateGetEndpointFunc()` if the private field is not available, can we test if the new public property you mentioned is available and access it that way (still via expression compilation I imagine). Or maybe even check for the public property first, then the private field second.

src/DurableTask.AzureStorage/Storage/TableQueryResponse.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Azure;
+
+    class TableQueryResponse<T> : IAsyncEnumerable<T> where T : notnull
```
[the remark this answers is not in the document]
jviau: Is this a new or existing metric?

src/DurableTask.AzureStorage/Tracking/InstanceStoreBackedTrackingStore.cs
```diff
@@ -54,37 +56,33 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
...
-        public override async Task<InstanceStatus> FetchInstanceStatusAsync(string instanceId)
+        public override async Task<InstanceStatus> FetchInstanceStatusAsync(string instanceId, CancellationToken cancellationToken = default)
         {
             OrchestrationState state = await this.GetStateAsync(instanceId, executionId: null);
             return state != null ? new InstanceStatus(state) : null;
         }
 
         /// <inheritdoc />
-        public override async Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions, bool fetchInput = true)
+        public override async IAsyncEnumerable<OrchestrationState> GetStateAsync(string instanceId, bool allExecutions, bool fetchInput = true, [EnumeratorCancellation] CancellationToken cancellationToken = default)
```
jviau, line 71: I don't know if this is recommended or not, but I tend to _not_ expose the cancellation token publicly on async IAsyncEnumerable methods. Instead, I split into a public and private method:

``` CSharp
public IAsyncEnumerable<T> MethodName(...args)
{
    return MethodNameCore(args);
}

private async IAsyncEnumerable<T> MethodNameCore(...args, [EnumeratorCancellation] CancellationToken cancellationToken = default)
{
    // actual implementation
}
```

The reasoning is dotnet analyzers has a warning about always forwarding cancellation token. Which gets redundant with `System.Linq.Async` APIs that take in cancellation and also `IAsyncEnumerable.WithCancellation()` exists. Just a couple lines down that analyzer would flag and require you to write `await this.GetStateAsync(instanceId, false, cancellationToken).FirstOrDefaultAsync(cancellationToken);` A bit redundant right? By hiding the cancellation token from the public method, we get rid of the redundancy.

src/DurableTask.AzureStorage/Tracking/ODataCondition.cs
```diff
@@ -0,0 +1,47 @@
...
+#nullable enable
+namespace DurableTask.AzureStorage.Tracking
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Represents an OData condition as specified by a filter and a projection.
+    /// </summary>
+    public readonly struct ODataCondition
```
jviau, line 22: Do we have access to record structs? Would be a good opportunity here.

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
```diff
@@ -69,21 +68,19 @@ public class OrchestrationInstanceStatusQueryCondition
...
 
         /// <summary>
-        /// Get the TableQuery object
+        /// Get the corresponding OData filter.
         /// </summary>
-        /// <typeparam name="T"></typeparam>
         /// <returns></returns>
-        public TableQuery<T> ToTableQuery<T>()
-            where T : ITableEntity, new()
+        public ODataCondition ToOData()
```
jviau, line 74: Can this method, and the `ODataCondition` struct be `internal`?

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
```diff
@@ -69,111 +73,108 @@ interface ITrackingStore
...
 
         /// <summary>
         /// Get The Orchestration State for the Latest or All Executions
         /// </summary>
         /// <param name="instanceId">Instance Id</param>
         /// <param name="allExecutions">True if states for all executions are to be fetched otherwise only the state for the latest execution of the instance is fetched</param>
         /// <param name="fetchInput">If set, fetch and return the input for the orchestration instance.</param>
-        Task<IList<OrchestrationState>> GetStateAsync(string instanceId, bool allExecutions, bool fetchInput);
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+        IAsyncEnumerable<OrchestrationState> GetStateAsync(string instanceId, bool allExecutions, bool fetchInput, CancellationToken cancellationToken = default);
```
jviau, line 86: Do the `IAsyncEnumerable` returning methods have to declare `CancellationToken` on their public contract? Callers can always use `.WithCancellation()`. Curious if a precedence has been set by the dotnet runtime or any other core libraries.

jviau (commented):
Looking great! Nice to see it narrowed down to just AzureStorage.

Not fully done with the review, but some comments for now.

src/DurableTask.AzureStorage/Storage/TableClientExtensions.cs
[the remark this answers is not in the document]
wsugarman: I've updated the package `Azure.DataTables` to the latest version (as of 4 days ago) and it now includes this new property. The caveat is that it required an update to the `Azure.Core` package which is implicit referenced by the webjobs SDK (since it explicitly depends on `Azure.Storage.Blobs`).

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -271,22 +276,19 @@ public string GetNewLargeMessageBlobName(MessageData message)
...
-                foreach (Blob blob in blobList)
+                await foreach (Page<Blob> page in this.blobContainer.ListBlobsAsync(sanitizedInstanceId, cancellationToken).AsPages())
                 {
-                    blobForDeletionTaskList.Add(blob.DeleteIfExistsAsync());
-                }
+                    storageOperationCount++;
 
-                await Task.WhenAll(blobForDeletionTaskList);
-                storageOperationCount += blobForDeletionTaskList.Count;
+                    await Task.WhenAll(page.Values.Select(b => b.DeleteIfExistsAsync(cancellationToken)));
```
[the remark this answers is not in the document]
wsugarman: I may have thought something earlier, which I've now forgotten, but I think I'll instead keep the logic the as it was where it attempts to delete all of the blobs at once.

src/DurableTask.AzureStorage/Partitioning/BlobPartitionLeaseManager.cs
```diff
@@ -57,44 +59,37 @@ sealed class BlobLeaseManager : ILeaseManager<BlobLease>
...
-
-            var downloadTasks = new List<Task<BlobLease>>();
-            foreach (Blob blob in blobs)
+            await foreach (Page<Blob> page in this.taskHubContainer.ListBlobsAsync(this.blobDirectoryName, cancellationToken: cancellationToken).AsPages())
             {
-                downloadTasks.Add(this.DownloadLeaseBlob(blob));
-            }
-
-            await Task.WhenAll(downloadTasks);
+                BlobPartitionLease[] leases = await Task.WhenAll(page.Values.Select(b => this.DownloadLeaseBlob(b, cancellationToken)));
```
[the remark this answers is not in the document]
wsugarman: Yeah, we can definitely do that. I think the only disadvantage is that we'll execute the downloads serially instead of batching them like they are done today on the thread pool. However, given that the max partitioning is 16 (and is usually the default 4), this probably has a minimal impact.

src/DurableTask.AzureStorage/Partitioning/Lease.cs
wsugarman: I think in this case, the runtime is actually taking care of the cache given that the domain of possible boolean values is small (2). I previously had read about it [here](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#task) when learning about `ValueTask`:

> And since there are only two possible Boolean results (`true` and `false`), there are only two possible Task<bool> objects needed to represent all possible result values, and so the runtime is able to cache two such objects and simply return a cached Task<bool> with a Result of true, avoiding the need to allocate. Only if the operation completes asynchronously does the method then need to allocate a new Task<bool>, because it needs to hand back the object to the caller before it knows what the result of the operation will be, and needs to have a unique object into which it can store the result when the operation does complete.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
wsugarman: Ah whoops! Good catch

wsugarman: After looking at it, I think we may want to postpone this as I think I'd need to understand the implications of cancellation for the balancer, as it really only seems to start up and shut down 😅

src/DurableTask.AzureStorage/Storage/AsyncPageableAsyncProjection.cs
wsugarman: I didn't see anything for this or the synchronous variant. That is mostly because I am trying to maintain the `AsyncPageable<T>` API while also performing the projections and `System.Linq.Async` doesn't know about this Azure-only Pageable API. Although I'm open to suggestions too!

src/DurableTask.AzureStorage/Storage/TableQueryResponse.cs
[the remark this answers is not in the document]
wsugarman: These are existing metrics. I introduced a new type `TableQueryResults<T>` that replaced `TableEntitiesResponseInfo<T>`. Previously `TableEntitiesResponseInfo<T>` represented the values and metrics for a paginated table query, but `TableQueryResults<T>` represents the entire result set.

`TableQueryResponse<T>` on the other hand is a flexible type that may be treated as an `IAsyncEnumerable<T>`, `AsyncPageable<T>`, or can be used to retrieve the `TableQueryResults<T>` object if metrics are desired about the query execution.

src/DurableTask.AzureStorage/Tracking/InstanceStoreBackedTrackingStore.cs
wsugarman: I think that makes perfect sense for conciseness. However, I am also seeing recent BCL additions that include an explicit token parameter, like in the `ChannelReader` type that was introduced with .NET 6 that defines the method [`ReadAllAsync`](https://github.com/dotnet/runtime/blob/d099f075e45d2aa6007a22b71b45a08758559f80/src/libraries/System.Threading.Channels/src/System/Threading/Channels/ChannelReader.netcoreapp.cs#L18). There is also [`DeserializeAsyncEnumerable`](https://github.com/dotnet/runtime/blob/5ed24429e72ea7ed7348a5b3698c82d1a13a602f/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Read.Stream.cs#L366) in the `JsonSerializer` class. Although, there are not too many examples of `IAsyncEnumerable<T>` yet in the BCL.

src/DurableTask.AzureStorage/Tracking/ODataCondition.cs
wsugarman: We can use `record struct` if we move the project to use `<LangVersion>10.0</LangVersion>`. That's no big deal, but in this case I don't think we need the automatic equality comparisons. The nicest thing we could use would be the `init` keyword, but we would need a shim to use that in .NET Standard and .NET Framework. This is actually possible! I do this in my own personal projects, but unsure if we want to include it here lol

src/DurableTask.AzureStorage/Tracking/OrchestrationInstanceStatusQueryCondition.cs
wsugarman: Yeah totally!

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
wsugarman: As mentioned in the earlier comment, an explicit public cancellation token appears to be the growing pattern in the .NET BCLs and in the Azure libraries.

src/DurableTask.AzureStorage/Tracking/TableEntityConverter.cs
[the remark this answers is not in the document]
wsugarman: I tried to ensure correctness by including new backward and forward-compatible tests where I assert the same object can be read and written between the versions. I also tried to emulate the current behavior, but it is possible I missed a scenario. However, I don't see any handling of `IExtensibleDataObject` in the existing `TableEntityConverter`. I thought maybe JSON.NET does some handling, but it seems that it relies on its own `JsonExtensionData` attribute (which the new converter will also leverage).

wsugarman:
@jviau, @cgillum, @davidmrdavid - I have added a `<VersionSuffix>preview.1</VersionSuffix>` to the AzureStorage project to keep this in prerelease. I just looked at what the Azure client sdks did for their prerelease versions a few years ago, but please let me know if you want to use some other convention for the name or change how it is specified in the build.

jviau:
`2.0.0-preview.1` is perfect for the version. Thank you!

src/DurableTask.AzureStorage/Tracking/TableEntityConverter.cs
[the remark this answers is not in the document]
jviau: Can you link that code? I have look through the Newtonsoft.Json repo before, and [not a single reference](https://github.com/JamesNK/Newtonsoft.Json/search?q=IExtensibleDataObject) to `IExtensibleDataObject`. Maybe it is custom code in Antares doing this? I could see it done via a custom contract resolver.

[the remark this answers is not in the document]
cgillum: Oops, nevermind, I was wrong. It looks like we were actually using `JsonExtensionData`, and the places where we were using `IExtensibleDataObject` we were still using the old `DataContractSerializer` types.

wsugarman:
> `2.0.0-preview.1` is perfect for the version. Thank you!
> 
> @wsugarman can you let us know when you consider this ready for merge? It looks all good to me but wondering if you are still validation or working on any further tweaks.

I've merged with the latest code in `main` and have re-run the tests. I think it's good-to-go.

markphillips100:
I've not used Durable Tasks before but I'd like to give this PR a try if possible as I'm dependent on v12 storage deps in my host (not using Azure Functions).  Is the preview going to be available soon?

markphillips100:
FYI @wsugarman I've been using this PR and the hosting provided by @jviau with package reference modification to make use of it.  I only build locally and published packages to a local nuget server to try out some things, namely can it help me with a real world use case of managing the creation of long running azure cloud resources.

I am only using the azure storage for all DT processing and had no problems so far. I asked before but any ETA on a merge as I'd really like to use this?

wsugarman:
> FYI @wsugarman I've been using this PR and the hosting provided by @jviau with package reference modification to make use of it. I only build locally and published packages to a local nuget server to try out some things, namely can it help me with a real world use case of managing the creation of long running azure cloud resources.
> 
> I am only using the azure storage for all DT processing and had no problems so far. I asked before but any ETA on a merge as I'd really like to use this?

I suspect we will merge this in February. I was discussing with the Durable Task team before the holiday, but there is no firm ETA yet. I'll let the others provide a more authoritative timeline.

markphillips100:
@wsugarman Thanks for the update.  Do you anticipate any further changes?

I think for now I can make do with private feed packages for my own needs and swap over to official in feb (or whenever).

PR #764, opened 2022-07-13
Enable suspend-resume of orchestrators.

Julio-Arroyo:
Introducing the first draft of an implementation of a suspend-resume feature for orchestrations.

Some highlights of the design include the decision **not** to introduce new storage types to handle incoming messages while an orchestration is suspended. Instead, we treat incoming messages as a piece of state that gets re-built during replay of the history.


cgillum (commented):
Initial batch of comments. I'll try to add more over the next couple of days.

aueloka:
Just curious about the design for this feature. How would sub orchestrations be handled? Would a "suspend" on the parent fan out "suspend" events to all running sub orchestrations? Or would any running sub orchestration still proceed while the parent is suspended?

cgillum:
@aueloka the current design only suspends the target orchestration. Sub-orchestrations would continue running.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -466,7 +467,7 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
             AnalyticsEventSource.SetLogicalTraceActivityId(traceActivityId);
             
             try
-            {
+            {
```
davidmrdavid, line 470: Let's remove this whitespace :)

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -957,6 +957,30 @@ public override Task StartAsync()
...
+                    case EventType.ExecutionSuspended:
+                        runtimeStatus = OrchestrationStatus.Suspended;
+                        ExecutionSuspendedEvent executionSuspendedEvent = (ExecutionSuspendedEvent)historyEvent;
+                        instanceEntity.Properties["RuntimeStatus"] = new EntityProperty(OrchestrationStatus.Suspended.ToString());
+                        instanceEntity.Properties["CompletedTime"] = new EntityProperty(DateTime.UtcNow);
+                        this.SetInstancesTablePropertyFromHistoryProperty(
+                            historyEntity,
+                            instanceEntity,
+                            historyPropertyName: nameof(executionTerminatedEvent.Input),
+                            instancePropertyName: OutputProperty,
```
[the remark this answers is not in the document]
davidmrdavid: I also noticed we're using `executionTerminatedEvent.Input` in this block, which seems like an event for terminate instead of suspend.

src/DurableTask.Core/SuspendedOrchestrationMessageInfo.cs
```diff
@@ -0,0 +1,28 @@
...
+using DurableTask.Core.History;
+
+namespace DurableTask.Core
+{
+    class SuspendedOrchestrationMessageInfo
+    {
+        public HistoryEvent historyEvent;
+
+        // Had to put this here to handle raised event case. Otherwise, if I did it in TaskOrchestrationExecutor, I can't access the OpenTasks.
+        public TaskOrchestration taskOrchestration;
```
davidmrdavid, line 13: Can you clarify this comment some more? What was the alternative that did not work?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -34,6 +34,10 @@ internal class TaskOrchestrationContext : OrchestrationContext
         private bool executionCompletedOrTerminated;
         private int idCounter;
 
+        // Buffer to hold metadata associated with messages that arrive while orchestration is in suspended state
+        public List<SuspendedOrchestrationMessageInfo> suspendedOrchestrationMessages;
+        public bool executionSuspended;
```
davidmrdavid, line 39: nit: maybe rename to `isSuspended`?

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -110,7 +111,6 @@ void ProcessEvents(IEnumerable<HistoryEvent> events)
                                 this.context.CurrentUtcDateTime = decisionStartedEvent.Timestamp;
                                 continue;
                             }
-
```
davidmrdavid, line 113: can we add this whitespace back?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -538,6 +552,50 @@ public void CompleteOrchestration(string result)
...
+                            var eventRaisedEvent = (EventRaisedEvent)currMsgInfo.historyEvent;
+                            currMsgInfo.taskOrchestration.RaiseEvent(this, eventRaisedEvent.Name, eventRaisedEvent.Input);
+                        }
+                        else
+                        {
+                            this.AddEventToNextIteration(currMsgInfo.historyEvent);
+                        }
+                        break;
+                    default:
+                        throw new NotImplementedException("Found unkown event when resuming orchestration.");
```
davidmrdavid, line 590: typo: unkown -> unknown

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -538,6 +552,50 @@ public void CompleteOrchestration(string result)
             CompleteOrchestration(result, null, OrchestrationStatus.Completed);
         }
 
+        public void HandleExecutionSuspendedEvent(ExecutionSuspendedEvent suspendedEvent)
+        {
+            this.executionSuspended = true;
+            Debug.Assert(this.suspendedOrchestrationMessages == null);
+            this.suspendedOrchestrationMessages = new List<SuspendedOrchestrationMessageInfo>();
+        }
```
davidmrdavid, line 560: Hmm, I would like us to consider this logic some more. It seems that this assumes that orchestrators only receive a "suspend" resume once before getting resumed, but what happens if they receive multiple suspend-requests before receiving any resume-requests?

From the current logic, it would seem each subsequent resume would clear the buffer of messages, or am I reading this wrong?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -538,6 +552,50 @@ public void CompleteOrchestration(string result)
...
+        public void HandleExecutionResumedEvent(ExecutionResumedEvent resumedEvent)
+        {
+            Debug.Assert(suspendedOrchestrationMessages != null);
+
+            while (suspendedOrchestrationMessages.Count > 0)
+            {
+                var currMsgInfo = suspendedOrchestrationMessages[0];
+                switch (currMsgInfo.historyEvent.EventType)
+                {
+                    case EventType.TaskCompleted:
```
davidmrdavid, line 571: I worry about this method duplicating some of the logic we already have in `ProcessEvent`. Is it possible at all to re-use that code directly?

davidmrdavid (commented):
Also left a first batch of comments!

src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
```diff
@@ -151,6 +151,16 @@ public async Task ForceTerminateTaskOrchestrationAsync(string instanceId, string
             }
         }
 
+        public Task ForceSuspendTaskOrchestrationAsync(string instanceId, string reason)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task ForceResumeTaskOrchestrationAsync(string instanceId, string reason)
```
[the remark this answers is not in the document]
Julio-Arroyo: Done.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -479,6 +480,14 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
                     batch.LastCheckpointTime = history.LastCheckpointTime;
                 }
 
+                // query instances table to determine if the orchestration is suspended
+                IList<OrchestrationState> instances = await this.trackingStore.GetStateAsync(batch.OrchestrationInstanceId, false, fetchInput: true);
+                OrchestrationState instanceState = instances.FirstOrDefault();
+                if (instanceState.OrchestrationStatus == OrchestrationStatus.Suspended)
+                {
+                    batch.OrchestrationState.OrchestrationStatus = OrchestrationStatus.Suspended;
+                }
```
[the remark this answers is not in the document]
Julio-Arroyo: You are right, this is some logic I forgot to remove from the prototype implementation.

src/DurableTask.Core/SuspendedOrchestrationMessageInfo.cs
Julio-Arroyo: When implementing this, I had to consider two things: I needed access to the `isSuspended` flag (which is in `TaskOrchestrationContext.cs`) and needed the `taskOrchestrationContext` (which is in `TaskOrchestrationExecutor`) to make the `RaiseEvent()` call. Since I decided to use handlers in `TaskOrchestrationExecutor.cs` for `ExecutionResumedEvent` and `ExecutionSuspendedEvent`, then the logic is in `TaskOrchestrationContext.cs`.

Notice that the only `EventType` in `ProcessEvent` that does not make use of a handler is the `EventRaised`. Thus, I encountered the challenge that if I was dealing with `EventRaised` when resuming the orchestration, I could not make the `RaiseEvent()` call as usual.

src/DurableTask.Core/TaskOrchestrationContext.cs
Julio-Arroyo: I agree that these assumptions are not safe to make. I want to argue, though, that those cases are dealt with.

The suspend (resp. resume) request will go through if and only if the orchestration is in the running (resp. suspended) state. This is implemented in `DurableClient.cs` in the DF extension, in the `SuspendAsync` (resp. `ResumeAsync)` methods.

Let me know @davidmrdavid if that sounds reasonable, if you think that this logic should be enough to deal with the cases you mentioned.

src/DurableTask.Core/TaskOrchestrationContext.cs
Julio-Arroyo: Good point. With some refactoring, it may be possible to call `ProcessEvent()` directly? The issue is that all this logic is spread across two classes `TaskOrchestrationContext.cs` and `TaskOrchestrationExecutor.cs`. I left some discussion about this in your comment asking for further clarification on the `SuspendedOrchestrationMessageInfo.cs`.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -538,6 +552,50 @@ public void CompleteOrchestration(string result)
...
+                        }
+                        break;
+                    default:
+                        throw new NotImplementedException("Found unknown event when resuming orchestration.");
+
+                }
+                suspendedOrchestrationMessages.RemoveAt(0);
+            }
+            this.isSuspended = false;
+            suspendedOrchestrationMessages = null;
```
cgillum, line 596: Consider making this field never null - i.e. initialize it in the constructor and mark it as `readonly`. That may help reduce some of the overall complexity of your change.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -161,7 +167,11 @@ public OrchestrationStatus OrchestrationStatus
                     return ExecutionCompletedEvent.OrchestrationStatus;
                 }
 
-                return OrchestrationStatus.Running;
+                return _orchestrationStatus;
+            }
+            set
+            {
+                _orchestrationStatus = value;
```
cgillum, line 174: I don't feel good about this change because it fundamentally changes some of the assumptions we can make about the OrchestrationRuntimeState class. The design of this class, from my perspective, is to derive all properties from the history events. Making this property settable breaks that and will likely introduce new complexity. I suggest instead that you remove the setter and update the orchestration status field at the point in time where you're adding new history events. I think the `SetMarkerEvents` method would be a good place to check for your new history events and update the internal orchestration status accordingly.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -83,6 +88,7 @@ public OrchestrationRuntimeState(IList<HistoryEvent>? events)
             PastEvents = events != null ? new List<HistoryEvent>(events.Count) : new List<HistoryEvent>();
             NewEvents = new List<HistoryEvent>();
             completedEventIds = new HashSet<int>();
+            // OrchestrationStatus = orchStatus;
```
cgillum, line 91: Can this be removed?

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -65,6 +65,11 @@ public class OrchestrationRuntimeState
         /// </summary>
         public string? Status;
 
+        /// <summary>
+        /// The OrchestrationStatus corresponding to the OrchestrationRuntimeStatus
+        /// </summary>
+        public OrchestrationStatus _orchestrationStatus;
```
cgillum, line 71: This is clearly an internal field, so you should remove `public`.  I also recommend removing the `_` prefix since we don't use that convention elsewhere in this code.

cgillum (commented):
A few more comments from me, this time focused on the core of the implementation.

src/DurableTask.Core/TaskOrchestrationContext.cs
davidmrdavid: Just writing down the takeaways of our discussion, @Julio-Arroyo.

So I'm not certain that we can **enforce** the invariant that orchestrations only receive suspend requests when they're running. It could be the case that two clients send suspend requests at the same time, and each of them would think the orchestration is running, but the orchestrator would receive two requests. As a result, I suggest making the code resilient to this possibility :)

src/DurableTask.AzureServiceFabric/FabricOrchestrationServiceClient.cs
```diff
@@ -151,6 +151,16 @@ public async Task ForceTerminateTaskOrchestrationAsync(string instanceId, string
             }
         }
 
+        public Task SuspendTaskOrchestrationAsync(string instanceId, string reason)
+        {
+            throw new NotImplementedException();
```
davidmrdavid, line 156: We should make a note to notify the ServiceFabric backend maintainers about this change.

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
```diff
@@ -321,6 +321,68 @@ public override string OnGetStatus()
...
+        private bool hasSetCustomStatus = false;
+        private string customStatus = null;
+
+        private async Task SetCustomStatus(OrchestrationContext context, string status)
+        {
+            this.hasSetCustomStatus = true;
+            this.customStatus = status;
+            // Dummy timer to persist state.
+            await context.CreateTimer<int>(context.CurrentUtcDateTime, 0);
+        }
```
davidmrdavid, line 336: Interesting, I was under the impression that `setCustomStatus` was already a DurableTask-level API. From this change, it seems that it's only implemented at the Durable Extension level. _The more you know :)_

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
```diff
@@ -321,6 +321,68 @@ public override string OnGetStatus()
...
+        public override async Task<string> RunTask(OrchestrationContext context, string status)
+        {
+            await this.SetCustomStatus(context, status);
+
+            string newStatus = await this.WaitForTrigger();
+
+            if (newStatus == null)
+            {
+                return string.Empty;
+            }
```
davidmrdavid, line 347: Do we really need this return statement? Isn't it safe to also just set the customStatus to the empty string?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +920,32 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+                await client.TerminateAsync("terminate");
+
+                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(10));
+
+                Assert.AreEqual(OrchestrationStatus.Terminated, status?.OrchestrationStatus);
+                Assert.AreEqual("terminate", status?.Output);
+
+                await host.StopAsync();
+            }
+        }
```
Julio-Arroyo, line 947: @amdeel Added this basic test. We didn't explicitly discuss it before, but this was some core functionality that was giving me issues when testing other things, so I figured I better test for this explicitly.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +920,180 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+                /*                string suspendReason = "Deactivate_bomb";
+
+                                await host.StartAsync();
+                                var client = await host.StartOrchestrationAsync(typeof(Orchestrations.Bomb), "bomb");
+                                await client.WaitForStartupAsync(TimeSpan.FromSeconds(10));
+                                await client.SuspendAsync(suspendReason);
+
+                                // if this event goes through, the orchestration will throw an exception
+                                await client.RaiseEventAsync("bombTrigger", "tryToDeactivate");
+                                await Task.Delay(2000);
```
Julio-Arroyo, line 1065: @amdeel I decided that for this test scenario, I would rather go with the custom status approach (rather than the "bomb", or throwing exception, which you can find commented out). For one, it follows the template of other tests in this file, so I am using patterns already in place. Another reason is that checking the expected observed behavior explicitly might be better than trying to avoid an exception.

src/DurableTask.Core/History/ExecutionResumedEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// Gets the event type
+        /// </summary>
+        public override EventType EventType => EventType.ExecutionResumed;
+
+        /// <summary>
+        /// Gets or sets the serialized input for the the resuming event
+        /// </summary>
+        [DataMember]
+        public string Input { get; set; }
```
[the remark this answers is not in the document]
davidmrdavid: @Julio-Arroyo: just to be sure - if we wanted to name this field `Reason` instead of `Input`, then would it be possible to rename this field and its accessors to make that work? I think the answer is _yes_, and I believe that's what Chris was suggesting here.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -161,7 +166,7 @@ public OrchestrationStatus OrchestrationStatus
                     return ExecutionCompletedEvent.OrchestrationStatus;
                 }
 
-                return OrchestrationStatus.Running;
+                return orchestrationStatus;
```
davidmrdavid, line 169: Given that we've returned this field into a "getter", do we still need this new `return` statement in it?

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
```diff
@@ -321,6 +321,68 @@ public override string OnGetStatus()
         }
     }
 
+    public sealed class ChangeStatusOrchestration2 : TaskOrchestration<string, string>
```
davidmrdavid, line 324: Let's make sure to drop the `2` at the end of this name once we're done reviewing the tests :)

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
```diff
@@ -321,6 +321,68 @@ public override string OnGetStatus()
...
+        TaskCompletionSource<string> waitForTriggerHandle;
+        private bool hasSetCustomStatus = false;
+        private string customStatus = null;
+
+        private async Task SetCustomStatus(OrchestrationContext context, string status)
+        {
+            this.hasSetCustomStatus = true;
+            this.customStatus = status;
+            // Dummy timer to persist state.
+            await context.CreateTimer<int>(context.CurrentUtcDateTime, 0);
```
davidmrdavid, line 335: Interesting, I also didn't expect the "dummy timer" to be required. I'd love @amdeel's confirmation on this one.

davidmrdavid (commented):
Left some comments. Overall, I like the testing strategy with the custom status and I think the tests are fine as-is. 

My main blocker before merging this PR is the code duplication in `HandleExecutionResumedEvent`. I'd love to reach a resolution on that :)

src/DurableTask.Core/TaskOrchestrationContext.cs
Julio-Arroyo: Using your ideas, I refactored the signature of `HandleExecutionResumed` method to take a function pointer, so that the same code can be reused to process the events.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -406,10 +417,20 @@ public void HandleTaskCompletedEvent(TaskCompletedEvent completedEvent)
...
             {
-                OpenTaskInfo info = this.openTasks[taskId];
-                info.Result.SetResult(completedEvent.Result);
+                if (this.IsSuspended)
+                {
+                    Debug.Assert(this.suspendedOrchestrationMessages != null);
 
-                this.openTasks.Remove(taskId);
+                    var msgInfo = new SuspendedOrchestrationMessageInfo(completedEvent);
+                    suspendedOrchestrationMessages.Enqueue(msgInfo);
```
amdeel, line 446: I don't love this design. I think we could enqueue messages in ProcessEvent instead. This would also make sure we handle all incoming messages when the orchestrator is in the suspended state without missing any, including Timer messages. It also future-proofs the design to handle any new event type we may eventually add.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -214,16 +215,29 @@ void ProcessEvent(HistoryEvent historyEvent)
...
                 case EventType.EventRaised:
-                    if (this.skipCarryOverEvents || !this.context.HasContinueAsNew)
+                    if (this.context.IsSuspended)
                     {
-                        var eventRaisedEvent = (EventRaisedEvent)historyEvent;
-                        this.taskOrchestration.RaiseEvent(this.context, eventRaisedEvent.Name, eventRaisedEvent.Input);
-                    }
-                    else
+                        var msgInfo = new SuspendedOrchestrationMessageInfo(historyEvent, this.taskOrchestration, this.skipCarryOverEvents);
+                        this.context.AddSuspendedOrchestrationMessage(msgInfo);
```
amdeel: See my comment above. We can remove this code duplication.

src/DurableTask.Core/TaskOrchestrationContext.cs
amdeel: I agree, you should call ProcessEvent for each of the queue messages instead of adding this method. Reach out if you're struggling with the refactor.

amdeel (commented):
Left some initial feedback. I'll get to the testing component later today.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +920,180 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+
+                await host.StopAsync();
+            }
+        }
+
+        /// <summary> 
+        /// Tests that the status of an orchestration changes when it is suspended and resumed.
+        /// </summary>
+        [TestMethod]
+        public async Task SuspendResumeCheckStatus()
```
amdeel, line 953: We want to make sure we cover all the test cases, you can even add comments in code to be clear about what you're testing if you want to, but I think we can consolidate this code and get rid of the code duplication by merging these tests.

davidmrdavid: @amdeel: I'm not entirely clear on which tests you're suggesting should get merged :) . Can you please clarify?

src/DurableTask.Core/TaskOrchestrationContext.cs
davidmrdavid: I agree with this, let's see if we can generalize this enqueue'ing to work for all events in `ProcessEvent`. Otherwise, we have code duplication that will be harder to maintain.

src/DurableTask.Core/SuspendedOrchestrationMessageInfo.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using DurableTask.Core.History;
+
+namespace DurableTask.Core
+{
+    class SuspendedOrchestrationMessageInfo
```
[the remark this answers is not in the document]
davidmrdavid: Yep - now that `ProcessEvent` has been refactored to be recursive, I think you can safely remove this class, right?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
amdeel: Most of the test methods can be merged into one.

src/DurableTask.Core/TaskOrchestrationContext.cs
Julio-Arroyo: I love this suggestion! Based on this, I have made the following changes, which I think make the implementation easier to follow and avoid code duplication:

- In `ProcessEvent` I have added a generalized handler for the case when an orchestration is suspended.
- Remove dealing with suspended orchestrations from all other event handlers.

src/DurableTask.Core/SuspendedOrchestrationMessageInfo.cs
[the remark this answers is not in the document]
Julio-Arroyo: You are right! It's not needed, I've removed it.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -33,10 +33,8 @@ internal class TaskOrchestrationContext : OrchestrationContext
         private OrchestrationCompleteOrchestratorAction continueAsNew;
         private bool executionCompletedOrTerminated;
         private int idCounter;
-        // Buffer to hold metadata associated with messages that arrive while orchestration is in suspended state
-        private readonly Queue<SuspendedOrchestrationMessageInfo> suspendedOrchestrationMessages;
+        private readonly Queue<HistoryEvent> eventsWhileSuspended;
```
Julio-Arroyo, line 36: Now that I got rid of the unnecessary class SuspendedOrchestrationMessageInfo and have centralized the logic of dealing with events while the orchestration is suspended to the ProcessEvent class, I have modified the naming of this buffer and I consider that everything is much more readable now so no need for comments that clutter the code.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -177,52 +177,59 @@ void ProcessEvents(IEnumerable<HistoryEvent> events)
 
         void ProcessEvent(HistoryEvent historyEvent)
```
Julio-Arroyo, line 178: Now the individual handlers do not deal with the suspended or running question, and all is done in a single place here. To keep the pattern of using handlers, I kept the function pointer to `ProcessEvent` in `HandleExecutionResumed`. @amdeel perhaps I misunderstood your question earlier about not passing along the function pointer, did you want me to recursively call `ProcessEvent` here instead?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
Julio-Arroyo: Ok! I will consolidate them into a single test.

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
davidmrdavid: @amdeel: is this required?

src/DurableTask.Core/OrchestrationRuntimeState.cs
Julio-Arroyo: You are suggesting having no return value?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
davidmrdavid: Resolving this to help narrow down the remaining feedback.

src/DurableTask.AzureServiceFabric/Remote/RemoteOrchestrationServiceClient.cs
```diff
@@ -127,6 +127,26 @@ public async Task ForceTerminateTaskOrchestrationAsync(string instanceId, string
             }
         }
 
+        /// <summary>
+        /// Suspends the specified orchestration instance
```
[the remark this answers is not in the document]
davidmrdavid: Since this doesn't seem to be a convention we've followed well across this file / project (I see this from just reviewing the methods right above and below), I'd personally suggesting not blocking on this.

davidmrdavid (approved):
LGTM / Looks good to me.
No further blockers on my part, but please wait for @amdeel's final word as she's the SME on this repo.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -180,7 +180,10 @@ void ProcessEvent(HistoryEvent historyEvent)
             bool isWakingEvent = historyEvent.EventType == EventType.ExecutionResumed || historyEvent.EventType == EventType.ExecutionTerminated;
             if (this.context.IsSuspended && !isWakingEvent)
             {
-                this.context.HandleEventWhileSuspended(historyEvent);
+                if (historyEvent.EventType != EventType.ExecutionSuspended)
+                {
+                    this.context.HandleEventWhileSuspended(historyEvent);
+                }
```
Julio-Arroyo, line 186: I added this condition based on @davidmrdavid 's suggestion, and as a fix to @amdeel 's concern about being resilient - at the Durable Task Framework level - that suspend "behaves appropriately when misused". In particular, here I avoid adding a Suspend event to the queue if the orchestrator is already suspended.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -176,54 +177,63 @@ void ProcessEvents(IEnumerable<HistoryEvent> events)
...
-                    {
-                        var eventRaisedEvent = (EventRaisedEvent)historyEvent;
-                        this.taskOrchestration.RaiseEvent(this.context, eventRaisedEvent.Name, eventRaisedEvent.Input);
-                    }
-                    else
-                    {
-                        this.context.AddEventToNextIteration(historyEvent);
-                    }
-                    break;
+                if (historyEvent.EventType != EventType.ExecutionSuspended)
```
amdeel, line 183: You can make this one conditional instead of two.

nit: I'm not sure about the name "isWakingEvent." Aren't all events waking? But all not events override a suspended state.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -27,6 +27,7 @@ namespace DurableTask.AzureStorage.Tests
     using DurableTask.Core;
     using DurableTask.Core.Exceptions;
     using DurableTask.Core.History;
+    using DurableTask.Test.Orchestrations;
```
amdeel, line 30: The orchestrations should exist in the same class, you shouldn't need this.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +921,73 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
             }
         }
 
+        /// <summary>
+        /// End-to-end test which validates the Suspend-Resume functionality.
+        /// </summary>
+        [TestMethod]
+        public async Task SuspendResumeOrchestration()
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
```
amdeel, line 930: Make this test run when enableEtendedSessions is true and false by adding attributes to this test like "[DataRow(true)] [DataRow(false)]" instead of defaulting the parameter to true.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +921,73 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+            }
+        }
+
+        /// <summary>
+        /// Test that a suspended orchestration can be terminated.
+        /// </summary>
+        [TestMethod]
+        public async Task TerminateSuspendedOrchestration()
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
```
amdeel, line 971: Same comment about extended sessions.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +921,73 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+        /// </summary>
+        [TestMethod]
+        public async Task SuspendResumeOrchestration()
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions: true))
+            {
+                string originalStatus = "OGstatus";
+                string newStatus = "newStatus";
+                string suspendMsg = "sleepyOrch";
+                string resumeMsg = "wakeUp";
```
amdeel, line 935: These suspend and resume variables don't seem very important and are only used once. You can just use these strings in-line.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +921,73 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+                status = await client.GetStatusAsync();
+                Assert.AreEqual(originalStatus, JToken.Parse(status?.Status));
+
+                // Test case 3: external event now goes through
+                await client.ResumeAsync(resumeMsg);
+                await Task.Delay(2000);
+                status  = await client.GetStatusAsync();
+                Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);
+                Assert.AreEqual(newStatus, JToken.Parse(status?.Status));
+
```
amdeel, line 960: You should call WaitForCompletionAsync before stopping.

amdeel (commented):
This looks great! Just a few more small changes.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
Julio-Arroyo: It seems to me that this method should be in charge of re-directing the flow of execution based on the context, whereas the handlers should be the ones to incorporate the logic of dealing with messages. So, I have moved the second check into `HandleEventsWhileSuspended` (since its goal is to ignore suspend messages when an orchestrator is already suspended). As a result, we don't have a double check in this method.

I have renamed the `isWakingEvent` flag to `overrideSuspension` in an attempt to make it more descriptive. The reason we have this flag rather than putting all conditions in the same if statement, is based on @davidmrdavid 's feedback that if you do put it all together it becomes too long and hard to read and reason about.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
Julio-Arroyo: done

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
Julio-Arroyo: Sure

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
Julio-Arroyo: Got it.

src/DurableTask.Core/TaskOrchestrationExecutor.cs
davidmrdavid: @Julio-Arroyo: Since Amanda is the owner of this part of the code, please feel free to override my comment with her suggestion on this one.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -925,38 +924,35 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
             {
-                string originalStatus = "OGstatus";
-                string newStatus = "newStatus";
-                string suspendMsg = "sleepyOrch";
-                string resumeMsg = "wakeUp";
-
                 await host.StartAsync();
-                var client = await host.StartOrchestrationAsync(typeof(Test.Orchestrations.NextExecution), originalStatus);
+                var client = await host.StartOrchestrationAsync(typeof(Test.Orchestrations.NextExecution), "OGstatus");
```
amdeel, line 934: Sorry for the confusion, you can pull out the definitions for the reused strings. WakeUp appears to be the only non-reused string here.

src/DurableTask.Core/History/EventType.cs
```diff
@@ -11,6 +11,9 @@
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
 
+
+// NOTE: Changing the order of variables may cause bugs in OOProc SDKs. When introducing
+//       new EventTypes, append them at the end.
```
amdeel, line 16: This sounds like an OOProc bug, though in general I agree that we would be adding new types to the end of this file.

amdeel (approved):
It's great to see this PR done! We'll want to wait for Chris's approval before merging. Congratulations on your first feature Julio!

Test/DurableTask.Stress.Tests/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "WSL": {
+      "commandName": "WSL2",
+      "distributionName": ""
+    }
+  }
+}
```
cgillum, line 8: Does this file need to be included in the PR or can it be excluded? It doesn't seem related.

src/DurableTask.Core/IOrchestrationServiceClient.cs
```diff
@@ -75,6 +75,20 @@ public interface IOrchestrationServiceClient
...
+        /// <param name="instanceId">Instance to suspend</param>
+        /// <param name="reason">Reason for suspending</param>
+        Task SuspendTaskOrchestrationAsync(string instanceId, string reason);
+
+        /// <summary>
+        /// Resumes the specified orchestration instance
+        /// </summary>
+        /// <param name="instanceId">Instance to resume</param>
+        /// <param name="reason">Reason for resuming </param>
+        Task ResumeTaskOrchestrationAsync(string instanceId, string reason);
```
cgillum, line 90: Adding methods to an interface is a breaking change in .NET. For example, if we release this change, it will cause compile errors to the MSSQL and Netherite backends if they try to update the version of DurableTask.Core that they use. This seems pretty dangerous to me, so I suggest we introduce this in a non-breaking way by doing the following:

1. Create a new interface called `IOrchestrationServiceSuspendResumeClient` that contains these two method signatures.
2. Update AzureStorageOrchestrationService to implement this new interface (should be a one-line change)
3. Remove the `NotImplementedException`s for the services that don't support this, like Service Fabric now that they're not required to implement the interface.
4. In TaskHubClient, instead of calling into `IOrchestrationServiceClient` directly, try to first cast the `IOrchestrationServiceClient` object into an `IOrchestrationServiceSuspendResumeClient` using the `is` or `as` operator. If the cast succeeds, call the appropriate Suspend/Resume method on the object. Otherwise, throw a `NotImplementedException`.
5. Depending on how you call this functionality from the Durable Extension (I haven't looked at that PR yet), you may need to do the same thing as (4) in the Durable extension code.

This is a bit more work but will allow us to release the feature as a non-breaking change.

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -18,7 +18,7 @@
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>10</MinorVersion>
-    <PatchVersion>0</PatchVersion>
+    <PatchVersion>1</PatchVersion>
```
cgillum, line 21: The `PatchVersion` is incremented only for internal bug fixes. For new (non-breaking) features, we need to update the `MinorVersion` instead.

cgillum (changes requested):
Some initial comments related to versioning that we'll need to take care of before merging this change.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -259,6 +262,14 @@ void SetMarkerEvents(HistoryEvent historyEvent)
 
                 ExecutionCompletedEvent = completedEvent;
             }
+            else if (historyEvent is ExecutionSuspendedEvent)
+            {
+                orchestrationStatus = OrchestrationStatus.Suspended;
+            }
+            else if (historyEvent is ExecutionResumedEvent)
+            {
+                orchestrationStatus = OrchestrationStatus.Running;
```
cgillum, line 271: I think we should first check to see whether the orchestration is in a non-completed state before changing the status to `Running`. Otherwise, an `ExecutionResumedEvent` that gets received by a completed orchestration could potentially result in the orchestration set to the `Running` state, which would be incorrect and confusing. See the checks further above for other examples of where we do status sanity checks. I don't think we need to throw exceptions for this, but we should avoid setting the status to Running.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -720,6 +720,56 @@ public async Task TerminateInstanceAsync(OrchestrationInstance orchestrationInst
             await this.ServiceClient.ForceTerminateTaskOrchestrationAsync(orchestrationInstance.InstanceId, reason);
         }
 
+        /// <summary>
+        ///     Suspend the specified orchestration instance.
+        /// </summary>
+        /// <param name="orchestrationInstance">Instance to suspend.</param>
+        public Task SuspendInstanceAsync(OrchestrationInstance orchestrationInstance)
```
cgillum, line 727: Rather than creating two overloads, make a single overload with an optional `reason` parameter. This is the preferred design for modern .NET APIs.

Something like this:
```csharp
public async Task SuspendInstanceAsync(OrchestrationInstance orchestrationInstance, string reason = null)
{
    // ...
}
```

Same thing for `ResumeInstanceAsync`.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -41,6 +44,14 @@ public void AddEventToNextIteration(HistoryEvent he)
             continueAsNew.CarryoverEvents.Add(he);
         }
 
+        public void HandleEventWhileSuspended(HistoryEvent historyEvent)
```
cgillum, line 47: code style nit: please move this method below the constructor. This is a common practice for C# code (whoever added the `AddEventToNextIteration` method should have also followed this style guideline). Ideally it would be grouped with the other suspend/resume-related events.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -400,6 +412,19 @@ public void HandleEventSentEvent(EventSentEvent eventSentEvent)
             this.orchestratorActionsMap.Remove(taskId);
         }
 
+        public void HandleEventRaisedEvent(HistoryEvent historyEvent, bool skipCarryOverEvents, TaskOrchestration taskOrchestration)
```
cgillum, line 415: Shouldn't this method take `EventRaisedEvent` instead of `HistoryEvent`?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +920,76 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+                string suspendReason = "sleepyOrch";
+                string changedStatus = "newStatus";
+
+                await host.StartAsync();
+                var client = await host.StartOrchestrationAsync(typeof(Test.Orchestrations.NextExecution), originalStatus);
+                await client.WaitForStartupAsync(TimeSpan.FromSeconds(10));
+
+                // Test case 1: Suspend changes the status Running->Suspended
+                await client.SuspendAsync(suspendReason);
+                await Task.Delay(2000);
```
cgillum, line 943: nit: it's best to avoid sleep/delays in test code as much as possible since they are a common source of test reliability issues. Instead, consider writing a method that checks the status in a loop and breaks only when the orchestration status is "Suspended" (or throw an exception if the status doesn't change after X amount of time).

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +920,76 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+
+                // Test case 2: external event does not go through
+                await client.RaiseEventAsync("changeStatusNow", changedStatus);
+                await Task.Delay(2000);
+                status = await client.GetStatusAsync();
+                Assert.AreEqual(originalStatus, JToken.Parse(status?.Status));
+
+                // Test case 3: external event now goes through
+                await client.ResumeAsync("wakeUp");
+                await Task.Delay(2000);
```
cgillum, line 956: Same comment as the above.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -920,6 +920,76 @@ public async Task TerminateOrchestration(bool enableExtendedSessions)
...
+        public async Task TerminateSuspendedOrchestration(bool enableExtendedSessions)
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
+            {
+                await host.StartAsync();
+                var client = await host.StartOrchestrationAsync(typeof(Orchestrations.Counter), 0);
+                await client.WaitForStartupAsync(TimeSpan.FromSeconds(10));
+
+                await client.SuspendAsync("suspend");
+                await Task.Delay(2000);
```
cgillum, line 980: Same comment as above

test/DurableTask.AzureStorage.Tests/TestOrchestrationClient.cs
```diff
@@ -136,6 +136,22 @@ public Task TerminateAsync(string reason)
             return this.client.TerminateInstanceAsync(instance, reason);
         }
 
+        public Task SuspendAsync(string reason)
```
cgillum, line 139: Related to my previous comments, consider adding a flag parameter for waiting until the status changes before returning from this method.

cgillum (commented):
Completed a full pass of the code. Really excited to see this feature come in! I made some changes to a previous comment I made about `IOrchestrationServiceClient` so please take a look at that and let me know if you have any questions.

src/DurableTask.Core/History/ExecutionSuspendedEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// A history event for orchestration pause
+    /// </summary>
+    [DataContract]
+    public class ExecutionSuspendedEvent : HistoryEvent
+    {
+        /// <summary>
+        /// Creates a new ExecutionSuspendedEvent with the supplied params
+        /// </summary>
+        /// <param name="eventId">The event id of the history event</param>
+        /// <param name="reason">The serialized input of the termination event</param>
```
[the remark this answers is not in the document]
Julio-Arroyo: Fixed.

src/DurableTask.Core/History/ExecutionSuspendedEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.Core.History
+{
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// A history event for orchestration pause
```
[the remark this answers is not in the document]
Julio-Arroyo: Done.

src/DurableTask.Core/TaskOrchestrationContext.cs
Julio-Arroyo: Done.

src/DurableTask.Core/TaskHubClient.cs
Julio-Arroyo: fixed.

src/DurableTask.Core/TaskOrchestrationContext.cs
Julio-Arroyo: Got it.

src/DurableTask.Core/OrchestrationRuntimeState.cs
Julio-Arroyo: `ExecutionCompletedEvent` should be null as long as the orchestration is not completed right? I have used it for this check. Let me know if you don't think that would be resilient enough.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -937,7 +937,8 @@ bool IsExecutableInstance(OrchestrationRuntimeState runtimeState, IList<TaskMess
 
             if (runtimeState.ExecutionStartedEvent != null &&
                 runtimeState.OrchestrationStatus != OrchestrationStatus.Running &&
-                runtimeState.OrchestrationStatus != OrchestrationStatus.Pending)
+                runtimeState.OrchestrationStatus != OrchestrationStatus.Pending &&
+                runtimeState.OrchestrationStatus != OrchestrationStatus.Suspended)
```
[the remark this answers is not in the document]
Julio-Arroyo: Oh, great point. Thanks for bringing it up.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -161,7 +164,7 @@ public OrchestrationStatus OrchestrationStatus
                     return ExecutionCompletedEvent.OrchestrationStatus;
                 }
 
-                return OrchestrationStatus.Running;
+                return orchestrationStatus;
             }
```
cgillum, line 168: It doesn't feel right to me to have two sources of truth for what the orchestration status is (this new `orchestrationStatus` variable and also the `ExecutionCompletedEvent` property). To reduce overall complexity, I suggest you get rid of the ExecutionCompletedEvent != null check (and the return statement inside it) and always return the value of `orchestrationStatus` here. Also, update `SetMarkerEvents` with logic that will update `orchestrationStatus` when `historyEvent is ExecutionCompletedEvent completedEvent`. This way, `orchestrationStatus` becomes the single source of truth which is only updated inside `SetMarkerEvents`.

src/DurableTask.Core/DurableTask.Core.csproj
cgillum: Friendly reminder to update this. The new version should be 2.11.0 since we're adding a new feature.

cgillum (commented):
Looking much better! Just a couple more things.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -252,15 +247,9 @@ void SetMarkerEvents(HistoryEvent historyEvent)
...
-            else if (historyEvent is ExecutionCompletedEvent completedEvent)
+            else if (historyEvent is ExecutionCompletedEvent)
             {
-                if (ExecutionCompletedEvent != null)
-                {
-                    throw new InvalidOperationException(
-                        "Multiple ExecutionCompletedEvent found, potential corruption in state storage");
-                }
-
-                ExecutionCompletedEvent = completedEvent;
```
cgillum, line 260: I think this code you removed needs to be added back.

Julio-Arroyo: Added it back.

src/DurableTask.Core/OrchestrationRuntimeState.cs
Julio-Arroyo: Got it. Changes should be in place.

src/DurableTask.Redis/DurableTask.Redis.csproj
```diff
@@ -5,7 +5,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <RunCodeAnalysis>false</RunCodeAnalysis>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <FileVersion>0.1.9</FileVersion>
+    <FileVersion>0.1.10</FileVersion>
```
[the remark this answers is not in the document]
Julio-Arroyo: Done.

src/DurableTask.Core/DurableTask.Core.csproj
Julio-Arroyo: Got it.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -258,6 +256,18 @@ void SetMarkerEvents(HistoryEvent historyEvent)
                 }
 
                 ExecutionCompletedEvent = completedEvent;
+                orchestrationStatus = OrchestrationStatus.Completed;
```
cgillum, line 259: I believe this needs to be `completedEvent.OrchestrationStatus`, just like it was in the previous implementation of `SetMarkerEvents()`. Otherwise, orchestration failure won't be handled correctly. This is because `ExecutionCompletedEvent` handles multiple forms of completion, including failure.

cgillum (approved):
Changes look good to me and the CI is passing, so signing off on this! Since this is a big change that touches the core engine, let's verify that the scenario is working as expected in the extension repo before merging this change.

davidmrdavid:
<3 thanks @amdeel and @cgillum for the detailed reviews. Will work with @Julio-Arroyo to make sure this works as expected in the Extension prior to the merge

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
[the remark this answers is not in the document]
Julio-Arroyo: @cgillum  Actually, I want to argue that we _do_ need this change: `runtimeState.OrchestrationStatus != OrchestrationStatus.Suspended` must be included. (@davidmrdavid this is the bug we were unable to explain earlier today).

`isExecutableInstance` has one single reference. It is used by `AzureStorageOrchestrationService.cs` to check whether the batch of messages dequeued should be discarded or abandoned. If we do not include this extra check, there are many scenarios where this feature won't work. Consider the following two examples:

1. Consider a suspended orchestration. The client tries to raise an event. Thus, that message is enqueued, later dequeued. If we don't add `runtimeState.OrchestrationStatus != OrchestrationStatus.Suspended`, `isExecutableInstance` will return false. Which means that the `RaiseEvent` will be discarded and will not make it to the history.
2. Similarly, if you have a suspended orchestration, and try to terminate it, the `TerminateEvent` message will be also discarded. So the terminate operation wouldn't work on suspended orchestrations.

@cgillum I know you mentioned you worry about potential problems this change could introduce. Could you elaborate on that?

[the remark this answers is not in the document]
cgillum: @Julio-Arroyo Ah, I see. I was reading the logic backwards and thought we'd discard the message if it was in a suspended state. You're right - we need the suspended check here to avoid unnecessarily discarding work items for suspended instances.

cgillum:
I just took a look at the CI results, and it looks like tests never actually ran!  We need to investigate what's going on, and whether it's just this PR or if other PRs have the same problem.

cgillum:
I think the CI may have been silently broken since this PR back in April: https://github.com/Azure/durabletask/pull/707. Basically, our CI filters the list of test assemblies using folder paths. When we changed the TFM from `net461` to `net462` for the test projects, that changed the folder paths. As a result, the CI wouldn't find any test assemblies and simply marked the run as "passed" since nothing failed (problem is, no tests passed either).

I've fixed the problem with the CI configuration and will try re-running the CI. If we're lucky, it will pass, but it's possible that a regression may have snuck in from some other PR during the time that the tests were silently disabled.

cgillum:
Looks like some existing tests not related to this PR is failing.

image

My guess is that the regression was caused by https://github.com/Azure/durabletask/pull/750. I verified by debugging locally that at least one of these tests fail consistently. The regression has not shipped yet so the problem doesn't exist out in the wild, but it's unfortunately a blocker for this release. I'll contact the contributor to see if they can fix the issue. Worst case scenario is that we have to revert their changes to get the CI green again.

PR #768, opened 2022-07-14
Added .editorconfig and forced policies within it

xperiandri:
Also fixed race condition warnings from Parallel Checker
And added Directory.Directory.Build.props
And set C# 10 for all projects

I will tag you in each real code change, @cgillum


src/DurableTask.AzureServiceFabric/FabricProviderClient.cs
```diff
@@ -18,17 +18,20 @@ namespace DurableTask.AzureServiceFabric
...
+
     using Newtonsoft.Json;
 
     internal class FabricProviderClient : IFabricProviderClient
     {
-        readonly IReliableStateManager stateManager;
-        readonly SessionProvider orchestrationProvider;
-        readonly JsonDataConverter FormattingConverter = new JsonDataConverter(new JsonSerializerSettings() { Formatting = Formatting.Indented });
+        // TOOD: Check why is it imported if never used
+        private readonly IReliableStateManager stateManager;
```
xperiandri, line 32: This field is unused

src/DurableTask.Core/EventHandlingOnContinueAsNew.cs
```diff
@@ -24,7 +24,7 @@ public enum BehaviorOnContinueAsNew
         Ignore,
 
         /// <summary>
-        /// 
+        /// All pending EventRaisedEvents will be carried
```
xperiandri, line 27: Is this description correct?

src/DurableTask.Core/Middleware/DispatchMiddlewarePipeline.cs
```diff
@@ -14,19 +14,18 @@
...
-    using System.Linq;
+    using System.Collections.Concurrent;
     using System.Threading.Tasks;
 
     internal class DispatchMiddlewarePipeline
     {
-        readonly IList<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>> components =
-            new List<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>>();
+        private readonly ConcurrentStack<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>> components =
+            new ConcurrentStack<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>>();
```
xperiandri, line 23: Changed to `ConcurrentStack` to fix Parallel Checker race condition warning

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -465,9 +448,9 @@ async Task<long> GetQueueCount(string entityName)
         /// <param name="runtimeState">The Orchestration runtime state this message count is associated with</param>
         public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState)
         {
-            return currentMessageCount
-                + ((this.InstanceStore != null) ? runtimeState.NewEvents.Count + 1 : 0) // one history message per new message + 1 for the orchestration
-                > MaxMessageCount;
+            // one history message per new message + 1 for the orchestration
+            var newMessages = (this.InstanceStore is not null) ? runtimeState.NewEvents.Count + 1 : 0;
+            return currentMessageCount + newMessages > MaxMessageCount;
```
xperiandri, line 453: Refactored for readability

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -1563,34 +1528,32 @@ string GetNormalizedWorkItemEvent(int index, string message, OrchestrationWorkIt
...
+            static string GetMessageInfo(MessageContainer m) => $"{m.Message.MessageId} <{m.Action?.Event.EventId.ToString()}>";
+            var messagesInfo = string.Join(",", messages.Select(GetMessageInfo));
+            var format = GetFormattedLog($@"{messages.Count} messages queued for {messageType}: {messagesInfo}");
             TraceHelper.TraceSession(
                 TraceEventType.Information,
                 "ServiceBusOrchestrationService-SentMessageLog",
-                session.SessionId,
-            GetFormattedLog($@"{messages.Count.ToString()} messages queued for {messageType}: {
-                        string.Join(",", messages.Select(m => $"{m.Message.MessageId} <{m.Action?.Event.EventId.ToString()}>"))}"));
+                session.SessionId, format);
```
xperiandri, line 1542: Refactored for readability

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -1673,47 +1636,51 @@ async Task ServiceMonitorAsync(CancellationToken cancellationToken)
...
+                {
+                    await Task.FromResult(0);
+                }
+                catch (MessagingEntityNotFoundException)
+                {
+                    await Task.FromResult(0);
+                }
+            }
+
+            await Utils.ExecuteWithRetries(DeleteAsync, null, "SafeDeleteQueueAsync", 3, 5);
```
xperiandri, line 1657: Refactored to local method for readability

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -1673,47 +1636,51 @@ async Task ServiceMonitorAsync(CancellationToken cancellationToken)
...
-                }, null, "SafeCreateQueueAsync", 3, 5);
+                    await CreateQueueAsync(managementClient, path, requiresSessions, requiresDuplicateDetection, maxDeliveryCount, maxSizeInMegabytes);
+                }
+                catch (MessagingEntityAlreadyExistsException)
+                {
+                    await Task.FromResult(0);
+                }
+            }
+
+            await Utils.ExecuteWithRetries(CreateAsync, null, "SafeCreateQueueAsync", 3, 5);
```
xperiandri, line 1680: Refactored to local method for readability

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -165,23 +164,15 @@ TableQuery<AzureTableOrchestrationJumpStartEntity> CreateJumpStartQuery(DateTime
...
-            OrchestrationStateQueryFilter primaryFilter = null;
-            IEnumerable<OrchestrationStateQueryFilter> secondaryFilters = null;
-            Tuple<OrchestrationStateQueryFilter, IEnumerable<OrchestrationStateQueryFilter>> filters =
-                stateQuery.GetFilters();
-            if (filters != null)
-            {
-                primaryFilter = filters.Item1;
-                secondaryFilters = filters.Item2;
-            }
+            (OrchestrationStateQueryFilter primaryFilter, IEnumerable<OrchestrationStateQueryFilter> secondaryFilters) = stateQuery.GetFilters();
```
xperiandri, line 167: Refactored to use `ValueTuple` instead of `Tuple`

src/DurableTask.ServiceBus/Tracking/AzureTableCompositeTableEntity.cs
```diff
@@ -64,7 +65,7 @@ public abstract class AzureTableCompositeTableEntity : ITableEntity
         internal abstract IEnumerable<ITableEntity> BuildDenormalizedEntities();
 
         /// <summary>
-        /// 
+        /// Gets entity property by its name
```
xperiandri, line 68: Is this description correct?

test/DurableTask.ServiceBus.Tests/OrchestrationHubTableClientTests.cs
```diff
@@ -203,10 +206,10 @@ public async Task OrchestrationEventHistoryTest()
...
         {
-            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst0Gen0 = CreateStateEntities(this.tableClient, "0", "0");
-            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst0Gen1 = CreateStateEntities(this.tableClient, "0", "1");
-            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst1Gen0 = CreateStateEntities(this.tableClient, "1", "0");
-            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst1Gen1 = CreateStateEntities(this.tableClient, "1", "1");
+            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst0Gen0 = OrchestrationHubTableClientTests.CreateStateEntities(this.tableClient, "0", "0");
+            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst0Gen1 = OrchestrationHubTableClientTests.CreateStateEntities(this.tableClient, "0", "1");
+            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst1Gen0 = OrchestrationHubTableClientTests.CreateStateEntities(this.tableClient, "1", "0");
+            IEnumerable<AzureTableOrchestrationStateEntity> entitiesInst1Gen1 = OrchestrationHubTableClientTests.CreateStateEntities(this.tableClient, "1", "1");
```
xperiandri, line 212: Made methods static as analyzer required

test/DurableTask.SqlServer.Tests/QueryEntitiesTests.cs
```diff
@@ -113,8 +122,15 @@ public async Task VerifyOrchestrationStateQueryByInstanceIdTest()
...
+#pragma warning disable CA1806 // Do not ignore method results
             //ensure each status exists in the collection and they all have the same InstanceId
-            entities.Select((e, i) => { e.State.OrchestrationStatus = values[i]; e.State.OrchestrationInstance.InstanceId = instanceId; return e; }).ToList();
+            entities.Select((e, i) =>
+                     {
+                         e.State.OrchestrationStatus = values[i];
+                         e.State.OrchestrationInstance.InstanceId = instanceId;
+                         return e;
+                     }).ToList();
+#pragma warning restore CA1806 // Do not ignore method results
```
xperiandri, line 133: Just changed formatting fro readability

test/DurableTask.Test.Orchestrations/Performance/ExecutionCountingActivity.cs
```diff
@@ -28,8 +29,7 @@ public sealed class ExecutionCountingActivity : AsyncTaskActivity<int, int>
         protected override async Task<int> ExecuteAsync(TaskContext context, int taskId)
         {
             await Task.Delay(new Random().Next(5, 10));
-            Interlocked.Increment(ref Counter);
-            return Counter;
+            return Interlocked.Increment(ref Counter);
```
xperiandri, line 32: This way I fixed racing condition warning from Parallel Checker

test/TestFabricApplication/TestApplication.Common/Orchestrations/CounterException.cs
```diff
@@ -19,17 +19,23 @@ namespace TestApplication.Common.Orchestrations
...
+        private CounterException() { }
+
+        /// <summary>Initializes a new instance of the <see cref="CounterException" /> class with a specified error message.</summary>
+        /// <param name="message">The message that describes the error.</param>
+        private CounterException(string message) : base(message) { }
+
+        /// <summary>Initializes a new instance of the <see cref="CounterException" /> class with a specified error message and a reference to the inner exception that is the cause of this exception.</summary>
+        /// <param name="message">The error message that explains the reason for the exception.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception, or a null reference (<see langword="Nothing" /> in Visual Basic) if no inner exception is specified.</param>
+        private CounterException(string message, Exception innerException) : base(message, innerException) { }
```
xperiandri, line 38: All exceptions now have missing constructors as private like analyzer required

src/DurableTask.AzureServiceFabric/AsyncManualResetEvent.cs
```diff
@@ -33,18 +33,17 @@ public async Task<bool> WaitAsync(TimeSpan timeout, CancellationToken cancellati
...
-        }
+        public void Set() => this.taskCompletionSource.TrySetResult(true);
 
         public void Reset()
         {
             while (true)
             {
-                var thisTcs = this.taskCompletionSource;
+                TaskCompletionSource<bool> thisTcs;
+                thisTcs = this.taskCompletionSource;
```
xperiandri, line 43: This way I fixed race condition Parallel Checker warning

xperiandri (commented):
@cgillum these are all the real changes.
Everything else is just policies enforcement.

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -26,8 +26,8 @@ namespace DurableTask.AzureStorage.Tests
     [TestClass]
     public class StressTests
     {
-        int originalMinWorkerThreads;
-        int originalMinIoThreads;
+        private int originalMinWorkerThreads;
+        private int originalMinIoThreads;
```
cgillum, line 30: Please revert all changes to add default visibility modifiers. We want the opposite of this, which is to always *omit* the default visibility modifiers.

Test/DurableTask.AzureStorage.Tests/TestInstance.cs
```diff
@@ -58,7 +58,7 @@ public async Task<OrchestrationState> WaitForStart(TimeSpan timeout = default)
             do
             {
                 OrchestrationState state = await this.GetStateAsync();
-                if (state != null && state.OrchestrationStatus != OrchestrationStatus.Pending)
+                if (state is not null && state.OrchestrationStatus != OrchestrationStatus.Pending)
```
cgillum, line 61: Please revert the change from `!= null` to `is not null` and `== null` to `is null`. This change seems unnecessary, and I'd prefer we minimize the size of this PR. I noticed it also seems to be making changes to generated code, which makes me feel like it's going to cause more frustration than benefit.

Test/DurableTask.Core.Tests/RetryInterceptorTests.cs
```diff
@@ -47,6 +48,7 @@ public async Task Invoke_WithFailingRetryCall_ShouldHaveCorrectNumberOfCalls(int
                     callCount++;
                     throw new Exception();
                 });
+#pragma warning disable CA1031 // Do not catch general exception types
```
cgillum, line 51: Please revert these `#pragma warning disable CA1031 // Do not catch general exception types` warnings. I understand the purpose behind them, but I don't agree that they should be flagged. IMO whether to catch general exception types should be a case-by-case decision during code review.

samples/Correlation.Samples/HelloOrchestrator.cs
```diff
@@ -13,18 +13,20 @@
 
 namespace Correlation.Samples
 {
+#pragma warning disable CA1812 // Internal classes instantiated indirectly
```
cgillum, line 16: Let's revert this warning too. There are too many valid cases where it might need to be suppressed and I don't find it valuable enough to be worth enforcing.

samples/Correlation.Samples/TraceContextBaseExtensions.cs
```diff
@@ -13,25 +13,31 @@
...
         /// </summary>
         /// <param name="context">TraceContext</param>
         /// <returns></returns>
         public static RequestTelemetry CreateRequestTelemetry(this TraceContextBase context)
         {
-            var telemetry = new RequestTelemetry { Name = context.OperationName };
-            telemetry.Duration = context.Duration;
-            telemetry.Timestamp = context.StartTime;
-            telemetry.Id = context.TelemetryId;
+            Contract.Assume(context is not null);
```
cgillum, line 33: What does this do?

xperiandri:
I think it worse to merge this before https://github.com/Azure/durabletask/pull/763
Then I will also migrate the project to Packages.props before https://github.com/Azure/durabletask/pull/763 and apply that to it

cgillum (commented):
Some initial comments. My hope is that we can radically reduce the size of this PR by not changing previously established coding conventions.

Test/DurableTask.AzureStorage.Tests/TestInstance.cs
xperiandri: As far as I know `is null` and `is not null` are safer because they do not rely on equality operators. That is why I changed that

Test/DurableTask.Core.Tests/RetryInterceptorTests.cs
xperiandri: But if this is an exclusion from the general rule you must specify it explicitly and describe the reason, don't you?

xperiandri: OK, I'll fix that

Test/DurableTask.AzureStorage.Tests/StressTests.cs
xperiandri: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0040
`omit_if_default` ?

xperiandri:
Does it looks like you want now?

cgillum (commented):
It's going to take a while to review all 243 changed files, and I'll be going on vacation soon, which will delay things even further. Here are a few more comments after going through the first 10% of changed files (most of which is unfortunately code that I don't maintain).

samples/Correlation.Samples/HelloOrchestrator.cs
```diff
@@ -19,7 +19,7 @@ namespace Correlation.Samples
     using DurableTask.Core;
 
     [KnownType(typeof(Hello))]
-    internal class HelloOrchestrator : TaskOrchestration<string, string>
+class HelloOrchestrator : TaskOrchestration<string, string>
```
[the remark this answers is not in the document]
xperiandri: Ah, I'll fix that

Test/DurableTask.AzureStorage.Tests/StressTests.cs
```diff
@@ -19,8 +19,11 @@ namespace DurableTask.AzureStorage.Tests
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+
```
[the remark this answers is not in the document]
xperiandri: I'll try. It is what CodeRush does by default

samples/Correlation.Samples/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -313,10 +313,7 @@ internal static void UpdateTelemetry(ITelemetry telemetry, Activity activity, bo
             if (initializeFromCurrent)
             {
                 opTelemetry.Id = activity.SpanId.ToHexString();
-                if (activity.ParentSpanId != null)
-                {
-                    opTelemetry.Context.Operation.ParentId = activity.ParentSpanId.ToHexString();
-                }
+                opTelemetry.Context.Operation.ParentId = activity.ParentSpanId.ToHexString();
```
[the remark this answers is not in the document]
xperiandri: Because struct cannot be null 🙂

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -115,20 +102,13 @@ public Task DeleteAsync(bool deleteInstanceStore)
...
-        {
-            return false;
-        }
+        public bool IsMaxMessageCountExceeded(int currentMessageCount, OrchestrationRuntimeState runtimeState) => false;
 
         public int GetDelayInSecondsAfterOnProcessException(Exception exception)
-        {
-            return GetDelayForFetchOrProcessException(exception);
-        }
+         => GetDelayForFetchOrProcessException(exception);
```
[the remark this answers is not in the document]
xperiandri: Would you like me to force the opposite style?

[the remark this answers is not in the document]
xperiandri: Or switch back to suggestion but not warning?

src/DurableTask.AzureServiceFabric/OrchestrationStatusExtensions.cs
```diff
@@ -17,14 +17,9 @@ namespace DurableTask.AzureServiceFabric
...
-        {
-            return !status.IsRunningOrPending();
-        }
+        public static bool IsTerminalState(this OrchestrationStatus status) => !status.IsRunningOrPending();
 
         public static bool IsRunningOrPending(this OrchestrationStatus status)
-        {
-            return status == OrchestrationStatus.Pending || status == OrchestrationStatus.Running;
-        }
+         => status == OrchestrationStatus.Pending || status == OrchestrationStatus.Running;
```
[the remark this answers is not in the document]
cgillum: We always use 4 spaces.

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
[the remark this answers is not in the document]
cgillum: I'd prefer switch back to suggestion (or none at all, if possible) and revert. This is a case where I think the best style totally depends on the context, so I don't want to enforce either.

src/DurableTask.Core/OrchestrationInstance.cs
```diff
@@ -16,6 +16,7 @@ namespace DurableTask.Core
     using System.Diagnostics.CodeAnalysis;
     using System.Runtime.Serialization;
 
+    // TODO: Consider declaring as a record
```
xperiandri, line 19: Maybe it is better to convert this class to a record?

src/DurableTask.AzureServiceFabric/AsyncManualResetEvent.cs
xperiandri: Reverted

samples/Correlation.Samples/HelloOrchestrator.cs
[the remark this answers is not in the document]
xperiandri: Fixed

Test/DurableTask.AzureStorage.Tests/StressTests.cs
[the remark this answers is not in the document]
xperiandri: fixed

src/DurableTask.AzureServiceFabric/Exceptions/InvalidInstanceIdException.cs
```diff
@@ -11,6 +11,9 @@
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
 
+using System.Runtime.Serialization;
+using System.Security;
```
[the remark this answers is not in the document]
xperiandri: Enforced policy and fixed all

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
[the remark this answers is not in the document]
xperiandri: Reverted major expression-bodied members

src/DurableTask.AzureServiceFabric/OrchestrationStatusExtensions.cs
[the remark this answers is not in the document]
xperiandri: fixed

xperiandri (commented):
So I reverted everything as much as possible

xperiandri:
@shankarsama you've broken something in your last commit

src/DurableTask.AzureServiceFabric/Exceptions/InvalidInstanceIdException.cs
```diff
@@ -26,8 +28,23 @@ public class InvalidInstanceIdException : Exception
...
+        InvalidInstanceIdException(string message, Exception innerException) : base(message, innerException) { }
+
+        /// <summary>Initializes a new instance of the <see cref="InvalidInstanceIdException" /> class with serialized data.</summary>
+        /// <param name="info">The <see cref="System.Runtime.Serialization.SerializationInfo" /> that holds the serialized object data about the exception being thrown.</param>
+        /// <param name="context">The <see cref="System.Runtime.Serialization.StreamingContext" /> that contains contextual information about the source or destination.</param>
+        /// <exception cref="System.ArgumentNullException">
+        /// <paramref name="info" /> is <see langword="null" />.</exception>
+        /// <exception cref="System.Runtime.Serialization.SerializationException">The class name is <see langword="null" /> or <see cref="System.Exception.HResult" /> is zero (0).</exception>
+        [SecuritySafeCritical]
+        protected InvalidInstanceIdException(SerializationInfo info, StreamingContext context) : base(info, context) { }
```
cgillum, line 48: Do you feel confident that introducing this new constructor won't change the existing serialization behavior of this class? It's critical that no change in this PR results in any serialization changes for data that may appear in orchestration history.

src/DurableTask.AzureServiceFabric/Service/FabricOrchestrationServiceController.cs
```diff
@@ -27,13 +27,14 @@ namespace DurableTask.AzureServiceFabric.Service
...
     {
-        private readonly IOrchestrationServiceClient orchestrationServiceClient;
+        readonly IOrchestrationServiceClient orchestrationServiceClient;
 
         /// <summary>
         /// Creates an instance of FabricOrchestrationServiceController for given OrchestrationServiceClient
         /// </summary>
         /// <param name="orchestrationServiceClient">IOrchestrationServiceClient instance</param>
         public FabricOrchestrationServiceController(IOrchestrationServiceClient orchestrationServiceClient)
+
```
cgillum, line 37: Accidental newline?

src/DurableTask.AzureServiceFabric/TaskMessageItem.cs
```diff
@@ -21,6 +21,7 @@ namespace DurableTask.AzureServiceFabric
     sealed class TaskMessageItem : IExtensibleDataObject
     {
         public TaskMessageItem(TaskMessage taskMessage)
+
```
cgillum, line 24: Accidental newline?

src/DurableTask.Core/OrchestrationInstance.cs
xperiandri: It will only affected equality and hash code implementation

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -106,25 +103,44 @@ public Table GetTableReference(string tableName)
             return new Table(this, this.tableClient, tableName);
         }
 
-        public Task<T> MakeBlobStorageRequest<T>(Func<OperationContext, CancellationToken, Task<T>> storageRequest, string operationName, string? clientRequestId = null) =>
-            this.MakeStorageRequest<T>(storageRequest, BlobAccountName, operationName, clientRequestId);
+        public Task<T> MakeBlobStorageRequest<T>(
+            Func<OperationContext, CancellationToken, Task<T>> storageRequest, string operationName, string? clientRequestId = null)
+        {
+            return this.MakeStorageRequest<T>(storageRequest, BlobAccountName, operationName, clientRequestId);
+        }
```
cgillum, line 110: I'm surprised to see these changes. What's the rule your tool is using to decide whether to use `=>` or full method definitions? I don't see any reason to change the syntax for these methods.

Also, this mechanism for wrapping parameters goes against the general style for DurableTask.AzureStorage, which is to put each parameter on its own line (or keep them all on the method declaration line).

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -684,26 +686,28 @@ async Task<PurgeHistoryResult> DeleteAllDataForOrchestrationInstance(Orchestrati
...
-                RowKey = string.Empty,
-                ETag = "*"
-            }));
+                this.InstancesTable.DeleteAsync(new DynamicTableEntity
+                {
+                    PartitionKey = orchestrationInstanceStatus.PartitionKey,
+                    RowKey = string.Empty,
+                    ETag = "*"
+                })
+            };
```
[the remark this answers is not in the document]
cgillum: TBH, I'm not a big fan of this change because I often choose to *not* use collection initializers. I don't think it helps with readability in many cases. This is one of those cases.

src/DurableTask.Core/EventHandlingOnContinueAsNew.cs
cgillum: Looks like the author forgot to finish the sentence. My suggestion is "All pending EventRaisedEvents will be carried over to the new execution".

src/DurableTask.Core/Exceptions/SubOrchestrationFailedException.cs
```diff
@@ -59,8 +51,8 @@ public SubOrchestrationFailedException(string reason, Exception innerException)
         /// <param name="version">Version of the Type Instance that experienced the error.</param>
         /// <param name="reason">The message that describes the error.</param>
         /// <param name="cause">The exception that is the cause of the current exception, or a null reference if no cause is specified.</param>
-        public SubOrchestrationFailedException(int eventId, int scheduleId, string name, string version, string reason,
-            Exception cause)
+        public SubOrchestrationFailedException(
+            int eventId, int scheduleId, string name, string version, string reason, Exception cause)
```
cgillum, line 55: I prefer we not adopt this style of parameter wrapping. Either keep the parameter list on the same line as the constructor/method name or put each on its own line.

src/DurableTask.Core/Exceptions/TaskFailedException.cs
```diff
@@ -73,14 +65,22 @@ public TaskFailedException(string reason, Exception innerException)
...
         }
 
+        /// <summary>
+        /// Initializes an new instance of the TaskFailedException class with a specified event id and error message
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="eventId">EventId of the error.</param>
+        /// <param name="message">The message that describes the error.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception, or a null reference if no inner exception is specified.</param>
+        TaskFailedException(int eventId, string message, Exception innerException) : base(eventId, message, innerException) { }
```
cgillum, line 82: What is this new constructor for? Is anyone calling it?

src/DurableTask.Core/Exceptions/TaskFailureException.cs
```diff
@@ -95,11 +82,11 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
-                Environment.NewLine,
-                FailureSource,
-                Details,
-                Message,
-                base.ToString());
+                                 Environment.NewLine,
+                                 FailureSource,
+                                 Details,
+                                 Message,
+                                 base.ToString());
```
cgillum, line 89: Please revert this change. It's not consistent with how we prefer to do argument wrapping.

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -15,7 +15,9 @@ namespace DurableTask.Core
 {
     using System;
     using System.Runtime.Serialization;
+
```
cgillum, line 18: Please revert these newlines, per a comment I made in a previous review iteration.

src/DurableTask.Core/History/GenericEvent.cs
```diff
@@ -31,8 +31,7 @@ public class GenericEvent : HistoryEvent
         /// </summary>
         /// <param name="eventId">The integer event id</param>
         /// <param name="data">The data for the event</param>
-        public GenericEvent(int eventId, string data)
-            : base(eventId)
+        public GenericEvent(int eventId, string data) : base(eventId)
```
cgillum, line 34: What is motivating the desire to put the `base()` call on the same line as the constructor? The existing pattern was based on older StyleCop rules and we tried to follow it pretty consistently.

src/DurableTask.Core/History/SubOrchestrationInstanceFailedEvent.cs
```diff
@@ -40,14 +40,10 @@ public SubOrchestrationInstanceFailedEvent(int eventId, int taskScheduledId, str
 
         /// <inheritdoc cref="SubOrchestrationInstanceFailedEvent(int, int, string?, string?, FailureDetails?)"/>
         public SubOrchestrationInstanceFailedEvent(int eventId, int taskScheduledId, string? reason, string? details)
-            : this(eventId, taskScheduledId, reason, details, null)
-        {
-        }
+            : this(eventId, taskScheduledId, reason, details, null) { }
```
cgillum, line 43: I'm also not a fan of the move to put the `{ }` on the previous line. We previously avoided doing this to be consistent with StyleCop rule enforcements. I personally prefer that we keep `{ }` on its own line so that it can be more clearly seen.

src/DurableTask.Core/Middleware/DispatchMiddlewarePipeline.cs
```diff
@@ -14,19 +14,18 @@
...
+    using System.Collections.Concurrent;
     using System.Threading.Tasks;
 
-    internal class DispatchMiddlewarePipeline
+    class DispatchMiddlewarePipeline
     {
-        readonly IList<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>> components =
-            new List<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>>();
+        readonly ConcurrentStack<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>> components =
+            new ConcurrentStack<Func<DispatchMiddlewareDelegate, DispatchMiddlewareDelegate>>();
```
cgillum, line 23: What's the purpose of making this a `ConcurrentStack`? There are no concurrency concerns with this code, as far as I'm concerned.

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -121,6 +118,7 @@ public abstract class OrchestrationContext
         /// <param name="retryOptions">Retry policies</param>
         /// <returns>Dynamic proxy that can be used to schedule the remote tasks</returns>
         public virtual T CreateRetryableClient<T>(RetryOptions retryOptions) where T : class
+
```
cgillum, line 121: Accidental newline?

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -288,13 +285,15 @@ public OrchestrationRuntimeStateDump GetOrchestrationRuntimeStateDump()
 
             return runtimeStateDump;
 #else
+#pragma warning disable IDE0022 // Use expression body for methods
```
cgillum, line 288: I'd prefer we remove this rule. It should be a choice that's left up by the developer and not enforced by the tooling.

src/DurableTask.Core/OrchestrationStateQuery.cs
```diff
@@ -52,12 +52,12 @@ public OrchestrationStateQuery()
         /// <summary>
         /// Gets the primary_filter, collection_of(secondary_filters) for the query
         /// </summary>
-        public Tuple<OrchestrationStateQueryFilter, IEnumerable<OrchestrationStateQueryFilter>> GetFilters()
+        public (OrchestrationStateQueryFilter Primary, IEnumerable<OrchestrationStateQueryFilter> Secondary) GetFilters()
```
cgillum, line 55: This is technically a breaking change since it was previously a `Tuple<T1, T2>` and now you're making it a `ValueTuple<T1, T2>`. Please revert.

samples/Correlation.Samples/TestOrchestrationHost.cs
```diff
@@ -33,7 +33,7 @@ public TestOrchestrationHost(AzureStorageOrchestrationServiceSettings settings)
         {
             try
             {
-                var service = new AzureStorageOrchestrationService(settings);
+                using var service = new AzureStorageOrchestrationService(settings);
```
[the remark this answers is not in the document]
xperiandri: Agree, I missed that. I'll add a suppression instead

src/DurableTask.AzureServiceFabric/Exceptions/InvalidInstanceIdException.cs
xperiandri: Well, it is what the analyzer recommends. Without it, binary serialization will not work at all, as far as I know.
And it looks like you do not use binary serialization

src/DurableTask.AzureServiceFabric/Service/FabricOrchestrationServiceController.cs
xperiandri: Yes, I'll remove

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
xperiandri: OK, I will take that into account.
I thought that you want expression-bodied members only for single-line methods (if the signature and body fit into a single line within display width)

src/DurableTask.Core/Exceptions/TaskFailedException.cs
xperiandri: Again, the analyzer requires that the derived exception has constructors for each of 4 default constructors of `Exception`

src/DurableTask.Core/FailureDetails.cs
xperiandri: Just missed, I'll fix

src/DurableTask.Core/History/GenericEvent.cs
xperiandri: The idea is that if you just pass a parameter you don't want to occupy a whole line for such unimportant thing

src/DurableTask.Core/OrchestrationContext.cs
xperiandri: Yes

src/DurableTask.Core/OrchestrationRuntimeState.cs
xperiandri: I did that but forgot to remove this pragma

src/DurableTask.Core/OrchestrationStateQuery.cs
xperiandri: What if we make this method obsolete and add a new one. So no breaking change.
You are going to bump a version to 2.0 anyway

cgillum: The version bump to 2.0 is only for DurableTask.AzureStorage. This code is in DurableTask.Core. Also, the only consumer I found in my quick scan was the DurableTask.ServiceBus package, which isn't actively maintained. For those reasons, I think we should skip making this change.

cgillum:
Just a heads-up that I'll be on vacation for the next three weeks and won't be able to continue reviewing while I'm gone. You can try reaching out to some of the other repo maintainers if you want to continue moving this PR forward.

.editorconfig
```diff
@@ -0,0 +1,991 @@
...
+
+# Naming Rules
+# Constant fields must be PascalCase
+dotnet_naming_rule.constant_fields_must_be_pascal_case.severity = warning
+dotnet_naming_rule.constant_fields_must_be_pascal_case.symbols               = constant_fields
+dotnet_naming_rule.constant_fields_must_be_pascal_case.style = pascal_case
+# Public, internal fields must be PascalCase
+dotnet_naming_rule.non_private_readonly_fields_must_be_pascal_case.severity = warning
+dotnet_naming_rule.non_private_readonly_fields_must_be_pascal_case.symbols   = public_internal_fields
+dotnet_naming_rule.non_private_readonly_fields_must_be_pascal_case.style = pascal_case
```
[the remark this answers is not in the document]
davidmrdavid: I agree that the spacing for the `=` is a bit weird here, and I don't fully grok what caused it. I think it might be best to just let the editor enforce a default one space.

davidmrdavid:
Hi @xperiandri, thank you for this change. I'm a fan of having an `.editorconfig` moving forward.
Just a thought: is it possible to split this PR into smaller changes? I wonder if we can have multiple PR rounds of editorconfig policities when enforced, which would make reviewing and accepting these changes a lot more manageable. Is that possible?

xperiandri:
@davidmrdavid, in my opinion, the main point is to define rules that you accept
In theory, it is possible, but you can notice that changes are primarily caused by 3-4 policies.
If you want to go one policy by one, that can be done

davidmrdavid:
@xperiandri - I think that, given the size of this change (200+ files), breaking it down into more atomic changes (one policy at a time) will help us review it faster / validate that the policies are doing the right thing. Right now it's a lot to keep track of, in my opinion. So my vote is to split this into smaller PRs. Otherwise, I worry this will take a long time to review and be de-prioritized.

I'll let others chime in about the strategy. Any thoughts: @amdeel, @bachuv?

xperiandri:
OK, I agree. And I will try to do that.
But please confirm that all the values of policies satisfy you.

davidmrdavid:
@xperiandri:

I assume you mean to review the contents of `.editorconfig`. I did take a glance at them, but frankly I can't say with certainty that I agree with all of them until I see them in action: there's just too many policies for me to fully grok how they'll interact with each other and the rest of the code.

Just checking: how did you come to this set of policies? Are these originating from some .NET-wide set of conventions or something like that? If so, it would be good if you could post your source, for our reference.

So, to summarize: I can't fully predict how these `.editorconfig` policies will look like. If they're coming from some community-wide .NET standard of config, then I think it's fair to assume we'll be ok with adopting most of them. That said, enabling them all at once in a single PR is going to make it really difficult for us to review the change, so we would benefit from several smaller PRs, with progressively more policies enabled, to help us keep the change at an understandable level. Hope that makes sense, I realize it's a fair bit of work, but hopefully you understand that this PR is rather large at the moment.

xperiandri:
> That said, enabling them all at once in a single PR is going to make it really difficult for us to review the change

I will leave them all but comment.
Then we will uncomment them one by one

xperiandri:
> Just checking: how did you come to this set of policies?

Just set values as I want to see code in my company. Using method of experiments and mistakes fixing

xperiandri:
> Are these originating from some .NET-wide set of conventions or something like that?

What was not aligned with my point of view I adjusted for your repository reality.
Where something was inconsistent I asked above.

xperiandri:
I've opened separate PRs for non-related fixes. Please review

xperiandri:
After they are merged I am ready to create a new smaller PR for .gitconfig

davidmrdavid:
Thanks @xperiandri, we'll give them a look as soon as we can. @cgillum  should also be back from a short break starting next week, so ideally he'll also be able to give a review to these smaller PRs.

PR #770, opened 2022-07-17
Improved performance of PerformanceHeartbeat's ToString() call

saguiitay:
This PR improved the performance of `PerformanceHeartbeat.ToString()` method by avoiding calls to `StringBuilder.Append(String.Join())`. Instead, we should call `StringBuilder.Append()` in a loop, thus avoiding the allocation of the intermediate strings.

### DIFF
--- a/src/DurableTask.AzureStorage/Monitoring/PerformanceHeartbeat.cs
+++ b/src/DurableTask.AzureStorage/Monitoring/PerformanceHeartbeat.cs
@@ -67,14 +67,38 @@ namespace DurableTask.AzureStorage.Monitoring
             sb.Append(nameof(this.PartitionCount)).Append(": ").Append(this.PartitionCount).Append(", ");
             if (this.ControlQueueLengths != null)
             {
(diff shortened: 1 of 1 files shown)


cgillum:
While I appreciate performance improvements in the code, I wonder if this might be a little overkill? `string.Join()` already uses a string builder, so the number of allocations we're saving seems quite limited. It's also not performance sensitive code, and it's suddenly become harder to read because of the extra verbosity.

Can you elaborate a bit more on why you wanted to make this optimization?

saguiitay:
Using `string.Join` results in the allocation of a new StringBuilder. In addition, it results in the allocation of the intermediate string (the result of the join).
This indeed is probably not in the hot-path, but it's a low hanging fruit, especially if I'll start gather statistics at a high frequency (I'm currently planning on gathering them every 5 minutes)

davidmrdavid:
@saguiitay - thanks for your contribution! Like @cgillum, I'm also a bit concerned about the trade-off here between simplicity and performance. I think that two things would help strengthen the case for incorporating this change.

(1) Potentially organizing this code onto a well-commented helper method, which could improve readability.
(2) Data showing the performance benefits of this change. Any statistics during DF execution of how much this helps would help us greatly.

Thanks!

PR #775, opened 2022-07-22
Ignore JsonConvert.DefaultSettings in serialization logic

davidmrdavid:
Would fix: https://github.com/Azure/azure-functions-durable-extension/issues/2135

**Problem:** Customers who change `Newtonsoft`'s `JsonConvert.DefaultSettings` object may affect DurableTask's own serialization logic. This has the ability cause stuck orchestrators.

**DF-specific Example:**

Assume a customer has the following block of code in their application:
```C#
JsonSerializerSettings globalSettings = new()
{
    DefaultValueHandling = DefaultValueHandling.Ignore,
};

JsonConvert.DefaultSettings = () => globalSettings;
```

This will make `Newtonsoft`'s `JsonConvert.SerializeObject(...)` drop all `0`-valued fields on an object when serializing to JSON. For example, a message with field `TaskScheduledId = 0` will have its `TaskScheduledId` field omitted in the resulting JSON-string.

This can cause issues during replay because, when de-serializing the message, the TaskScheduledId field will be missing and thus its value will be whatever the constructor sets as the default. As of today, that default is `-1` which in turn causes the orchestrator replay to get stuck.

**Root Cause:**

According to [these](http://james.newtonking.com/archive/2013/05/08/json-net-5-0-release-5-defaultsettings-and-extension-data) release notes, `JsonConvert.DefaultSettings` is designed to affect all invocations to `JsonConvert.SerializeObject` and `JsonConvert.DeserializeObject` across a project and its dependencies.

To avoid this behaviour, serialization-sensitive packages like ours need to instantiate their own `JsonSerializer` object using `JsonSerializer.Create(...)` and serialize using that.

**Minimal reproducer:**

The following `ConsoleApp` snippet shows how global serialization settings affect all invocations of `JsonConvert`, _even those that specify a `settings` parameter_ explicitly. In this latter case, all default settings that aren't explicitly specified in the `settings` parameter will be applied.

```C#
using Newtonsoft.Json;
using System.Text;

JsonSerializerSettings globalSettings = new()
{
    DefaultValueHandling = DefaultValueHandling.Ignore,
};

JsonConvert.DefaultSettings = () => globalSettings;

var settings = new JsonSerializerSettings
{
    DefaultValueHandling = DefaultValueHandling.Include
};

var settings2 = new JsonSerializerSettings
{
    TypeNameHandling = TypeNameHandling.Objects,
};

// set EventId to 0 to see if it gets dropped
Message messageData = new Message(0, "payload");



string out1 = JsonConvert.SerializeObject(messageData, settings);
string out2 = JsonConvert.SerializeObject(messageData, settings2);

var serializer = JsonSerializer.Create(settings);
StringBuilder sb = new StringBuilder();
StringWriter sw = new StringWriter(sb);

serializer.Serialize(sw, messageData);
var out3 = sw.ToString();

Console.WriteLine(out1); // outputs - {"Name":"payload","EventId":0}, contains EventId=0 because we explicitely contradict the global settings
Console.WriteLine(out2); // outputs - {"$type":"Message, JSONSettingsRepro","Name":"payload"}, no EventId field because we didn't explicitely contradict the global settings
Console.WriteLine(out3); // outputs - {"Name":"payload","EventId":0} because the third approach ignores the global settings.


// message to serialize
[DataContract]
class Message
{
    public Message(int eventId, string name)
    {
        this.Name = name;
        this.EventId = eventId;
    }

    public Message()
    { }

    [DataMember]
    public string Name { get; private set; }

    [DataMember]
    public int EventId { get; private set; }
}
```

**This PR:**
This PR changes every reference to `JsonConvert.SerializeObject` and `JsonConvert.DeserializeObject` in DTFx.Core and DTFx.AzureStorage to instead utilize custom `JsonSerializer` objects, thus ignoring global settings`.

New utilities have been introduced in DTFx.Core and DTFx.AzureStorage to do this with minimal code repetition. However, the utilities are repeated in both packages to avoid exporting these helpers to end-users.

A test was also added.

**Assumptions:**
1. I didn't apply this fix on our pre-existing tests because it seemed overkill.
2. I didn't apply this fix to other projects in this repo (servicebus backend, emulator, redis) as I think those need to be handled separately.

**Next steps:**
1. We may want to notify the Azure Functions Host of this edge case, as a quick code search shows that they're not protected against it.
2. We may want to apply this change in Netherite, MSSQL, and the servicebus backend.
3. Let's be careful with this change, it affects all our serialization paths after all :-)

### DIFF
--- a/src/DurableTask.AzureStorage/Utils.cs
+++ b/src/DurableTask.AzureStorage/Utils.cs
@@ -209,5 +209,34 @@ namespace DurableTask.AzureStorage
         {
             return DeserializeFromJson<T>(serializer, jsonString);
         }
+
+        /// <summary>
+        /// Deserialize a JSON-string into an object of type T
+        /// </summary>
+        /// <param name="jsonString">The JSON-string to deserialize.</param>
+        /// <param name="type">The expected de-serialization type.</param>
+        /// <returns></returns>
+        public static object DeserializeFromJson(string jsonString, Type type)
+        {
+            return DeserializeFromJson(serializer, jsonString, type);
+        }
+
+        /// <summary>
+        /// Deserialize a JSON-string into an object of type T
+        /// </summary>
+        /// <param name="serializer">The serializer whose config will guide the deserialization.</param>
+        /// <param name="jsonString">The JSON-string to deserialize.</param>
+        /// <param name="type">The expected de-serialization type.</param>
+        /// <returns></returns>
+        public static object DeserializeFromJson(JsonSerializer serializer, string jsonString, Type type)
+        {
+            object obj;
+            using (var reader = new StringReader(jsonString))
+            using (var jsonReader = new JsonTextReader(reader))
+            {
+                obj = serializer.Deserialize(jsonReader, type);
+            }
+            return obj;
+        }
     }
 }
--- a/src/DurableTask.AzureStorage/Utils.cs
+++ b/src/DurableTask.AzureStorage/Utils.cs
@@ -180,5 +180,34 @@ namespace DurableTask.AzureStorage
             var jsonStr = stringBuilder.ToString();
             return jsonStr;
         }
+
+        /// <summary>
+        /// Deserialize a JSON-string into an object of type T
+        /// </summary>
+        /// <typeparam name="T">The type to deserialize the JSON string into.</typeparam>
+        /// <param name="serializer">The serializer whose config will guide the deserialization.</param>
+        /// <param name="jsonString">The JSON-string to deserialize.</param>
+        /// <returns></returns>
+        public static T DeserializeFromJson<T>(JsonSerializer serializer, string jsonString)
+        {
+            T obj;
+            using (var reader = new StringReader(jsonString))
+            using (var jsonReader = new JsonTextReader(reader))
+            {
+                obj = serializer.Deserialize<T>(jsonReader);
+            }
+            return obj;
+        }
+
+        /// <summary>
+        /// Deserialize a JSON-string into an object of type T
+        /// </summary>
+        /// <typeparam name="T">The type to deserialize the JSON string into.</typeparam>
+        /// <param name="jsonString">The JSON-string to deserialize.</param>
+        /// <returns></returns>
+        public static T DeserializeFromJson<T>(string jsonString)
+        {
+            return DeserializeFromJson<T>(serializer, jsonString);
+        }
     }
 }
--- a/src/DurableTask.AzureStorage/Utils.cs
+++ b/src/DurableTask.AzureStorage/Utils.cs
@@ -169,7 +169,7 @@ namespace DurableTask.AzureStorage
         /// Serialize some object payload to a JSON-string representation.
         /// This utility is resilient to end-user changes in the DefaultSettings of Newtonsoft.
         /// </summary>
-        /// <param name="serializer">The DefaultJsonSerializer to use.</param>
+        /// <param name="serializer">The serializer to use.</param>
         /// <param name="payload">The object to serialize.</param>
         /// <returns>The JSON-string representation of the payload</returns>
         public static string SerializeToJson(JsonSerializer serializer, object payload)
@@ -188,7 +188,7 @@ namespace DurableTask.AzureStorage
         /// This utility is resilient to end-user changes in the DefaultSettings of Newtonsoft.
         /// </summary>
         /// <typeparam name="T">The type to deserialize the JSON string into.</typeparam>
-        /// <param name="serializer">The DefaultJsonSerializer whose config will guide the deserialization.</param>
+        /// <param name="serializer">The serializer whose config will guide the deserialization.</param>
         /// <param name="jsonString">The JSON-string to deserialize.</param>
         /// <returns></returns>
         public static T DeserializeFromJson<T>(JsonSerializer serializer, string jsonString)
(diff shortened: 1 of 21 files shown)


src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -148,5 +151,98 @@ public static string GetTargetClassName(this string s)
             int index = s.IndexOf('+');
             return s.Substring(index + 1, s.Length - index - 1);
         }
+
+        private static readonly JsonSerializer serializer = JsonSerializer.Create();
```
cgillum, line 155: Let's rename this to `DefaultJsonSerializer` and move it to the top of the class, which is where fields should go.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -62,6 +62,90 @@ public static class Utils
             Binder = new PackageUpgradeSerializationBinder()
 #endif
         };
+        private static readonly JsonSerializer objectJsonSerializer = JsonSerializer.Create(ObjectJsonSettings);
```
cgillum, line 65:
```suggestion
        private static readonly JsonSerializer DefaultObjectJsonSerializer = JsonSerializer.Create(ObjectJsonSettings);
```

And please move this up with the other `static readonly` fields.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -62,6 +62,90 @@ public static class Utils
...
+
+
+        private static readonly JsonSerializer serializer = JsonSerializer.Create();
```
cgillum, line 68:
```suggestion
        private static readonly JsonSerializer DefaultSerializer = JsonSerializer.Create();
```
And please move this up with the other `static readonly` fields.

src/DurableTask.Core/TraceContextBase.cs
```diff
@@ -42,8 +43,12 @@ static TraceContextBase()
                 PreserveReferencesHandling = PreserveReferencesHandling.Objects,
                 ReferenceLoopHandling = ReferenceLoopHandling.Serialize,
             };
+
+            serializer = JsonSerializer.Create(CustomJsonSerializerSettings);   
         }
-       
+
+        private static readonly JsonSerializer serializer;
```
cgillum, line 50: Please move this to the top of the class, above the constructors.

test/DurableTask.Core.Tests/CommonTests.cs
```diff
@@ -37,6 +38,38 @@ public void DateTimeExtensionsIsSetTest()
...
+            // serialize with global settings, and then with the serialization utility
+            var jsonStrFromGlobalSettings = JsonConvert.SerializeObject(messageData);
+            var jsonStrFromUtils = Utils.SerializeToJson(messageData);
+
+            // global serializer is expected to drop EventId, but the utility serializer doesn't.
+            Assert.IsFalse(jsonStrFromGlobalSettings.Contains("EventId"));
+            Assert.IsTrue(jsonStrFromUtils.Contains("EventId"));
+
+            // restore default settings
+            JsonConvert.DefaultSettings = previousDefaultSettings;
```
cgillum, line 69: I suggest putting this in a `finally` block to ensure we restore default settings even if the test throws an exception midway through.

test/DurableTask.Core.Tests/TestUtils.cs
```diff
@@ -29,5 +30,27 @@ public static string GenerateRandomString(int length)
...
+        public class DummyMessage
+        {
+            public DummyMessage(int eventId, string name)
+            {
+                this.Name = name;
+                this.EventId = eventId;
+            }
+
+            public DummyMessage()
+            { }
```
cgillum, line 47: If you want to make this more representative of the actual issue you're fixing, consider initializing `EventId` to -1, or some other non-zero value, and then testing the actual value in your test case.

cgillum (commented):
Initial feedback, mostly focused on code style.

test/DurableTask.Core.Tests/CommonTests.cs
davidmrdavid: Good idea, I've added a try-finally block here.

src/DurableTask.Core/TraceContextBase.cs
davidmrdavid: done!

src/DurableTask.AzureStorage/Utils.cs
davidmrdavid: I moved it right below the `public static readonly` fields. Let me know if that's not quite right.

test/DurableTask.Core.Tests/TestUtils.cs
davidmrdavid: I was thinking about that but, in my mind, the root problem starts as soon as the serializer decides to drop the `EventId` field. I'm less concerned with the de-serialization step, which is what this addition would test, as by that point we've already broken the expectation we have regarding the input string.

I'm willing to modify this test if you feel strongly about it, but I wanted to note down my reasoning for leaving it as is.

src/DurableTask.Core/Common/Utils.cs
davidmrdavid: I applied the renaming, but I was unsure about your suggestion on moving this further up, as it seems to me that all the fields preceding `DefaultSerializer` are already `static readonly`, and near the top of the class. Am I missing something?

src/DurableTask.Core/Common/Utils.cs
davidmrdavid: I have the same question as in another comment, which I'll paste here:

I applied the renaming, but I was unsure about your suggestion on moving this further up, as it seems to me that all the fields preceding DefaultSerializer are already static readonly, and near the top of the class. Am I missing something?

cgillum (approved):
LGTM! Nice work getting to the bottom of this tough issue!

src/DurableTask.Core/Common/Utils.cs
cgillum: I think I misread the code, thinking that the line above this was a method when really it was an object initializer for other field (I should make more use of the github.dev PR experience so I can better see the full context of these changes).

kirk-marple:
Awesome @davidmrdavid.  Great to have this fix see the light of day.

PR #777, opened 2022-07-24
Added GetFiltersTuple to return ValueTuple instead of allocatable Tuple

xperiandri:


src/DurableTask.Core/OrchestrationStateQuery.cs
```diff
@@ -84,6 +84,17 @@ public OrchestrationStateQuery()
                 }
             }
 
+            return (primaryFilter, secondaryFilters);
+        }
+
+        /// <summary>
+        /// Gets the primary_filter, collection_of(secondary_filters) for the query
+        /// </summary>
+        [Obsolete("Use GetFiltersTuple() instead")]
```
jviau, line 93: I think adding `Obsolete` tag is still a breaking change, as you are introducing a new warning that may fail consumer builds.

xperiandri: What would you recommend instead?

torosent:

PR #778, opened 2022-07-24
Fixed race condition for DispatchMiddlewarePipeline detected by Parallel Checker

xperiandri:


sebastianburckhardt:
It is unclear to me why there would be a race here.

The current code assumes all pipeline stages are added before the pipeline is run. That seems to be consistent with the useage I have seen.

xperiandri:
Maybe, that is just what my analyzer told me.
Anyway, the collection was selected incorrectly in my opinion. As the Stack is the most appropriate one.
And the other similar class uses concurrent collection for similar purposes.

torosent:

PR #779, opened 2022-07-24
Fixed race condition warning in NonBlockingCountdownLock detected by Parallel Checker

xperiandri:


sebastianburckhardt:
Thanks @xperiandri, adding a `volatile` keyword is definitely the right thing to do here.

This is a one-line change, but the PR contains many whitespace changes. Could you revert those? Not sure if there is an official policy about whitespace in this repo, but in my experience fixing whitespace is usually not worth the extra noise and potential merge conflicts.

xperiandri:
We will apply policy in my other PR #768. I will make it smaller after #779 #777 #778 #780

xperiandri:
So I would recommend to accept this one completely with whitespaces

torosent:

PR #782, opened 2022-08-09
Make ReflectionBasedTaskActivity.InvokeActivity Public to Match Overload

aueloka:
This change simply exposes the generic overload of `ReflectionBasedTaskActivity.InvokeAsync`, the overloaded version is already public so this simply achieves parity.

### DIFF
--- a/src/DurableTask.Core/ReflectionBasedTaskActivity.cs
+++ b/src/DurableTask.Core/ReflectionBasedTaskActivity.cs
@@ -161,7 +161,12 @@ namespace DurableTask.Core
             return MethodInfo.Invoke(ActivityObject, inputParameters);
         }
 
-        private object InvokeActivity(object[] inputParameters, Type[] genericTypeParameters)
+        /// <summary>
+        /// Invokes the target method on the activity object with supplied parameters
+        /// </summary>
+        /// <param name="inputParameters">The methods' input parameters.</param>
+        /// <param name="genericTypeParameters">The methods' generic parameters.</param>
+        public virtual object InvokeActivity(object[] inputParameters, Type[] genericTypeParameters)
         {
             if (genericTypeParameters.Any())
             {


aueloka:
@davidmrdavid , @cgillum . Would it be possible for someone to help review this?

davidmrdavid:
Thanks @aueloka. This seems like a reasonable change to me, but I'll defer to @cgillum or @amdeel as they oversee this codebase more often.

jviau:
Adding `virtual` to a method is _technically_ a binary breaking change. But it often has no effect on C# - but it theoretically could.

> ❌ DISALLOWED: Adding the [virtual](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/virtual) keyword to a member
While this often is not a breaking change because the C# compiler tends to emit [callvirt](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.callvirt) Intermediate Language (IL) instructions to call non-virtual methods (callvirt performs a null check, while a normal call doesn't), this behavior is not invariable for several reasons:
C# is not the only language that .NET targets.
The C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and is probably not null (such as a method accessed through the [?. null propagation operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)).
Making a method virtual means that the consumer code would often end up calling it non-virtually.

https://docs.microsoft.com/en-us/dotnet/core/compatibility/

cgillum:
From the description, it sounds like adding `virtual` to an _existing_ member might be breaking?  I was thinking that adding a *new* member as virtual should be non-breaking, though I haven't done the work to confirm this.

aueloka:
@cgillum , @jviau , this API is not yet part of the public SDK. It was added after the last release. So, I think this wouldn't be breaking if they form part of the same release. Also, as pointed out, it was previously private, so there wouldn't be any dependencies on it anyway.

aueloka:
> From the description, it sounds like adding `virtual` to an _existing_ member might be breaking? I was thinking that adding a _new_ member as virtual should be non-breaking, though I haven't done the work to confirm this.

@cgillum , is this a blocker for this PR? Or is it okay to merge given that the changes could follow the same release?

cgillum:
@aueloka good point. In that case, no blockers from my POV.

cgillum (approved):

cgillum:
@davidmrdavid let's include this PR in the upcoming DurableTask.Core release.

PR #786, opened 2022-08-18
Fix Retry Proxy Bug Causing Test Failures

aueloka:
Bug fix for RetryProxy.


cgillum:
There seems to be an issue with the CI timing out. One container in particular is repeatedly failing with various problems, including inability to establish connections with nuget.org. I'm going to go ahead and proactively merge this PR since AFAICT, these problems don't seem to be related to this PR. With luck, the issue will resolve itself when the CI runs against subsequent PRs.

cgillum (approved):

PR #787, opened 2022-08-22
Add support for cascading operations (terminate, suspend, resume).

Julio-Arroyo:
This is intended to be a proof-of-concept demo of how a simple, functional, and performant design of a cascading operations feature.

It exposes a new optional parameter so that operations (which at the moment include terminate, suspend, and resume), can "cascade down" or "fan-out" from the targeted orchestration to all suborchestrations.

The most important elements of this design include using history events for creating suborchestrations to maintain a collection of strings consisting of the instance IDs of suborchestrations. Therefore, when an event for terminate, suspend, or resume comes along, we can use the client operations to send the corresponding message to the suborchestrations.


davidmrdavid:
Thanks @Julio-Arroyo :-)

Same as in my comment here: https://github.com/Azure/azure-functions-durable-extension/pull/2256#issuecomment-1223465259

It would be great if you could pull in the latest `main` into this PR, so that this PR is more easily mergeable :-) . Thanks!

PR #790, opened 2022-08-24
Enable redacting user-code exception messages

cgillum:
### Context
DTFx code is often hosted in Azure App Service and Azure Functions. When running on these platforms, certain framework traces are captured into an internal log capturing service. One of the things logged by DTFx is exception messages for activity failures. These exception messages are often generated by user code, which may contain sensitive information (PII, secrets, etc.). 

### Change
As a safety mechanism, this PR will redact exception messages that come from activity failures. This redaction is enabled by default when hosted in an Azure App Service environment, which we infer by checking for the `WEBSITE_SITE_NAME` environment variable. This policy of redaction can be overridden by setting a `DTFX_REDACT_USER_CODE_EXCEPTIONS` environment variable to `1` or `true`.

This PR also adds additional `#nullable enable` compiler directives, which is something we routinely do as we touch the code to help make it safe from unexpected null-ref exceptions.

### Testing
I tested the exception redaction logic using the following LINQPad script:

```csharp
try
{
	Foo();
}
catch (Exception e)
{
	Console.WriteLine(GetSafeExceptionDetails(e, false));
	Console.WriteLine();
	Console.WriteLine(GetSafeExceptionDetails(e, true));
}

static string GetSafeExceptionDetails(Exception exception, bool redact)
{
	if (redact)
	{
		// Redact the exception message since its possible to contain sensitive information (PII, secrets, etc.)
		// Exception.ToString() code: https://referencesource.microsoft.com/#mscorlib/system/exception.cs,e2e19f4ed8da81aa
		var sb = new StringBuilder(capacity: 1024);
		sb.Append(exception.GetType().FullName).Append(": ").Append("[Redacted]");
		if (exception.InnerException != null)
		{
			// Recursive
			sb.AppendLine().Append(" ---> ").AppendLine(GetSafeExceptionDetails(exception.InnerException, redact));
			sb.Append("   --- End of inner exception stack trace ---");
		}

		sb.AppendLine().Append(exception.StackTrace);
		return sb.ToString();
	}
	else
	{
		return exception.ToString();
	}
}

static void Foo()
{
	try
	{
		Bar();
	}
	catch (Exception e)
	{
		throw new ApplicationException("Oh no!", e);
	}
}

static void Bar()
{
	try
	{
		Baz();
	}
	catch (Exception e)
	{
		throw new Exception("Umm...", e);
	}
}

static void Baz()
{
	throw new InvalidOperationException("Kah-BOOOOM!!!");
}
```

The results when redaction IS NOT enabled:
```
System.ApplicationException: Oh no!
 ---> System.Exception: Umm...
 ---> System.InvalidOperationException: Kah-BOOOOM!!!
   at UserQuery.<Main>g__Baz|4_3() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 63
   at UserQuery.<Main>g__Bar|4_2() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 53
   at UserQuery.<Main>g__Bar|4_2() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 57
   at UserQuery.<Main>g__Foo|4_1() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 41
   at UserQuery.<Main>g__Foo|4_1() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 45
   at UserQuery.Main() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 4
```

This is what would be seen if redaction IS enabled:
```
System.ApplicationException: [Redacted]
 ---> System.Exception: [Redacted]
 ---> System.InvalidOperationException: [Redacted]
   at UserQuery.<Main>g__Baz|4_3() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 63
   at UserQuery.<Main>g__Bar|4_2() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 53
   at UserQuery.<Main>g__Bar|4_2() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 57
   at UserQuery.<Main>g__Foo|4_1() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 41
   at UserQuery.<Main>g__Foo|4_1() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 45
   at UserQuery.Main() in C:\Users\cgillum\AppData\Local\Temp\LINQPad7\_wrmpjfpn\otarjw\LINQPadQuery:line 4
```

### DIFF
--- a/src/DurableTask.Core/Logging/LogEvents.cs
+++ b/src/DurableTask.Core/Logging/LogEvents.cs
@@ -1279,13 +1279,13 @@ namespace DurableTask.Core.Logging
                 OrchestrationInstance instance,
                 string name,
                 TaskFailedEvent taskEvent,
-                Exception exception)
+                string exceptionDetails)
             {
                 this.InstanceId = instance.InstanceId;
                 this.ExecutionId = instance.ExecutionId;
                 this.Name = name;
                 this.TaskEventId = taskEvent.EventId;
-                this.Details = exception.ToString();
+                this.Details = exceptionDetails;
             }
 
             [StructuredLogField]
--- a/src/DurableTask.Core/Logging/LogEvents.cs
+++ b/src/DurableTask.Core/Logging/LogEvents.cs
@@ -1275,17 +1275,20 @@ namespace DurableTask.Core.Logging
 
         internal class TaskActivityFailure : StructuredLogEvent, IEventSourceEvent
         {
+            readonly Exception exception;
+
             public TaskActivityFailure(
                 OrchestrationInstance instance,
                 string name,
                 TaskFailedEvent taskEvent,
-                string exceptionDetails)
+                Exception exception)
             {
                 this.InstanceId = instance.InstanceId;
                 this.ExecutionId = instance.ExecutionId;
                 this.Name = name;
                 this.TaskEventId = taskEvent.EventId;
-                this.Details = exceptionDetails;
+                this.Details = exception.ToString();
+                this.exception = exception;
             }
 
             [StructuredLogField]
@@ -1318,7 +1321,7 @@ namespace DurableTask.Core.Logging
                     this.ExecutionId,
                     this.Name,
                     this.TaskEventId,
-                    this.Details,
+                    Utils.RedactUserCodeExceptions ? LogHelper.GetRedactedExceptionDetails(this.exception) : this.Details, // We have to be extra guarded about logging user exceptions to EventSource (ETW)
                     Utils.AppName,
                     Utils.PackageVersion);
         }
--- a/src/DurableTask.Core/Logging/LogHelper.cs
+++ b/src/DurableTask.Core/Logging/LogHelper.cs
@@ -10,20 +10,22 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-
+#nullable enable
 namespace DurableTask.Core.Logging
 {
     using System;
     using System.Collections.Generic;
+    using System.Text;
     using DurableTask.Core.Command;
+    using DurableTask.Core.Common;
     using DurableTask.Core.History;
     using Microsoft.Extensions.Logging;
 
     class LogHelper
     {
-        readonly ILogger log;
+        readonly ILogger? log;
 
-        public LogHelper(ILogger log)
+        public LogHelper(ILogger? log)
         {
             // null is okay
             this.log = log;
@@ -92,7 +94,6 @@ namespace DurableTask.Core.Logging
                 this.WriteStructuredLog(new LogEvents.DispatcherStarting(context));
             }
         }
-
 
         /// <summary>
         /// Logs that a work item dispatcher has stopped running.
@@ -360,7 +361,7 @@ namespace DurableTask.Core.Logging
         /// </summary>
         /// <param name="instanceId">The ID of the instance.</param>
         /// <param name="executionId">The execution ID of the instance, if applicable.</param>
-        internal void FetchingInstanceState(string instanceId, string executionId = null)
+        internal void FetchingInstanceState(string instanceId, string? executionId = null)
         {
             if (this.IsStructuredLoggingEnabled)
             {
@@ -567,8 +568,11 @@ namespace DurableTask.Core.Logging
         {
             if (this.IsStructuredLoggingEnabled)
             {
+                // This is a user-code exception which could contain sensitive information. Depending on the
+                // environment, some of the exception details may need to be redacted.
+                string exceptionDetails = GetSafeExceptionDetails(exception);
                 this.WriteStructuredLog(
-                    new LogEvents.TaskActivityFailure(instance, name, failedEvent, exception),
+                    new LogEvents.TaskActivityFailure(instance, name, failedEvent, exceptionDetails),
                     exception);
             }
         }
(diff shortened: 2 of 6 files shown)


src/DurableTask.Core/Logging/LogHelper.cs
```diff
@@ -639,9 +643,50 @@ internal void RenewActivityMessageFailed(TaskActivityWorkItem workItem, Exceptio
...
 
-        void WriteStructuredLog(ILogEvent logEvent, Exception exception = null)
+        void WriteStructuredLog(ILogEvent logEvent, Exception? exception = null)
         {
             this.log?.LogDurableEvent(logEvent, exception);
         }
+
+        static string GetSafeExceptionDetails(Exception? exception)
+        {
+            if (exception == null)
```
jviau, line 653: Tiny suggestion: I like using `is null` and `is not null` for null checks. There is minor difference: `== null` will resolve any operator overloads, so its implementation may change based on the type. `is null` is a pattern match against `null` itself. So it is more akin to `Object.ReferenceEquals(target, null);`

jviau (approved):
Any concerns on this being a breaking behavior change for AppService users?

src/DurableTask.Core/Logging/LogHelper.cs
```diff
@@ -567,8 +568,11 @@ internal void TaskActivityCompleted(OrchestrationInstance instance, string name,
         {
             if (this.IsStructuredLoggingEnabled)
             {
+                // This is a user-code exception which could contain sensitive information. Depending on the
+                // environment, some of the exception details may need to be redacted.
+                string exceptionDetails = GetSafeExceptionDetails(exception);
                 this.WriteStructuredLog(
-                    new LogEvents.TaskActivityFailure(instance, name, failedEvent, exception),
+                    new LogEvents.TaskActivityFailure(instance, name, failedEvent, exceptionDetails),
```
jviau, line 575: Can we stick with passing in the exception here and have this logic decide to extract safe exception details or not?

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1279,13 +1279,13 @@ internal class TaskActivityFailure : StructuredLogEvent, IEventSourceEvent
                 OrchestrationInstance instance,
                 string name,
                 TaskFailedEvent taskEvent,
-                Exception exception)
+                string exceptionDetails)
```
jviau, line 1282: Not sure if you want to consider this for this PR, but with `ILogger`, exception is an explicit parameter you pass in. This is so all the logger providers have access to the exception and can do their own logic with it. For example, app insights will by default send it to the `exceptions` table with some app insights specific formatting so it is presented in a structured way.

Other reasons is just that exception `.ToString()` is expensive - so by default many loggers will opt to only include minimal details and forgo the stack trace.

I actually find it weird DurableTask is calling `ToString()` on this at all. After this change, we should evaluate removing the ToString _altogether_. We can look at what other frameworks do and take those as examples. We can instead include an optional middleware to log thrown exceptions.

src/DurableTask.Core/Logging/LogHelper.cs
cgillum: Yeah, that's probably a better way to go, and may help me resolve your concern about breaking changes for AppService users.

src/DurableTask.Core/Logging/LogEvents.cs
cgillum: We do pass the exception as-is to `ILogger.Log`, which you can see in `LoggingExtensions.LogDurableEvent()`, so that capability is there.

In the past, I've always regretted it when we didn't log the `ToString()` output of exceptions, either because we miss out on critical information about the exception that's not contained in the `Message` property (StorageException from the Azure Storage SDK and SqlException from ADO.NET are good examples of this) or because not having the stack trace made it impossible to determine where exactly the problem occurred (especially with things like ArgumentNullException or NullReferenceException). I'm open to looking at other options, but my experience tells me that this is the best way to go unless there's a very specific reason not to. Exceptions being the exceptional case, I don't worry as much about the performance impact.

jviau: I am fine leaving it as is in this PR. For the future, we should consider making this customizable.

jviau (approved):

src/DurableTask.Core/Logging/LogHelper.cs
cgillum: For the sake of code consistency, I think I'll keep using `==` and `!=` for now, but I see the benefit of your suggestion. Let's make a point of doing this at least in the newer codebases (like the new out-of-proc .NET SDK).

PR #793, opened 2022-09-06
Remove duplicate PackageReference

jviau:
Started getting a warning for duplicate packages with a newer .NET SDK. Turns out this package is brought in at `DurableTask.props`.


cgillum:
Hmm...I thought I already did this? https://github.com/Azure/durabletask/pull/788

jviau:
You totally did, guess I was working on a stale branch.

PR #798, opened 2022-09-22
Fix logging for large queue message sizes

### LINKED ISSUE #797: DT.AzureStorage: Large message sizes are not correctly logged
Large messages (those that get uploaded to blob storage) appear to not have their sizes logged correctly. This applies to both the `SendingMessage` and `ReceivedMessage` logs. This needs to be fixed so that automation can be written that detects large message usage and flags for potential performance issues.

cgillum:
Fixes #797 

The existing traces for message sizes tracked the actual queue message sizes but didn't account for large message scenarios where the payload is offloaded to blob storage. It's really important that we log the actual payload sizes since the impact on performance is significant.


davidmrdavid (approved):
:shipit:

PR #800, opened 2022-09-26
Upgraded Newtonsoft and added ServiceBus Changes for MSI Auth on NET …

namanjainv:
…461 and greater


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -35,7 +35,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="7.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
```
cgillum, line 38: We can't accept this change because it will break existing Azure Functions users.

cgillum (changes requested):
Hi @namanjainv, can you reduce the scope of this PR to only touch Service Bus-related code? We can't accept the changes for DurableTask.AzureStorage, for example, because those will break existing customers in Azure Functions. We have fewer restrictions on the Service Bus-related projects.

src/DurableTask.ServiceBus/Common/Abstraction/ServiceBusAbstraction.cs
```diff
@@ -8,7 +8,7 @@ namespace DurableTask.ServiceBus.Common.Abstraction
     using System.Runtime.Serialization;
     using System.Threading.Tasks;
     using System.Xml;
-#if !NETSTANDARD2_0
+#if !(NETSTANDARD2_0 || NET461_OR_GREATER)
```
cgillum, line 11: Can all the `#if !(NETSTANDARD2_0 || NET461_OR_GREATER)` lines be removed? I think this conditional now covers everything.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
namanjainv: Removed changes from other packages. Have this change only in ServiceBus project

src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
```diff
@@ -13,10 +13,11 @@
     <PackageReference Include="Microsoft.Data.OData" version="5.8.5" />
     <PackageReference Include="Microsoft.Data.Services.Client" version="5.8.5" />
     <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" version="3.2.1" />
-    <PackageReference Include="Newtonsoft.Json" version="7.0.1" />
+    <PackageReference Include="Newtonsoft.Json" version="10.0.3" />
     <PackageReference Include="System.Spatial" version="5.8.5" />
     <PackageReference Include="WindowsAzure.ServiceBus" version="4.1.3" />
     <PackageReference Include="WindowsAzure.Storage" version="7.0.0" />
+    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="4.1.3" />
```
cgillum, line 20: Is it okay that you're now referencing both `WindowsAzure.ServiceBus` and `Microsoft.Azure.ServiceBus`?

cgillum: Related, do we need to continue to distinguish between this target (`net462`) and the target mentioned below this (`netstandard2.0`)? I'm wondering whether we can just target `netstandard2.0` instead of trying to maintain two different targets.

cgillum (commented):
Missing from this PR is a change to the package version. Normally for this package it comes from [/tools/DurableTask.props](https://github.com/Azure/durabletask/blob/a0ed8efefb880fda070e274ee170ad68dd53882a/tools/DurableTask.props#L42-L44). However, we're trying to instead keep version numbers in the package that we're actually versioning rather than having a common version configuration. Can you add the following markup to DurableTask.ServiceBus.csproj?

```xml
  <!-- Version Info -->
  <PropertyGroup>
    <MinorVersion>7</MinorVersion>
    <PatchVersion>0</PatchVersion>

    <FileVersion>$(VersionPrefix).0</FileVersion>
    <!-- FileVersionRevision is expected to be set by the CI. This is useful for distinguishing between multiple builds of the same version. -->
    <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevision)</FileVersion>
    <!-- The assembly version is only the major/minor pair, making it easier to do in-place upgrades -->
    <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
    <!-- This version is used as the nuget package version -->
    <Version>$(VersionPrefix)</Version>
  </PropertyGroup>
```

Note that I suggested this to be v2.7.0, but we may want to consider it to be 3.0.0 since we're making potentially breaking changes by upgrading dependencies. See some of my other comments too, which might impact the final version we go with.

src/DurableTask.ServiceBus/Common/Abstraction/ServiceBusAbstraction.cs
namanjainv: Made required changes with commit # d01cfb655657c00d5b1df8e16bfb5184a87fb76d

src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
namanjainv: Made required changes with commit # d01cfb655657c00d5b1df8e16bfb5184a87fb76d.

I removed the old library. Now we use only Microsoft.Azure.ServiceBus for all versions. Also made discussed changes to support all versions.

src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
```diff
@@ -6,27 +6,38 @@
...
+    <FileVersion>$(VersionPrefix).0</FileVersion>
+    <!-- FileVersionRevision is expected to be set by the CI. This is useful for distinguishing between multiple builds of the same version. -->
+    <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevision)</FileVersion>
+    <!-- The assembly version is only the major/minor pair, making it easier to do in-place upgrades -->
+    <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
+    <!-- This version is used as the nuget package version -->
+    <Version>$(VersionPrefix)</Version>
+  </PropertyGroup>
+	
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
```
cgillum, line 25: With your latest changes, is it possible to only target `netstandard2.0` instead of both `net462` and `netstandard2.0`? That would be preferred since it will reduce the size of the package.

namanjainv: I tried to remove net462 dependencies, but this started throwing multiple build-time errors. For example System.Transaction was not available. And I dont think we need to remove net462 from TargetFrameworks right?

cgillum (approved):
I'm good with these changes now, and the CI is passing.

cgillum:
Actually, it looks like the CI didn't run any of the Service Bus tests due to a filter which is no longer finding them. I'll need to update the CI to account for this.

namanjainv:
> Actually, it looks like the CI didn't run any of the Service Bus tests due to a filter which is no longer finding them. I'll need to update the CI to account for this.

Sounds good

cgillum:
I assume this was closed accidentally?

namanjainv:
Yup. My bad

src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
```diff
@@ -2,41 +2,42 @@
...
-    <PackageReference Include="Microsoft.Data.Services.Client" version="5.8.5" />
-    <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" version="3.2.1" />
-    <PackageReference Include="Newtonsoft.Json" version="7.0.1" />
-    <PackageReference Include="System.Spatial" version="5.8.5" />
-    <PackageReference Include="WindowsAzure.ServiceBus" version="4.1.3" />
-    <PackageReference Include="WindowsAzure.Storage" version="7.0.0" />
-    <Reference Include="System.Transactions" />
-  </ItemGroup>
+  <!-- Version Info -->
+  <PropertyGroup>
```
jviau, line 10: Can we start with this as a `-preview` version? If we are investing in ServiceBus, we might as well move to trace 2 SDKs.

DurableTask.sln
```diff
@@ -27,8 +27,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DurableTask.Core.Tests", "t
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DurableTask.Samples", "samples\DurableTask.Samples\DurableTask.Samples.csproj", "{BC204374-927C-4A48-8473-F5BF8C97EAEA}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DurableTask.Stress.Tests", "test\DurableTask.Stress.Tests\DurableTask.Stress.Tests.csproj", "{96980359-8003-4500-B97B-949A7FE01F31}"
```
jviau, line 30: Another PR has just gone in which changes up a lot of the same test code. Instead of removing, I believe it migrated this to use DurableTask.AzureStorage. Please ensure this PR will not undo those changes.

namanjainv: I restaged the main with my branch. Made sure Stress Tests are removed. Please review the changes when you find the time.

jviau:
@namanjainv to make sure I understand this change correctly: it looks like you are removing the `net461` TFM and all of its ifdef logic. Is the end result that the MSI features you want were behind netstandard2.0 ifdefs and thus not available when targeting netfx?

If the above is correct, this is really interesting that you did not have to _add_ any code, just remove that `net461` specifics. I am curious if there was a reason this was never added to net461 in the first place? @cgillum do you know why?

cgillum:
@jviau my memory is a bit fuzzy, but I think the main reason we originally had multiple TFMs was because DT.Core, DT.ServiceBus, and some other dependencies were targeting `net451`, which was not compatible with `netstandard2.0`. Since then, the older TFMs have been replaced with newer ones, but I think we simply never got around to just collapsing everything to `netstandard2.0`, like what is being done in this PR.

cgillum:
This change requires us to update our CI pipeline because the location of the test DLLs will change. Unfortunately, this CI was created before it was possible to use YAML files in the repo (and we haven't converted it to YAML yet), so we have to use the ADO UX to make these updates.

@namanjainv can you tell me where I'll be able to find the Service Bus test DLLs in the build output (paths relative to the repo root)?

namanjainv:
@cgillum,  The DLL is generated at "Test\DurableTask.ServiceBus.Tests\bin\Debug\netcoreapp3.1\DurableTask.ServiceBus.dll"

torosent:

PR #804, opened 2022-10-12
Propagating Orchestration Tags to TaskActivity

RamjotSingh:
Propagating OrchestrationRuntimeState to TaskActivity dispatcher middleware. This allows Tags and other historical events to be accessible from TaskActivity dispatcher middleware

### DIFF
--- a/src/DurableTask.Core/OrchestrationExecutionContext.cs
+++ b/src/DurableTask.Core/OrchestrationExecutionContext.cs
@@ -10,7 +10,6 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
-#nullable enable
 namespace DurableTask.Core
 {
     using System.Collections.Generic;
@@ -26,6 +25,6 @@ namespace DurableTask.Core
         /// Gets the orchestration tags
         /// </summary>
         [DataMember]
-        public IDictionary<string, string> OrchestrationTags { get; internal set; } = new Dictionary<string, string>();
+        public IReadOnlyDictionary<string, string> OrchestrationTags { get; internal set; }
     }
 }
--- a/src/DurableTask.Core/OrchestrationExecutionContext.cs
+++ b/src/DurableTask.Core/OrchestrationExecutionContext.cs
@@ -1,0 +1,31 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+#nullable enable
+namespace DurableTask.Core
+{
+    using System.Collections.Generic;
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// Context associated with the orchestration being executed.
+    /// </summary>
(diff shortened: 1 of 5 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -842,6 +842,7 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
 
             taskMessage.Event = scheduledEvent;
             taskMessage.OrchestrationInstance = runtimeState.OrchestrationInstance;
+            taskMessage.OrchestrationRuntimeState = runtimeState;
```
cgillum, line 845: The runtime state can potentially be huge, so we should never attach it to messages. Instead, we should attach the most limited amount of data we need for the scenario.

RamjotSingh: Created a new object and put just Tags in it for now. Made it available in both Orchestration and Activity.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -565,6 +565,11 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
             return isCompleted || continuedAsNew || isInterrupted;
         }
 
+        static OrchestrationExecutionContext GetOrchestrationExecutionContext(OrchestrationRuntimeState runtimeState)
+        {
+            return new OrchestrationExecutionContext { OrchestrationTags = runtimeState.Tags ?? new Dictionary<string, string>() };
```
cgillum, line 570: I wonder if it would be better to allow this property to be `null` to avoid unnecessary allocations. Alternatively, we should initialize it with a capacity of zero (`new Dictionary<string, string>(capacity: 0)`).

src/DurableTask.Core/OrchestrationExecutionContext.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// Context associated with the orchestration being executed.
+    /// </summary>
+    [DataContract]
+    public class OrchestrationExecutionContext
+    {
+        /// <summary>
+        /// Gets the orchestration tags
+        /// </summary>
+        [DataMember]
+        public IDictionary<string, string> OrchestrationTags { get; internal set; } = new Dictionary<string, string>();
```
cgillum, line 29: It looks like we're always assigning this property eternally at construction time, so the default initialization is being wasted. I suggest removing that. I also suggest we make this an `IReadOnlyDictionary<string, string>` since developers should never be adding to it themselves.
```csharp
public IReadOnlyDictionary<string, string>? OrchestrationTags { get; internal set; }
```

jviau (changes requested):
This looks good. But can you add a test case for the migration window? Where an existing activity does not have `OrchestrationExecutionContext` serialized - what does that look like in middleware? Ensure it doesn't cause any regressions.

src/DurableTask.Core/OrchestrationExecutionContext.cs
RamjotSingh: Made it IReadonlyDictionary but havent made it nullable since we will always assign value from public access standpoint

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
```diff
@@ -170,6 +177,40 @@ public async Task ActivityDispatcherMiddlewareContextFlow()
             Assert.AreEqual("01234567899876543210", output?.ToString());
         }
 
+        [TestMethod]
+        public async Task EnsureActivityDispatcherMiddlewareHasAccessToRuntimeState()
```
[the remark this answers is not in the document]
RamjotSingh: Done.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
RamjotSingh: Fixed.

jviau (approved):
Looks good!

cgillum (approved):
LGTM!

PR #808, opened 2022-10-24
Replace Service Bus Backend in Samples and Stress Tests

wsugarman:
Given the deprecation of the Service Bus backend, changes to the Durable Task Framework and its related packages should not be back-ported to the Service Bus package (unless explicitly desired). However, there are a number of project references to the Service Bus backend including samples, integration test, and stress tests. To avoid making future unnecessary changes to the Service Bus backend, I have created this PR where I isolate it from the other projects with the following changes:
- Replace Service Bus with Azure Storage in both the sample and stress tests projects
- Removal of dependencies on other projects from the Service Bus test project
    - This involved the duplication of one class, one method, and the inclusion of test orchestrations for compilation
- .NET Framework refactors including:
    - Use of `AutoGenerateBindingRedirects`
    - Removal of unused packages / dependencies
- Normalize indent to 2 spaces in `.csproj` files


cgillum (approved):
This is great - thank you doing this work!

PR #817, opened 2022-11-01
Fix wrong blob URL on orchestration inputs

cgillum:
An internal partner team noticed an issue where the Instances table gets populated with a wrong blob URL when an orchestration is started with a large input (one that doesn't fit into table storage). Specifically, the `Input` column/property was getting populated with the blob URL of the queue message instead of with a blob URL for the orchestration's input.  This was causing orchestration status queries to return the wrong input data.

Upon investigation, it looks like this issue has existed for a while but just went undetected. I was able to observe the problem by updating some existing unit tests to validate the reported orchestration input when large inputs are used.

~~The fix was to remove some code that was causing us to populate the `Input` column/property incorrectly.~~

EDIT: The originally submitted fix ended up creating a new blob, resulting in further duplication of large payloads. I decided this would not be a good fix as it would increase the performance penalty associated with large messages. I've submitted an updated fix which instead examines the previous payload and parses out the input contained within it. This allows us to keep the same number of blobs, ensuring there's no performance regression. The previous fix would only have worked for new orchestrations, but this new fix works for both old and new orchestrations.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -130,14 +130,7 @@ namespace DurableTask.AzureStorage
 
         internal static bool TryGetLargeMessageReference(string messagePayload, out Uri blobUrl)
         {
-            if (Uri.IsWellFormedUriString(messagePayload, UriKind.Absolute))
-            {
-                blobUrl = new Uri(messagePayload);
-                return true;
-            }
-
-            blobUrl = null;
-            return false;
+            return Uri.TryCreate(messagePayload, UriKind.Absolute, out blobUrl);
         }
 
         public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMessage, string queueName)
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -118,9 +118,9 @@ namespace DurableTask.AzureStorage
         /// <returns>Actual string representation of message.</returns>
         public async Task<string> FetchLargeMessageIfNecessary(string message)
         {
-            if (Uri.IsWellFormedUriString(message, UriKind.Absolute))
+            if (TryGetLargeMessageReference(message, out Uri blobUrl))
             {
-                return await this.DownloadAndDecompressAsBytesAsync(new Uri(message));
+                return await this.DownloadAndDecompressAsBytesAsync(blobUrl);
             }
             else
             {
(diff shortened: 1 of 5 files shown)


jviau (approved):
Super interesting this went on undetected for so long. I guess oversized messages are not common?

cgillum:
Oversized messages aren't necessarily uncommon. I think what's uncommon is folks using the status query to examine the contents of the input after the orchestration is started.

To be clear, everything works fine from a runtime perspective. The only thing that was broken was the orchestration status query results.

cgillum:
These are the tests that are currently failing, and likely need to be updated.
image

cgillum:
I changed my strategy for how to fix this issue after observing the new behavior through several of the "large message" tests. See the updated PR description for more details.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -118,24 +118,36 @@ public async Task<string> SerializeMessageDataAsync(MessageData messageData)
...
-                return await this.DownloadAndDecompressAsBytesAsync(new Uri(message));
+                return await this.DownloadAndDecompressAsBytesAsync(blobUrl);
             }
             else
             {
                 return message;
             }
         }
 
+        internal static bool TryGetLargeMessageReference(string messagePayload, out Uri blobUrl)
```
jviau, line 131: Can `Uri.TryCreate` be used here instead? Or is there a difference in behavior I am missing?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -534,7 +534,34 @@ async Task<OrchestrationState> ConvertFromAsync(OrchestrationInstanceStatus orch
 
             if (this.settings.FetchLargeMessageDataEnabled)
             {
-                orchestrationState.Input = await this.messageManager.FetchLargeMessageIfNecessary(orchestrationState.Input);
+                if (MessageManager.TryGetLargeMessageReference(orchestrationState.Input, out Uri blobUrl))
+                {
+                    string json = await this.messageManager.DownloadAndDecompressAsBytesAsync(blobUrl);
+                    if (blobUrl.AbsolutePath.EndsWith("ExecutionStarted.json.gz"))
```
jviau, line 540: Just so I am clear on the changes - this appears to be the only functional change? Checking if the JSON is the input itself or an `ExecutionStartedEvent` which contains the input?

jviau (approved):

src/DurableTask.AzureStorage/MessageManager.cs
cgillum: Ha, I didn't even notice that. Yeah, let me do that.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: Yes, that's correct. This is the only functional change: to determine whether this blob is the input itself or the event/message that contains the input.

PR #818, opened 2022-11-02
Add utilities for estimating the size of orchestration states and history events

sebastianburckhardt:
Computes an approximation of the number of bytes consumed.

### DIFF
--- a/src/DurableTask.Core/Common/SizeUtils.cs
+++ b/src/DurableTask.Core/Common/SizeUtils.cs
@@ -1,0 +1,150 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using DurableTask.Core;
+    using DurableTask.Core.History;
+
+    /// <summary>
+    /// Functionality for estimating the number of bytes of memory taken up by some commonly used objects.
+    /// </summary>
+    public class SizeUtils
```
jviau, line 25: Does this need to be public?

amdeel: It doesn't because we should be able to use this only in Core to store the accurate history size. That value will then be able to be used by whoever has access to the OrchestrationRuntimeState.

PR #825, opened 2022-11-16
OpenTelemetry instrumentation for sub-orchestrations

bachuv:
This PR updates OTel instrumentation for sub-orchestrations. There are now separate spans for scheduling a sub-orchestration and executing a sub-orchestration.

**Sample orchestration**
```
class HelloSubOrch : TaskOrchestration<string, string>
{
    {
        string result = "";
        result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
        result += await context.ScheduleTask<string>(typeof(SayHello), "Tokyo");
        Task<string> fanOut = context.CreateSubOrchestrationInstance<string>(typeof(HelloFanOut), null);
        result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
        result += await fanOut;

        return result;
    }
}
```
**Related Gantt charts**
Application Insights
image
Zipkin
image


jviau:
Couple questions regarding the Gantt diagram:

1. Do we need the index on the dependency telemetry? The `(#1)`, `(#2)` suffixes? The fact they are different spans (with different SpanId's) should already correctly differentiate them. What value do we get out of them?
    - We should be referencing the OTel spec for how to name spans anyways. So, we may have more changes then just this.
2.  Why is the duration for all the sub orchestration dependency (client) telemetry 0? This should be the duration between first enqueueing the sub orchestration and when you are finally replayed with the completed sub orchestration.
3. Are the `TaskActivity` server spans coming in a later PR?

cgillum:
> Do we need the index on the dependency telemetry? The (#1), (#2) suffixes? The fact they are different spans (with different SpanId's) should already correctly differentiate them. What value do we get out of them?

These are the task IDs. They are the same value as the `dtfx.task_id` properties. It's not critical for them to be in the name. We should definitely *not* include them when their values are `-1` though (-1 means that task ID doesn't apply).

jviau:
I think we should leave task ID off the span name. Here is some OTel guidance on span naming: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span With the relevant part being:

> The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable.

To me, including task ID starts encroaching into "individual Span instances" naming and not "class of Spans".

cgillum:
I'm fine with that. However, it begs another question: should the activity spans be named after the activity type being invoked or should it simply be "Activity"?

jviau:
Looking at the HTTP request example, I would say a third option: it should be named with the task name it was enqueued with. But then we also have version, and do we want to also include task type in the span name? We have three potential pieces to include: `taskType`, `taskName`, and `taskVersion`. Which ones do we want to include and in what format? Some ideas:

All three: `{taskType}|{taskName}(|{taskVersion})?` (so when `taskVersion` is `null`, we leave it off)
Type and name: `{taskType}|{taskName}` - version will still be an attribute.
Just name: `{taskName}` - version and type will be an attribute.

I am using `|` as a separator, but open to other separators. I wanted something that would not be in a type name (of any language), but is not confused with other spans (ie: http uses `/`, so `orchestration/myOrchestration/2` looks like a HTTP route span), and also is not going to cause encoding or support problems on various telemetry services.

cgillum:
I'm fine with `|`, though I find it's often hard to read. I do prefer `/` but I understand that you want to avoid confusing this with HTTP. Maybe `||`? I also like your suggestion about type + name + version, so something like `Activity||SayHello` (which has no version) seems good to me.

bachuv:
@jviau

> 1. Do we need the index on the dependency telemetry? The `(#1)`, `(#2)` suffixes? The fact they are different spans (with different SpanId's) should already correctly differentiate them. What value do we get out of them?
>     - We should be referencing the OTel spec for how to name spans anyways. So, we may have more changes then just this.

I'll remove the task IDs and update it to `{taskType}|{taskName}`.

> 2.  Why is the duration for all the sub orchestration dependency (client) telemetry 0? This should be the duration between first enqueueing the sub orchestration and when you are finally replayed with the completed sub orchestration.

I'll update the instrumentation so the duration includes the time from when it's enqueued to when it actually completes.

> 3. Are the `TaskActivity` server spans coming in a later PR?

Yes

jviau:
The duration of the sub-orchestration `client` span doesn't look correct to me. It should always be strictly _longer_ than the `server` span for that sub orchestration. The importance of this is you want to see the time delay difference between "here is how long I waited for my sub orchestration" vs "here is how long that sub orchestration actually ran". The gaps between `client` and `server` respective start times and their respective end times has significance.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -910,8 +919,13 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                 Version = createSubOrchestrationAction.Version
             };
 
+            startedEvent.SetParentTraceContext(parentTraceActivity);
+            using Activity activity = TraceHelper.StartTraceActivityForSubOrchestration(startedEvent);
+
             this.logHelper.SchedulingOrchestration(startedEvent);
 
+            activity.Stop();
```
jviau, line 927: As mentioned in my other comment, the duration of this activity should represent how long this orchestration was waiting on the sub orchestration complete.

jviau, line 927: This activity should also have success/error information populated as well `exception.*` tags as necessary depending on how the sub-orchestration failed (if it did fail).

jviau (changes requested):

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -910,8 +919,13 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                 Version = createSubOrchestrationAction.Version
             };
 
+            startedEvent.SetParentTraceContext(parentTraceActivity);
+            using Activity activity = TraceHelper.StartTraceActivityForSubOrchestration(startedEvent);
```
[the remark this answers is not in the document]
bachuv: I tried using Activity.Current to get the parent, but it doesn't exist in the context where the sub-orchestration span is created in ReconcileMessagesWithState(). The method is called after the end of the sub-orchestration execution and before the parent orchestration state is restored (which restores Activity.Current). I ended up storing the parent trace context instead of the entire ExecutionStartedEvent.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -149,6 +150,40 @@ static TraceHelper()
...
+                return null;
+            }
+
+            return ActivityTraceSource.StartActivity(
+                name: $"{startEvent.Name} (#{startEvent.EventId})",
+                kind: ActivityKind.Client,
+                parentContext: activityContext,
+                tags: new KeyValuePair<string, object?>[]
+                {
+                    new("dtfx.type", "orchestrator"),
```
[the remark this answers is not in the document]
bachuv: I'll make those changes in a separate PR.

src/DurableTask.Core/History/SubOrchestrationInstanceCompletedEvent.cs
```diff
@@ -50,5 +52,23 @@ public SubOrchestrationInstanceCompletedEvent(int eventId, int taskScheduledId,
...
+        /// The W3C trace context associated with this event.
+        /// </summary>
+        [DataMember]
+        public DistributedTraceContext ParentTraceContext { get; set; }
+
+        /// <summary>
+        /// The start time of the sub-orchestration.
+        /// </summary>
+        [DataMember]
+        public DateTime StartTime { get; set; }
```
jviau, line 72: Is it possible for us to avoid storing this, and instead find the pairing `SubOrchestrationInstanceCreatedEvent` to get the start time?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -698,6 +699,12 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                 }
 
                 workItem.OrchestrationRuntimeState.AddEvent(message.Event);
+
+                if (message.Event is SubOrchestrationInstanceCompletedEvent completedEvent && completedEvent.Name != null && completedEvent.ParentTraceContext != null && completedEvent.StartTime != null)
```
jviau, line 703: May want to simplify this if statement a bit, especially if we also need to start also considering `SubOrchestrationInstanceFailedEvent`.

1. Can use pattern matching, syntax may not be 100%, but try: `if (message.Event is SubOrchestrationInstanceCompletedEvent completedEvent { Name: not null, ParentTraceContext: not null, StartTime: not null })`
2. Or, should we extract a common interface for completed and failed events which contain all the relevant information? Can test for that and pass to the `TraceHelper` method. Something like:

``` CSharp
if (message.Event is ISubOrchestratrionFinishedEvent orchestrationEvent)
{
    TraceHelper.StartTraceActivityForSubOrchestrationFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, orchestrationEvent)?.Dispose();
}
}
```

src/DurableTask.Core/History/SubOrchestrationInstanceCompletedEvent.cs
bachuv: I tried to find the corresponding `SubOrchestrationInstanceCreatedEvent`, but wasn't able to. @cgillum do you know if it's possible to get a sub-orchestration's `SubOrchestrationInstanceCreatedEvent` by looking at its `SubOrchestrationInstanceCompletedEvent`?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I created the common interface and kept the validation checks in the if statement before it calls `StartTraceActivityForSubOrchestrationFinished()`. I kept the validation checks there to avoid calling the TraceHelper method any time there's a `SubOrchestrationInstanceCompletedEvent` or `SubOrchestrationInstanceFailedEvent`. Let me know if any improvements can be made here.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -775,6 +782,13 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                         new SubOrchestrationInstanceCompletedEvent(-1, runtimeState.ParentInstance.TaskScheduleId,
                             completeOrchestratorAction.Result);
 
+                    if (Activity.Current != null)
```
bachuv, line 785: Adding a note here to handle the fire and forget scenario.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
jviau: Looks great! I do recommend moving the check into the `TraceHelper` method. This keeps the details of when/what/how we perform this trace together in one place. But it is up to you.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -149,6 +149,50 @@ static TraceHelper()
...
+            {
+                return null;
+            }
+
+            ActivityContext.TryParse(
+                finishedEvent.ParentTraceContext.TraceParent,
+                finishedEvent.ParentTraceContext.TraceState,
+                out ActivityContext parentActivityContext);
+
+            List<KeyValuePair<string, object?>> activityTags = new List<KeyValuePair<string, object?>>
```
jviau, line 174: For optimization, we should create the activity first (line 188), then append tags only if non-null. If the activity is null, we can return immediately and avoid allocations.

jviau (changes requested):

src/DurableTask.Core/History/SubOrchestrationInstanceCompletedEvent.cs
jviau: I think you can find it via:

``` CSharp
workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == {TaskScheduledId});
```

Where `TaskScheduledId` is the property found on both `SubOrchestrationInstanceFailedEvent` and `SubOrchestrationInstanceCompletedEvent` (so can move it to the interface). And `workItem` is the parameter in `ReconcileMessagesWithState`

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -149,6 +149,50 @@ static TraceHelper()
...
+                new("dtfx.type", "orchestrator"),
+                new("dtfx.instance_id", orchestrationInstance.InstanceId),
+                new("dtfx.execution_id", orchestrationInstance.ExecutionId),
+            };
+
+            // Adding additional tags for a SubOrchestrationInstanceFailedEvent
+            if (finishedEvent is SubOrchestrationInstanceFailedEvent failedEvent)
+            {
+                activityTags.Add(new KeyValuePair<string, object?>("otel.status_code", "ERROR"));
+                activityTags.Add(new KeyValuePair<string, object?>("dtfx.details", failedEvent.Details));
```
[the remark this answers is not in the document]
bachuv: That makes sense. I will update this. I remember that we decided against stack trace because there could be sensitive information.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: Ok, I can move the validation into the `TraceHelper` method.

src/DurableTask.Core/Tracing/TraceHelper.cs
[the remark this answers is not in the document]
jviau: Yeah it might, so we want to make sure it is configurable.

1. Introduce some way to configure durable task instrumentation. We can discuss this later. We may want other things configurable. We can definitely include a "options" callback for the OTel extension method we eventually include.
2. OTel exporters typically also have their own way to control stack trace exporting - as long as we use the correct attribute name, they typically have ways to ignore exporting of stack trace.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: How should we handle instrumenting/showing the execution of a fire and forget sub-orchestration? Since it's fire and forget, we don't await the execution completion which means that the orchestration execution might complete or might not within the same process. 

I'm noticing that if there's an await after a fire and forget sub-orchestration call, then the orchestration execution is captured and emitted (in the simple test code). If there is no await (the fire and forget call is the last step in the orchestration), then we either see nothing about the orchestration execution or we see a partial execution (one activity call span, but this has the incorrect parent [highlighted in image below]). 

I might be misunderstanding how fire and forget works so please correct me if I'm wrong. I also pushed these changes as a recent commit if anyone wants to refer to the instrumentation code I added for fire and forget sub-orchestrations.

```
class FireAndForgetSubOrchestration : TaskOrchestration<string, string>
{
    {
        string result = "";
        result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
        context.CreateSubOrchestrationInstance<object>(NameVersionHelper.GetDefaultName(typeof(HelloFanOut)), "", "1", null, new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
        context.CreateSubOrchestrationInstance<object>(NameVersionHelper.GetDefaultName(typeof(HelloSequence)), "", "2", null, new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
        context.CreateSubOrchestrationInstance<object>(NameVersionHelper.GetDefaultName(typeof(HelloSequence)), "", "3", null, new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
        return result;
    }
}
```

image

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -305,10 +305,10 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
             ExecutionStartedEvent startEvent =
                 runtimeState.ExecutionStartedEvent ??
                 workItem.NewMessages.Select(msg => msg.Event).OfType<ExecutionStartedEvent>().FirstOrDefault();
-            Activity traceActivity = TraceHelper.StartTraceActivityForExecution(startEvent);
-
             OrchestrationState instanceState = null;
 
+            Activity traceActivity = TraceHelper.StartTraceActivityForExecution(startEvent);
```
bachuv, line 310: This is used to restore the current Activity before going through the new messages. It's restored here to give the spans that are created in `ReconcileMessagesWithState()` an accurate parent Activity.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -325,6 +325,8 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
             {
                 do
                 {
+                    traceActivity = TraceHelper.StartTraceActivityForExecution(runtimeState.ExecutionStartedEvent);
```
bachuv, line 328: This is used to restore the current Activity for sub-orchestrations. I added it to the end of the do clause (after line 555) since it's already restored on line 310, but it looks like the orchestration state gets modified and the spans don't emit correctly.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: I would expect that a fire-and-forget sub-orchestration would show as completed after the sub-orchestration completed, which means it might live outside the scope of the parent orchestration. I'm not sure what that translates to in terms of implementation, but that's the picture I'd expect to see. If you're seeing that nothing gets emitted ever, then I wonder if we need to revisit something more fundamental about how we're tracking the start and end of trace spans.

That said, I'm okay with punting that to a separate PR. Fire-and-forget is not a common scenario (the Durable Task Framework doesn't really support it, actually, so the behavior is a bit undefined). I'm okay if we get the main sub-orchestration scenario working and do the edge cases separately.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -323,7 +325,7 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
             {
                 do
                 {
-                    Activity traceActivity = TraceHelper.StartTraceActivityForExecution(runtimeState.ExecutionStartedEvent);
+                    traceActivity = TraceHelper.StartTraceActivityForExecution(runtimeState.ExecutionStartedEvent);
```
jviau, line 328: Slightly worried about this line in combination with line `310`, where `traceActivity` is originally set. Mostly worried about a case where we run line 310 then immediately run this line. What are the implications of essentially starting this activity twice? Do we need to clean up the old one first? Or should we skip this line on the first pass here?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
jviau: Agree with punting it to another PR.

I also think it does have to do with _when_ we are emitting the span. I can say though it is incorrect to _not_ `await` the fire & forget. Even if it is fire & forget, you should still `await` it in case the framework is doing _some_ work behind the scenes.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -171,26 +168,32 @@ static TraceHelper()
...
+            activity.SetTag("dtfx.type", "orchestrator");
+            activity.SetTag("dtfx.instance_id", orchestrationInstance.InstanceId);
+            activity.SetTag("dtfx.execution_id", orchestrationInstance.ExecutionId);
 
             // Adding additional tags for a SubOrchestrationInstanceFailedEvent
             if (finishedEvent is SubOrchestrationInstanceFailedEvent failedEvent)
             {
-                activityTags.Add(new KeyValuePair<string, object?>("otel.status_code", "ERROR"));
-                activityTags.Add(new KeyValuePair<string, object?>("dtfx.details", failedEvent.Details));
+                Exception cause = Utils.RetrieveCause(failedEvent.Details, new JsonDataConverter());
```
jviau, line 189: Oh, this may be problematic. Customers can change how error details are converted, but we probably do not have access to that converter at this point. Not entirely sure what to do here. Do we leave this in? Or only retain `otel.status_code` for now and address this later? There is a legitimate concern where the customization of error data conversion would cause this to throw if we try and deserialize without having the proper converter.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -918,8 +947,15 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                 Version = createSubOrchestrationAction.Version
             };
 
+            startedEvent.SetParentTraceContext(parentTraceActivity);
+
             this.logHelper.SchedulingOrchestration(startedEvent);
 
+            if (OrchestrationTags.IsTaggedAsFireAndForget(createSubOrchestrationAction.Tags))
+            {
+                TraceHelper.CreateActivityForFireAndForgetSubOrchestration(historyEvent, runtimeState.OrchestrationInstance, parentTraceActivity)?.Dispose();
```
cgillum, line 956: This seems like a really strange thing to do. I wonder if there's a better way to represent this - perhaps an OTel event of some kind? Having a span feels meaningless since there is no end time associated with this.

cgillum (commented):
Added some comments. One problem I'm anticipating is that this branch is really out of date and missing several other changes that have been made. We may need to rebase these PRs.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
jviau: While the duration will be very small, nearly 0, it does have meaning still. It will be a `Producer` span to convey meaning that this orchestration has kicked off a fire & forget. I think it is important to keep this as a span as it will maintain the chain of `Orchestration (Server) -> Fire & Forget (Producer) -> Orchestration (Server)`.

For OTel event, spans do have events, but I think a producer span here is still important so that users can distinguish between _multiple_ fire & forgets in the same orchestration, because then we get the parent-child relationship I mentioned. It would be ideal for this producer span, and several other spans, to include the actual work of the `IOrchestrationServicer` in it, but that will be challenging to do with the current structure of this code.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I reverted those changes and will open a new PR for fire and forget scenarios once this PR is merged. I'll look into where we are emitting that span in the new PR and look for a more accurate place.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I was actually able to remove it, so we don't need to worry about this.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: I pulled the latest from `main` and added details based on the `FailureDetails` property.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -143,6 +144,58 @@ static TraceHelper()
...
+        /// </summary>
+        /// <param name="orchestrationInstance">The associated orchestration instance metadata.</param>
+        /// <param name="finishedEvent">The sub-orchestration event. Represents a sub-orchestration completing or failing.</param>
+        /// <param name="createdEvent">The related sub-orchestration created event.</param>
+        /// <returns>
+        /// Returns a newly started <see cref="Activity"/> with (task) activity and orchestration-specific metadata.
+        /// </returns>
+        internal static Activity? StartTraceActivityForSubOrchestrationFinished(
+            OrchestrationInstance orchestrationInstance,
+            ISubOrchestrationFinishedEvent finishedEvent,
```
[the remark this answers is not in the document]
bachuv: This interface isn't necessary anymore since we are restoring the "current" `Activity` (parent orchestration) before creating spans for completed or failed sub-orchestrations. It was used in a previous iteration to manually get the parent's trace context. I deleted the interface.

jviau (approved):
I take it the fire & forget logic is being left for another time?

bachuv:
Yes, I removed it from this PR and will work on it at another time.

src/DurableTask.Core/History/SubOrchestrationInstanceCompletedEvent.cs
```diff
@@ -13,7 +13,9 @@
 
 namespace DurableTask.Core.History
 {
+    using System;
     using System.Runtime.Serialization;
+    using DurableTask.Core.Tracing;
```
cgillum, line 18: Non-blocking, but it looks like this whole file can be reverted (I assume these new `using` statements aren't necessary).

src/DurableTask.Core/History/SubOrchestrationInstanceFailedEvent.cs
```diff
@@ -13,7 +13,9 @@
 #nullable enable
 namespace DurableTask.Core.History
 {
+    using System;
     using System.Runtime.Serialization;
+    using DurableTask.Core.Tracing;
```
cgillum, line 18: Non-blocking, but it looks like this whole file can be reverted (I assume these new `using` statements aren't necessary).

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -737,10 +739,28 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
...
+                    {
+                        // We immediately publish the activity span for this timer by creating the activity and immediately calling Dispose() on it.
+                        TraceHelper.CreateActivityForTimer(workItem.OrchestrationRuntimeState.OrchestrationInstance, message.Event.Timestamp, (timerFiredEvent).FireAt)?.Dispose();
+                    }
+                    else if (historyEvent is SubOrchestrationInstanceCompletedEvent subOrchestrationInstanceCompletedEvent)
+                    {
+                        SubOrchestrationInstanceCreatedEvent subOrchestrationCreatedEvent = (SubOrchestrationInstanceCreatedEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == subOrchestrationInstanceCompletedEvent.TaskScheduledId);
+
+                        // We immediately publish the activity span for this sub-orchestration by creating the activity and immediately calling Dispose() on it.
+                        TraceHelper.StartTraceActivityForSubOrchestrationFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, subOrchestrationCreatedEvent)?.Dispose();
```
cgillum, line 755: Given that we will _always_ immediately call dispose on the trace activity that's returned from this method, I suggest we rename this method to `EmitTraceActivityForSubOrchestrationFinished`, have it return `void`, and do the `Dispose()` inside the implementation. That will make the intent of the code more clear.

cgillum (commented):
A could tiny cleanup things and then just one small refactor and I'm ready to sign off on this PR.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I made the same changes for Timers since they follow the same pattern.

cgillum (approved):
Latest iteration looks good to me!

PR #826, opened 2022-11-16
OpenTelemetry instrumentation for Timers

bachuv:
This PR adds instrumentation for Durable timers.

**Sample orchestration**
```
class HelloSequenceWithTimer : TaskOrchestration<string, string>
{
    {
        string output = "";
        output += await context.ScheduleTask<string>(typeof(SayHello), "Tokyo") + ", ";
        await context.CreateTimer(context.CurrentUtcDateTime.AddSeconds(15), true);
        output += await context.ScheduleTask<string>(typeof(SayHello), "London") + ", ";
        output += await context.ScheduleTask<string>(typeof(SayHello), "Seattle");
        return output;
    }
}
```

**Related Gantt charts**
Application Insights
image

image

Zipkin
image

image

### DIFF
--- a/src/DurableTask.Core/History/TimerCreatedEvent.cs
+++ b/src/DurableTask.Core/History/TimerCreatedEvent.cs
@@ -13,7 +13,9 @@
 
 namespace DurableTask.Core.History
 {
+    using DurableTask.Core.Tracing;
     using System;
+    using System.Diagnostics;
     using System.Runtime.Serialization;
 
     /// <summary>
@@ -41,5 +43,39 @@ namespace DurableTask.Core.History
         /// </summary>
         [DataMember]
         public DateTime FireAt { get; set; }
+
+        /// <summary>
+        /// The W3C trace context associated with this event.
+        /// </summary>
+        [DataMember]
+        public DistributedTraceContext ParentTraceContext { get; set; }
+
+        /// <summary>
+        /// Gets the W3C distributed trace parent context from this event, if any.
+        /// </summary>
+        public bool TryGetParentTraceContext(out ActivityContext parentTraceContext)
+        {
+            if (this.ParentTraceContext?.TraceParent == null)
+            {
+                parentTraceContext = default;
+                return false;
+            }
+
+            return ActivityContext.TryParse(
+                this.ParentTraceContext.TraceParent,
+                this.ParentTraceContext.TraceState,
+                out parentTraceContext);
+        }
+
+        // Used for new orchestration and sub-orchestration
+        internal void SetParentTraceContext(Activity traceActivity)
+        {
+            if (traceActivity != null)
+            {
+                this.ParentTraceContext = new DistributedTraceContext(
+                    traceActivity.Id,
+                    traceActivity.TraceStateString);
+            }
+        }
     }
 }
--- a/src/DurableTask.Core/History/TimerCreatedEvent.cs
+++ b/src/DurableTask.Core/History/TimerCreatedEvent.cs
@@ -13,10 +13,10 @@
 
 namespace DurableTask.Core.History
 {
-    using DurableTask.Core.Tracing;
     using System;
     using System.Diagnostics;
     using System.Runtime.Serialization;
+    using DurableTask.Core.Tracing;
 
     /// <summary>
     /// A history event for a new timer creation
@@ -43,39 +43,5 @@ namespace DurableTask.Core.History
         /// </summary>
         [DataMember]
         public DateTime FireAt { get; set; }
-
-        /// <summary>
-        /// The W3C trace context associated with this event.
-        /// </summary>
-        [DataMember]
-        public DistributedTraceContext ParentTraceContext { get; set; }
-
-        /// <summary>
-        /// Gets the W3C distributed trace parent context from this event, if any.
-        /// </summary>
-        public bool TryGetParentTraceContext(out ActivityContext parentTraceContext)
-        {
-            if (this.ParentTraceContext?.TraceParent == null)
-            {
-                parentTraceContext = default;
-                return false;
-            }
-
-            return ActivityContext.TryParse(
-                this.ParentTraceContext.TraceParent,
-                this.ParentTraceContext.TraceState,
-                out parentTraceContext);
-        }
-
-        // Used for new orchestration and sub-orchestration
-        internal void SetParentTraceContext(Activity traceActivity)
-        {
-            if (traceActivity != null)
-            {
-                this.ParentTraceContext = new DistributedTraceContext(
-                    traceActivity.Id,
-                    traceActivity.TraceStateString);
-            }
-        }
     }
 }
--- a/src/DurableTask.Core/IDurableTraceContextWrapper.cs
+++ b/src/DurableTask.Core/IDurableTraceContextWrapper.cs
@@ -1,0 +1,59 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using DurableTask.Core.Tracing;
+
+namespace DurableTask.Core
+{
+    /// <summary>
+    /// IDurableTraceContextWrapper
+    /// </summary>
+    public interface IDurableTraceContextWrapper
+    {
+        /// <summary>
+        /// ParentTraceContext
+        /// </summary>
+        public DistributedTraceContext ParentTraceContext { get; set; }
+    }
+
+    internal static class DurableTraceContextWrapperExtensions
+    {
+        internal static bool TryGetParentTraceContext(this IDurableTraceContextWrapper wrapper, out ActivityContext parentTraceContext)
+        {
+            if (wrapper.ParentTraceContext?.TraceParent == null)
+            {
+                parentTraceContext = default;
+                return false;
+            }
+
+            return ActivityContext.TryParse(
+                wrapper.ParentTraceContext.TraceParent,
+                wrapper.ParentTraceContext.TraceState,
+                out parentTraceContext);
+        }
+
+        internal static void SetParentTraceContext(this IDurableTraceContextWrapper wrapper, Activity activity)
+        {
+            if (activity != null)
+            {
+                wrapper.ParentTraceContext = new DistributedTraceContext(
+                    activity.Id,
+                    activity.TraceStateString);
+            }
+        }
+    }
+}
--- a/src/DurableTask.Core/IDurableTraceContextWrapper.cs
+++ b/src/DurableTask.Core/IDurableTraceContextWrapper.cs
@@ -20,12 +20,12 @@ using DurableTask.Core.Tracing;
 namespace DurableTask.Core
 {
     /// <summary>
-    /// IDurableTraceContextWrapper
+    /// Interface to store the Trace Context for events. Used for Distributed Tracing.
     /// </summary>
     public interface IDurableTraceContextWrapper
     {
         /// <summary>
-        /// ParentTraceContext
+        /// The trace context associated with an event.
         /// </summary>
         public DistributedTraceContext ParentTraceContext { get; set; }
     }
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -484,12 +484,6 @@ namespace DurableTask.Core
             {
                 LogDuplicateEvent("TimerFired", timerFiredEvent, taskId);
             }
-
-            if (!this.IsReplaying)
-            {
-                Activity activity = TraceHelper.CreateActivityForTimer(this.OrchestrationInstance, timerFiredEvent.Timestamp, timerFiredEvent.FireAt);
-                activity.Stop();
-            }
         }
 
         private void LogDuplicateEvent(string source, HistoryEvent historyEvent, int taskId)
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -484,6 +484,12 @@ namespace DurableTask.Core
             {
                 LogDuplicateEvent("TimerFired", timerFiredEvent, taskId);
             }
+
+            if (!this.IsReplaying)
+            {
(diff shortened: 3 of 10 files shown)


cgillum:
I'd like to get other people's opinion on this (feel free to chime in @jviau), but I think it would be more intuitive if the span for the timer lasted for the full duration, similar to activities, rather than having it cover only the few microseconds required to create the timer in memory, which the current implementation appears to be doing.

Here's what I'd expect this to look like:
image

Also, do we really need the timer history event to include the trace context? I didn't expect this to be necessary since we're not actually calling into any other code or services. Timers are just a way for an orchestration to call itself. If we could implement this in a way that didn't require trace context in the timer event, that would be ideal.

jviau:
Yes, I agree regarding duration. Duration should be the full time this span is "active". We should also make sure to include cancelled timers and relevant attributes to identify it is canceled (maybe set it to success=false)

bachuv:
To get the duration to reflect the time from when the timer is created to when it's fired, I was thinking of adding the trace context to the TimerFired event. That way the Activity could start when the timer is created [here in TaskOrchestrationDispatcher ](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L876) and get assigned to the TimerFiredEvent. Then, the Activity would stop when the timer is fired [here in TaskOrchestrationContext](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/TaskOrchestrationContext.cs#L519). This involves storing the trace context in the TimerFired event though.

I'm continuing to look through the code to avoid saving the trace context to the timer history event. Any pointers on where I could look would be helpful!

cgillum:
@bachuv I think the approach you're suggesting is mostly right, and I agree we'll need to persist the trace context in the TimerFiredEvent payload. However, stopping the trace activity in TaskOrchestrationContext.cs won't work for modern out-of-proc because that code path will never be executed. Instead, I think we'll want to add code to stop the activity trace somewhere in TaskOrchestrationDispatcher.cs (the same file we start it in, which is also nice for symmetry), which should be executed always for all supported laguages. [ReconcileMessagesWithState](https://github.com/Azure/durabletask/blob/4ef9609d43881de4bfec0f851631028ddf5ec1b9/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L656) might be a good place to do this since we're iterating through all the incoming messages/events here anyways.

cgillum:
Actually, I want to revisit this point:

> To get the duration to reflect the time from when the timer is created to when it's fired, I was thinking of adding the trace context to the TimerFired event.

I changed my mind (again) and think we might not need to embed any trace context into TimerFiredEvent. We don't need it to get the duration. Instead, we can use the `TimerFiredEvent.Timestamp` value to get the start time of the activity span.

bachuv:
This makes sense. We would create the `Activity` for the Timer when it's fired and wouldn't have to store any information in the history event. This seems like a clean implementation. I'll update the PR with this implementation.

bachuv:
Looking at it again, we still need to set the parent of the Timer span which leads to me to think we still need information from TaskOrchestrationDispatcher.

cgillum:
@bachuv I think it depends on what we want the parent to be. If the parent is a common, long-running orchestration span (similar to my [fake screenshot above](https://github.com/Azure/durabletask/pull/826#issuecomment-1317894523)) then it might not be needed. However, I've since realized that there may be problems with the idea of a long-running orchestration span. If we instead decide to go with a new orchestration span for each replay (which I think the code currently does?) then I agree we need to save some activity context state so that we can parent the timer fired event to the correct replay. I ended up needing to do this in the `durabletask-go` implementation of distributed tracing for timers.

jviau:
> Looking at it again, we still need to set the parent of the Timer span which leads to me to think we still need information from TaskOrchestrationDispatcher.

This depends where / how this activity is started. The tracing code we have so far has a lot of explicit setting of parent context. However, parent will also automatically be populated if there is a current activity (`Activity.Current` is non-null). So, if the timer activity is started in a context where `Activity.Current` is already set to be the parent orchestration, then there is no work to be done! It will be set for us.

jviau:
I'd like to bring up a comment I have in our spec document:

I'd like to discuss how timer spans look in a variety of scenarios. We are focusing on the "standard" scenario where orchestration starts and waits on a timer. How about other scenarios?

1. Timer is in parallel with other work. IE: orchestration starts a timer, does some other work, then eventually waits on it.
2. Timer is waited on in a `Task.WhenAny`? When the timer is the task the on yielded, vs when it is not.
3. Timer is cancelled.

I want to make sure a dedicated span is the correct choice for a timer. The alternative is an event on the orchestration running span.

bachuv:
> @bachuv I think it depends on what we want the parent to be. If the parent is a common, long-running orchestration span (similar to my [fake screenshot above](https://github.com/Azure/durabletask/pull/826#issuecomment-1317894523)) then it might not be needed. However, I've since realized that there may be problems with the idea of a long-running orchestration span. If we instead decide to go with a new orchestration span for each replay (which I think the code currently does?) then I agree we need to save some activity context state so that we can parent the timer fired event to the correct replay. I ended up needing to do this in the `durabletask-go` implementation of distributed tracing for timers.

@cgillum We currently have the parent as the common, long-running orchestration span. [This PR](https://github.com/Azure/durabletask/pull/692) made the update from showing orchestration replays to a common span. 

You had mentioned there might be issues with having the common span like not being able to see traces for in-progress orchestrations. Should we re-instrument orchestration spans then? We should probably also keep the implementation consistent with `durabletask-go`. Do you have a PR for the changes you're making that I can take a look at?

For now, I will see if I can use `Activity.Current` as the parent (thanks @jviau for the suggestion!)

bachuv:
> That said, I can see how they can be noisy in certain situations too, like when they are used for timeouts (often as part of scenario 2). If a timer is cancelled (scenario 3), then I could see us simply not emitting the span since it isn't actually relevant to the orchestration's execution. We'd only emit the span for completed timers. Thoughts?

I think we should only emit the span when it adds value for the customer. For scenario 2, I think it makes sense if we emit the timer span if the timer is fired. So I agree that we should emit the span for completed timers. If a timer doesn't fire then I think the customer should be able to infer why because of the related spans (like another task completed like in scenario 2) around that time, but let me know if you think otherwise.

> @bachuv I had a side chat with Jacob on this topic and I think he convinced me that it's actually better to have a single span for the entire orchestration execution, for more than just visualization purposes (for example, customers can get more useful metrics with one big span). So if that's actually already implemented, let's keep it as-is. I'll update my `durabletask-go` implementation to match this (though part of me wonders if we should do both - one big span and have another child span for each replay - but let's not worry about that for the sake of moving this PR forward).

Ok I'll leave it how it is then. I'll add a comment about adding replay spans in the spec doc.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -484,6 +484,12 @@ public void HandleTimerFiredEvent(TimerFiredEvent timerFiredEvent)
             {
                 LogDuplicateEvent("TimerFired", timerFiredEvent, taskId);
             }
+
+            if (!this.IsReplaying)
+            {
+                Activity activity = TraceHelper.CreateActivityForTimer(this.OrchestrationInstance, timerFiredEvent.Timestamp, timerFiredEvent.FireAt);
+                activity.Stop();
```
cgillum, line 491: This won't work for Java because we never run any code in `TaskOrchestrationContext`. The same will be true for any future language SDKs that are based on the out-of-proc V2 protocol. I suggest you instead do this in TaskOrchestrationDispatcher.cs when we're looping through all the new messages we received. This is similar to what I'm doing in `durabletask-go`.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -165,6 +165,24 @@ static TraceHelper()
...
+                name: "Timer",
+                kind: ActivityKind.Internal,
+                startTime: startTime,
+                parentContext: Activity.Current?.Context ?? default,
+                tags: new KeyValuePair<string, object?>[]
+                {
+                    new("dtfx.type", "timer"),
+                    new("dtfx.instance_id", instance.InstanceId),
+                    new("dtfx.execution_id", instance.ExecutionId),
+                    new("dtfx.fire_at", fireAt.ToString("o")),
```
cgillum, line 180: I suppose we can do this as part of a separate PR, but just a reminder that we'll be replacing "dtfx" with "durabletask" in these attributes.

cgillum (commented):
Updated screenshots look great! Only one blocking comment from me regarding where we emit this trace activity event.

src/DurableTask.Core/TaskOrchestrationContext.cs
bachuv: Right, you had mentioned that before. Sorry about that! I'll move it to TaskOrchestrationDispatcher.cs.

src/DurableTask.Core/History/TimerCreatedEvent.cs
```diff
@@ -41,5 +43,39 @@ public TimerCreatedEvent(int eventId)
         /// </summary>
         [DataMember]
         public DateTime FireAt { get; set; }
+
+        /// <summary>
+        /// The W3C trace context associated with this event.
+        /// </summary>
+        [DataMember]
+        public DistributedTraceContext ParentTraceContext { get; set; }
```
jviau, line 51: Does `ParentTraceContext` need to be stored? I am wondering if `Activity.Current` will already be re-hydrated to the appropriate parent when timer event span is to be emitted. And as for storing this timers span itself, I don't think it needs to be stored at all? We only need to create the `Acitivty` and emit it right when the timer completes. (I think at least).

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -484,6 +484,12 @@ public void HandleTimerFiredEvent(TimerFiredEvent timerFiredEvent)
             {
                 LogDuplicateEvent("TimerFired", timerFiredEvent, taskId);
             }
+
+            if (!this.IsReplaying)
+            {
+                Activity activity = TraceHelper.CreateActivityForTimer(this.OrchestrationInstance, timerFiredEvent.Timestamp, timerFiredEvent.FireAt);
```
jviau, line 490: If you add `using` here you can remove the `activity.Stop()`. The disposal will call `Stop()` for you. Plus, `activity` might be `null` - the using will handle that case for you.

```suggestion
                using Activity activity = TraceHelper.CreateActivityForTimer(this.OrchestrationInstance, timerFiredEvent.Timestamp, timerFiredEvent.FireAt);
```

src/DurableTask.Core/History/TimerCreatedEvent.cs
```diff
@@ -41,5 +43,39 @@ public TimerCreatedEvent(int eventId)
...
+            }
+
+            return ActivityContext.TryParse(
+                this.ParentTraceContext.TraceParent,
+                this.ParentTraceContext.TraceState,
+                out parentTraceContext);
+        }
+
+        // Used for new orchestration and sub-orchestration
+        internal void SetParentTraceContext(Activity traceActivity)
```
jviau, line 71: This code is duplicated in many history events. Can we consolidate it a bit?

1. First, make sure it is absolutely necessary. Can some of these events rely on `Activity.Current` being available instead of storing them? There is a cost to storing this extra property on these events, lets only pay it if we absolutely have to.
2. You could make an interface and extension methods:

``` CSharp
internal interface IDurableContextHolder
{
    DistributedTraceContext TraceContext { get; set; }
}

internal static DistributedContextHolderExtensions
{
    public static bool TryGetParentTraceContext(this IDurableContextHolder holder, out ActivityContext parentContext)
    {
        // implementation
    }

    public static void SetParentTraceContext(this IDurableContextHolder holder, Activity activity)
    {
        // implementation
    }
}
```

bachuv:
@jviau do you have any comments/feedback about the responses that @cgillum and I gave regarding the different timer scenarios?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -697,6 +697,11 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                     continue;
                 }
 
+                if (Activity.Current != null && message.Event is TimerFiredEvent)
```
cgillum, line 700: You can avoid the cast on line 702 by declaring a variable in your `is` expression.
```suggestion
                if (Activity.Current != null && message.Event is TimerFiredEvent timerFired)
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -697,6 +697,11 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
...
+                {
+                    using Activity activity = TraceHelper.CreateActivityForTimer(workItem.OrchestrationRuntimeState.OrchestrationInstance, message.Event.Timestamp, ((TimerFiredEvent)message.Event).FireAt);
```
cgillum, line 702: nit: Now that I'm looking at this, it feels like a very unintuitive use of the C# `using` feature. I think it would be easier for readers of the to understand what we're doing if we disposed the activity explicitly, instead of implicitly. That also reduces the chance that someone accidentally introduces a bug by moving curly braces around. I also suggest a comment explaining what we're doing here since it's a bit counterintuitive.
```suggestion
                    // We immediately publish the activity span for this timer by creating the activity and immediately calling Dispose() on it.
                    TraceHelper.CreateActivityForTimer(workItem.OrchestrationRuntimeState.OrchestrationInstance, message.Event.Timestamp, ((TimerFiredEvent)message.Event).FireAt).Dispose();
```

src/DurableTask.Core/IDurableTraceContextWrapper.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Diagnostics;
+using System.Text;
+using DurableTask.Core.Tracing;
+
+namespace DurableTask.Core
+{
+    /// <summary>
+    /// Interface to store the Trace Context for events. Used for Distributed Tracing.
+    /// </summary>
+    public interface IDurableTraceContextWrapper
```
cgillum, line 25: I was initially wondering if this could be internal instead of public, but I'm guessing we need this to be public so that backends (which are in separate projects) can access the trace context...

Unrelated, but I don't really like the name "Wrapper" since we're not actually wrapping anything. Another suggestion would be `ISupportsDurableTraceContext`.

cgillum (commented):
A few small comments.

cgillum (approved):
LGTM!

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -165,6 +165,24 @@ static TraceHelper()
...
         }
 
+        internal static Activity? CreateActivityForTimer(OrchestrationInstance instance, DateTime startTime, DateTime fireAt)
+        {
+            Activity? newActivity = ActivityTraceSource.StartActivity(
+                name: "Timer",
+                kind: ActivityKind.Internal,
+                startTime: startTime,
+                parentContext: Activity.Current?.Context ?? default,
+                tags: new KeyValuePair<string, object?>[]
```
jviau, line 175: Small suggestion (I think Rohit first mentioned this): We can have avoid allocating this array if you add the tags after activity creation and only if it is non-null.

``` CSharp
 Activity? newActivity = ActivityTraceSource.StartActivity(...);

if (newActivity is not null)
{
    newActivity.AddTag(...);
    newActivity.AddTag(...);
    newActivity.AddTag(...);
}

return newActivity;
```

jviau (approved):

PR #828, opened 2022-11-19
Increment version of DT.AzureStorage to v1.13.2

cgillum:
This is in preparation for a bug-fix release of DurableTask.AzureStorage.

I also slipped in a new unit test that had been on my local machine but never contributed as a PR (it seemed too insignificant to create a separate PR just for that, hence the bundling).


jviau (approved):
👍

Issue #840, opened 2022-12-10
DT.AzureStorage to support querying orchestration tags


cgillum:
The Azure Storage backend doesn't currently support querying orchestration tags. Tags can be added to a new orchestration and are visible to the runtime, but not to `TaskHubClient`.

The fix likely needs to go into `AzureStorageTrackingStore.cs` around the code where we interact with the "Instances" table.

There's an existing `LargeOrchestrationTags` test in the DurableTask.AzureStorage.Tests project. As part of adding support for querying tags, we should uncomment some of the additional validation that exists for this test.

PR #841, opened 2022-12-10
DT.AzureStorage: Fixed exception when orchestration tags are too large

cgillum:
If an orchestration is started with tag data that exceeds the maximum allowed size for table storage columns when using the DurableTask.AzureStorage backend, the orchestration will cause the following exception (possibly repeatedly):

```
DurableTask.AzureStorage.Storage.DurableTaskStorageException: Element 1 in the batch returned an unexpected response code.
 ---> Microsoft.WindowsAzure.Storage.StorageException: Element 1 in the batch returned an unexpected response code.
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ExecuteAsyncInternal[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext, CancellationToken token)
   at DurableTask.AzureStorage.TimeoutHandler.ExecuteWithTimeout[T](String operationName, String account, AzureStorageOrchestrationServiceSettings settings, Func`3 operation, AzureStorageOrchestrationServiceStats stats, String clientRequestId) in /_/src/DurableTask.AzureStorage/TimeoutHandler.cs:line 136
   at DurableTask.AzureStorage.Storage.AzureStorageClient.MakeStorageRequest[T](Func`3 storageRequest, String accountName, String operationName, String clientRequestId, Boolean force) in /_/src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs:line 149
Request Information
RequestID:32d7747d-b002-000a-0929-0c3b5a000000
RequestDate:Fri, 09 Dec 2022 23:56:37 GMT
StatusMessage:1:The property value exceeds the maximum allowed size (64KB). If the property value is a string, it is UTF-16 encoded and the maximum number of characters should be 32K or less.
ErrorCode:
ErrorMessage:1:The property value exceeds the maximum allowed size (64KB). If the property value is a string, it is UTF-16 encoded and the maximum number of characters should be 32K or less.
RequestId:32d7747d-b002-000a-0929-0c3b5a000000
Time:2022-12-09T23:56:38.3192478Z
```

This PR fixes this by adding `"Tags"` to the list of variable-length properties.

NOTE: There is no impact to Durable Functions customers as Durable Functions doesn't support the Tags feature (and this also explains why this bug has existed for so long - the scenario was simply never considered when this backend was first written).

FYI @RamjotSingh 

Related: https://github.com/Azure/durabletask/issues/840

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <MajorVersion>1</MajorVersion>
     <MinorVersion>13</MinorVersion>
-    <PatchVersion>2</PatchVersion>
+    <PatchVersion>3</PatchVersion>
 
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <FileVersion>$(VersionPrefix).0</FileVersion>
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -57,6 +57,7 @@ namespace DurableTask.AzureStorage.Tracking
             "Details",
             "Correlation",
             "FailureDetails",
+            "Tags",
         };
 
         readonly string storageAccountName;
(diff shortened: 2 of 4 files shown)


RamjotSingh (approved):

cgillum:
The CI is failing with an error that looks like it's related to .NET Core 3.1 support. We may need to update the pipeline to explicitly install .NET Core 3.1 not that it's officially retired.

cgillum:
Actually, the problem may be a .NET 7 regression: https://github.com/dotnet/msbuild/issues/8003. We don't currently control the version of the .NET SDK used to build our solution, but that might be what's impacting us.

PR #845, opened 2023-01-11
Added Orchestration Memory Manager

amdeel:
Added Orchestration Memory Manager to prevent OutOfMemoryExceptions when loading in orchestration histories.

This memory manager checks currently allocated process memory and attempts to reserve the necessary memory or wait until the memory is available.

Added orchestration history size calculation stored in the instances table and an instances table lookup before pulling in orchestration history, adding 1 storage transaction per orchestration history load.

### DIFF
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -1071,6 +1071,28 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
+        internal void ThrottlingOrchestrationHistory(
+            string Account, 
+            string TaskHub, 
+            string InstanceId, 
+            string ExecutionId, 
+            long SizeInBytes,
+            string Details, 
+            string AppName, 
+            string ExtensionVersion)
+        {
+            this.WriteEvent(
+                EventIds.ThrottlingOrchestrationHistory,
+                Account,
+                TaskHub,
+                InstanceId,
+                ExecutionId ?? string.Empty,
+                SizeInBytes,
+                Details,
+                AppName,
+                ExtensionVersion);
+        }
+
         // Specifying tasks is necessary when using WriteEventWithRelatedActivityId
         // or else the "TaskName" property written to ETW is the name of the opcode instead
         // of the name of the trace method.
--- a/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
+++ b/src/DurableTask.AzureStorage/AnalyticsEventSource.cs
@@ -1083,12 +1083,12 @@ namespace DurableTask.AzureStorage
                 ExtensionVersion);
         }
 
+        [Event(EventIds.ThrottlingOrchestrationHistory, Level = EventLevel.Informational, Version = 1)]
         internal void ThrottlingOrchestrationHistory(
             string Account, 
             string TaskHub, 
             string InstanceId, 
             string ExecutionId, 
-            long SizeInBytes,
             string Details, 
             string AppName, 
             string ExtensionVersion)
@@ -1097,9 +1097,8 @@ namespace DurableTask.AzureStorage
                 EventIds.ThrottlingOrchestrationHistory,
                 Account,
                 TaskHub,
-                InstanceId,
+                InstanceId ?? string.Empty,
                 ExecutionId ?? string.Empty,
-                SizeInBytes,
                 Details,
                 AppName,
                 ExtensionVersion);
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
@@ -251,6 +251,11 @@ namespace DurableTask.AzureStorage
         public TimeSpan StorageRequestsTimeoutCooldown { get; set; } = TimeSpan.FromMinutes(5);
 
         /// <summary>
+        /// Gets or sets the memory limit in bytes.
+        /// </summary>
+        public long MemoryLimitBytes { get; set; } = (long)(1.5 * 1024 * 1024 * 1024);
+
+        /// <summary>
         /// Returns bool indicating is the TrackingStoreStorageAccount has been set.
         /// </summary>
         public bool HasTrackingStoreStorageAccount => this.TrackingStoreStorageAccountDetails != null;
--- a/src/DurableTask.AzureStorage/Logging/LogEvents.cs
+++ b/src/DurableTask.AzureStorage/Logging/LogEvents.cs
@@ -2654,61 +2654,55 @@ namespace DurableTask.AzureStorage.Logging
                 Utils.AppName,
                 Utils.ExtensionVersion);
         }
-    }
 
-    internal class ThrottlingOrchestrationHistory : StructuredLogEvent, IEventSourceEvent
-    {
-        public ThrottlingOrchestrationHistory(
-            string account,
-            string taskHub,
-            string instanceId,
-            string executionId,
-            long sizeInBytes,
-            string details)
+        internal class ThrottlingOrchestrationHistory : StructuredLogEvent, IEventSourceEvent
         {
-            this.Account = account;
-            this.TaskHub = taskHub;
-            this.InstanceId = instanceId;
-            this.ExecutionId = executionId;
-            this.SizeInBytes = sizeInBytes;
-            this.Details = details;
+            public ThrottlingOrchestrationHistory(
+                string account,
+                string taskHub,
+                string instanceId,
+                string executionId,
+                string details)
+            {
+                this.Account = account;
+                this.TaskHub = taskHub;
+                this.InstanceId = instanceId;
+                this.ExecutionId = executionId;
+                this.Details = details;
+            }
+
+            [StructuredLogField]
+            public string Account { get; }
+
+            [StructuredLogField]
+            public string TaskHub { get; }
+
+            [StructuredLogField]
+            public string InstanceId { get; }
+
+            [StructuredLogField]
+            public string ExecutionId { get; }
+
+            [StructuredLogField]
+            public string Details { get; }
+
+            public override EventId EventId => new EventId(
+                EventIds.ThrottlingOrchestrationHistory,
+                nameof(EventIds.ThrottlingOrchestrationHistory));
+
+            public override LogLevel Level => LogLevel.Information;
+
+            protected override string CreateLogMessage() =>
+                $"Throttling orchestration history load for instance {this.InstanceId}: {this.Details})";
+
+            void IEventSourceEvent.WriteEventSource() => AnalyticsEventSource.Log.ThrottlingOrchestrationHistory(
+                this.Account,
+                this.TaskHub,
+                this.InstanceId,
+                this.ExecutionId,
+                this.Details,
+                Utils.AppName,
+                Utils.ExtensionVersion);
         }
-
-        [StructuredLogField]
-        public string Account { get; }
-
-        [StructuredLogField]
-        public string TaskHub { get; }
-
-        [StructuredLogField]
-        public string InstanceId { get; }
-
-        [StructuredLogField]
-        public string ExecutionId { get; }
-
-        [StructuredLogField]
-        public long SizeInBytes { get; }
-
-        [StructuredLogField]
-        public string Details { get; }
-
-        public override EventId EventId => new EventId(
-            EventIds.ThrottlingOrchestrationHistory,
-            nameof(EventIds.ThrottlingOrchestrationHistory));
-
-        public override LogLevel Level => LogLevel.Information;
-
-        protected override string CreateLogMessage() =>
-            $"Throttling orchestration history load for instance {this.InstanceId}: {this.Details})";
-
-        void IEventSourceEvent.WriteEventSource() => AnalyticsEventSource.Log.ThrottlingOrchestrationHistory(
-            this.Account,
-            this.TaskHub,
-            this.InstanceId,
-            this.ExecutionId,
-            this.SizeInBytes,
-            this.Details,
-            Utils.AppName,
-            Utils.ExtensionVersion);
     }
 }
--- a/src/DurableTask.AzureStorage/Logging/LogEvents.cs
+++ b/src/DurableTask.AzureStorage/Logging/LogEvents.cs
@@ -2619,4 +2619,60 @@ namespace DurableTask.AzureStorage.Logging
                 Utils.ExtensionVersion);
         }
     }
+
+    internal class ThrottlingOrchestrationHistory : StructuredLogEvent, IEventSourceEvent
+    {
+        public ThrottlingOrchestrationHistory(
+            string account,
+            string taskHub,
+            string instanceId,
+            string executionId,
+            long sizeInBytes,
+            string details)
+        {
+            this.Account = account;
+            this.TaskHub = taskHub;
+            this.InstanceId = instanceId;
+            this.ExecutionId = executionId;
+            this.SizeInBytes = sizeInBytes;
+            this.Details = details;
+        }
+
+        [StructuredLogField]
+        public string Account { get; }
+
+        [StructuredLogField]
+        public string TaskHub { get; }
+
+        [StructuredLogField]
+        public string InstanceId { get; }
+
+        [StructuredLogField]
+        public string ExecutionId { get; }
+
+        [StructuredLogField]
+        public long SizeInBytes { get; }
+
+        [StructuredLogField]
+        public string Details { get; }
+
+        public override EventId EventId => new EventId(
+            EventIds.ThrottlingOrchestrationHistory,
+            nameof(EventIds.ThrottlingOrchestrationHistory));
+
+        public override LogLevel Level => LogLevel.Information;
+
+        protected override string CreateLogMessage() =>
+            $"Throttling orchestration history load for instance {this.InstanceId}: {this.Details})";
+
+        void IEventSourceEvent.WriteEventSource() => AnalyticsEventSource.Log.ThrottlingOrchestrationHistory(
+            this.Account,
+            this.TaskHub,
+            this.InstanceId,
+            this.ExecutionId,
+            this.SizeInBytes,
+            this.Details,
+            Utils.AppName,
+            Utils.ExtensionVersion);
+    }
 }
--- a/src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
@@ -33,5 +33,6 @@ namespace DurableTask.AzureStorage
         public string RuntimeStatus { get; set; }
         public DateTime? ScheduledStartTime { get; set; }
         public int Generation { get; set; }
+        public string Size { get; set; }
     }
 }
--- a/src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
@@ -29,13 +29,14 @@ namespace DurableTask.AzureStorage
         private long pendingMemory;
         private readonly object lockObject = new object();
 
-        public OrchestrationMemoryManager(AzureStorageOrchestrationServiceSettings settings, string storageAccountName, long? memoryBufferBytes = null)
+        public OrchestrationMemoryManager(AzureStorageOrchestrationServiceSettings settings, string storageAccountName, long memoryBufferBytes = 500000000)
         {
             this.settings = settings;
             this.storageAccountName = storageAccountName;
             this.messageVisibilityTimeout = settings.ControlQueueVisibilityTimeout.TotalMilliseconds;
             this.totalMemoryBytes = settings.MemoryLimitBytes;
             this.adjustedTotalMemory = this.totalMemoryBytes - memoryBufferBytes;
+            this.pendingMemory = 0;
         }
 
         async Task<bool> WaitUntilMemoryReservedAsync(OrchestrationState state, CancellationToken cancellationToken)
@@ -48,7 +49,7 @@ namespace DurableTask.AzureStorage
             double elapsed = 0;
             while (elapsed < this.messageVisibilityTimeout && !cancellationToken.IsCancellationRequested)
             {
-                var currentlyAllocatedMemory = Process.GetCurrentProcess().PrivateMemorySize64;
+                long currentlyAllocatedMemory = Process.GetCurrentProcess().PrivateMemorySize64;
 
                 lock (this.lockObject)
                 {
@@ -58,7 +59,7 @@ namespace DurableTask.AzureStorage
                         return true;
                     }
                 }
-                //TODO: can make this an exponential backoff
+
                 int delayInMs = 1000;
 
                 this.settings.Logger.ThrottlingOrchestrationHistory(
@@ -66,8 +67,7 @@ namespace DurableTask.AzureStorage
                    this.settings.TaskHubName,
                    instanceId,
                    executonId,
-                   bytesNeeded,
-                   $"Not enough memory to load orchestrator history. Retrying in {delayInMs}ms. Total elapsed time: {elapsed}. Attempt number {attemptNumber}");
+                   $"Not enough memory to load orchestrator history. Retrying in {delayInMs}ms. Total elapsed time: {elapsed}, Attempt number {attemptNumber}, Instance size: {bytesNeeded}, Currently allocated memory: {currentlyAllocatedMemory}, Pending orchestrator history memory: {this.pendingMemory}, MemoryLimitBytes: {this.totalMemoryBytes}");
 
                 await Task.Delay(delayInMs, cancellationToken);
                 elapsed += delayInMs;
--- a/src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
@@ -1,0 +1,112 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using DurableTask.AzureStorage.Tracking;
+using DurableTask.Core;
+
+namespace DurableTask.AzureStorage
+{
+    class OrchestrationMemoryManager
+    {
+        private AzureStorageOrchestrationServiceSettings settings;
+        private string storageAccountName;
+        private double messageVisibilityTimeout;
+        private long totalMemoryBytes;
+        private long? adjustedTotalMemory;
+        private long pendingMemory;
+        private readonly object lockObject = new object();
+
+        public OrchestrationMemoryManager(AzureStorageOrchestrationServiceSettings settings, string storageAccountName, long? memoryBufferBytes = null)
+        {
+            this.settings = settings;
+            this.storageAccountName = storageAccountName;
+            this.messageVisibilityTimeout = settings.ControlQueueVisibilityTimeout.TotalMilliseconds;
+            this.totalMemoryBytes = settings.MemoryLimitBytes;
+            this.adjustedTotalMemory = this.totalMemoryBytes - memoryBufferBytes;
+        }
+
+        async Task<bool> WaitUntilMemoryReservedAsync(OrchestrationState state, CancellationToken cancellationToken)
+        {
+            long bytesNeeded = state.Size;
+            string instanceId = state.OrchestrationInstance.InstanceId;
+            string executonId = state.OrchestrationInstance.ExecutionId;
+
+            int attemptNumber = 1;
+            double elapsed = 0;
+            while (elapsed < this.messageVisibilityTimeout && !cancellationToken.IsCancellationRequested)
+            {
+                var currentlyAllocatedMemory = Process.GetCurrentProcess().PrivateMemorySize64;
+
+                lock (this.lockObject)
+                {
+                    if (bytesNeeded + currentlyAllocatedMemory + this.pendingMemory <= this.adjustedTotalMemory)
+                    {
+                        this.pendingMemory += bytesNeeded;
+                        return true;
+                    }
+                }
+                //TODO: can make this an exponential backoff
+                int delayInMs = 1000;
+
+                this.settings.Logger.ThrottlingOrchestrationHistory(
+                   this.storageAccountName,
+                   this.settings.TaskHubName,
+                   instanceId,
+                   executonId,
+                   bytesNeeded,
+                   $"Not enough memory to load orchestrator history. Retrying in {delayInMs}ms. Total elapsed time: {elapsed}. Attempt number {attemptNumber}");
+
+                await Task.Delay(delayInMs, cancellationToken);
+                elapsed += delayInMs;
+                attemptNumber++;
+            }
+
+            return false;
+        }
+
+        void FreeMemoryReserve(long bytesToFree)
+        {
+            lock (this.lockObject)
+            {
+                this.pendingMemory -= bytesToFree;
+            }
+        }
+
+        public async Task<OrchestrationHistory> GetHistoryEventsAsync(ITrackingStore trackingStore, OrchestrationState state, CancellationToken cancellationToken)
+        {
+            bool memoryReserved = false;
+
+            try
+            {
+                memoryReserved = await this.WaitUntilMemoryReservedAsync(state, cancellationToken);
+
+                OrchestrationHistory history = await trackingStore.GetHistoryEventsAsync(
+                   state.OrchestrationInstance.InstanceId,
+                   state.OrchestrationInstance.ExecutionId,
+                   cancellationToken);
+
+                return history;
+            }
+            finally
+            {
+                if (memoryReserved)
+                {
+                    this.FreeMemoryReserve(state.Size);
+                }
+            }
+        }
+    }
+}
--- a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.AzureStorage
 {
     using System;
+    using System.Buffers;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
@@ -40,6 +41,7 @@ namespace DurableTask.AzureStorage
         readonly AzureStorageOrchestrationServiceStats stats;
         readonly ITrackingStore trackingStore;
         readonly DispatchQueue fetchRuntimeStateQueue;
+        readonly OrchestrationMemoryManager memoryManager;
 
         public OrchestrationSessionManager(
             string queueAccountName,
@@ -53,6 +55,8 @@ namespace DurableTask.AzureStorage
             this.trackingStore = trackingStore;
 
             this.fetchRuntimeStateQueue = new DispatchQueue(this.settings.MaxStorageOperationConcurrency);
+
+            this.memoryManager = new OrchestrationMemoryManager(settings, queueAccountName);
         }
 
         internal IEnumerable<ControlQueue> Queues => this.ownedControlQueues.Values;
@@ -469,10 +473,9 @@ namespace DurableTask.AzureStorage
             {
                 if (batch.OrchestrationState == null)
                 {
-                    OrchestrationHistory history = await this.trackingStore.GetHistoryEventsAsync(
-                       batch.OrchestrationInstanceId,
-                       batch.OrchestrationExecutionId,
-                       cancellationToken);
+                    var state = await this.trackingStore.GetStateAsync(batch.OrchestrationInstanceId, batch.OrchestrationExecutionId, false);
+
+                    OrchestrationHistory history = await this.memoryManager.GetHistoryEventsAsync(this.trackingStore, state, cancellationToken);
 
                     batch.OrchestrationState = new OrchestrationRuntimeState(history.Events);
                     batch.ETag = history.ETag;
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -808,6 +808,7 @@ namespace DurableTask.AzureStorage.Tracking
                     ["ScheduledStartTime"] = new EntityProperty(executionStartedEvent.ScheduledStartTime),
                     ["ExecutionId"] = new EntityProperty(executionStartedEvent.OrchestrationInstance.ExecutionId),
                     ["Generation"] = new EntityProperty(executionStartedEvent.Generation),
+                    ["Size"] = new EntityProperty(0),
                 }
             };
 
@@ -921,6 +922,7 @@ namespace DurableTask.AzureStorage.Tracking
                     ["CustomStatus"] = new EntityProperty(newRuntimeState.Status ?? "null"),
                     ["ExecutionId"] = new EntityProperty(executionId),
                     ["LastUpdatedTime"] = new EntityProperty(newEvents.Last().Timestamp),
+                    ["Size"] = new EntityProperty(newRuntimeState.Size),
                 }
             };
            
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -531,6 +531,7 @@ namespace DurableTask.AzureStorage.Tracking
             orchestrationState.Output = orchestrationInstanceStatus.Output;
             orchestrationState.ScheduledStartTime = orchestrationInstanceStatus.ScheduledStartTime;
             orchestrationState.Generation = orchestrationInstanceStatus.Generation;
+            orchestrationState.Size = long.Parse(orchestrationInstanceStatus.Size);
 
             if (this.settings.FetchLargeMessageDataEnabled)
             {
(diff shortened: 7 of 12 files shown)


src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        private AzureStorageOrchestrationServiceSettings settings;
+        private string storageAccountName;
+        private double messageVisibilityTimeout;
+        private long totalMemoryBytes;
+        private long? adjustedTotalMemory;
+        private long pendingMemory;
+        private readonly object lockObject = new object();
+
+        public OrchestrationMemoryManager(AzureStorageOrchestrationServiceSettings settings, string storageAccountName, long memoryBufferBytes = 500000000)
```
amdeel, line 32: I used this memory buffer value in testing but this buffer value needs further discussion. My intention is to make this an AzureStorageOrchestrationServiceSettings setting. We need to discuss what we want the default buffer value to be and how to think about additional memory needed on the worker in addition to orchestration history loading.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,112 @@
...
+                lock (this.lockObject)
+                {
+                    if (bytesNeeded + currentlyAllocatedMemory + this.pendingMemory <= this.adjustedTotalMemory)
+                    {
+                        this.pendingMemory += bytesNeeded;
+                        return true;
+                    }
+                }
+
+                int delayInMs = 1000;
```
amdeel, line 63: This is currently set to 1000ms, but this was an arbitrary value and we could consider using an exponential backoff approach here instead.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -547,6 +547,7 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                 instanceState.Status = runtimeState.Status;
             }
 
+            runtimeState.Size += SizeUtils.GetEstimatedSize(runtimeState);
```
cgillum, line 550: Why do we need to recompute the size here? Shouldn't the size get computed automatically when the history events get added?

cgillum (commented):
Oh, one thing we should absolutely do is include the size of the orchestration runtime state in some of our existing traces, like the trace for updating the orchestration instance record. That will be really useful for debugging.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -250,6 +250,11 @@ public class AzureStorageOrchestrationServiceSettings
         /// </summary>
         public TimeSpan StorageRequestsTimeoutCooldown { get; set; } = TimeSpan.FromMinutes(5);
 
+        /// <summary>
+        /// Gets or sets the memory limit in bytes.
+        /// </summary>
+        public long MemoryLimitBytes { get; set; } = (long)(1.5 * 1024 * 1024 * 1024);
```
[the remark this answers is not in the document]
jviau: Why `int` instead of `long`?

src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
```diff
@@ -33,5 +33,6 @@ class OrchestrationInstanceStatus : TableEntity
         public string RuntimeStatus { get; set; }
         public DateTime? ScheduledStartTime { get; set; }
         public int Generation { get; set; }
+        public string Size { get; set; }
```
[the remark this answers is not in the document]
jviau: Should this also be a number-type?

jviau:
> Oh, one thing we should absolutely do is include the size of the orchestration runtime state in some of our existing traces, like the trace for updating the orchestration instance record. That will be really useful for debugging.

When we use metrics, this would be a good candidate metric as well.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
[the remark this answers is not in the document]
cgillum: You typically wouldn't use `long` for MBytes. It's not a strong reason to switch to `int` but it seems like a logical choice.

src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        /// <param name="state">The object.</param>
+        /// <returns>The estimated number of bytes.</returns>
+        public static long GetEstimatedSize(OrchestrationRuntimeState state)
+        {
+            long sum = 0;
+            if (state != null)
+            {
+                sum += 120; // sum of members of constant size
+                sum += 2 * ((state.Status?.Length ?? 0) + (state.Output?.Length ?? 0) + (state.Name?.Length ?? 0) + (state.Input?.Length ?? 0) + (state.Version?.Length ?? 0));
```
davidmrdavid, line 38: why is this multiplied by 2?
In general, I think the calculations here could use some comments. It would help us debug issues in the future

src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Returns the approximate number of bytes of memory used by the given history event object.
+        /// </summary>
+        /// <param name="historyEvent">The history event object.</param>
+        /// <returns>The estimated number of bytes.</returns>
+        public static long GetEstimatedSize(HistoryEvent historyEvent)
+        {
+            long estimate = 5 * 8; // estimated size of base class
+            void AddStringSize(string s) => estimate += 8 + ((s == null) ? 0 : 2 * s.Length);
```
davidmrdavid, line 62: why are we multiplying `s.Length` by 2 again? Is it because of array resizing doubling the amount of allocated space?

src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+                    AddStringSize(eventRaisedEvent.Name);
+                    break;
+                case EventSentEvent eventSentEvent:
+                    AddStringSize(eventSentEvent.InstanceId);
+                    AddStringSize(eventSentEvent.Name);
+                    AddStringSize(eventSentEvent.Input);
+                    break;
+                case ExecutionCompletedEvent executionCompletedEvent:
+                    AddStringSize(executionCompletedEvent.Result);
+                    estimate += 8;
```
davidmrdavid, line 79: why are we adding 8 here?

src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+                case ExecutionResumedEvent executionResumedEvent:
+                    AddStringSize(executionResumedEvent.Reason);
+                    break;
+                case ExecutionStartedEvent executionStartedEvent:
+                    AddStringSize(executionStartedEvent.Input);
+                    AddStringSize(executionStartedEvent.Name);
+                    AddStringSize(executionStartedEvent.OrchestrationInstance.InstanceId);
+                    AddStringSize(executionStartedEvent.OrchestrationInstance.ExecutionId);
+                    estimate += 8 + (executionStartedEvent.Tags == null ? 0
+                        : executionStartedEvent.Tags.Select(kvp => 20 + 2 * (kvp.Key.Length + kvp.Value.Length)).Sum());
```
davidmrdavid, line 93: can you please add a comment explaining this calculation?

src/DurableTask.Core/Common/SizeUtils.cs
davidmrdavid: Same question for the other `estimate += 8` scattered in this file.

davidmrdavid (commented):
Left some comments, looking forward to this feature

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -1083,6 +1083,27 @@ public void GeneralWarning(string Account, string TaskHub, string Details, strin
                 ExtensionVersion);
         }
 
+        [Event(EventIds.ThrottlingOrchestrationHistory, Level = EventLevel.Informational, Version = 1)]
```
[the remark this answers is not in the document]
amdeel: No, if this was a warning we'd be sending a lot of false alarm signals to customers. This log can be written many times and isn't indicative of an actual issue.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+                    memorySize = state.Size;
+                }
+
+                OrchestrationHistory history = await trackingStore.GetHistoryEventsAsync(
+                   instanceId,
+                   executionId,
+                   cancellationToken);
+
+                return history;
+            }
```
davidmrdavid, line 124: if `memoryReserved` is set to `false`, perhaps due to the message visibility timeout expiring, should we not check for that before calling `GetHistoryEventAsync` in line 118?

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            return false;
+        }
+
+        void FreeMemoryFromReserve(long bytesToFree)
+        {
+            lock (this.lockObject)
+            {
+                this.pendingMemory -= bytesToFree;
+            }
```
davidmrdavid, line 175: To minimize lock contention, which can affect perf: could we remove the lock from here so that only `WaitUntilMemoryReservedAsync` threads use it, and instead update `pendingMemory` here using `Interlocked.Add` (using a negative 2nd parameter should turn it into substract)?

In think that, for this to work correctly, you also need to use `Interlocked.Add` (and perhaps `Interlocked.Read` too) inside `WaitUntilMemoryReservedAsync`, which would now would be the only method requesting the lock.

davidmrdavid (commented):
Left some thoughts and suggestions!

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -250,6 +250,21 @@ public class AzureStorageOrchestrationServiceSettings
...
 
+        /// <summary>
+        /// Enables OrchestrationHistoryLoadThrottle feature, which attempts to throttle Orchestration history load when non enough memory is available.
+        /// </summary>
+        public bool UseOrchestrationHistoryLoadThrottle { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the total process memory in MBytes when using OrchestrationHistoryLoadThrottle.
+        /// </summary>
+        public int TotalProcessMemoryMBytes { get; set; } = (int)1.5 * 1024;
```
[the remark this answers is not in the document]
amdeel: Developers know this as "process memory." Your wording sounds far more complicated than it needs to be. I'm going to keep total process memory.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -250,6 +250,21 @@ public class AzureStorageOrchestrationServiceSettings
...
+
+        /// <summary>
+        /// Gets or sets the memory buffer in MBytes when using OrchestrationHistoryLoadThrottle, representing additional memory needed in an application at a given time when loading orchestration histories.
+        /// </summary>
+        public int MemoryBufferMBytes { get; set; } = (int)0.25 * 1024;
```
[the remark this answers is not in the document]
amdeel: I'll add the words "other than." Other than that, I prefer my comment, both for clarity and sentence structure.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+                throw new InvalidOperationException($"{nameof(OrchestrationMemoryManager)} has already started");
+            }
+
+            this.currentMemoryCancellationTokenSource = new CancellationTokenSource();
+
+            this.currentMemoryUpdaterTask = await Task.Factory.StartNew(() => this.CurrentMemoryUpdater());
+        }
+
+        public async Task StopAsync()
+        {
```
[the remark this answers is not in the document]
amdeel: There's telemetry on the updater task, which is what we're actually concerned with starting and stopping, this telemetry is much more useful so we are sure the task started and stopped.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
amdeel: A lock is necessary for this use. We need to protect more than just the variable writes. I use a lock because I only want one thread to read and update the value of pendingMemory at a time. A lock protects both of those operations. If I didn't have a lock, many threads could get past the conditional in WaitUntilMemoryReservedAsync and would be able to load in orchestrator histories, even when there's only enough memory for one. A lock ensures we are working with the most accurate value of pendingMemory and threads don't get past the conditional when there isn't enough memory

Luckily, this lock doesn't affect performance in practice, there aren't enough threads to compete for the lock at the same time and the content in the lock is extremely fast.

davidmrdavid: Just want to check my thinking here:

I agree a lock is needed in WaitUntilMemoryReservedAsync because you want a read and then an increase of pendingMemory to occur without exceeding the limit.

But in FreeMemoryFromReserve you are only making pendingMemory smaller, which cannot negate your conditional check. So irrespective of how many times you call this method inbetween your conditional check and the write to pendingMemory, your conditionals result won't become invalidated (only more memory can become available every time). So while I agree you need a lock when reserving memory, I dont think you need one to free it. Am I missing something?

jviau:
I am worried about separation of concerns here - particularly some durable functions logic being inserted into the core SDK. But I think we can find a way to accomplish this. What if we make some kind of pre-fetch throttler contract for AzureStorage. The default implementation in `DurableTask.AzureStorage` will either do no throttling (todays behavior), or we could implement it with that dotnet memory-check API. Then Durable Functions can provide its own implementation which performs the more bespoke / manual memory calculations to throttle the pre-fetch of orchestrations.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+            string instanceId = state.OrchestrationInstance.InstanceId;
+            string executonId = state.OrchestrationInstance.ExecutionId;
+
+            int attemptNumber = 1;
+            double elapsed = 0;
+            while (elapsed < this.messageVisibilityTimeout && !cancellationToken.IsCancellationRequested)
+            {
+                lock (this.lockObject)
+                {
+                    if (bytesNeeded + this.currentlyAllocatedMemory + this.pendingMemory <= this.adjustedTotalMemory)
```
jviau, line 146: Because `this.currentlyAllocatedMemory` is only updated at set intervals, we could in theory allocate more than we intend to, correct? As we are always working on a 'stale' memory value. I don't see us solving this without checking memory before _every_ history load. And I am already concerned about performance impact of the getting current process memory as is.

Would be good to call out this possibility in our documentation for this feature.

src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using DurableTask.Core;
+    using DurableTask.Core.History;
+
+    /// <summary>
+    /// Functionality for estimating the number of bytes of memory taken up by some commonly used objects.
+    /// </summary>
+    internal class SizeUtils
```
jviau, line 25: Theoretically all of this size calculation is CPU architecture dependent - 32bit vs 64bit, or whatever else CPUs decide to have. Is this worth factoring in? I think in cases like this, you have a static 'size' before account for architecture, then factor in architecture via `IntPtr.Size` after.

jviau:
Something that would make me feel safer with this change: can we introduce a _minimum_ prefetch? Like if there is nothing in the `readyToProcessQueue` we will force-prefetch at least one regardless of memory. This way we can prevent throughput going to 0.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        async Task<bool> WaitUntilMemoryReservedAsync(OrchestrationState state, CancellationToken cancellationToken)
+        {
+            long bytesNeeded = state.Size;
+            string instanceId = state.OrchestrationInstance.InstanceId;
+            string executonId = state.OrchestrationInstance.ExecutionId;
+
+            int attemptNumber = 1;
+            double elapsed = 0;
+            while (elapsed < this.messageVisibilityTimeout && !cancellationToken.IsCancellationRequested)
```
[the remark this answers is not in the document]
amdeel: What would the benefit be of adding an extra queue layer here? We already don't make guarantees about message order, so I'm not seeing any benefit.

src/DurableTask.AzureStorage/Logging/LogEvents.cs
```diff
@@ -2654,5 +2659,94 @@ internal class PurgeInstanceHistory : StructuredLogEvent, IEventSourceEvent
...
+            public string ExecutionId { get; }
+
+            [StructuredLogField]
+            public string Details { get; }
+
+            public override EventId EventId => new EventId(
+                EventIds.ThrottlingOrchestrationHistoryLoad,
+                nameof(EventIds.ThrottlingOrchestrationHistoryLoad));
+
+            public override LogLevel Level => LogLevel.Information;
```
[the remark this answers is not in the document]
amdeel: I don't think we want a warning here. It would add a lot of false alarm for customers and clutter logs when investigating issues.

amdeel:
@jviau I strongly disagree with that suggestion. If memory is being throttled it means there is no memory available and OutOfMemoryExceptions are being avoided. If we ignore those limits and load another orchestrator into memory we're not getting value from having the memory manager. We would be knowingly causing exceptions. The benefit of having this feature is to wait for available memory. The readyToProcessQueue could have 0 messages before the previous messages have even been processed.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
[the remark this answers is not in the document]
davidmrdavid: Sounds good, I think it would help me to see the suggested rewording in full, so I'll wait for your edit.

src/DurableTask.AzureStorage/Tracking/InstancesTableCache.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                foreach (KeyValuePair<string, EntityProperty> property in entity.Properties)
+                {
+                    if (storedEntity.Properties.ContainsKey(property.Key))
+                    {
+                        storedEntity.Properties[property.Key] = property.Value;
+                    }
+                    else
+                    {
+                        storedEntity.Properties.Add(property.Key, property.Value);
```
davidmrdavid, line 49: should we have a check avoiding "input" from being added as a property?

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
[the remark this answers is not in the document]
davidmrdavid: I see what you mean. Sounds good, I'll resolve this comment then.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            this.currentMemoryCancellationTokenSource = new CancellationTokenSource();
+
+            this.currentMemoryUpdaterTask = await Task.Factory.StartNew(() => this.CurrentMemoryUpdater());
+        }
+
+        public async Task StopAsync()
+        {
+            if (Interlocked.CompareExchange(ref this.isStarted, 0, 1) != 1)
```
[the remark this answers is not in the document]
davidmrdavid: @jviau - are you saying that, for thread-safety, Start needs to run to completion before an invocation Stop begins? That makes sense to me and, now that I think about it, there might be other places in the Extension code where we have this race condition. I think there's similar logic on the partition manager Start and Stop methods: https://github.com/Azure/durabletask/blob/68a4a0a80e6dc5d7744d67eac6e56fcad472dcf9/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs#L113-L154

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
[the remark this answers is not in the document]
davidmrdavid: Sounds to me like the main benefit is not in message ordering, but in avoiding the "busy work" of looping here until memory is available. Instead, this thread would be unblocked by the thread that calculates the current process memory whenever there's sufficient memory available for the request.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -416,6 +419,23 @@ public override async Task<IList<OrchestrationState>> GetStateAsync(string insta
...
+                    this.settings.Logger.FetchedInstanceStatusFromCache(
+                        this.storageAccountName,
+                        this.taskHubName,
+                        instanceId,
+                        orchestrationState.OrchestrationInstance.ExecutionId ?? string.Empty,
+                        orchestrationState.Size);
+
+                    return orchestrationState;
+                }
+            }
```
davidmrdavid, line 437: Why is this under "!fetchInput"? Not sure I understand what this parameter is controlling.

src/DurableTask.AzureStorage/MRUCache.cs
```diff
@@ -0,0 +1,92 @@
...
+                    this.list.AddFirst(node);
+                    node.Value = (key, value);
+                }
+                else
+                {
+                    if (this.cache.Count == capacity)
+                    {
+                        var last = this.list.Last;
+                        this.list.RemoveLast();
+                        this.cache.Remove(last.Value.key);
```
davidmrdavid, line 48: I haven't though about eviction policies in a bit but - if we're Add ang Get moves the node to the beginning of the list, and removing elements from the end whenever we reach capacity, isn't this an LRU cache?

davidmrdavid (commented):
A few more thoughts and questions :)

src/DurableTask.AzureStorage/Logging/LogEvents.cs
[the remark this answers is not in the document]
amdeel: My hesitation is that I expect these logs to get written multiple times when a throttling event occurs. Not only will multiple messages likely get affected at a time, but they'll continue to get written until the message is no longer blocked. We will likely see many of these logs for each message being blocked.

src/DurableTask.AzureStorage/Tracking/InstancesTableCache.cs
```diff
@@ -0,0 +1,67 @@
...
+        public InstancesTableCache(AzureStorageOrchestrationServiceSettings settings)
+        {
+            this.settings = settings;
+            this.numberOfRecords = 2400; //settings.InstanceTableCacheSize;
+            this.mruCache = new MRUCache<string, DynamicTableEntity>(numberOfRecords);
+        }
+
+        public void AddToCache(DynamicTableEntity entity)
+        {
+            entity.Properties.Remove("Input");
```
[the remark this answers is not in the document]
amdeel: The input is large. Do you still think we need a comment for that?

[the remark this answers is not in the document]
davidmrdavid: Yeah, I'd appreciate a comment. I don't think it's immediately why we want to remove it. I can imagine someone undoing this change in a future refactoring if we don't write it down.

src/DurableTask.AzureStorage/Logging/LogEvents.cs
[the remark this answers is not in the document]
davidmrdavid: I see, that's a good point. I agree that it wouldn't be good to have this as a warning if we see this for multiple messages **and** also multiple times per message. 

For that second part, i.e logging warnings multiple times per message, I think we can mitigate that concern with @jviau's suggestion here: https://github.com/Azure/durabletask/pull/845#discussion_r1152561280  assuming I understand it correctly.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
amdeel: If we don't need the input when accessing the instances table we can use the cache instead.

src/DurableTask.AzureStorage/MRUCache.cs
amdeel: I actually wondered the same thing. When I've heard the term LRU/MRU cache I've heard it based on the eviction policy. Chris wrote this. I'll update this.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
amdeel: Locking this simply protects both writes to this variable.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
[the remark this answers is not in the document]
amdeel: Jacob's suggesting that instead of using both total memory and memory buffer as a setting, we only give them one setting and have them calculate their own "adjusted total" memory.

However, I think the approach to keep these separate works best for a few reasons. It might be more confusing for customers to understand an adjusted total memory and potentially make a mistake when inputting this value. It also allows us to input our own value for total process memory based on the worker they're running on, but still allows customers to manually enter their own buffer value depending on their app needs.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
davidmrdavid: I see. I think to intuitively understand this, the reader needs to know in advance that the cache explicitly drops the input, which isn't immediately obvious to me: I'd expect the cache to contain the data unmodified. Can you please add a guiding comment here reminding that the cache drops the input field? I think it'll help with readability.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
davidmrdavid: Hmm, I think we're talking past each other here. Since performance is already being addressed by the cache, let's dismiss this suggestion for now. Thank you for considering it.

src/DurableTask.AzureStorage/Logging/LogEvents.cs
```diff
@@ -2654,5 +2660,134 @@ internal class PurgeInstanceHistory : StructuredLogEvent, IEventSourceEvent
...
+            public string ExecutionId { get; }
+
+            [StructuredLogField]
+            public long Size { get; }
+
+            public override EventId EventId => new EventId(
+                EventIds.FetchedInstanceStatusFromCache,
+                nameof(EventIds.FetchedInstanceStatusFromCache));
+
+            public override LogLevel Level => LogLevel.Information;
```
cgillum, line 2778: This feels like a verbose log to me? Reading something from a cache doesn't seem particularly noteworthy.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,193 @@
...
+                    memoryReserved = await this.WaitUntilMemoryReservedAsync(state, cancellationToken);
+                    memorySize = state.Size;
+                }
+
+                if (memoryReserved || !this.settings.UseOrchestrationHistoryLoadThrottle)
+                {
+                    OrchestrationHistory history = await trackingStore.GetHistoryEventsAsync(
+                   instanceId,
+                   executionId,
+                   cancellationToken);
```
cgillum, line 127: nit: indentation needs to be fixed here.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -470,10 +474,15 @@ bool IsScheduledAfterInstanceUpdate(MessageData msg, OrchestrationState? remoteI
...
-                       batch.OrchestrationExecutionId,
-                       cancellationToken);
+                    OrchestrationHistory history = await this.memoryManager.GetHistoryEventsAsync(this.trackingStore,
+                            batch.OrchestrationInstanceId,
+                            batch.OrchestrationExecutionId,
+                            cancellationToken);
+
+                    if (history == null)
+                    {
+                        return;
```
cgillum, line 484: Does a `null` history mean that we were throttled? If so, can you add a comment above this `return` statement confirming this?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -117,6 +118,8 @@ Table instancesTable
 
         internal Table InstancesTable { get; }
 
+        private InstancesTableCache InstancesTableCache;
```
cgillum, line 121: nit: this should be moved up, above the constructor, alongside all the other private fields. It should also start with a lowercase letter (`instancesTableCache`).

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
cgillum: I wonder if naming the parameter `canUseCache` and reversing the Boolean logic might make it less confusing? That might also help future callers think more carefully about how they use this parameter since I assume there could be more differences than just whether or not an input is included.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -837,6 +864,7 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string
                     ["ScheduledStartTime"] = new EntityProperty(executionStartedEvent.ScheduledStartTime),
                     ["ExecutionId"] = new EntityProperty(executionStartedEvent.OrchestrationInstance.ExecutionId),
                     ["Generation"] = new EntityProperty(executionStartedEvent.Generation),
+                    ["Size"] = new EntityProperty((long)0),
```
cgillum, line 867: Pro-tip:
```suggestion
                    ["Size"] = new EntityProperty(0L),
```

src/DurableTask.Core/Common/SizeUtils.cs
sebastianburckhardt: Just wanted to mention that all the code in this class is highly approximate and based on guessing.

Computing an accurate size of everything we have in memory is prohibitely complicated or even impossible since a lot of it is opaque, i.e. depends on internals of libraries or the CLR.

I think for our purpose it does not have to be all that accurate, "ballpark" should be good enough - the important part is that we can detect if something is particularly large, and handle it more carefully.

I believe (hope) I used 8 bytes as the assumed pointer size throughout.

src/DurableTask.Core/Common/SizeUtils.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// Returns the approximate number of bytes of memory used by the given orchestration state object.
+        /// </summary>
+        /// <param name="state">The object.</param>
+        /// <returns>The estimated number of bytes.</returns>
+        public static long GetEstimatedSize(OrchestrationRuntimeState state)
+        {
+            long sum = 0;
+            if (state != null)
+            {
+                sum += 120; // sum of members of constant size
```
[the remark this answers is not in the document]
sebastianburckhardt: yes and yes.

(see note above on "ballpark" estimates).

Most fields would be harmless to miss, though I do worry a bit about the new distributed tracing fields which could be large.

src/DurableTask.AzureStorage/Tracking/InstancesTableCache.cs
```diff
@@ -0,0 +1,64 @@
...
+        private MRUCache<string, DynamicTableEntity> mruCache;
+
+        public InstancesTableCache()
+        {
+            this.mruCache = new MRUCache<string, DynamicTableEntity>(NumberOfRecords);
+        }
+
+        public void AddToCache(DynamicTableEntity entity)
+        {
+            entity.Properties.Remove("Input");
```
cgillum, line 30: This is dangerous because you're mutating an object in a way that could impact the caller of this method without them knowing. Consider instead creating a copy of this item and removing fields only from the copy.

src/DurableTask.Core/Common/SizeUtils.cs
sebastianburckhardt: 8 is the assumed size of a pointer.

src/DurableTask.AzureStorage/LRUCache.cs
```diff
@@ -0,0 +1,92 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage
+{
+    using System.Collections.Generic;
+
+    internal class LRUCache<TKey, TValue>
```
cgillum, line 18: Is this in use? I see that the InstancesTableCache is using `MRUCache` and not `LRUCache`.

src/DurableTask.AzureStorage/Tracking/InstancesTableCache.cs
```diff
@@ -0,0 +1,64 @@
...
+//  limitations under the License.
+
+namespace DurableTask.AzureStorage.Tracking
+{
+    using System.Collections.Generic;
+    using Microsoft.WindowsAzure.Storage.Table;
+
+    class InstancesTableCache
+    {
+        private const int NumberOfRecords = 2400;
```
cgillum, line 20: I suggest adding a comment saying that the number 2400 is based on performance testing, and that the impact of having a large number shouldn't be too high (I assume these statements are true 😊).

src/DurableTask.AzureStorage/Tracking/InstancesTableCache.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            if (this.mruCache.TryGet(entity.PartitionKey, out DynamicTableEntity storedEntity))
+            {
+                foreach (KeyValuePair<string, EntityProperty> property in entity.Properties)
+                {
+                    if (storedEntity.Properties.ContainsKey(property.Key))
+                    {
+                        storedEntity.Properties[property.Key] = property.Value;
+                    }
+                    else if (property.Key != "Input")
```
cgillum, line 44: I suggest we make a static list of fields we want to exclude and use a `Contains` operation here on that list. We can also use that list to filter in `AddToCache`. That way if we need to add more fields to check against, we can do it in one place.

For example, it looks like we also need to filter out `CustomStatus` and `Output`.

src/DurableTask.AzureStorage/Tracking/InstancesTableCache.cs
```diff
@@ -0,0 +1,64 @@
...
+                    {
+                        storedEntity.Properties.Add(property.Key, property.Value);
+                    }
+                }
+
+                this.mruCache.Add(storedEntity.PartitionKey, storedEntity);
```
cgillum, line 50: An entity with this key already exists in the cache so it looks like you're adding a redundant entry into the cache here. Is this intentional or am I misunderstanding the code? It seems to me that you've already updated the existing entry.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1097,6 +1131,7 @@ public override Task StartAsync()
 
             Stopwatch orchestrationInstanceUpdateStopwatch = Stopwatch.StartNew();
             await this.InstancesTable.InsertOrMergeAsync(instanceEntity);
+            this.InstancesTableCache.AddToCache(instanceEntity);
```
cgillum, line 1134: Why are we adding to the cache instead of updating?  And if the orchestration has completed, failed, or terminated, should we be removing the entry from the cache?

cgillum (changes requested):
I re-reviewed everything except for OrchestrationMemoryManager, which seems to have a bunch of unresolved comments already. The main things to highlight are some questions on how we manage the cache itself as there may be some leaks/bugs. Please see my comments.

src/DurableTask.AzureStorage/LRUCache.cs
amdeel: It looks like the variable name in InstanceTableCache wasn't updated. Done.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
amdeel: Add will replace the current value if that instance already exists. I chose add here because the insert or merge led me to believe that there's a chance the instance doesn't exist in the Instances Table yet. Looking at the code, it doesn't look like this method gets called anywhere but on orchestration completion actions. However, there's technically the case where partitions have been rebalanced and this worker doesn't yet know about this instance, so I think add makes sense here.

Currently adding to the instances table or updated the value is the only thing that triggers an add to the instances table cache.

I think it does make sense to remove the instances in completed states from the cache. I'll add that.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,178 @@
...
+                await Task.Delay(10, this.currentMemoryCancellationTokenSource.Token);
+            }
+
+            this.settings.Logger.OrchestrationMemoryManagerInfo(
+                this.storageAccountName,
+                this.taskHub,
+                $"Background updater for currenly allocated memory completed.");
+        }
+
+        public async Task<OrchestrationHistory> GetHistoryEventsAsync(ITrackingStore trackingStore, string instanceId, string executionId, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
amdeel: Initially I had it there but I thought it made sense to push it to the memory manager. The added benefit is that the memory manager can handle its own pieces, like freeing memory, and ATTS can keep things as simple as possible.

src/DurableTask.AzureStorage/AnalyticsEventSource.cs
```diff
@@ -1115,14 +1115,15 @@ public void GeneralWarning(string Account, string TaskHub, string Details, strin
...
-               ExtensionVersion);
+                EventIds.OrchestrationMemoryManagerInfo,
+                Account,
+                TaskHub,
+                Details,
+                AppName,
+                ExtensionVersion);
         }
 
+        [Event(EventIds.FetchedInstanceStatusFromCache, Level = EventLevel.Informational, Version = 1)]
```
cgillum, line 1126: Consider making this verbose. Most customers won't be interested in this trace. It's mostly for our own investigation purposes.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -78,15 +79,16 @@ public async Task StopAsync()
                 await this.currentMemoryUpdaterTask;
             }
 
+            this.currentMemoryCancellationTokenSource.Dispose();
```
cgillum, line 82: I recommend we guard against null-refs here.
```suggestion
            this.currentMemoryCancellationTokenSource?.Dispose();
```

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -119,12 +121,12 @@ public async Task<OrchestrationHistory> GetHistoryEventsAsync(ITrackingStore tra
...
-                if (memoryReserved || !this.settings.UseOrchestrationHistoryLoadThrottle)
+                if (memoryReserved || !this.settings.MemoryThrottleSettings.UseOrchestrationHistoryLoadThrottle)
                 {
                     OrchestrationHistory history = await trackingStore.GetHistoryEventsAsync(
-                   instanceId,
-                   executionId,
-                   cancellationToken);
+                    instanceId,
+                    executionId,
+                    cancellationToken);
```
cgillum, line 129: nit: these arguments need to be indented further.

cgillum (commented):
A few more comments based on the latest iteration. There are still a few unresolved comments from my previous iteration, though.

src/DurableTask.AzureStorage/MRUCache.cs
davidmrdavid: @cgillum: I'd appreciate it if you could chime in on this: if Add and Get moves the node to the beginning of the list, and removing nodes from the end whenever we reach capacity, isn't this a LRU cache?

cgillum: Yeah, I just looked up "LRU vs MRU" and it seems like this should be considered an LRU cache. It was 12+ years ago when I acquired the original version of this code, and somebody (maybe me?) might have had it backwards.

src/DurableTask.AzureStorage/OrchestrationMemoryManager.cs
```diff
@@ -0,0 +1,195 @@
...
+                {
+                    OrchestrationHistory history = await trackingStore.GetHistoryEventsAsync(
+                    instanceId,
+                    executionId,
+                    cancellationToken);
+
+                    return history;
+                }
+
+                return null;
```
davidmrdavid, line 134: When do we expect to return `null` here?
If it's expected in a valid scenario: can we please add a comment documenting it.
If it's not supposed to happen, should we just throw an exception?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -909,7 +916,8 @@ public override async Task UpdateStatusForRewindAsync(string instanceId)
                 string.Empty,
                 OrchestrationStatus.Pending,
                 currentEpisodeNumber,
-                stopwatch.ElapsedMilliseconds);
+                stopwatch.ElapsedMilliseconds,
+                null);
```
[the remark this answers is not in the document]
davidmrdavid: Thanks for the explanation.  For context: the log where we pass in `0` is this one: https://github.com/Azure/durabletask/blob/2cf569afa7188d08228bfdf7052377421ba37ac1/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L901-L911

With that log in mind -  why shouldn't both of these be `0` or both be `null`? :)

davidmrdavid (commented):
Left a few more comments :)

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
davidmrdavid: I like `canUseCache`

torosent:

PR #846, opened 2023-01-13
.Net core 6.0 support

vandanavisw:
This is a diff for supporting DurableTask.AzureServiceFabric nuget with .net core. This was tested for .net6.0.
Following major changes were made:

- Changed listener from Owin to Kestrel, as Owin does not have .net core support.

- Changed Fabric Controller's inherited class from ApiController to Base controller

- Added NewtonSoft.Json formatter in controller to parse data from Client, as Client was parsing it with NewtonSoft.Json

- Changed https listener url to IP:Port\{partitionId}\{instanceId}


src/DurableTask.AzureServiceFabric/DurableTask.AzureServiceFabric.csproj
```diff
@@ -34,13 +35,21 @@
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
```
lchew7, line 38: Why is this condition necessary if this project is targeting net6.0?  Is it because it might be included by other projects which target different frameworks?

src/DurableTask.AzureServiceFabric/FabricOrchestrationProviderSettings.cs
```diff
@@ -42,6 +42,11 @@ public FabricOrchestrationProviderSettings()
             };
         }
 
+        /// <summary>
+        ///     Settings to configure the SSLSubjectName
```
lchew7, line 46: Minor: I know you're just following the existing convention, but this comment is not adding much.  I would either remove it or add more details (e.g., client cert subject name that is used for SSL communication)

src/DurableTask.AzureServiceFabric/DurableTask.AzureServiceFabric.csproj
vandanavisw: Yes this is mainly for backward compatibility

src/DurableTask.AzureServiceFabric/FabricOrchestrationProviderSettings.cs
vandanavisw: Can make the comment more explicit.

PR #847, opened 2023-01-14
Add Distributed Tracing spec

bachuv:
This PR adds a spec that outlines the semantic conventions we are using to implement distributed tracing in the DurableTask Framework. It follows the [Open Telemetry Trace Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions "Open Telemetry Trace Semantic Conventions").

Additional information:
The image below is an example of a distributed trace created from the current instrumentation implementation. Each blue or red bar (solid and dashed) is a **span**. The properties on the right are the span's **attributes**. The group of spans in this image make up a **trace**.

image

### DIFF
--- a/docs/telemetry/traces/semantic-conventions.md
+++ b/docs/telemetry/traces/semantic-conventions.md
@@ -10,6 +10,17 @@ This document serves as a specification for Distributed Tracing in the DurableTa
 ## Specification
 The following only applies to `DurableTask.Core`
 
+## Table of Contents
+1. [Attributes](#attributes)
+2. Spans <br>
+    a. [Client: Starting an Orchestration](#client-starting-an-orchestration)<br>
+    b. [Client: Starting a sub-orchestration](#client-starting-a-sub-orchestration)<br>
+    c. [Worker: Running an Orchestration](#worker-running-an-orchestration)<br>
+    d. [Worker: Starting an Activity](#worker-starting-an-activity)<br>
+    e. [Worker: Running an Activity](#worker-running-an-activity)<br>
+    f. [Worker: Timer](#worker-timer)<br>
+    g. [Sending an Event](#sending-an-event)<br>
+
 ### Attributes
 These attributes are all new, defined by us, and specific to DurableTask. All attributes here begin with the “durabletask” prefix.
 
@@ -30,7 +41,7 @@ durabletask.event.instance_id | string | The target orchestration instance ID of
 [2] When instance ID is not available (e.g. activities) use the parent orchestration's instance ID
 
 ### Spans
-#### Client – Starting an Orchestration
+#### Client: Starting an Orchestration
 
 Represents enqueueing a new orchestration via TaskHubClient to the backend.
 
@@ -58,7 +69,7 @@ Attributes:
 | durabletask.task.execution_id  | Execution ID of the enqueued orchestration  |
 | exception.*  | Exception details on failure. |
 
-#### Client - Starting a sub-orchestration
+#### Client: Starting a sub-orchestration
 
 Represents enqueueing and waiting for a sub-orchestration to complete 
 
@@ -85,7 +96,7 @@ Attributes:
 | exception.*   | Exception details on failure |
 
 
-#### Worker – Running an Orchestration
+#### Worker: Running an Orchestration
 
 Represents enqueueing a new orchestration via TaskHubClient to the backend
 
@@ -113,7 +124,7 @@ Attributes:
 | exception.*   | Exception details on failure |
 
 
-#### Worker - Starting an Activity
+#### Worker: Starting an Activity
 
 Represents enqueueing an activity
 
@@ -140,7 +151,7 @@ Attributes:
 | durabletask.task.task_id   | ID of the current task  |
 | exception.*   | Exception details on failure |
 
-#### Worker - Running an Activity
+#### Worker: Running an Activity
 
 Represents the activity executing
 
@@ -168,7 +179,7 @@ Attributes:
 | exception.*   | Exception details on failure |
 
 
-#### Worker - Timer
+#### Worker: Timer
 
 Represents the Durable Timer
 
--- a/docs/telemetry/traces/semantic-conventions.md
+++ b/docs/telemetry/traces/semantic-conventions.md
@@ -33,8 +33,9 @@ durabletask.event.instance_id | string | The target orchestration instance ID of
 
 Represents enqueueing a new orchestration via TaskHubClient to the backend.
 
-| Name  | create\_orchestration\|\|\{orchestrationName\}\(\|\|\{orchestrationVersion\}\)?  |
+| Span Property | Span Value |
 |---|---|
+| Name  | `create_orchestration||{orchestrationName}(||{orchestrationVersion})?`  |
 | Kind  | Producer  |
 | Start  | Before submitting the new orchestration to the backend  |
 | End  | After the orchestration has been submitted to the backend  |
@@ -56,19 +57,16 @@ Attributes:
 | durabletask.task.execution_id  | Execution ID of the enqueued orchestration  |
 | exception.*  | Exception details on failure. |
 
-Note: Backends SHOULD enrich this activity with additional attributes, links, or events as appropriate.
-
 #### Client - Starting a sub-orchestration
 
 Represents enqueueing and waiting for a sub-orchestration to complete 
 
-| Name   | orchestration\|\|\{orchestrationName\}\(\|\| \{orchestrationVersion\}\)?   |
+| Span Property | Span Value |
 |---|---|
-| Kind   | Standard: Client   |
-|  | Fire and forget: Producer  |
+| Name   | `orchestration||{orchestrationName}(||{orchestrationVersion})?`   |
+| Kind   | Standard: Client<br />Fire and forget: Producer  |
 | Start   | When the parent orchestration enqueues a message to execute the sub\-orchestration  |
-| End   | Standard: When the parent orchestration is notified that the sub\-orchestration completed\.  |
-|  | Fire and forget: After the sub\-orchestration started event is enqueued  |
+| End   | Standard: When the parent orchestration is notified that the sub\-orchestration completed\.<br />Fire and forget: After the sub\-orchestration started event is enqueued  |
 | Status   | OK – successfully enqueued<br />Error – failed to enqueue for any reason   |
 | Links   | None   |
 | Events   | TBD   |
@@ -90,8 +88,9 @@ Attributes:
 
 Represents enqueueing a new orchestration via TaskHubClient to the backend
 
-| Name   | orchestration\|\|\{orchestrationName\}\(\|\|\{orchestrationVersion\}\)?   |
+| Span Property | Span Value |
 |---|---|
+| Name   | `orchestration||{orchestrationName}(||{orchestrationVersion})?`   |
 | Kind   | Server   |
 | Start   | Before the orchestration starts executing\.  |
 | End   | After the orchestration has finished executing\.   |
@@ -117,8 +116,9 @@ Attributes:
 
 Represents enqueueing an activity
 
-| Name   | activity\|\|\{activityName\}\(\|\|\{orchestrationVersion\}\)?   |
+| Span Property | Span Value |
 |---|---|
+| Name   | `activity||{activityName}(||{orchestrationVersion})?`   |
 | Kind   | Client   |
 | Start   | Before enqueuing the activity  |
 | End   | After the activity has finished executing\.   |
@@ -143,8 +143,9 @@ Attributes:
 
 Represents the activity executing
 
-| Name   | activity\|\|\{activityName\}\(\|\| \{orchestrationVersion\}\)?   |
+| Span Property | Span Value |
 |---|---|
+| Name   | `activity||{activityName}(|| {orchestrationVersion})?`   |
 | Kind   | Server   |
 | Start   | Before the activity starts executing\.  |
 | End   | After the activity has finished executing\.   |
@@ -170,8 +171,9 @@ Attributes:
 
 Represents the Durable Timer
 
-| Name    | timer    |
+| Span Property | Span Value |
 |---|---|
+| Name    | `timer`    |
 | Kind    | Internal    |
 | Start    | When the timer is created   |
 | End    | When the TimerFired event is processed    |
@@ -196,8 +198,9 @@ Attributes:
 
 Represents sending an event to an orchestration
 
-| Name    | orchestration\_event\|\|\{orchestration\_name\}  |
+| Span Property | Span Value |
 |---|---|
+| Name    | `orchestration_event||{orchestration_name}`  |
 | Kind    | Client  |
 | Start    | Before sending the event  |
 | End    | From client: when the event has been sent <br />From worker: when the event message has been created \(this will have a negligible duration, but we want a span for a link\) |
@@ -216,27 +219,3 @@ Attributes:
 | durabletask.event.instance_id  | The instance ID of the target orchestration (the one receiving the event).    |
 | durabletask.task.instance_id    | Instance ID of the orchestration that is sending the event. Not present if sent from the client.  |
 | durabletask.task.execution_id   | Execution ID of the orchestration that is sending the event. Not present if sent from the client. |
-
-#### Receiving an Event
-
-Represents an orchestration receiving an event
-
-| Name    | orchestration\_event\|\|\{orchestration\_name\}  |
-|---|---|
-| Kind    | Consumer  |
-| Start    | When the event message is raised  |
-| End    | After the event is processed  |
-| Status    | OK – event received<br />Error – event failed to be read  |
-| Links    | The orchestration span that produced this event  |
-| Events    | None  |
-| Parent    | Current active span\. |
-
-&nbsp;
-Attributes:
-
-| Name    | Value    |
-|---|---|
-| durabletask.type    | “event”    |
-| durabletask.event.name  | The name of the event being received.  |
-| durabletask.task.instance_id    | Instance ID of the orchestration that received the event.  |
-| durabletask.task.execution_id   | Execution ID of the orchestration that received the event. |
--- a/docs/telemetry/traces/semantic-conventions.md
+++ b/docs/telemetry/traces/semantic-conventions.md
@@ -1,26 +1,27 @@
 # DurableTask Distributed Tracing Specification
+**Status**: [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
 
 ## Overview
-This document is to serve as a specification for distributed tracing in DurableTask Framework. This will include a list of all spans, their attributes, and detailed information about what constitutes each respective span.
+This document serves as a specification for distributed tracing in the DurableTask Framework. It includes a list of all spans, their attributes, and detailed information about what constitutes each respective span. We want to gather community feedback and encourage any comments with suggestions and questions. 
 
 ## References
-- [Open Telemetry Trace Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions "Open Telemetry Trace Semantic Conventions") – various trace semantic conventions we will be referencing throughout this document. Particularly, we will be pulling in some attributes defined in these files.
+- [Open Telemetry Trace Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions "Open Telemetry Trace Semantic Conventions") – various trace semantic conventions we  reference throughout this document. Particularly, we pull in some attributes defined in these files.
 
 ## Specification
 The following only applies to DurableTask.Core
 
 ### Attributes
-These attributes are all new, defined by us, and specific to DurableTask Framework. All attributes here will begin with the “durabletask” prefix.
+These attributes are all new, defined by us, and specific to DurableTask Framework. All attributes here begin with the “durabletask” prefix.
 
 Name | Type | Description | Requirement Level
 ---|---|---|---
 durabletask.type | string | The type of durable event occurring. Options: “activity”, “orchestration”, “timer”, “event” | Required
-durabletask.task.name | string | The name of the durable task being invoked. This is the name it was queued with, not necessarily the name of the handler class/type. | Required
-durabletask.task.version | string | The version of the task being ran.  | Conditionally Required [1]
-durabletask.task.instance_id |  string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
-durabletask.task.execution_id | string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
+durabletask.task.name | string | The name of the durable task being invoked. This is the name it was scheduled with, not necessarily the name of the handler class/type. | Required
+durabletask.task.version | string | The version of the task being executed.  | Conditionally Required [1]
+durabletask.task.instance_id |  string | The execution ID of the task being ran. This is unique for all tasks. | Required [2]
+durabletask.task.execution_id | string | The execution ID of the task being ran. This is unique for all tasks. | Required [2]
 durabletask.task.task_id | integer | The ID / index of the task within the orchestration. | Conditionally required
-durabletask.task.result | string | The terminal status of a task. Options: “Succeeded”, “Failed”, “Terminated”, "Suspended" | Required
+durabletask.task.result | string | The terminal status of a task. Options: “Succeeded”, “Failed”, “Terminated” | Required
 durabletask.event.name | string | The name of the event being raised. | Required
 durabletask.event.instance_id | string | The target orchestration instance ID of a raised event. | Required
 
@@ -37,8 +38,8 @@ Represents enqueueing a new orchestration via TaskHubClient to the backend.
 |---|---|
 | Name  | `create_orchestration\|\|{orchestrationName}(\|\|{orchestrationVersion})?`  |
 | Kind  | Producer  |
-| Start  | Before submitting the new orchestration to the backend  |
-| End  | After the orchestration has been submitted to the backend  |
+| Start  | Timestamp before submitting the new orchestration to the backend  |
+| End  | Timestamp after the orchestration has been submitted to the backend  |
 | Status  | OK – successfully enqueued<br />Error – failed to enqueue for any reason  |
 | Status Description  | On failure – exception message which caused failure  |
 | Links  | None  |
@@ -65,8 +66,8 @@ Represents enqueueing and waiting for a sub-orchestration to complete
 |---|---|
 | Name   | `orchestration\|\|{orchestrationName}(\|\|{orchestrationVersion})?`   |
 | Kind   | Standard: Client<br />Fire and forget: Producer  |
-| Start   | When the parent orchestration enqueues a message to execute the sub\-orchestration  |
-| End   | Standard: When the parent orchestration is notified that the sub\-orchestration completed\.<br />Fire and forget: After the sub\-orchestration started event is enqueued  |
+| Start   | Timestamp of when the parent orchestration enqueues a message to execute the sub\-orchestration  |
+| End   | Standard: Timestamp of when the parent orchestration is notified that the sub\-orchestration completed\.<br />Fire and forget: Timestamp after the sub\-orchestration started event is enqueued  |
 | Status   | OK – successfully enqueued<br />Error – failed to enqueue for any reason   |
 | Links   | None   |
 | Events   | TBD   |
@@ -92,8 +93,8 @@ Represents enqueueing a new orchestration via TaskHubClient to the backend
 |---|---|
 | Name   | `orchestration\|\|{orchestrationName}(\|\|{orchestrationVersion})?`   |
 | Kind   | Server   |
-| Start   | Before the orchestration starts executing\.  |
-| End   | After the orchestration has finished executing\.   |
+| Start   | Timestamp before the orchestration starts executing\.  |
+| End   | Timestamp after the orchestration has finished executing\.   |
 | Status   | OK – successfully completed<br />Error – failed  |
 | Links   | Spans from “Sending an Event”\.  |
 | Events   | Received events – name of the event, and timestamp of the first time this event has been played  |
@@ -120,8 +121,8 @@ Represents enqueueing an activity
 |---|---|
 | Name   | `activity\|\|{activityName}(\|\|{orchestrationVersion})?`   |
 | Kind   | Client   |
-| Start   | Before enqueuing the activity  |
-| End   | After the activity has finished executing\.   |
+| Start   | Timestamp before enqueuing the activity  |
+| End   | Timestamp after the activity has finished executing\.   |
 | Status   | OK – successfully enqueued<br />Error – failed to enqueue for any reason   |
 | Links   | None   |
 | Events   | TBD   |
@@ -147,8 +148,8 @@ Represents the activity executing
 |---|---|
 | Name   | `activity\|\|{activityName}(\|\| {orchestrationVersion})?`   |
 | Kind   | Server   |
-| Start   | Before the activity starts executing\.  |
-| End   | After the activity has finished executing\.   |
+| Start   | Timestamp before the activity starts executing\.  |
+| End   | Timestamp after the activity has finished executing\.   |
 | Status   | OK – successfully completed<br />Error – failed  |
 | Links   | None   |
 | Events   | TBD   |
@@ -175,8 +176,8 @@ Represents the Durable Timer
 |---|---|
 | Name    | `timer`    |
 | Kind    | Internal    |
-| Start    | When the timer is created   |
-| End    | When the TimerFired event is processed    |
+| Start    | Timestamp of when the timer is created   |
+| End    | Timestamp of when the TimerFired event is processed    |
 | Status    | OK – timer successfully completed<br />Error – timer cancellation   |
 | Links    | None    |
 | Events    | TBD    |
@@ -202,8 +203,8 @@ Represents sending an event to an orchestration
 |---|---|
 | Name    | `orchestration_event\|\|{orchestration_name}`  |
 | Kind    | Producer |
-| Start    | Before sending the event  |
-| End    | From client: when the event has been sent <br />From worker: when the event message has been created \(this will have a negligible duration, but we want a span for a link\) |
+| Start    | Timestamp efore sending the event  |
+| End    | From client: Timestamp of when the event has been sent <br />From worker: Timestamp of when the event message has been created \(this has a negligible duration, but we want a span for a link\) |
 | Status    | OK – event sent<br />Error – event failed to send  |
 | Links    | None    |
 | Events    | None  |
@@ -216,6 +217,6 @@ Attributes:
 |---|---|
 | durabletask.type    | “event”    |
 | durabletask.event.name  | The name of the event being sent.  |
-| durabletask.event.instance_id  | The instance ID of the target orchestration (the one receiving the event).    |
+| durabletask.event.target_instance_id  | The instance ID of the target orchestration (the one receiving the event).    |
 | durabletask.task.instance_id    | Instance ID of the orchestration that is sending the event. Not present if sent from the client.  |
 | durabletask.task.execution_id   | Execution ID of the orchestration that is sending the event. Not present if sent from the client. |
--- a/docs/telemetry/traces/semantic-conventions.md
+++ b/docs/telemetry/traces/semantic-conventions.md
@@ -162,8 +162,8 @@ Attributes:
 |---|---|
 | durabletask.type   | “activity”   |
 | durabletask.task.name   | Name of the activity   |
-| durabletask.task.version   | Version of the invoked orchestration enqueued. Omitted if null version.   |
-| durabletask.task.instance_id   | Instance ID of the invoked orchestration   |
+| durabletask.task.version   | Version of the invoking orchestration enqueued. Omitted if null version.   |
+| durabletask.task.instance_id   | Instance ID of the invoking orchestration   |
 | durabletask.task.task_id   | ID of the current task  |
 | exception.*   | Exception details on failure |
 
@@ -190,8 +190,8 @@ Attributes:
 |---|---|
 | durabletask.type    | “timer”    |
 | durabletask.fire_at  | Configured FireAt time displayed in ISO 8601 format  |
-| durabletask.task.version    | Version of the invoked orchestration enqueued. Omitted if null version.    |
-| durabletask.task.instance_id    | Instance ID of the invoked orchestration    |
+| durabletask.task.version    | Version of the invoking orchestration enqueued. Omitted if null version.    |
+| durabletask.task.instance_id    | Instance ID of the invoking orchestration    |
 | durabletask.task.task_id   | ID of the current task |
 
 
--- a/docs/telemetry/traces/semantic-conventions.md
+++ b/docs/telemetry/traces/semantic-conventions.md
@@ -35,7 +35,7 @@ Represents enqueueing a new orchestration via TaskHubClient to the backend.
 
 | Span Property | Span Value |
 |---|---|
-| Name  | `create_orchestration||{orchestrationName}(||{orchestrationVersion})?`  |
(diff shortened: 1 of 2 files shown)


semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+[1] durabletask.task.version may be omitted when version is null. 
+
+[2] When instance ID or execution ID is not available (e.g. activities) use the parent orchestration's instance ID
+
+### Spans
+#### Client – Starting an Orchestration
+
+Represents enqueueing a new orchestration via TaskHubClient to the backend.
+
+| Name  | create\_orchestration\|\|\{orchestrationName\}\(\|\|\{orchestrationVersion\}\)?  |
```
jviau, line 36: Alternative naming format suggestion for all our spans:

`{span_name}:{task_name}(@{task_version})?`

jviau, line 36: small suggestion: use inline code blocks to avoid having to escape all the characters. It also gives the added benefit of giving an emphasis to this text to represent it as something users may see in-code.

`create_orchestration||{orchestrationName}(||{orchestrationVersion})?`

jviau, line 36: Small formatting nit. Move this header to be a row, and then add a separate header:

| Span Property | Span Value |
| - | - |
| Name  | create\_orchestration\|\|\{orchestrationName\}\(\|\|\{orchestrationVersion\}\)?  |

semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+| Name  | Value  |
+|---|---|
+| durabletask.type | "orchestration" |
+| durabletask.task.name  | Name of the enqueued orchestration  |
+| durabletask.task.version  | Version of the orchestration enqueued. Omitted if null version.  |
+| durabletask.task.instance_id  | Instance ID of the enqueued orchestration  |
+| durabletask.task.execution_id  | Execution ID of the enqueued orchestration  |
+| exception.*  | Exception details on failure. |
+
+Note: Backends SHOULD enrich this activity with additional attributes, links, or events as appropriate.
```
jviau, line 59: Maybe we should validate how this experience works on backends before we include this note.

semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+
+Represents enqueueing and waiting for a sub-orchestration to complete 
+
+| Name   | orchestration\|\|\{orchestrationName\}\(\|\| \{orchestrationVersion\}\)?   |
+|---|---|
+| Kind   | Standard: Client   |
+|  | Fire and forget: Producer  |
+| Start   | When the parent orchestration enqueues a message to execute the sub\-orchestration  |
+| End   | Standard: When the parent orchestration is notified that the sub\-orchestration completed\.  |
+|  | Fire and forget: After the sub\-orchestration started event is enqueued  |
```
jviau, line 71: You have multi-line table entries expressed two ways in this document. Please pick just one.

You have this above (my slight preference for this one):

| ... | ... |
| - | - |
| Status  | OK – successfully enqueued<br />Error – failed to enqueue for any reason  |

and then this here:

| ... | ... |
| - | - |
| End   | Standard: When the parent orchestration is notified that the sub\-orchestration completed\.  |
|  | Fire and forget: After the sub\-orchestration started event is enqueued  |

semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+| durabletask.event.name  | The name of the event being sent.  |
+| durabletask.event.instance_id  | The instance ID of the target orchestration (the one receiving the event).    |
+| durabletask.task.instance_id    | Instance ID of the orchestration that is sending the event. Not present if sent from the client.  |
+| durabletask.task.execution_id   | Execution ID of the orchestration that is sending the event. Not present if sent from the client. |
+
+#### Receiving an Event
+
+Represents an orchestration receiving an event
+
+| Name    | orchestration\_event\|\|\{orchestration\_name\}  |
```
jviau, line 224: Are we going to have receiving an event as an independent span, or as a span _event_ on the orchestration server span?

My vote is for the span event.

semantic-conventions.md
```diff
@@ -0,0 +1,242 @@
...
+| durabletask.task.task_id   | ID of the current task |
+
+
+#### Sending an Event
+
+Represents sending an event to an orchestration
+
+| Name    | orchestration\_event\|\|\{orchestration\_name\}  |
+|---|---|
+| Kind    | Client  |
```
jviau, line 201: I believe this should be `Producer`, as you _produce_ the event on the backend, but do not wait for it to be consumed.

jviau (commented):
Can we move this file to something like `docs/telemetry/traces/semantic_conventions.md`?

semantic-conventions.md
bachuv: Makes sense. I removed it.

semantic-conventions.md
bachuv: I think span event makes more sense since it's something that happens once and doesn't have an "end" time. I removed this section.

semantic-conventions.md
bachuv: I updated it to use inline code blocks, but it seems like markdown doesn't support `|` in code blocks. I added the inline code block and just escaped the pipes.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
+# DurableTask Distributed Tracing Specification
+
+## Overview
+This document is to serve as a specification for distributed tracing in DurableTask Framework. This will include a list of all spans, their attributes, and detailed information about what constitutes each respective span.
```
davidmrdavid, line 4: I was under the impression this document was placed here, in part, to encourage community feedback. I think it would be worthwhile to make that explicit in the document itself :)

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+## Specification
+The following only applies to DurableTask.Core
+
+### Attributes
+These attributes are all new, defined by us, and specific to DurableTask Framework. All attributes here will begin with the “durabletask” prefix.
+
+Name | Type | Description | Requirement Level
+---|---|---|---
+durabletask.type | string | The type of durable event occurring. Options: “activity”, “orchestration”, “timer”, “event” | Required
+durabletask.task.name | string | The name of the durable task being invoked. This is the name it was queued with, not necessarily the name of the handler class/type. | Required
```
davidmrdavid, line 18: is the verb "queued" not specific to the Azure Storage backend designed? I think that things like the mssql backend do not have a notion of queues, although perhaps I'm wrong about that. If I'm right - I would suggest a verb such as "scheduled" over "queued"

docs/telemetry/traces/semantic-conventions.md
davidmrdavid, line 4: I'm under the impression this document was placed here, in part, to encourage community feedback. I think it would be worthwhile to make that explicit in the document itself :)

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Specification
+The following only applies to DurableTask.Core
+
+### Attributes
+These attributes are all new, defined by us, and specific to DurableTask Framework. All attributes here will begin with the “durabletask” prefix.
+
+Name | Type | Description | Requirement Level
+---|---|---|---
+durabletask.type | string | The type of durable event occurring. Options: “activity”, “orchestration”, “timer”, “event” | Required
```
davidmrdavid, line 17: Just to be sure: there's no notion of "Enum" in these types, right? It would be nice to have this field be more constrained, if possible.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+durabletask.task.name | string | The name of the durable task being invoked. This is the name it was queued with, not necessarily the name of the handler class/type. | Required
+durabletask.task.version | string | The version of the task being ran.  | Conditionally Required [1]
```
davidmrdavid, line 19: being nitpicky: consider replacing "ran" wtih "executed"?

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+durabletask.task.instance_id |  string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
+durabletask.task.execution_id | string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
```
davidmrdavid, line 21: I noticed part of this document are written in the future tense. For example: "... This _will_ be unique for all tasks".

Given that this is a spec, I would expect a more declarative/imperative tense. I would suggest writing this as "This is a unique ID across all tasks" instead of "This _will_". I have similar feedback for other parts of this document where the verb "will" is used.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+
+&nbsp;
+Attributes:
+
+| Name    | Value    |
+|---|---|
+| durabletask.type    | “event”    |
+| durabletask.event.name  | The name of the event being sent.  |
+| durabletask.event.instance_id  | The instance ID of the target orchestration (the one receiving the event).    |
+| durabletask.task.instance_id    | Instance ID of the orchestration that is sending the event. Not present if sent from the client.  |
```
davidmrdavid, line 220: Not sure if this is the right place to share feedback on the spec (or if that time window is behind us). If it isn't, I did want to point out that I wish that the target instanceId and the task's instanceId were a bit more differentiated. Instead of `durabletask.event.instance_id` I think it would be more descriptive to have `durabletask.event.target_instance_id`. That way, I wouldn't accidentally get it confused with `durabletask.task.instance_id`

davidmrdavid (commented):
Left some non-blocking feedback. Excited to see this spec out there soon!

docs/telemetry/traces/semantic-conventions.md
jviau: Either is fine. `Queued` can be taken to mean it is placed _somewhere_ for later processing. But I am also not against `scheduled`.

docs/telemetry/traces/semantic-conventions.md
jviau: It is an enum only in the sense it is a convention. But no, it won't be forced in code (we don't have a way to force it even if we wanted to).

docs/telemetry/traces/semantic-conventions.md
jviau: Sounds good to me to replace usages as you suggested.

docs/telemetry/traces/semantic-conventions.md
jviau: I am fine with `durabletask.event.target_instance_id`

jviau (approved):

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
+# DurableTask Distributed Tracing Specification
+
+## Overview
```
jviau, line 3: Can we add this right under the header:

```
**Status**: [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
```

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+durabletask.task.instance_id |  string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
+durabletask.task.execution_id | string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
+durabletask.task.task_id | integer | The ID / index of the task within the orchestration. | Conditionally required
+durabletask.task.result | string | The terminal status of a task. Options: “Succeeded”, “Failed”, “Terminated”, "Suspended" | Required
+durabletask.event.name | string | The name of the event being raised. | Required
+durabletask.event.instance_id | string | The target orchestration instance ID of a raised event. | Required
+
+[1] durabletask.task.version may be omitted when version is null. 
+
+[2] When instance ID or execution ID is not available (e.g. activities) use the parent orchestration's instance ID
```
davidmrdavid, line 29: when the execution ID is not available, will we also use the parent's _**instance**_ ID, or will we use the parent's execution ID in that case?

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+
+Name | Type | Description | Requirement Level
+---|---|---|---
+durabletask.type | string | The type of durable event occurring. Options: “activity”, “orchestration”, “timer”, “event” | Required
+durabletask.task.name | string | The name of the durable task being invoked. This is the name it was queued with, not necessarily the name of the handler class/type. | Required
+durabletask.task.version | string | The version of the task being ran.  | Conditionally Required [1]
+durabletask.task.instance_id |  string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
+durabletask.task.execution_id | string | The execution ID of the task being ran. This will be unique for all tasks. | Required [2]
+durabletask.task.task_id | integer | The ID / index of the task within the orchestration. | Conditionally required
+durabletask.task.result | string | The terminal status of a task. Options: “Succeeded”, “Failed”, “Terminated”, "Suspended" | Required
```
davidmrdavid, line 23: When I think of "result", I think of the task's output. I wonder if we can come up with a more differentiated term here to avoid that confusion.

Also, a nitpick: "Suspended" is not a terminal state, as a task may be resumed :).

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+#### Client – Starting an Orchestration
+
+Represents enqueueing a new orchestration via TaskHubClient to the backend.
+
+| Span Property | Span Value |
+|---|---|
+| Name  | `create_orchestration\|\|{orchestrationName}(\|\|{orchestrationVersion})?`  |
+| Kind  | Producer  |
+| Start  | Before submitting the new orchestration to the backend  |
+| End  | After the orchestration has been submitted to the backend  |
```
davidmrdavid, line 41: So we will only receive Either "Start" or "End" but not both properties at once, right? This makes sense to me but the table doesn't make it immediately clear to me.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+
+&nbsp;
+Attributes:
+
+| Name   | Value   |
+|---|---|
+| durabletask.type   | “activity”   |
+| durabletask.task.name   | Name of the activity   |
+| durabletask.task.version   | Version of the related orchestration enqueued. Omitted if null version.   |
+| durabletask.task.instance_id   | Instance ID of the related orchestration   |
```
davidmrdavid, line 165: nitpick: should we replace "related orchestration" with "invoking orchestration" or "caller orchestration"? Just trying to be precise.

davidmrdavid: Same feedback for other places where "related" is used.

davidmrdavid (commented):
A few more non-blocking comments.

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Thanks for clarifying, I'll resolve this feedback then.

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Actually, how is this field populated? If Suspended is here, I could see states like "Running" or "Resumed" populating it, which would make this a more general "state" field, which I think would be a good name.

If this field contains terminal states _and_ Suspended, then I'm having a harder time thinking of a good name.

jviau: Good catch, you are correct `Suspended` does not belong here. This field is actually not _supper_ useful as it is largely redundant with activity status. The only benefit is it differentiates failure states `Failed` vs `Terminated` - but maybe there is a better way to show that. I think we may want something similar to `http.status_code`, which gives a more insight on _how_ an HTTP request completed. Is there an equivalent for Durable?

On my previous team, I found that a sub-status for Durable traces is very useful. As we would want to not just know success/failure but also _why_ something failed. As there were some acceptable failures or essentially client errors we would want to ignore in our alerts.

docs/telemetry/traces/semantic-conventions.md
jviau: I am actually not sure this is entirely necessary to have on activities at all. A user can easily go from a span representing a `TaskActivity` to the parent orchestration's instance ID and execution span ID by looking at the parent span itself.

docs/telemetry/traces/semantic-conventions.md
jviau: You get both properties. Well, it depends on how the selected exporter works. But all spans represent spans of time, so these are the start and end timestamps.

docs/telemetry/traces/semantic-conventions.md
jviau: I agree `related` may not be the best, but `invoking` or `caller` is not right to me either. What we are trying to get across here is that this span represents a single orchestration. And this attribute _is_ that orchestration's instance ID. What would be the appropriate word to use here?

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Oh, these are timestamps! In that case, can we update the description (Span Value column) to say this? I think the current text assumes that you already know these are timestamp-types.

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: If "invoking" works for everyone, it works for me. I do think "related" is a bit too underspecified. @bachuv: what do you think?

docs/telemetry/traces/semantic-conventions.md
bachuv: This is the right place and this is the exact type of feedback we're looking for :)

docs/telemetry/traces/semantic-conventions.md
bachuv: Yes, `Suspended` shouldn't be in this list, good catch! I removed it for now.

@jviau can you give some examples of how you would see the sub-status being populated? For failures, I would expect the details (the why) to be in `FailureDetails` and emitted through `otel.status_description`.

docs/telemetry/traces/semantic-conventions.md
bachuv: I think "invoking" works instead of "related." I'll leave this comment unresolved for now in case others have opinions.

docs/telemetry/traces/semantic-conventions.md
jviau: Ah don't worry about my comment for a sub-status code. Not worth getting into it - because as you pointed out, would require design around how it is populated. And for now, customers can just populate their own attribute for a sub-status code on the server span for now.

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: So, do we need this in here then? @bachuv?

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,222 @@
+# DurableTask Distributed Tracing Specification
+**Status**: [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
+
+## Overview
+This document serves as a specification for distributed tracing in the DurableTask Framework. It includes a list of all spans, their attributes, and detailed information about what constitutes each respective span. We want to gather community feedback and encourage any comments with suggestions and questions.
```
davidmrdavid, line 5: tiniest nitpick:

```suggestion
This document serves as a specification for Distributed Tracing in the DurableTask Framework. It includes a list of all spans, their attributes, and detailed information about what constitutes each respective span. We want to gather community feedback and encourage any comments with suggestions and questions. 
```

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,222 @@
+# DurableTask Distributed Tracing Specification
+**Status**: [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
+
+## Overview
+This document serves as a specification for distributed tracing in the DurableTask Framework. It includes a list of all spans, their attributes, and detailed information about what constitutes each respective span. We want to gather community feedback and encourage any comments with suggestions and questions. 
+
+## References
+- [Open Telemetry Trace Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions "Open Telemetry Trace Semantic Conventions") – various trace semantic conventions we  reference throughout this document. Particularly, we pull in some attributes defined in these files.
```
davidmrdavid, line 8:
```suggestion
- [Open Telemetry Trace Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions "Open Telemetry Trace Semantic Conventions") – various trace semantic conventions we reference throughout this document. Particularly, we pull in some attributes defined in these files.
```

davidmrdavid (commented):
A few more thoughts

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: I think this is my only soft-blocking question. After reaching a conclusion here, I'd be ready to approve :)

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,222 @@
...
+| exception.*  | Exception details on failure. |
+
+#### Client - Starting a sub-orchestration
+
+Represents enqueueing and waiting for a sub-orchestration to complete 
+
+| Span Property | Span Value |
+|---|---|
+| Name   | `orchestration\|\|{orchestrationName}(\|\|{orchestrationVersion})?`   |
+| Kind   | Standard: Client<br />Fire and forget: Producer  |
```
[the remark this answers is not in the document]
jviau: Here is a description of span kinds: https://opentelemetry.io/docs/concepts/signals/traces/#span-kind

A strict reading of that would mean most of our spans should be `Producer` and `Consumer`, because DurableTask is all about putting messages on some backend for _later_ processing. However, I think that the way users write code in orchestrations makes `Client` and `Server` more fitting - as it _looks_ like you are waiting on something to finish before resuming. But I may be wrong.

Anyways - for this case, I suggested `Client` for regular processing as the parent orchestration will wait on the sub orchestration to complete. But in fire & forget, the parent orchestration does not wait - so `Producer` is more appropriate.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+&nbsp;
+Attributes:
+
+| Name  | Value  |
+|---|---|
+| durabletask.type | "orchestration" |
+| durabletask.task.name  | Name of the enqueued orchestration  |
+| durabletask.task.version  | Version of the orchestration enqueued. Omitted if null version.  |
+| durabletask.task.instance_id  | Instance ID of the enqueued orchestration  |
+| durabletask.task.execution_id  | Execution ID of the enqueued orchestration  |
```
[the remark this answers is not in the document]
jviau: We already use the term `ExecutionId` throughout the code base (property names, etc). So I think we should line up with that.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,222 @@
...
+
+Name | Type | Description | Requirement Level
+---|---|---|---
+durabletask.type | string | The type of durable event occurring. Options: “activity”, “orchestration”, “timer”, “event” | Required
+durabletask.task.name | string | The name of the durable task being invoked. This is the name it was scheduled with, not necessarily the name of the handler class/type. | Required
+durabletask.task.version | string | The version of the task being executed.  | Conditionally Required [1]
+durabletask.task.instance_id |  string | The execution ID of the task being ran. This is unique for all tasks. | Required [2]
+durabletask.task.execution_id | string | The execution ID of the task being ran. This is unique for all tasks. | Required [2]
+durabletask.task.task_id | integer | The ID / index of the task within the orchestration. | Conditionally required
+durabletask.task.result | string | The terminal status of a task. Options: “Succeeded”, “Failed”, “Terminated” | Required
```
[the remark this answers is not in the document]
jviau: I think `durabletask.task.result` is more granular than `status_code`. For example, when `status_code` is `ERROR`, `durabletask.task.result` maybe be any of `Failed`, `Terminated`, or `Canceled` (well, when we add cooperative cancellation, which is a long request feature).

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,221 @@
...
+| Events   | TBD   |
+| Parent   | “Running an orchestration” span for parent orchestration |
+
+&nbsp;
+Attributes:
+
+| Name   | Value   |
+|---|---|
+| durabletask.type   | “orchestration”   |
+| durabletask.task.name   | Name of the enqueued sub-orchestration   |
```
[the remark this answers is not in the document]
jviau: It is a bit redundant. But I am not sure what we can do instead. We need to keep a constant prefix as our unique namespace (`durabletask`). And then we also want to group all orchestration / activity instance attributes into their own sub-namespace as well (`.task`). I think we must keep the two separate namespaces, but I am not against finding better names for one or the either.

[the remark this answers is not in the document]
cgillum: I'm not sure I understand the need for the `.task` sub-namespace. Is there a particular ambiguity we're hoping to avoid?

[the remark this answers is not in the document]
RohitRanjanMS: if we can come-up with something more generic, we have an opportunity to push this as FaaS standard.

docs/telemetry/traces/semantic-conventions.md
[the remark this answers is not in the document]
RohitRanjanMS: We can still use status_description to capture more granular details. 
I am fine with it as long as we clearly define the relationship between task.result and status_code. All the APM tools are going to use status_code to report success and failures.

docs/telemetry/traces/semantic-conventions.md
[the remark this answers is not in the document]
jviau: > I'm not sure I understand the need for the .task sub-namespace. Is there a particular ambiguity we're hoping to avoid?

Yes, it is to avoid ambiguity. Such as `durabletask.version` - is this the task version? The framework version? Is it a backend providers version? The version of the schema?

> if we can come-up with something more generic, we have an opportunity to push this as FaaS standard.

I don't think DurableTask is FaaS. It is its own thing.

docs/telemetry/traces/semantic-conventions.md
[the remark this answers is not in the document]
jviau: I think `status_description` is more of an opaque string - we are planning on putting the failure message there (the exception message in C#). I'd rather not put semantically meaningful values there.

> I am fine with it as long as we clearly define the relationship between task.result and status_code.

Yes, we will have a clearly defined relationship between these two.

docs/telemetry/traces/semantic-conventions.md
bachuv: I don't think activities need to save this either. Should we keep it consistent with instance ID though? If a `TaskActivity` span sets the instance ID then should it by default also set execution ID as a tag?

docs/telemetry/traces/semantic-conventions.md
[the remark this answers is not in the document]
bachuv: @davidmrdavid do you have any follow up questions about this or did Jacob's explanation make sense? For 'Client', the docs say that "it is not queued for later processing" and for `Producer`, it says "the creation of a job which may be asynchronously processed later."

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,222 @@
...
+[2] When instance ID is not available (e.g. activities) use the parent orchestration's instance ID
+
+### Spans
+#### Client – Starting an Orchestration
+
+Represents enqueueing a new orchestration via TaskHubClient to the backend.
+
+| Span Property | Span Value |
+|---|---|
+| Name  | `create_orchestration\|\|{orchestrationName}(\|\|{orchestrationVersion})?`  |
```
jviau, line 39: Question about the naming. Which is preferred?

1. `{label}||{task_name}(||{task_version})?` - `||` separator for both.
2. `{label}:{task_name}(@{task_version})?` - `:` separator for label -> name, `@` version separator
3. `{label}/{task_name}(/{task_version})?` - `/` for both
4. Some combination of any of the above separators

bachuv: Just updating here that a conversation in a separate internal thread led to agreeing on option 2 - `{label}:{task_name}(@{task_version})?`

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: I don't know enough to answer this question. @jviau, what are your thoughts?

docs/telemetry/traces/semantic-conventions.md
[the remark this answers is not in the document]
davidmrdavid: Thanks for the ping @bachuv.
@jviau's explanation makes sense. I can definitely be swayed both ways here so I won't block here. I think this is something where customer feedback would be helpful, and that's something we can only get by merging this!

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Can we update the PR to reflect this?

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Sounds like we have consensus for "invoking". @bachuv: can we update the PR so that I can resolve this comment?

davidmrdavid (approved):
No further blockers on my end. LGTM!
Also good to remember that the purpose of this document is, in part, to gather feedback from users so if there's anything that we can't decide on right now, we can always punt it as something to gather feedback on :)

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,222 @@
+# DurableTask Distributed Tracing Specification
+**Status**: [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
+
+## Overview
+This document serves as a specification for Distributed Tracing in the DurableTask Framework. It includes a list of all spans, their attributes, and detailed information about what constitutes each respective span. We want to gather community feedback and encourage any comments with suggestions and questions. 
+
```
cgillum, line 6: I recommend we add a table of contents here with links the the various sections. I have specific things that I want to look for but I don't know how far down to scroll, or if the content I care about exists at all.

cgillum:
Currently we have a single span that tracks the full lifecycle of the orchestration. This is nice because it gives a holistic picture of the time frame over which the orchestration ran. However, problematic in that it can hide various details, such as events (something I don't think this is currently called out in the spec).

Suppose for example that I want to add events to the orchestration span for whenever an orchestration is suspended, resumed, or when an external event is received. With a single span, we won't see these events unless the orchestration completes.

I wonder if we should instead do both a big span and also a child span for each orchestration replay. That would make it easier to view orchestration-specific events in real time. It might also make certain types of problems more obvious, such as long delays in orchestration execution which don't involve activities or sub-orchestrations. Something like this:

image

jviau:
@cgillum - for orchestration events I think we do want to emit a span event for them on the `orchestration1` span in your example. Regarding the replay spans, I do agree we should ensure we have _something_ to help customers follow individual replays. Child spans may be useful or maybe just logs are good enough? I'd like for us to spend more time exploring what is optimal here. However, I don't think this needs to be done right away. We can focus on the spec without the per-replay spans, then decide on adding them later. And if it is a breaking change, we can always give customers a way to control keeping the previous behavior or adding the new per-replay spans.

cgillum:
I'm fine with not solving this right now, but I think we'll need to carefully consider the user experience when we do get around to that. For example, the problem with adding events to `orchestration1` in my example is that you'll never see it until the orchestration completes. Not only that, in the case of multiple replays, you have to ensure that those events get added _on each replay_ to ensure they actually get published by the time the orchestration completes. Placing them on the replay span might be both more user friendly (they can be seen right away) and more performant.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,233 @@
+# DurableTask Distributed Tracing Specification
+**Status**: [Experimental](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
+
```
davidmrdavid, line 3: _I wasn't sure where to leave this suggestion, so I chose this blank line._

I noticed that the event names for every trace are a bit long. For example, the name for an Activity execution on the built-in sample reads as follows: `activity:OpenTelemetrySample.Program+SayHello`.

That's pretty verbose. I was wondering if users have a way to customize these names so that they're shorter. For example, I think it would suffice to _just_ display `activity:SayHello`, at least for the built-in sample.

These long names also make the gantt chart a bit hard to parse. See, for example, the length of the event names as displayed by AI:

verbose

So, to summarize, my question are:
1. is there a reason for the names to be always this long? 
2. are we giving users the ability to shorten these names?
3. given the above, should we consider giving the built-in sample more concise event names?

Thanks! This is a non-blocking suggestion.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,233 @@
...
+| durabletask.type   | “activity”   |
+| durabletask.task.name   | Name of the enqueued sub-orchestration   |
+| durabletask.task.version   | Version of the sub-orchestration enqueued. Omitted if null version.   |
+| durabletask.task.instance_id   | Instance ID of the enqueued orchestration   |
+| durabletask.task.task_id   | ID of the current task  |
+| exception.*   | Exception details on failure |
+
+#### Worker: Running an Activity
+
+Represents the activity executing
```
davidmrdavid, line 156: When looking at the generated gantt chart on AI, I'm not able to easily distinguish between the enqueueing and running of an Activity.  _Assuming, I'm looking at the right data on AI, please see below how they both share the same event name:

samename

I'm curious if others agree this is a bit confusing to read. The fact that there's two spans showing me the same name makes me think that the same kind of work is being represented, but that's not the case afaik.

I was hoping we could consider changing the event names so that users can immediately distinguish between these two spans.

For example: what if the enqueing and running activity spans had the following naming convention: `activity:{enqueing || execution}:{activityName}`.

Then, the spans could be named `activity:enqueing:SayHello` and `activity:running:SayHello` respectively. This is just one idea, I'm sure there's others.

davidmrdavid (commented):
Just leaving some new thoughts after trying out the built-in sample.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,233 @@
...
+| durabletask.task.execution_id  | Execution ID of the enqueued orchestration  |
+| exception.*  | Exception details on failure. |
+
+#### Client: Starting a sub-orchestration
+
+Represents enqueueing and waiting for a sub-orchestration to complete 
+
+| Span Property | Span Value |
+|---|---|
+| Name   | `orchestration:{orchestrationName}(@{orchestrationVersion})?`   |
```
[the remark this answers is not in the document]
davidmrdavid: I think I just left some similar feedback here: https://github.com/Azure/durabletask/pull/847#discussion_r1119335596

docs/telemetry/traces/semantic-conventions.md
jviau: 1. is there a reason for the names to be always this long?
Yes, you want to ensure span names are unique. For representing code, we include the namespace - otherwise you may have two "SayHello" orchestrations in different namespaces which start colliding. Also, uniqueness is important for aggregation of on span names. So even if a human could tell which `SayHello` this is for based on manual investigation, tooling may not know the difference.

2. are we giving users the ability to shorten these names?
Yes, customers have a couple ways they can customize names. (1) they can listen to activity created and modify `Activity.DisplayName`, or (2) they can use SDK-specific methods to modify the name. For OTel, this would be wrapping the exporter in their own implementation that modifies the activity as needed. For AppInsights, `ITelemetryProcessor` can be used to change the name. But we can always introduce other extension points to control span-name creation, but I wouldn't prioritize (or even document) this for now.

3. given the above, should we consider giving the built-in sample more concise event names?
I do not think we should for the reasons above - we need to err on the side of verbosity to avoid collisions. HTTP span names with OTel will be similarly verbose - they are the un-parameterized URL template. Span names in HTTP are as long as whatever route template you define. Take an Azure management route for example, it would be something like `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/{provider}/someResource/{resourceName}` - that isn't exactly short.

docs/telemetry/traces/semantic-conventions.md
```diff
@@ -0,0 +1,233 @@
...
+| Kind   | Standard: Client<br />Fire and forget: Producer  |
+| Start   | Timestamp of when the parent orchestration enqueues a message to execute the sub\-orchestration  |
+| End   | Standard: Timestamp of when the parent orchestration is notified that the sub\-orchestration completed\.<br />Fire and forget: Timestamp after the sub\-orchestration started event is enqueued  |
+| Status   | OK – successfully enqueued<br />Error – failed to enqueue for any reason   |
```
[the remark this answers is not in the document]
jviau: Ah yeah this `Error` case should probably read something like "Error - sub-orchestration failed". Basically, if the call to the `OrchestrationContext` for the sub-orchestration throws, this should be `Error`.

docs/telemetry/traces/semantic-conventions.md
jviau: @davidmrdavid it may be a bit vague in ApplicationInsights, but you should be able to tell by the telemetry type: `request` is "execution" and `dependency` is "enqueueing". Additionally, the icon to the left of `MySample` should indicate request vs dependency.

davidmrdavid: Thanks @jviau. Yes, I noticed that there's a few clues in the UI indicating that these spans represent different things despite them sharing the same name.

My main concern is that although I can tell there's _some_ difference between the spans (due to the UI elements changing and so on), I can't easily _what_ that difference is. I imagine that customers might also pick up that there's a difference but be unable to tell what that is.

I like the `call` and `run` prefixes you suggest.  I'd like to push for that if @bachuv and others also find it sensible.

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Thanks for clarifying. These seem like good reasons to keep the names this verbose. I think that for demo'ing purposes it would be neat to have the shorter names but, at this time, I agree that's not a priority. I'm good to put this in the backburner for now.

docs/telemetry/traces/semantic-conventions.md
jviau: I originally didn't include those prefixes because there are other examples of span names being identical, only differentiated by kind. HTTP for example - _sending_ (client) and _receiving_ (server) both have the same span naming semantics by default in OTel. So I am assuming we can rely on tooling to correctly convey the difference of identically named spans when `kind` is different.

jviau: I am not against the prefix. I believe it is understood (as far as my interpretation of OTel semantics) that span names need not be unique on their own, but rather other attributes of a span should be considered for determining unique-ness (`kind` for one, but also resource & span source information).

docs/telemetry/traces/semantic-conventions.md
jviau: I say lets leave these attributes off of `TaskActivity` spans for now. We can always add it later if it is requested.

docs/telemetry/traces/semantic-conventions.md
davidmrdavid: Understandable. Not high priority on my end either. Thanks for providing context

jviau: Also, span names are customizable. You can create an `ActivityListener` yourself and change `Activity.DisplayName` - so customers could add their own prefix to distinguish the spans further if they desire.

PR #849, opened 2023-01-20
Add Activity client spans

bachuv:
This PR adds spans that represent the time between when an Activity is called and finished. These are client spans.

Example:

The highlighted span is an example of the new client span for Activities.
Hello Sequence
image

Hello Fan Out
image

Sub-orchestrations
image


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -714,6 +714,14 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                     TraceHelper.StartTraceActivityForSubOrchestrationFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, finishedEvent, subOrchestrationScheduledEvent)?.Dispose();
                 }
 
+                if (Activity.Current != null && message.Event is TaskCompletedEvent taskCompletedEvent)
+                {
+                    TaskScheduledEvent taskScheduledEvent = (TaskScheduledEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == taskCompletedEvent.TaskScheduledId);
```
jviau, line 719: Will also need to handle `TaskFailedEvent`, similar to `SubOrchestrationFailedEvent`.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -248,6 +248,24 @@ static TraceHelper()
             return newActivity;
         }
 
+        internal static Activity? CreateActivityforTaskCompleted(OrchestrationInstance instance, TaskScheduledEvent taskScheduledEvent)
+        {
+            Activity? newActivity = ActivityTraceSource.StartActivity(
+                name: taskScheduledEvent.Name,
```
cgillum, line 254: According to the spec, this should be `activity||{activityName}(||{orchestrationVersion})?`, right?

cgillum (commented):
I like how small this PR is. :). Just one comment in addition to Jacob's.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -737,10 +739,43 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
...
+                    else if (historyEvent is SubOrchestrationInstanceFailedEvent subOrchestrationInstanceFailedEvent)
+                    {
+                        SubOrchestrationInstanceCreatedEvent subOrchestrationCreatedEvent = (SubOrchestrationInstanceCreatedEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == subOrchestrationInstanceFailedEvent.TaskScheduledId);
+
+                        // We immediately publish the activity span for this sub-orchestration by creating the activity and immediately calling Dispose() on it.
+                        TraceHelper.EmitTraceActivityForSubOrchestrationFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, subOrchestrationCreatedEvent, subOrchestrationInstanceFailedEvent);
+                    }
+                }
+
+                if (message.Event is TaskCompletedEvent taskCompletedEvent)
```
bachuv, line 766: I noticed that when I was testing these changes that there are missing spans when workItem.NewMessages > 1 (e.g. multiple activities are scheduled using Task.WhenAll() for the Fan Out Fan In pattern).



This is the flow that I was observing:

1. For the first message in the list, Activity.Current != null so the parent context is set correctly. 
2. Then, the Activity is created (e.g. Activity for TaskCompletedEvent). After the Activity is disposed (newActivity.Dispose()), then Activity.Current is null. 
3. For all remaining messages, they will not go in the if block on line 742. Even if I remove that check for Activity.Current != null, I don't think the parent would be set correctly since Activity.Current is null and the parent trace context isn't stored anywhere.



When I was testing a Fan Out Fan In scenario, I was only seeing a client span for the first activity since the Activity was being disposed after the getting emitted. To work around this, I stored the ParentTraceContext to TaskScheduledEvent and read that in ReconcileMessagesWithState(). I know we are trying to avoid storing the parent trace context in the HistoryEvent, but I wasn't sure how to get around this without Activity.Current. I don't think restoring Activity.Current while iterating through the NewMessages is good practice.



I have a couple of questions:

1. I handled the case for activities by not checking for Activity.Current and manually setting the parent trace context, but this could happen in many other situations right? For any Task.WhenAllI() in an orchestration for example?
2. Does anyone have any suggestions/pointers on how we can avoid storing the parent trace context in TaskScheduledEvent?

bachuv:
The commit history got automatically updated because the target branch got deleted btw. It was originally vabachu/suborchupdates, but then that branch got deleted since the related PR was merged so the target branch is now cgillum/opentelemetry. That automatically added some commits to this PR which I didn't realize would happen, but makes sense. (I'm learning a lot about how git works through these distributed tracing PRs (: ) The diff between cgillum/opentelemetry and this PR should be accurate now though so ignore the number of commits please (:

jviau:
> The commit history got automatically updated because the target branch got deleted btw. It was originally vabachu/suborchupdates, but then that branch got deleted since the related PR was merged so the target branch is now cgillum/opentelemetry. That automatically added some commits to this PR which I didn't realize would happen, but makes sense. (I'm learning a lot about how git works through these distributed tracing PRs (: ) The diff between cgillum/opentelemetry and this PR should be accurate now though so ignore the number of commits please (:

Did you do a squash merge of `vabachu/suborchupdates` to `cgillum/opentelemetry.`? If so, that would explain this. The squash merge doesn't merge in each individual commit, but a new commit which is all the commits squashed down. If you try and merge those same commits back in, git doesn't know that squashed commit contains them. This isn't a big deal. If you merge in again with another squash commit it will all be fine. If it does bother you, you can use `git rebase -i` on your branch and drop all the commits already contained in the previous PR and only pick your new ones.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: I'll go back and take a look at how I solved this in `durabletask-go`, but intuitively, I think the only place you need to store the parent trace context is in the `ExecutionStartedEvent` event. The fact that .NET uses thread-local variables probably makes tracking spans a lot harder than it needs to be, but I suspect each time we do a Dispose(), we need to reset `Activity.Current` to be the parent context that we already have in the `ExecutionStartedEvent`.

cgillum: Actually, thinking about this further, I think the problem you're trying to solve is a bit more complicated compared to what I did because you're emitting two spans for each activity, one for how long the activity waits for the activity and the other for the actual activity execution. I my case, I only did the latter, which made things much simpler.  If we want to have two spans for each activity execution, with one the parent of the other, then I think you're right that we'll need to store a parent trace context directly in `TaskScheduledEvent`.

jviau: Weird that `newActivity.Dispose()` makes `Activity.Current == null`. Disposing an activity should set `Activity.Current` to the recently disposed activity's parent. Is the parent activity (`Activity.Current`) not correctly captured when creating `newActivity`?

jviau: > Actually, thinking about this further, I think the problem you're trying to solve is a bit more complicated compared to what I did because you're emitting two spans for each activity, one for how long the orchestration waits for the activity and the other for the actual activity execution. I my case, I only did the latter, which made things much simpler. If we want to have two spans for each activity execution, with one the parent of the other, then I think you're right that we'll need to store a parent trace context directly in TaskScheduledEvent.

Yes, we want two spans. I also recommend you do the same for the go SDK. There is significance to having both, as you can identify gaps in execution- e.g.: if you see a big delta between the start of the `Client` span and the `Server` span, then you probably have a throughput issue and it may be worth investigating.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -763,6 +763,21 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
...
+                {
+                    TaskScheduledEvent taskScheduledEvent = (TaskScheduledEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == taskCompletedEvent.TaskScheduledId);
+
+                    // We immediately publish the activity span for this task by creating the activity and immediately calling Dispose() on it.
+                    TraceHelper.EmitActivityforTaskFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, taskScheduledEvent);
+                }
+                else if (message.Event is TaskFailedEvent taskFailedEvent)
+                {
+                    TaskScheduledEvent taskScheduledEvent = (TaskScheduledEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == taskFailedEvent.TaskScheduledId);
```
cgillum, line 775: Consider using `LastOrDefault(...)` instead of `FirstOrDefault(...)`. I suspect this will be more efficient in the common case (activity sequences). Also, for defense-in-depth, I wonder if we should do a null check in case we're in a situation where storage is somehow corrupted. It might be better to try and continue the best we can rather than throw a null-ref exception later.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: > Weird that newActivity.Dispose() makes Activity.Current == null. Disposing an activity should set Activity.Current to the recently disposed activity's parent. Is the parent activity (Activity.Current) not correctly captured when creating newActivity?

That was my expectation too, but not what I observed when I stepped through the code. I believe the parent `Activity` is set correctly for `newActivity` because it works fine for the first message in the `NewMessages` list.

bachuv:
> Did you do a squash merge of `vabachu/suborchupdates` to `cgillum/opentelemetry.`? If so, that would explain this. The squash merge doesn't merge in each individual commit, but a new commit which is all the commits squashed down. If you try and merge those same commits back in, git doesn't know that squashed commit contains them. This isn't a big deal. If you merge in again with another squash commit it will all be fine. If it does bother you, you can use `git rebase -i` on your branch and drop all the commits already contained in the previous PR and only pick your new ones.

Yes, I did a squash merge. Thanks for the explanation! This makes more sense now.

jviau (approved):

samples/DistributedTraceSample/Program.cs
```diff
@@ -46,10 +46,12 @@ static async Task Main(string[] args)
...
             worker.AddTaskOrchestrations(typeof(HelloSequence));
             worker.AddTaskOrchestrations(typeof(HelloFanOut));
+            worker.AddTaskOrchestrations(typeof(HelloFanOutWithException));
             worker.AddTaskOrchestrations(typeof(ExceptionSubOrch));
             worker.AddTaskOrchestrations(typeof(ExceptionOrchestration));
             worker.AddTaskActivities(typeof(SayHello));
             worker.AddTaskActivities(typeof(GetRequestResultMessageActivity));
+            worker.AddTaskActivities(typeof(ExceptionActivity));
             worker.AddTaskOrchestrations(typeof(GetRequestionDecisionOrchestration));
             worker.AddTaskOrchestrations(typeof(EventConversationOrchestration));
```
cgillum, line 56: Is the `FireAndForgetSubOrchestration` supposed to be part of this PR?

cgillum (changes requested):
It looks like this PR is reverting a bunch of changes that were made after the PR was initially opened. Does this PR need to be rebased on top of the latest changes from `main`?

cgillum:
Or maybe there's something wrong with the diff I'm looking at...let me look again using a different diff comparer.

cgillum:
Weird. The diffs I see on github.dev are very different from the diffs I see on github.com. Not sure why.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -763,6 +763,21 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                     }
                 }
 
+                if (message.Event is TaskCompletedEvent taskCompletedEvent)
+                {
+                    TaskScheduledEvent taskScheduledEvent = (TaskScheduledEvent)workItem.OrchestrationRuntimeState.Events.LastOrDefault(x => x.EventId == taskCompletedEvent.TaskScheduledId);
+
+                    // We immediately publish the activity span for this task by creating the activity and immediately calling Dispose() on it.
```
cgillum, line 770: This comment no longer makes sense here. It might make more sense to move it into the implementation of `EmitActivityforTaskFinished()`

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -763,6 +763,21 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
...
+                    TraceHelper.EmitActivityforTaskFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, taskScheduledEvent);
+                }
+                else if (message.Event is TaskFailedEvent taskFailedEvent)
+                {
+                    TaskScheduledEvent taskScheduledEvent = (TaskScheduledEvent)workItem.OrchestrationRuntimeState.Events.LastOrDefault(x => x.EventId == taskFailedEvent.TaskScheduledId);
+
+                    // We immediately publish the activity span for this task by creating the activity and immediately calling Dispose() on it.
```
cgillum, line 777: Same thing for this comment.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -912,6 +925,8 @@ private void ResetDistributedTraceActivity(OrchestrationRuntimeState runtimeStat
                     version: scheduleTaskOrchestratorAction.Version);
             }
 
+            scheduledEvent.SetParentTraceContext(parentTraceActivity);
```
cgillum, line 928: Didn't we already do this a few lines up?

cgillum (commented):
Just a couple small things.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: Yes, this line is supposed to handle the case where `scheduledEvent` is reassigned when `includeParameters` is `false`. I moved this line into the if block now.

cgillum: Rather than moving it inside the `if` block and maintaining two different calls, shouldn't we just make this call in one place *after* the `if` block?

cgillum: Nevermind, I can see that the way you're doing it now is correct since we're dealing with two `TaskScheduledEvent` objects that get sent to different places.

cgillum (approved):
:shipit:

Issue #819, opened 2022-11-08
Changes in the version 2.12.0 break DurableTask.AzureServiceFabric


dsempi:
If upgrading to the latest DurableTask.Core (2.12.0) when using DurableTask.AzureServiceFabric, you get the following error:

```Type 'System.Collections.Generic.Dictionary`2[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' with data contract name 'ArrayOfKeyValueOfstringstring:http://schemas.microsoft.com/2003/10/Serialization/Arrays' is not expected. Add any types not known statically to the list of known types - for example, by using the KnownTypeAttribute attribute or by adding them to the list of known types passed to DataContractSerializer.'```

The issue is caused by the fact `TaskMessage` now contains `OrchestrationExecutionContext` which has a serializable member `IReadOnlyDictionary<string, string> OrchestrationTags` which unfortunately cannot be serialized by the standard `DataContractSerializer` used in Service Fabric environment.

jviau:
Thank you for pointing this out. We did not intend this to be a breaking change and will address it soon. For the time being, are you able to use the previous version of the core library?

dsempi:
Yes, it works fine with 2.11.1. Thanks!

RamjotSingh:
This is fixed in https://github.com/Azure/durabletask/pull/821

PR #853, opened 2023-01-25
Adding exception information to spans (Replaces #694)

bachuv:
Creating this PR as a replacement of #694 to clean up the commit history. The code changes are similar to the previous PR. This PR adds exception information to spans when an orchestration or activity throws an exception. It also adds `Activity` extension methods for `SetId()`, `SetSpanId()`, and `SetStatus()`. 

The following examples cover both `ErrorPropagationMode` settings

1) HelloSequenceException orchestration with the following `TaskHubWorker` settings
```
worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;
worker.TaskActivityDispatcher.IncludeDetails = true;
worker.TaskOrchestrationDispatcher.IncludeDetails = true;
```
image
image
image

2) ExceptionSubOrch orchestration with the following `TaskHubWorker` settings
worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;
worker.TaskActivityDispatcher.IncludeDetails = true;
worker.TaskOrchestrationDispatcher.IncludeDetails = true;

image
image
image

3) HelloSequenceException orchestration with the following `TaskHubWorker` setting
worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;

image
image
image

4) ExceptionSubOrch orchestration with the following `TaskHubWorker` setting
worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;

image
image
image

### DIFF
--- a/samples/DistributedTraceSample/Program.cs
+++ b/samples/DistributedTraceSample/Program.cs
@@ -43,6 +43,8 @@ namespace OpenTelemetrySample
                 await GetOrchestrationServiceAndClient();
 
             using TaskHubWorker worker = new TaskHubWorker(service);
+            worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;
+
             worker.AddTaskOrchestrations(typeof(HelloSequence));
             worker.AddTaskOrchestrations(typeof(HelloFanOut));
             worker.AddTaskActivities(typeof(SayHello));
@@ -51,6 +53,9 @@ namespace OpenTelemetrySample
             worker.AddTaskOrchestrations(typeof(EventConversationOrchestration));
             worker.AddTaskOrchestrations(typeof(EventConversationOrchestration.Responder));
             worker.AddTaskOrchestrations(typeof(HelloSequenceWithTimer));
+            worker.AddTaskOrchestrations(typeof(HelloSequenceException));
+            worker.AddTaskActivities(typeof(ThrowException));
+
             await worker.StartAsync();
 
             TaskHubClient client = new TaskHubClient(serviceClient);
@@ -63,6 +68,15 @@ namespace OpenTelemetrySample
             await client.WaitForOrchestrationAsync(helloSeqInstance, TimeSpan.FromMinutes(5));
 
             Console.WriteLine("Done with Hello Sequence!");
+
+            // Hello Sequence throws exception
+            OrchestrationInstance helloSeqExceptionInstance = await client.CreateOrchestrationInstanceAsync(
+                typeof(HelloSequenceException),
+                input: null);
+            await client.WaitForOrchestrationAsync(helloSeqExceptionInstance, TimeSpan.FromMinutes(5));
+
+            Console.WriteLine("Done with Hello Sequence Exception!");
+
 
             // Hello Sequence with Timer
             OrchestrationInstance helloSeqWithTimerInstance = await client.CreateOrchestrationInstanceAsync(typeof(HelloSequenceWithTimer), null);
@@ -134,6 +148,27 @@ namespace OpenTelemetrySample
             {
                 Thread.Sleep(1000);
                 return $"Hello, {input}!";
+            }
+        }
+
+        class HelloSequenceException : TaskOrchestration<string, string>
+        {
+            public override async Task<string> RunTask(OrchestrationContext context, string input)
+            {
+                string output = "";
+                output += await context.ScheduleTask<string>(typeof(SayHello), "Tokyo") + ", ";
+                output += await context.ScheduleTask<string>(typeof(SayHello), "London") + ", ";
+                output += await context.ScheduleTask<string>(typeof(ThrowException), "This is an invalid operation.") + ", ";
+                output += await context.ScheduleTask<string>(typeof(SayHello), "Seattle");
+                return output;
+            }
+        }
+
+        class ThrowException : TaskActivity<string, string>
+        {
+            protected override string Execute(TaskContext context, string input)
+            {
+                throw new InvalidOperationException(input);
             }
         }
 
--- a/samples/DistributedTraceSample/Program.cs
+++ b/samples/DistributedTraceSample/Program.cs
@@ -118,7 +118,7 @@ namespace OpenTelemetrySample
             await client.WaitForOrchestrationAsync(helloSubOrchFailedInstance, TimeSpan.FromMinutes(5));
 
             Console.WriteLine("Done with Hello Sub-orchestration with failed sub-orchestration!");
-            
+
             // Hello Sequence with Timer
             OrchestrationInstance helloSeqWithTimerInstance = await client.CreateOrchestrationInstanceAsync(typeof(HelloSequenceWithTimer), null);
             await client.WaitForOrchestrationAsync(helloSeqWithTimerInstance, TimeSpan.FromMinutes(5));
--- a/samples/DistributedTraceSample/Program.cs
+++ b/samples/DistributedTraceSample/Program.cs
@@ -43,6 +43,7 @@ namespace OpenTelemetrySample
                 await GetOrchestrationServiceAndClient();
 
             using TaskHubWorker worker = new TaskHubWorker(service);
+            // worker.ErrorPropagationMode = ErrorPropagationMode.SerializeExceptions;
             worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;
 
             worker.AddTaskOrchestrations(typeof(HelloSubOrch));
@@ -61,6 +62,12 @@ namespace OpenTelemetrySample
             worker.AddTaskOrchestrations(typeof(HelloSequenceException));
 
             await worker.StartAsync();
+
+            // Uncomment the next 2 lines if ErrorPropagationMode is SerializeExceptions and
+            // you would like to emit exception details
+
+            // worker.TaskActivityDispatcher.IncludeDetails = true;
+            // worker.TaskOrchestrationDispatcher.IncludeDetails = true;
 
             TaskHubClient client = new TaskHubClient(serviceClient);
 
@@ -111,7 +118,7 @@ namespace OpenTelemetrySample
             await client.WaitForOrchestrationAsync(helloSubOrchFailedInstance, TimeSpan.FromMinutes(5));
 
             Console.WriteLine("Done with Hello Sub-orchestration with failed sub-orchestration!");
-
+            
             // Hello Sequence with Timer
             OrchestrationInstance helloSeqWithTimerInstance = await client.CreateOrchestrationInstanceAsync(typeof(HelloSequenceWithTimer), null);
             await client.WaitForOrchestrationAsync(helloSeqWithTimerInstance, TimeSpan.FromMinutes(5));
@@ -173,7 +180,7 @@ namespace OpenTelemetrySample
                 string result = "";
                 result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
                 result += await context.ScheduleTask<string>(typeof(SayHello), "Tokyo");
-                Task<string> fanOut = context.CreateSubOrchestrationInstance<string>(typeof(ExceptionOrchestration), null);
+                Task<string> fanOut = context.CreateSubOrchestrationInstance<string>(typeof(ExceptionOrchestration), "activity threw an exception");
                 result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
                 result += await fanOut;
 
@@ -185,7 +192,7 @@ namespace OpenTelemetrySample
         {
             public override async Task<string> RunTask(OrchestrationContext context, string input)
             {
-                throw new Exception();
+                throw new Exception(input);
             }
         }
 
--- a/samples/DistributedTraceSample/Program.cs
+++ b/samples/DistributedTraceSample/Program.cs
@@ -65,12 +65,19 @@ namespace OpenTelemetrySample
 
             // Hello Sequence
             OrchestrationInstance helloSeqInstance = await client.CreateOrchestrationInstanceAsync(
-                typeof(HelloFanOut),
-                //typeof(HelloSequence),
+                typeof(HelloSequence),
                 input: null);
             await client.WaitForOrchestrationAsync(helloSeqInstance, TimeSpan.FromMinutes(5));
 
             Console.WriteLine("Done with Hello Sequence!");
+
+            // Hello Fan Out
+            OrchestrationInstance helloFanOutInstance = await client.CreateOrchestrationInstanceAsync(
+                typeof(HelloFanOut),
+                input: null);
+            await client.WaitForOrchestrationAsync(helloSeqInstance, TimeSpan.FromMinutes(5));
+
+            Console.WriteLine("Done with Hello Fan Out!");
 
             // Hello Sequence throws exception
             OrchestrationInstance helloSeqExceptionInstance = await client.CreateOrchestrationInstanceAsync(
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -130,7 +130,13 @@ namespace DurableTask.AzureStorage
 
         internal static bool TryGetLargeMessageReference(string messagePayload, out Uri blobUrl)
         {
-            return Uri.TryCreate(messagePayload, UriKind.Absolute, out blobUrl);
+            if (Uri.IsWellFormedUriString(messagePayload, UriKind.Absolute))
+            {
+                return Uri.TryCreate(messagePayload, UriKind.Absolute, out blobUrl);
+            }
+
+            blobUrl = null;
+            return false;
         }
 
         public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMessage, string queueName)
--- a/src/DurableTask.Core/TaskActivityDispatcher.cs
+++ b/src/DurableTask.Core/TaskActivityDispatcher.cs
@@ -206,7 +206,7 @@ namespace DurableTask.Core
                                 : null;
                             responseEvent = new TaskFailedEvent(-1, scheduledEvent.EventId, e.Message, details, new FailureDetails(e));
 
-                            traceActivity?.SetStatus(ActivityStatusCode.Error, details);
+                            traceActivity?.SetStatus(ActivityStatusCode.Error, e.Message);
 
                             this.logHelper.TaskActivityFailure(orchestrationInstance, scheduledEvent.Name, (TaskFailedEvent)responseEvent, e);
                         }
@@ -224,7 +224,7 @@ namespace DurableTask.Core
                     string? details = this.IncludeDetails ? e.Details : null;
                     var failureEvent = new TaskFailedEvent(-1, scheduledEvent.EventId, e.Message, details, e.FailureDetails);
 
-                    traceActivity?.SetStatus(ActivityStatusCode.Error, details);
+                    traceActivity?.SetStatus(ActivityStatusCode.Error, e.Message);
 
                     this.logHelper.TaskActivityFailure(orchestrationInstance, scheduledEvent.Name, failureEvent, e);
                     CorrelationTraceClient.Propagate(() => CorrelationTraceClient.TrackException(e));
--- a/src/DurableTask.Core/TaskActivityDispatcher.cs
+++ b/src/DurableTask.Core/TaskActivityDispatcher.cs
@@ -205,6 +205,9 @@ namespace DurableTask.Core
                                 ? $"Unhandled exception while executing task: {e}"
                                 : null;
                             responseEvent = new TaskFailedEvent(-1, scheduledEvent.EventId, e.Message, details, new FailureDetails(e));
+
+                            traceActivity?.SetStatus(ActivityStatusCode.Error, details);
+
                             this.logHelper.TaskActivityFailure(orchestrationInstance, scheduledEvent.Name, (TaskFailedEvent)responseEvent, e);
                         }
 
@@ -220,12 +223,17 @@ namespace DurableTask.Core
                     TraceHelper.TraceExceptionInstance(TraceEventType.Error, "TaskActivityDispatcher-ProcessTaskFailure", taskMessage.OrchestrationInstance, e);
                     string? details = this.IncludeDetails ? e.Details : null;
                     var failureEvent = new TaskFailedEvent(-1, scheduledEvent.EventId, e.Message, details, e.FailureDetails);
+
+                    traceActivity?.SetStatus(ActivityStatusCode.Error, details);
+
                     this.logHelper.TaskActivityFailure(orchestrationInstance, scheduledEvent.Name, failureEvent, e);
                     CorrelationTraceClient.Propagate(() => CorrelationTraceClient.TrackException(e));
                     result = new ActivityExecutionResult { ResponseEvent = failureEvent };
                 }
                 catch (Exception middlewareException) when (!Utils.IsFatal(middlewareException))
                 {
+                    traceActivity?.SetStatus(ActivityStatusCode.Error, middlewareException.Message);
+
                     // These are considered retriable
                     this.logHelper.TaskActivityDispatcherError(workItem, $"Unhandled exception in activity middleware pipeline: {middlewareException}");
                     throw;
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -752,26 +752,26 @@ namespace DurableTask.Core
                         SubOrchestrationInstanceCreatedEvent subOrchestrationCreatedEvent = (SubOrchestrationInstanceCreatedEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == subOrchestrationInstanceCompletedEvent.TaskScheduledId);
 
                         // We immediately publish the activity span for this sub-orchestration by creating the activity and immediately calling Dispose() on it.
-                        TraceHelper.EmitTraceActivityForSubOrchestrationFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, subOrchestrationCreatedEvent);
+                        TraceHelper.EmitTraceActivityForSubOrchestrationCompleted(workItem.OrchestrationRuntimeState.OrchestrationInstance, subOrchestrationCreatedEvent);
                     }
                     else if (historyEvent is SubOrchestrationInstanceFailedEvent subOrchestrationInstanceFailedEvent)
                     {
                         SubOrchestrationInstanceCreatedEvent subOrchestrationCreatedEvent = (SubOrchestrationInstanceCreatedEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == subOrchestrationInstanceFailedEvent.TaskScheduledId);
 
                         // We immediately publish the activity span for this sub-orchestration by creating the activity and immediately calling Dispose() on it.
-                        TraceHelper.EmitTraceActivityForSubOrchestrationFinished(workItem.OrchestrationRuntimeState.OrchestrationInstance, subOrchestrationCreatedEvent, subOrchestrationInstanceFailedEvent, this.errorPropagationMode);
+                        TraceHelper.EmitTraceActivityForSubOrchestrationFailed(workItem.OrchestrationRuntimeState.OrchestrationInstance, subOrchestrationCreatedEvent, subOrchestrationInstanceFailedEvent, this.errorPropagationMode);
                     }
                 }
 
                 if (message.Event is TaskCompletedEvent taskCompletedEvent)
                 {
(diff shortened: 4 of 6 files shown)


samples/DistributedTraceSample/Program.cs
```diff
@@ -64,6 +69,15 @@ static async Task Main(string[] args)
 
             Console.WriteLine("Done with Hello Sequence!");
 
+            // Hello Sequence throws exception
+            OrchestrationInstance helloSeqExceptionInstance = await client.CreateOrchestrationInstanceAsync(
+                typeof(HelloSequenceException),
+                input: null);
+            await client.WaitForOrchestrationAsync(helloSeqExceptionInstance, TimeSpan.FromMinutes(5));
```
bachuv, line 76: When I test the changes in this PR using the sample under the OpenTelemetrySample project, I get the error below. Has anyone seen this error before? I tried stepping through the code, but don't understand why it's getting in this path. 

The exception is thrown in Program.cs line 76:
`await client.WaitForOrchestrationAsync(helloSeqExceptionInstance, TimeSpan.FromMinutes(5)); Exception:System.Argument`

Exception: `address (Parameter 'Missing account name information inside path style uri. Path style uris should be of the form http://<IPAddressPlusPort>/<accountName>')`

When I step through the code, I can see that the `address` value is `durabletask.core.exceptions.taskfailedexception: Task 'OpenTelemetrySample.Program+ThrowException' (#1) failed with an unhandled exception: This is an invalid operation.)`

Stack trace:
```
at Microsoft.WindowsAzure.Storage.Core.Util.NavigationHelper.GetServiceClientBaseAddress(Uri addressUri, Nullable`1 usePathStyleUris)
   at Microsoft.WindowsAzure.Storage.Core.Util.NavigationHelper.GetServiceClientBaseAddress(StorageUri addressUri, Nullable`1 usePathStyleUris)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlob.ParseQueryAndVerify(StorageUri address, StorageCredentials credentials)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlob..ctor(StorageUri blobAbsoluteUri, Nullable`1 snapshotTime, StorageCredentials credentials)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob..ctor(StorageUri blobAbsoluteUri, Nullable`1 snapshotTime, StorageCredentials credentials)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob..ctor(Uri blobAbsoluteUri, Nullable`1 snapshotTime, StorageCredentials credentials)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob..ctor(Uri blobAbsoluteUri, StorageCredentials credentials)
   at DurableTask.AzureStorage.Storage.Blob..ctor(AzureStorageClient azureStorageClient, CloudBlobClient blobClient, Uri blobUri) in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\Storage\Blob.cs:line 42
   at DurableTask.AzureStorage.Storage.AzureStorageClient.GetBlobReference(Uri blobUri) in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\Storage\AzureStorageClient.cs:line 88
   at DurableTask.AzureStorage.MessageManager.DownloadAndDecompressAsBytesAsync(Uri blobUri) in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\MessageManager.cs:line 209
   at DurableTask.AzureStorage.MessageManager.<FetchLargeMessageIfNecessary>d__12.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\MessageManager.cs:line 123
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at DurableTask.AzureStorage.Tracking.AzureTableTrackingStore.<ConvertFromAsync>d__40.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\Tracking\AzureTableTrackingStore.cs:line 568
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at DurableTask.AzureStorage.Tracking.AzureTableTrackingStore.<FetchInstanceStatusInternalAsync>d__37.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\Tracking\AzureTableTrackingStore.cs:line 451
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at DurableTask.AzureStorage.Tracking.AzureTableTrackingStore.<GetStateAsync>d__35.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\Tracking\AzureTableTrackingStore.cs:line 420
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at DurableTask.AzureStorage.AzureStorageOrchestrationService.<GetOrchestrationStateAsync>d__105.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\AzureStorageOrchestrationService.cs:line 1665
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at DurableTask.AzureStorage.AzureStorageOrchestrationService.<WaitForOrchestrationAsync>d__119.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\src\DurableTask.AzureStorage\AzureStorageOrchestrationService.cs:line 1864
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at OpenTelemetrySample.Program.<Main>d__0.MoveNext() in C:\Users\vabachu\GitHub\durabletask-otel-suborchestrations\durabletask\samples\DistributedTraceSample\Program.cs:line 77
```

bachuv: I discussed this offline with Amanda and she said that it's storing the output as a blob (like when there are large messages) and she pointed me to PR #819 to see if that was the cause. The sample I was using didn't have large messages though.

To rule out that the new distributed tracing changes were causing this, I reproduced this in the DurableTask.Samples project (existing sample project) without any distributed tracing changes and used the repro steps below. I reproduced it with and without the changes in #819 and only observed the issue after that PR was merged.

Repro steps:
1. Make sure `taskHubWorker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;` is set
3. Throw an exception from an activity and do not handle it in the orchestration.

Expectation: The orchestration finishes execution and the state is Failed.
Actual: The ArgumentException is thrown about the `blobUri`

I checked these commits when reproducing the issue:
1. Before: changes in PR 819 - c6019e9095cd3453a5c891eee13ac65c1b90afdb
2. After changes in PR 819 - 7a5d1774135214ff30c74a236f1670425715e059

@cgillum, do you have any pointers to help understand why this is happening, anything else I can check, or if there's a workaround for this? Also, let me know if I'm misunderstanding or missing something. Thanks!

cgillum: I'm not sure why #819 would be related to something like this. I think the best path forward is to step through this in the debugger at the point right before we call `MessageManager.DownloadAndDecompressAsBytesAsync`. 

Specifically, we want to understand why the `DownloadAndDecompressAsBytesAsync` method is being called (it shouldn't be, since there's no URL for us to be looking at). That should give us some clues as far as what to look for next.

cgillum: Also, you may want to try with and without PR https://github.com/Azure/durabletask/pull/817 which made some changes related to how we deal with certain blob URLs.

bachuv: The commits I pasted are actually a before and after of #817, not 819. I accidentally put the wrong PR number. I corrected it in the original message.

I stepped through the code and noticed that in [`WaitForOrchestrationAsync()`](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L1850), once the orchestration is at a terminal state, it will enter the following if block:

```
if (this.settings.FetchLargeMessageDataEnabled)
{
    state.Input = await this.messageManager.FetchLargeMessageIfNecessary(state.Input);
    state.Output = await this.messageManager.FetchLargeMessageIfNecessary(state.Output);
}
```

Here, if `this.settings.FetchLargeMessageDataEnabled` is true (which is the default), then it will check if it needs to download the message from blob storage. It will go into [`FetchLargeMessageIfNecessary()`](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/MessageManager.cs#L119) which checks to see if it can create a blob Url with the output. For the issue I was running into, `state.Output` was `durabletask.core.exceptions.taskfailedexception: Task 'OpenTelemetrySample.Program+ThrowException' (#1) failed with an unhandled exception: This is an invalid operation.)` which isn't a Url, but is treated like one in [`TryGetLargeMessageReference()`](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/MessageManager.cs#L131). Then, an `ArgumentException` is thrown from the [`Blob` constructor](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/Storage/Blob.cs#L42) when it's creating a `CloudBlockBlob`.

To avoid this, I added a check to see if the Uri is formatted correctly to `FetchLargeMessageIfNecessary()`:

New addition: `Uri.IsWellFormedUriString(message, UriKind.Absolute)`

```
public async Task<string> FetchLargeMessageIfNecessary(string message)
{
    if (Uri.IsWellFormedUriString(message, UriKind.Absolute) && TryGetLargeMessageReference(message, out Uri blobUrl))
    {
        return await this.DownloadAndDecompressAsBytesAsync(blobUrl);
    }
    else
    {
        return message;
    }
}
```

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -205,6 +205,9 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
                                 ? $"Unhandled exception while executing task: {e}"
                                 : null;
                             responseEvent = new TaskFailedEvent(-1, scheduledEvent.EventId, e.Message, details, new FailureDetails(e));
+
+                            traceActivity?.SetStatus(ActivityStatusCode.Error, details);
```
jviau, line 209: `details` here is a fairly large string - it will include the stack trace. I think we would just want `e.Message`.

I also wonder if there is anything from [OTel exceptions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md) we should be including here.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -854,6 +854,11 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
                     subOrchestrationFailedEvent.FailureDetails = completeOrchestratorAction.FailureDetails;
 
                     taskMessage.Event = subOrchestrationFailedEvent;
+
+                    if (DistributedTraceActivity.Current != null && completeOrchestratorAction.OrchestrationStatus == OrchestrationStatus.Failed)
```
cgillum, line 858: These null checks seem redundant since `AddFailureDetailsToActivity` already guards for null.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -854,6 +854,11 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
...
+                    {
+                        AddFailureDetailstoActivity(runtimeState);
```
cgillum, line 860: Aren't we already doing this several lines below here?

src/DurableTask.Core/TaskActivityDispatcher.cs
cgillum: +1 about using `e.Message` instead of `details`.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -220,12 +223,17 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
                     TraceHelper.TraceExceptionInstance(TraceEventType.Error, "TaskActivityDispatcher-ProcessTaskFailure", taskMessage.OrchestrationInstance, e);
                     string? details = this.IncludeDetails ? e.Details : null;
                     var failureEvent = new TaskFailedEvent(-1, scheduledEvent.EventId, e.Message, details, e.FailureDetails);
+
+                    traceActivity?.SetStatus(ActivityStatusCode.Error, details);
```
cgillum, line 227: Same comment as above.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -130,7 +130,13 @@ public async Task<string> FetchLargeMessageIfNecessary(string message)
 
         internal static bool TryGetLargeMessageReference(string messagePayload, out Uri blobUrl)
         {
-            return Uri.TryCreate(messagePayload, UriKind.Absolute, out blobUrl);
+            if (Uri.IsWellFormedUriString(messagePayload, UriKind.Absolute))
```
cgillum, line 133: Was `Uri.TryCreate(...)` throwing an exception before this change?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: This line is for sub-orchestrations. The one several lines below is for the parent orchestration.

src/DurableTask.AzureStorage/MessageManager.cs
bachuv: No, it was returning true.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: But `runtimeState` always refers to the "current" orchestration, which may be a sub-orchestration or a parent orchestration, right? And doesn't the second call below (lines 873-876) get called regardless? Since the input is exactly the same in either case, I don't quite understand why we need this call here (I'd expect that lines 873-876 is all we need).

bachuv: I was noticing that for sub-orchestrations, that it was exiting the method on [line 869](https://github.com/Azure/durabletask/blob/vabachu/addexceptioninfo/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L869) where it was returning the `taskMessage`.

cgillum: I see. Ugh, the deeply nested if-checks were throwing me off and I somehow didn't notice this. Thanks for pointing it out.

cgillum (approved):

samples/DistributedTraceSample/Program.cs
```diff
@@ -43,6 +43,8 @@ static async Task Main(string[] args)
                 await GetOrchestrationServiceAndClient();
 
             using TaskHubWorker worker = new TaskHubWorker(service);
+            worker.ErrorPropagationMode = ErrorPropagationMode.UseFailureDetails;
```
jviau, line 46: I was doing some testing and I think Durable Functions always uses `ErrorPropagationMode.SerializeExceptions`. How will this logic behave in that mode?

jviau (approved):
Please validate the behavior with `ErrorPropagationMode.SerializeExceptions`. But if that works out well, then this looks good.

samples/DistributedTraceSample/Program.cs
bachuv: Right now, if `ErrorPropagationMode` is `SerializeExceptions` then we don't log the exception details for orchestrations and sub-orchestrations.

This is the current way we call `SetStatus`. We rely on `FailureDetails`.
```
private void AddFailureDetailstoActivity(OrchestrationRuntimeState runtimeState)
{
    FailureDetails? failureDetails = runtimeState.FailureDetails;
    DistributedTraceActivity.Current?.SetStatus(ActivityStatusCode.Error, failureDetails?.ErrorMessage);
}
```

What is the value we want to send in SetStatus() as the description (instead of `failureDetails?.ErrorMessage`) if `ErrorPropagationMode` is `SerializeExceptions`? Is it `completeOrchestratorAction.Result`?

jviau: If it is set to `ErrorPropagationMode.SerializeExceptions`, are we able to deserialize as an exception and pull out `Message` still?

bachuv: Yes, I added some changes to deserialize the exception. `completeOrchestratorAction.Details` has the serialized exception.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -898,10 +898,20 @@ private void ResetDistributedTraceActivity(OrchestrationRuntimeState runtimeStat
...
-            DistributedTraceActivity.Current?.SetStatus(ActivityStatusCode.Error, failureDetails?.ErrorMessage);
+            string? statusDescription = "";
+            if (errorPropagationMode == ErrorPropagationMode.UseFailureDetails)
+            {
+                FailureDetails? failureDetails = runtimeState.FailureDetails;
+                statusDescription = failureDetails?.ErrorMessage;
+            }
+            else if (errorPropagationMode == ErrorPropagationMode.SerializeExceptions)
+            {
+                statusDescription = JsonDataConverter.Default.Deserialize<Exception>(details).Message;
```
bachuv, line 911: I noticed that `JsonDataConverter` was used in other places in `TaskOrchestrationDispatcher` to serialize information so I used the same converter here, but I also noticed that in `TaskOrchestrationContext` that there is an `ErrorDataConverter`. Is there recommended way to use that converter in `TaskOrchestrationDispatcher`?

samples/DistributedTraceSample/Program.cs
bachuv: I also added screenshots in the issue description based on the `ErrorPropagationMode`.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Hmm...good question. I think the idea is that we're _supposed_ to use `ErrorDataConveter` for all exception serialization. It's concerning if we're not using it consistently. You may want to do some testing to see which one is generally being used for exceptions that come from user code.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -181,13 +173,21 @@ static TraceHelper()
             // Adding additional tags for a SubOrchestrationInstanceFailedEvent
             if (failedEvent != null)
             {
-                activity.SetTag("otel.status_code", "ERROR");
-
-                FailureDetails? failureDetails = failedEvent.FailureDetails;
-                if (failureDetails != null)
+                string? statusDescription = "";
```
cgillum, line 176: You can remove the `?` here since `statusDescription` is never null.
```suggestion
                string statusDescription = "";
```

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -228,7 +228,7 @@ internal static void EmitTraceActivityForTimer(OrchestrationInstance? instance,
             }
         }
 
-        internal static void EmitActivityforTaskFinished(OrchestrationInstance? instance, TaskScheduledEvent taskScheduledEvent, TaskFailedEvent? taskFailedEvent = null)
+        internal static void EmitActivityforTaskFinished(OrchestrationInstance? instance, TaskScheduledEvent taskScheduledEvent, TaskFailedEvent? taskFailedEvent = null, ErrorPropagationMode errorPropagationMode = ErrorPropagationMode.SerializeExceptions)
```
cgillum, line 231: I suggest we not have a default value for this and always be explicit.
```suggestion
        internal static void EmitActivityforTaskFinished(OrchestrationInstance? instance, TaskScheduledEvent taskScheduledEvent, ErrorPropagationMode errorPropagationMode, TaskFailedEvent? taskFailedEvent = null)
```

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -259,13 +259,22 @@ internal static void EmitActivityforTaskFinished(OrchestrationInstance? instance
 
             if (taskFailedEvent != null)
             {
-                newActivity.SetTag("otel.status_code", "ERROR");
+                string? statusDescription = "";
```
cgillum, line 262:
```suggestion
                string statusDescription = "";
```

cgillum (commented):
This is looking pretty good to me. I think we just need to answer the question about which serializer to use when deserializing exception payloads.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: We likely won't be able to solve this problem as part of this PR, so I'd be okay with us just picking one for now and opening a GitHub issue to make exception serialization more consistent. We may end up never fixing the inconsistency anyways since `ErrorPropagationMode.UseFailureDetails` is arguably a better long-term solution, but it would be good to at least document it in an issue.

bachuv: I'm noticing that the default is `JsonDataConverter.Default` for `ErrorDataConverter`.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: I made it an optional parameter because it is only used if the `TaskActivity` or sub-orchestration fails. This parameter is unnecessary when they complete successfully so I refactored the code a little bit. I created separate TraceHelper methods for completed and failed `TaskActivity`s and sub-orchestrations.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
jviau: Yeah, it would require a lot of refactoring for us to get access to the correct `ErrorDataConverter` here. What you have here will have to be good enough for now.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -249,18 +266,99 @@ internal static void EmitActivityforTaskFinished(OrchestrationInstance? instance
...
+            activity?.Dispose();
+        }
+
+        internal static void EmitTraceActivityForTaskFailed(
+            OrchestrationInstance? orchestrationInstance,
+            TaskScheduledEvent taskScheduledEvent,
+            TaskFailedEvent? taskFailedEvent,
+            ErrorPropagationMode errorPropagationMode)
+        {
+            Activity? activity = CreateTraceActivityForTaskFinished(orchestrationInstance, taskScheduledEvent);
```
jviau, line 337: nit: recommend checking for null and returning right away:

```suggestion
            Activity? activity = CreateTraceActivityForTaskFinished(orchestrationInstance, taskScheduledEvent);
            
            if (activity is null)
            {
                return;
            }
```

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -161,26 +151,53 @@ public class TraceHelper
...
             {
-                activity.SetTag("otel.status_code", "ERROR");
-
-                FailureDetails? failureDetails = failedEvent.FailureDetails;
-                if (failureDetails != null)
+                string statusDescription = "";
+                if (errorPropagationMode == ErrorPropagationMode.SerializeExceptions)
                 {
-                    activity.SetTag("otel.status_description", failureDetails.ErrorMessage);
+                    statusDescription = JsonDataConverter.Default.Deserialize<Exception>(failedEvent.Details).Message;
```
jviau, line 186: An extension method (internal only) would be nice for extracting status description. Something like `failedEvent.GetStatusDescription()` - it can perform the internal check for `ErrorPropagationMode` and retrieve the description as appropriate.

jviau (approved):
LGTM. Some minor suggestions for code organization.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
bachuv: I wanted to clarify that I was noticing serialization in other places in `TaskOrchestrationDispatcher`, but it wasn't related to exception serialization. I was referring to those examples to understand what converter was used so I don't believe there's an inconsistency. This is one example:

`JsonDataConverter.Default.Serialize(runtimeState.GetOrchestrationRuntimeStateDump(), true)`

https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L332

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: If I add an extension method to `SubOrchestrationInstanceFailedEvent`, how would it check the `ErrorPropagationMode`? I don't see a clean way of doing that.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -134,25 +134,13 @@ public class TraceHelper
...
-        /// </summary>
-        /// <param name="orchestrationInstance">The associated orchestration instance metadata.</param>
-        /// <param name="createdEvent">The related sub-orchestration created event.</param>
-        /// <param name="failedEvent">The sub-orchestration failed event.</param>
-        /// <param name="errorPropagationMode">Specifies how to handle exceptions.</param>
-        /// <returns>
-        /// Returns a newly started <see cref="Activity"/> with (task) activity and orchestration-specific metadata.
-        /// </returns>
-        internal static void EmitTraceActivityForSubOrchestrationFinished(
+        internal static Activity? CreateTraceActivityForSubOrchestrationFinished(
```
cgillum, line 137: Is this the right method name? The name ends with `*SubOrchestrationFinished` but I don't see anything about "finished", "completed", or "failed" in the implementation. It looks like we're just starting a trace activity.

cgillum (approved):
One small method naming nit:

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: The method name was trying to say that the sub-orchestration completed or failed, but I see how it's misleading. I removed "Finished" so now it's `CreateTraceActivityForSubOrchestration`. Is that better?

cgillum: Yes, I like that better. 👍🏽

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Thanks for clarifying. I think it's fine to go with what we have for now as best-effort. We should add some exception handling, though, because we don't want any deserialization issues to cause orchestration processing failures. If there's an exception trying to deserialize the exception, maybe we can just put some generic message in the trace details.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: I'm going to merge this PR, but I'm happy to address this in another PR.

PR #855, opened 2023-01-30
Prevent starvation in lease ownership and add more telemetry to load balancer

davidmrdavid:
Resolves: https://github.com/Azure/azure-functions-durable-extension/issues/2373

This PR adds a timeout check in the partition shutdown logic. If the timeout is met, the process is forcibly terminated.
It also adds a few extra logs in places where exceptions are swallowed, to help us gain more granular insights during investigations.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
@@ -159,7 +159,7 @@ namespace DurableTask.AzureStorage.Partitioning
             var winner = await Task.WhenAny(shutdownTask, timeoutTask);
             if (winner == timeoutTask)
             {
-                Environment.FailFast("Process will be forceably shut down after 30 minutes");
+                Environment.FailFast("Process will be forceably shut down");
             }
         }
 
--- a/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
@@ -167,7 +167,8 @@ namespace DurableTask.AzureStorage.Partitioning
             // Time out task puts allows us to recover from a stuck worker scenario.
             // If this timeout is met, we forcibly terminate the process to free up
             // its leases.
-            var timeoutTask = Task.Delay(options.OwnershipLeaseReleaseTimeout);
+            var timeout = options.LeaseBalancerShutdownTimeout;
+            var timeoutTask = Task.Delay(timeout);
 
             var winner = await Task.WhenAny(shutdownTask, timeoutTask);
             if (winner == timeoutTask)
@@ -177,9 +178,9 @@ namespace DurableTask.AzureStorage.Partitioning
                     this.taskHub,
                     this.workerName,
                     string.Empty,
-                    "Worker took too long to shut down. Initiating forceful shutdown to prevent lease balacning starvation");
+                    "Lease balancer took too long to shut down. Initiating forceful shutdown of the worker process");
 
-                Environment.FailFast("Process was forcibly shut down");
+                Environment.FailFast($"Process was forcibly shut down due to a {timeout} timeout expiration while shutting down the lease manager");
             }
             else
             {
@@ -188,7 +189,7 @@ namespace DurableTask.AzureStorage.Partitioning
                     this.taskHub,
                     this.workerName,
                     string.Empty,
-                    "Lease resources were gracefully released.");
+                    "Lease balancer was successfully stopped");
             }
         }
 
--- a/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
@@ -35,6 +35,7 @@ namespace DurableTask.AzureStorage.Partitioning
         readonly LeaseObserverManager leaseObserverManager;
         readonly Func<string, bool> shouldAquireLeaseDelegate;
         readonly Func<string, bool> shouldRenewLeaseDelegate;
+        readonly OrchestrationSessionManager orchestrationSessionManager;
 
         int isStarted;
         bool shutdownComplete;
@@ -50,7 +51,8 @@ namespace DurableTask.AzureStorage.Partitioning
             ILeaseManager<T> leaseManager, 
             LeaseCollectionBalancerOptions options,
             Func<string, bool> shouldAquireLeaseDelegate = null,
-            Func<string, bool> shouldRenewLeaseDelegate = null)
+            Func<string, bool> shouldRenewLeaseDelegate = null,
+            OrchestrationSessionManager orchestrationSessionManager = null)
 
         {
             this.leaseType = leaseType;
@@ -67,6 +69,7 @@ namespace DurableTask.AzureStorage.Partitioning
             this.currentlyOwnedShards = new ConcurrentDictionary<string, T>();
             this.keepRenewingDuringClose = new ConcurrentDictionary<string, T>();
             this.leaseObserverManager = new LeaseObserverManager(this);
+            this.orchestrationSessionManager = orchestrationSessionManager;
         }
 
         private static bool DefaultLeaseDecisionDelegate(string leaseId)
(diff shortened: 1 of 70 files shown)


src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -133,25 +133,38 @@ public async Task StopAsync()
...
-            this.shutdownComplete = true;
+            var timeoutTask = Task.Delay(options.OwnershipLeaseReleaseTimeout);
 
-            if (this.renewTask != null)
+            var winner = await Task.WhenAny(shutdownTask, timeoutTask);
+            if (winner == timeoutTask)
             {
-                this.leaseRenewerCancellationTokenSource.Cancel();
-                await this.renewTask;
+                Environment.FailFast("Process will be forcibly shut down");
```
cgillum, line 168: nit: this message reads as if something will happen in the future, but in reality the forceable shutdown has already happened (also note that the exact timeout is configurable whereas this message hardcodes 30 minutes)

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -133,25 +133,44 @@ public async Task StopAsync()
...
+
+                if (this.renewTask != null)
+                {
+                    this.leaseRenewerCancellationTokenSource.Cancel();
+                    await this.renewTask;
+                }
+
+                this.leaseTakerCancellationTokenSource = null;
+                this.leaseRenewerCancellationTokenSource = null;
+            });
```
cgillum, line 159: I expect this approach will probably work in cases where lease rebalancing happens as part of a shutdown. I'm not 100% certain whether all the cases we've seen involve shutdowns, or whether some of them could involve lease rebalancing (without any shutdowns). Maybe this is sufficient for the cases we care about, but are there any other places in the code where we wait before releasing the ownership lease?

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
davidmrdavid: Ah yes. I removed the 30 minute note but it seems it might not have updated on your end by the time you wrote this comment.
I'm still working on the wording of this message, but good call out in the phrasing tense. I'll make it past tense.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
davidmrdavid: Unfortunately, I don't know if there are other such cases and, even if I assumed there were, I would not be sure how to handle them yet.

I do see there's a few places in the code where exceptions are ignored. I'm looking to add some logs there to try and give us more of a chance to tackle these issues in the future. Hopefully that will make it easier to reason about the cases you bring up.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancerOptions.cs
```diff
@@ -37,6 +37,12 @@ class LeaseCollectionBalancerOptions
         /// </summary>
         public TimeSpan LeaseInterval { get; set; } = TimeSpan.FromSeconds(30);
 
+        /// <summary>
+        /// Maximum alloted time to wait for a worker to release its ownership lease. If this timeout is met,
+        /// the worker will be forcibly terminated so that its lease may be load balanced.
+        /// </summary>
+        public TimeSpan OwnershipLeaseReleaseTimeout { get; set; } = TimeSpan.FromMinutes(10);
```
cgillum, line 44: I think we need to rename this to something more accurate, like `LeaseBalancerShutdownTimeout`. AFAICT the code you wrote doesn't know anything about "ownership" leases, per-se.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -133,25 +133,67 @@ public async Task StopAsync()
...
             {
-                this.leaseRenewerCancellationTokenSource.Cancel();
-                await this.renewTask;
-            }
+                this.settings.Logger.PartitionManagerError(
+                    this.accountName,
+                    this.taskHub,
+                    this.workerName,
+                    string.Empty,
+                    "Worker took too long to shut down. Initiating forceful shutdown to prevent lease balacning starvation");
```
cgillum, line 180: I suggest we reword this a bit for clarity: "Lease balancer took too long to shutdown. Initiating forceful shutdown of the worker process.".

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -133,25 +133,67 @@ public async Task StopAsync()
...
 
-            this.leaseTakerCancellationTokenSource = null;
-            this.leaseRenewerCancellationTokenSource = null;
+                Environment.FailFast("Process was forcibly shut down");
```
cgillum, line 182: Consider this alternative text: `$"Process was forcibly shut down due to a {options.OwnershipLeaseReleaseTimeout} timeout expiration in shutting down the lease manager."

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -133,25 +133,67 @@ public async Task StopAsync()
...
+            }
+            else
+            {
+                this.settings.Logger.PartitionManagerInfo(
+                    this.accountName,
+                    this.taskHub,
+                    this.workerName,
+                    string.Empty,
+                    "Lease resources were gracefully released.");
```
cgillum, line 191: Consider this alternative text: "Lease balancer was successfully stopped"

cgillum (commented):
Just a few nit-picky error message comments but otherwise I think this looks good.

davidmrdavid:
Just for visibility, I've applied all your feedback @cgillum

davidmrdavid:
@cgillum: Hmm, I would rather not delay that Extension release at this time. I'd say let's keep that release going and, if necessary, make another faster release afterwards with this patch.

I also realize @amdeel may want to chime in on this one, so I think I'll wait for now. 

I'm starting some manual validation of this changes now and, if they pass, I'll publish these changes as a private package :)

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
```diff
@@ -133,25 +133,68 @@ public async Task StopAsync()
...
+                this.settings.Logger.PartitionManagerError(
+                    this.accountName,
+                    this.taskHub,
+                    this.workerName,
+                    string.Empty,
+                    "Lease balancer took too long to shut down. Initiating forceful shutdown of the worker process");
 
-            this.leaseTakerCancellationTokenSource = null;
-            this.leaseRenewerCancellationTokenSource = null;
+                Environment.FailFast($"Process was forcibly shut down due to a {timeout} timeout expiration while shutting down the lease manager");
```
jviau, line 183: I am not an expert with this, but it doesn't feel right to include a `FailFast` in a shutdown scenario. I would expect the external actor shutting this down to perform the fail-fast and not us.

jviau (commented):
I am not too familiar with the lease code. But generally, a timeout on shutdown is performed by taking in a `CancellationToken` on your stop method and letting the caller control timeout by cancelling that token when you have exceeded it. This is still cooperative cancellation but that is fine - a forceful termination is handled by the caller performing their own `FailFast` (either by code, or via external process termination).

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancerOptions.cs
```diff
@@ -37,6 +37,12 @@ class LeaseCollectionBalancerOptions
         /// </summary>
         public TimeSpan LeaseInterval { get; set; } = TimeSpan.FromSeconds(30);
 
+        /// <summary>
+        /// Maximum alloted time to wait for a worker to shut down its lease manager. If this timeout is met,
+        /// the worker will be forcibly terminated.
+        /// </summary>
+        public TimeSpan LeaseBalancerShutdownTimeout { get; set; } = TimeSpan.FromMinutes(10);
```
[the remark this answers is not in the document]
davidmrdavid: I agree 10 minutes is long. In general, this PR's goal is to terminate a worker that is perpetually stuck, and it's inspired by a recent IcM where we saw a linux DF app fail to shut down for **6 days** , until a manual application restart took place. So, while the 10 minute timeout is long, it is still shorter than potentially days of waiting which is what I'm trying to prevent :-) . I don't mind reducing the timeout further though, if other reviewers agree.

I think one of the scenarios where the lease balancer is expected to shutu down is during app shut down. I'm unsure if there's others.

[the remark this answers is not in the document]
davidmrdavid: So, ideally, this FailFast will almost never be needed. It's really just trying to mitigate a worst case scenario.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancer.cs
davidmrdavid: I agree that ideally we would have the Host do this for us, but we have seen DF apps be stuck in the shutdown step for indefinite time before (hours, days) so I think the Host isn't currently doing this for us. Perhaps that's a work item we can coordinate with them and I'd be happy to bring it to the team!

In the meantime, and given a conservative enough timeout on our end, I figured it is "ok" to have an auto-mitigating `FailFast` to prevent a DF instance that seems unable to make progress. I think we do this in other scenarios, like when we detect a Deadlock.

src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancerOptions.cs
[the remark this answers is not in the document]
jviau: If the scenario is a DF app failed to shutdown due to this, can we not implement something in the DF extension which does a timer and a `Task.WhenAny(stopTask, timerTask)` and will abort waiting on the lease shutdown itself?

For changes to the this SDK though, I see we have a `StopAsync()` chain starting in `AzureStorageOrchestrationService` -> `AppLeaseManager` -> `IPartitionManager` -> `LeaseCollectionBallancerOptions`. To improve shutdown, we should do one of the following:

1. Quickest: `AzureStorageOrchestrationService.StopAsync` already takes in a bool `isForced`. We can plumb this bool through the call chain mentioned above, so that callers can force a shutdown after a certain time by calling `StopAsync(true)` for a second time.
2. Best: we add a `StopAsync(CancellationToken)` overload to `AzureStorageOrchestrationService` and then plumb that token down through the above call chain. Now callers can cancel that token to force the shutdown.

davidmrdavid:
Closing this PR for now. I'll be re-using this branch simply to track the private packages I'm creating for the affected customers. Once we reach a resolution, I think it'll be easier to determine what changes from the private package to bring into main. Thanks all for the reviews.

PR #861, opened 2023-02-08
Add README for Distributed Tracing with OpenTelemetry sample

bachuv:
This PR adds a README to explain how the Distributed Tracing sample app is configured to emit traces to different telemetry exporters. It also simplifies the sample.


samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,58 @@
+# Distributed Tracing with Open Telemetry for Durable Task
+
```
bachuv, line 2: I have a few questions about this README that I would like to get feedback on:
1. Should there be more intro that explains OpenTelemetry and why we are using it?
2. Should the quickstart only focus on Application Insights or should it list out examples of using the most common telemetry exporters?
3. Should we keep this separate sample app for using Distributed Tracing with OpenTelemetry? The existing "DurableTask.Samples` sample app could also be configured with distributed tracing and we could remove this sample application. I want to get input to understand if this additional sample is adding value to customers. One option could be to simplify this additional sample to run one or two orchestrations and keep the focus on the set up.

jviau: 1. Recommend including at least a link to https://opentelemetry.io/
2. I like what you have: showing AppInsights, ZipKin, and Console
3. Having a separate sample that runs only 1 top-level orchestration but includes sub-orchestrations and activities which shows all the various possibilities would be good. We may want to remove the existing distributed tracing sample when this is shipped.

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+## Tracer Provider
+
+The following startup code is added to create a tracer provider. This code is necessary to add at startup to ensure that the traces are collected and emitted to the correct telemetry exporters. It specifies the service name for the app, which source that the traces should be collected from, and the telemetry exporters where the traces get emitted. "DurableTask" is the service name that will emit the Durable Task related traces.
+
+```
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .SetResourceBuilder(ResourceBuilder.CreateDefault().AddService("MySample"))
+    .AddSource("DurableTask")
```
jviau, line 27: This work may not be tracked, but we should have a `Microsoft.Azure.DurableTask.OpenTelemetry` package which includes helpers for the OTel SDK. With that package, a customer would be able to call:

```suggestion
    .AddDurableTask()
```

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,58 @@
+# Distributed Tracing with Open Telemetry for Durable Task
+
+This sample explains how the OpenTelemetrySample project is configured with Distributed Tracing and emits trace information to the following telemetry exporters: Console, Application Insights, and ZipKin. The following steps can be used to configure your own app as well.
+
+## Prerequisites
+If you would like to emit traces to Application Insights then create an Application Insights resource in the Azure Portal. If you would like to emit traces to Zipkin, then follow the instructions [here](https://zipkin.io/pages/quickstart.html) to start a local instance.
```
davidmrdavid, line 6: nitpick: can we link to Application Insights? https://learn.microsoft.com/en-us/azure/azure-monitor/app/app-insights-overview?tabs=net

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,58 @@
...
+The following package references are added to OpenTelemetrySample.csproj so that the app can emit traces to the console, Application Insights, and Zipkin. You can also configure other telemetry exporters by adding those packages.
+
+```
+<ItemGroup>
+    <PackageReference Include="OpenTelemetry.Exporter.Console" Version=""1.1.0 />
+    <PackageReference Include="OpenTelemetry.Exporter.Zipkin" Version="1.1.0" />
+    <PackageReference Include="Azure.Monitor.OpenTelemetry.Exporter" Version="1.0.0-beta.3" />
+</ItemGroup>
+```
+
```
davidmrdavid, line 19: _Super nitpick_: can we order these packages in the same order as they are introduced in your paragraph: console, then Application Insights, then Zipkin?

samples/DistributedTraceSample/README.md
davidmrdavid: 1. I think a small blurb would help. I think just a few sentences explaining why DF users benefit from OpenTelemetry and Distributed Tracing, which may revolve around diagnosing issues and monitoring their app health/perf.
2. +1 to what Jacob said
3. No major thoughts on this.

bachuv: Thanks for the feedback! I updated the description and simplified the sample.

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+## Run the application
+
+Run the sample app to execute the orchestrations and emit the traces.
+
+## Looking through the traces
+
+### Console Exporter
+Traces emitted to the console should appear in the Visual Studio Debug Console. This telemetry exporter is usually used for testing purposes.
+
+![Console Exporter](images/ConsoleExporter.png)
```
cgillum, line 52: Reminder: This screenshot needs to be updated with the latest span and attribute names from https://github.com/Azure/durabletask/pull/864.

samples/DistributedTraceSample/Program.cs
```diff
@@ -71,77 +61,10 @@ static async Task Main(string[] args)
 
             TaskHubClient client = new TaskHubClient(serviceClient);
 
-            // Hello Sequence
-            OrchestrationInstance helloSeqInstance = await client.CreateOrchestrationInstanceAsync(
-                typeof(HelloSequence),
+            OrchestrationInstance orchestrationInstance = await client.CreateOrchestrationInstanceAsync(
+                typeof(Orchestration),
                 input: null);
```
davidmrdavid, line 66: non-blocking nit: does the `input` show in our Dist. Tracing visualization? If so, I can see a benefit in having a non-null input in the sample.

samples/DistributedTraceSample/Program.cs
```diff
@@ -158,36 +81,21 @@ static async Task<(IOrchestrationService, IOrchestrationServiceClient)> GetOrche
...
         {
             public override async Task<string> RunTask(OrchestrationContext context, string input)
             {
                 string result = "";
                 result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
                 result += await context.ScheduleTask<string>(typeof(SayHello), "Tokyo");
-                Task<string> parallelTask = context.CreateSubOrchestrationInstance<string>(typeof(HelloFanOut), null);
+                Task<string> parallelTask = context.CreateSubOrchestrationInstance<string>(typeof(ExceptionOrchestration), "orchestration threw an exception");
                 result += await context.CreateSubOrchestrationInstance<string>(typeof(HelloSequence), null);
                 result += await parallelTask;
```
davidmrdavid, line 93: Won't calling the `ExceptionOrchestration` cause this orchestrator to fail before it gets to call `HelloSequence` again?

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,62 @@
+# Distributed Tracing with OpenTelemetry for Durable Task
+
+Enabling distributed tracing in Durable Task allows you to take a closer look at orchestrations and activities and investigate issues quicker.
+
+[OpenTelemetry](https://opentelemetry.io/) provides a way to collect and export telemetry to an observability tool. It is becoming the new standard for distributed tracing and is vendor neutral. Some examples of telemetry exporters that you could choose include Application Insights, Zipkin, Jaeger, and New Relic.
+
```
davidmrdavid, line 6: I noticed Distributed Tracing is lowercased here ("distributed tracing") but later capitalized in this readme. Can we always call it "Distributed Tracing" or just keep it consistent? :)

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+    .AddConsoleExporter()
+    .AddZipkinExporter()
+    .AddAzureMonitorTraceExporter(options =>
+    {
+        options.ConnectionString = Environment.GetEnvironmentVariable("AZURE_MONITOR_CONNECTION_STRING");
+    })
+    .Build();
+```
+
+For Application Insights, add an app setting with the name `AZURE_MONITOR_CONNECTION_STRING`. If you are running this locally, then add an environment variable. If you are running this in Azure, then add an app setting under **Configuration** in the portal. The value should be the Application Insights resource connection string.
```
davidmrdavid, line 41:
```suggestion
For Application Insights, add an app setting with the name `AZURE_MONITOR_CONNECTION_STRING`. If you are running this locally, then add it as an environment variable. If you are running this in Azure, then add an app setting under **Configuration** in the portal. The value should be the Application Insights resource connection string.
```

davidmrdavid (commented):
My only blocker is whether the top-level orchestration will error out before scheduling the final task. Otherwise, this looks good to me.

cgillum:
@bachuv were you planning on adding the new tracing nuget package to this PR or will that come as part of a separate PR?

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,62 @@
+# Distributed Tracing with OpenTelemetry for Durable Task
+
+Enabling distributed tracing in Durable Task allows you to take a closer look at orchestrations and activities and investigate issues quicker.
+
+[OpenTelemetry](https://opentelemetry.io/) provides a way to collect and export telemetry to an observability tool. It is becoming the new standard for distributed tracing and is vendor neutral. Some examples of telemetry exporters that you could choose include Application Insights, Zipkin, Jaeger, and New Relic.
+
+This sample explains how the OpenTelemetrySample project is configured with Distributed Tracing and emits trace information to the following telemetry exporters: Console, Application Insights, and ZipKin. The following steps can be used to configure your own app as well.
+
+## Prerequisites
+If you would like to emit traces to [Application Insights](https://learn.microsoft.com/en-us/azure/azure-monitor/app/app-insights-overview?tabs=net) then create an Application Insights resource in the Azure Portal. If you would like to emit traces to Zipkin, then follow the instructions [here](https://zipkin.io/pages/quickstart.html) to start a local instance.
```
nytiannn, line 10:
```suggestion
If you would like to emit traces to [Application Insights](https://learn.microsoft.com/en-us/azure/azure-monitor/app/app-insights-overview?tabs=net), then create an Application Insights resource in the Azure Portal. If you would like to emit traces to Zipkin, then follow the instructions [here](https://zipkin.io/pages/quickstart.html) to start a local instance.
```

samples/DistributedTraceSample/README.md
nytiannn, line 41:
```suggestion
For **Application Insights**, add an app setting with the name `AZURE_MONITOR_CONNECTION_STRING`. If you are running this locally, then add an environment variable. If you are running this in Azure, then add it under **Configuration** page in the portal. The value should be the Application Insights resource connection string.
```

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Looking through the traces
+
+### Console Exporter
+Traces emitted to the console should appear in the Visual Studio Debug Console. This telemetry exporter is usually used for testing purposes.
+
+![Console Exporter](images/ConsoleExporter.png)
+
+### Application Insights
+To view traces in Application Insights, navigate to your Application Insights resource, go to **Transaction Search**, click on an entry and a Gantt chart should appear showing a visual of the trace and the spans.
```
nytiannn, line 55:
```suggestion
To view traces in Application Insights, navigate to your Application Insights resource, go to **Transaction Search**, and click on an entry. Then a Gantt chart should appear showing a visual of the trace and the spans.
```

nytiannn (commented):
Just several small suggestions. :)

samples/DistributedTraceSample/Program.cs
bachuv: If I remember correctly, we decided against it since there could be sensitive information. @cgillum let me know if you think otherwise.

samples/DistributedTraceSample/Program.cs
bachuv: The task is created on line 91, but is actually executed on line 93 (after `HelloSequence` is called) so it won't fail before calling `HelloSequence`.

samples/DistributedTraceSample/README.md
bachuv: Thanks for the reminder! I'll update this PR once #864 is merged.

bachuv:
> @bachuv were you planning on adding the new tracing nuget package to this PR or will that come as part of a separate PR?

@cgillum, I'm going to add it in a separate PR.

samples/DistributedTraceSample/Program.cs
cgillum: Yeah, we wouldn't want to trace the actual input. We could decide to trace something like the size of the input (e.g. the number of bytes), but we should probably discuss that in the spec design first.

samples/DistributedTraceSample/README.md
```diff
@@ -0,0 +1,62 @@
+# Distributed Tracing with OpenTelemetry for Durable Task
+
+Enabling distributed tracing in Durable Task allows you to take a closer look at orchestrations and activities and investigate issues quicker.
+
+[OpenTelemetry](https://opentelemetry.io/) provides a way to collect and export telemetry to an observability tool. It is becoming the new standard for distributed tracing and is vendor neutral. Some examples of telemetry exporters that you could choose include Application Insights, Zipkin, Jaeger, and New Relic.
+
+This sample explains how the OpenTelemetrySample project is configured with distributed tracing and emits trace information to the following telemetry exporters: Console, Application Insights, and ZipKin. The following steps can be used to configure your own app as well.
```
cgillum, line 7:
```suggestion
This sample explains how the OpenTelemetrySample project is configured with distributed tracing and emits trace information to the following telemetry exporters: Console, Application Insights, and Zipkin. The following steps can be used to configure your own app as well.
```

cgillum (approved):
This is very nicely written! Just one character fix suggestion but I'm otherwise signing off on this.

samples/DistributedTraceSample/Program.cs
davidmrdavid: Fair enough, it was non-blocking nit anyways :) .

davidmrdavid (approved):
All my feedback was addressed, so LGTM on my end.

PR #862, opened 2023-02-11
Bring entity logic into DurableTask.Core

sebastianburckhardt:
Brings the basic logic (protocols and dispatching) used for durable entities into DurableTask.Core.

- Updating to this will have to be synchronized with the corresponding changes in azure-functions-durable-extension
- If done so, this remains compatible with current applications so users can update without breaking changes
- To make this possible I was careful to make no changes to the distributed protocol and keep the serialization formats compatible
- This PR does change the way the entity middleware is invoked to simplify the logic in the extension
- there is now a separate middleware dispatch pipeline for entities with a very simple input-output semantics (cf. activity functions)
-  input is `OperationBatchRequest`, result is `OperationBatchResult`
- OperationBatchResult may be a shorter array if for some reason only part of the batch was executed (e.g. timeout, host shutdown)

Places that require further consideration are marked `\\ DRAFT`. This includes tracing. We need to consider how much tracing we want in DurableTask.Core (vs. tracing in the extension that is already in place).


src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -516,12 +513,68 @@ public void HandleSubOrchestrationInstanceFailedEvent(SubOrchestrationInstanceFa
...
+            {
+                TaskCompletionSource<string> tcs = info.Result;
+
+                // if we can deserialize the original exception we rethrow it without any wrapper   
+                bool ThrowOriginalException()
+                {
+                    try
+                    {
+                        var type = Type.GetType(failedEvent.Error, true);
+                        var deserializedException = (Exception)this.ErrorDataConverter.Deserialize(failedEvent.Result, type);
```
cgillum, line 546: We have two mechanisms for handling exceptions depending on the value of `this.ErrorPropagationMode`. What you have here works for the original mode but it turns out that in practice serializing and deserializing exceptions can be very problematic since many exception types aren't actually serializeable. For the newer out-of-proc languages (Java and .NET Isoalted) the intent is to use `ErrorPropagationMode.UseFailureDetails`, where exceptions/errors are converted into a specific runtime-neutral data contract. It would be great if entities could support that as well.

src/DurableTask.Core/History/OperationFailedEvent.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// Gets the operation result (see <see cref="OperationResult.Result"/>)
+        /// </summary>
+        [DataMember]
+        public string? Result { get; private set; }
+
+        /// <summary>
+        /// Gets the operation error (see <see cref="OperationResult.Error"/>)
+        /// </summary>
+        [DataMember]
+        public string? Error { get; private set; }
```
cgillum, line 66: Consider using `FailureDetails` for error information, which is language agnostic and doesn't rely on error-prone exception serialization. We've also added this property to `TaskFailedEvent`.

src/DurableTask.Core/Exceptions/EntityOperationFailedException.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        /// <summary>
+        /// The serialized content of the original exception
+        /// </summary>
+        public string ExceptionType { get; set; }
+
+        /// <summary>
+        /// The serialized content of the original exception
+        /// </summary>
+        public string ExceptionContent { get; set; }
```
cgillum, line 130: Consider using `FailureDetails`, which we've also added to the `TaskFailureException` exception class.

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1177,6 +1180,99 @@ public DiscardingMessage(TaskOrchestrationWorkItem workItem, TaskMessage message
...
+            public string EntityState { get; }
+
+            public override EventId EventId => new EventId(
+                EventIds.EntityBatchExecuting,
+                nameof(EventIds.EntityBatchExecuting));
+
+            public override LogLevel Level => LogLevel.Debug;
+
+            protected override string CreateLogMessage() =>
+                $"{this.InstanceId}: executing batch of {this.OperationCount} operations on state {this.EntityState}.";
```
cgillum, line 1211: Counld `EntityState` contain sensitive information, like application state? If so, we should consider not logging it. We recently had the security team reach out to us when potentially sensitive information was discovered in our Kusto logs.

cgillum (commented):
This is exciting to see! I haven't reviewed everything, but have a few initial thoughts below.

src/DurableTask.Core/TaskOrchestrationContext.cs
sebastianburckhardt: I am well aware of this issue, and there is a mechanism to handle it automatically: note that the exception is being caught (both when trying to serialize and when trying to deserialize exceptions), with a fallback that propagates the error in a runtime-neutral data contract.

Essentially, what this mechanism does is to first try `ErrorPropagationMode.SerializeExceptions`, and if that does not work, it resorts to  `ErrorPropagationMode.UserFailureDetails`. The advantage of this automatic fallback is that even if the user knows nothing about this configuration options, they still get reasonable behaviors in all situations.

I understand that it would be nicer to use exactly the same error propagation mechanism and configuration across entities, activities, and suborchestrations; but I had this automatic fallback-mechanism already in place for entities, before the `ErrorPropagationMode` feature was introduced for activities and suborchestrations, so that is why we have two different mechanisms.

Regardless of what we end up deciding, this PR is purposefully designed to make zero changes to the existing entity semantics so I think changing the errorPropagationMode behavior for entities and/or activities+suborchestrations would  need to be a separate PR.

src/DurableTask.Core/History/OperationFailedEvent.cs
sebastianburckhardt: `Error` is already a language-agnostic encoding of the error as a string. The language SDK is solely responsible for interpreting the contents. For example, the Python OOProc runtime directly reads and produces these `Error` strings (I don't even know what it puts in there).

src/DurableTask.Core/Exceptions/EntityOperationFailedException.cs
sebastianburckhardt: I think we should indeed try to unify all the failure handling options and terminology (e.g. we could add a `ErrorPropagationMode.SerializeExceptionsWithFallback` to make this mechanism available to all contexts), but it should be a separate PR because this PR is only about moving the logic, without making any behavioral changes.

src/DurableTask.Core/Logging/LogEvents.cs
sebastianburckhardt: Yes, it could contain sensitive information. In terms of logging, I would put it into the same category as inputs and outputs to activities and suborchestrations, and I will change it to use the same convention for those (which is to only print the size).

Note that already, this is printing the entity state in Debug build only. However, relying on customers to not run Debug build is probably not a good idea, so I will still change it.

src/DurableTask.Core/Exceptions/EntityOperationFailedException.cs
sebastianburckhardt: I think I don't actually need to create my own custom exception for this at all, but can just reuse the TaskFailureException.

src/DurableTask.Core/TaskOrchestrationContext.cs
sebastianburckhardt: I only just realize that this code path is never even exercised when running code via the old C# SDK, so it is not a problem at all to copy all the existing conventions.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -274,5 +275,23 @@ internal LogHelper Logger
                 return this.logHelper;
             }
         }
+
+        /// <summary>
+        /// Gets or sets the maximum number of entity operations that are processed as a single batch.
+        /// </summary>
+        /// <remarks>
+        /// Reducing this number can help to avoid timeouts on consumption plans. If set to 1, batching is disabled, and each operation
+        /// message executes and is billed as a separate function invocation.
```
davidmrdavid, line 284: nitpick: I suppose we probably shouldn't include references to the Functions service in DurableTask, right? Can we either generalize this remark, or somehow move it to the Extension layer?

src/DurableTask.Core/Entities/OperationFormat/OperationActionType.cs
```diff
@@ -0,0 +1,31 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.Core.Entities.OperationFormat
+{
+    /// <summary>
+    /// Enumeration of orchestrator actions.
```
davidmrdavid, line 17: should this be "entity" actions?

davidmrdavid (commented):
_Slowly_ making my way through this. I wasn't sure if you opened this for some early reviews, please let me know if that was not the case.

Left some tiny nitpicks. I also took a look at the sorter behavior and was a tad confused, so I will probably need to give that more detailed pass. In the meantime, any extra comments on that component would help my review. Thanks!

jviau (commented):
Some first round of comments as I am still working through the PR as it has a lot to it.

My most immediate concern is the dependency on `Newtonsoft.Json` here. This would be the first instance of relying on a specific serialization library in the `DurableTask.Core`. Are we okay with that? Can we remove that dependency? How will it behave with different `IOrchestrationService` implementations, as well as when customers change their serialization approach?

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            this.EntityName = entityName.ToLowerInvariant();
+            this.EntityKey = entityKey ?? throw new ArgumentNullException(nameof(entityKey), "Invalid entity id: entity key must not be null.");
+            this.schedulerId = GetSchedulerId(this.EntityName, this.EntityKey);
+        }
+
+        /// <summary>
+        /// The name for this class of entities.
+        /// </summary>
+        [JsonProperty(PropertyName = "name", Required = Required.Always)]
```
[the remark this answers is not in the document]
sebastianburckhardt: I think I already did that in all cases where I am defining new classes.

However, some of the classes (including this one) need to remain compatible with existing task hubs (when DF users upgrade to a newer version of DF that depends on the DT-entity implementation). This is also why the Json field names do not always match the original field name. It is designed to be compatible. I am not sure if this is possible to achieve with generic serialization.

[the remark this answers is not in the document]
cgillum: I think Jacob is asking if you can use `[DataMember(...)]` instead of `[JsonProperty(...)]`.

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            return $"@{entityName.ToLowerInvariant()}@";
+        }
+
+        /// <summary>
+        /// Returns the entity ID for a given instance ID.
+        /// </summary>
+        /// <param name="instanceId">The instance ID.</param>
+        /// <returns>the corresponding entity ID.</returns>
+        public static EntityId GetEntityIdFromInstanceId(string instanceId)
```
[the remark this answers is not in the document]
sebastianburckhardt: Makes sense. Some of the methods in this section seem superfluous and more complicated than necessary. I will use just ToString and FromString.

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -0,0 +1,125 @@
...
+//  ----------------------------------------------------------------------------------
+namespace DurableTask.Core.Entities
+{
+    using System;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// A unique identifier for an entity, consisting of entity name and entity key.
+    /// </summary>
+    public struct EntityId : IEquatable<EntityId>, IComparable
```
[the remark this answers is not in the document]
sebastianburckhardt: Sounds good. Also, I will remove schedulerId altogether. This was meant to be a caching optimization but I think the value is dubious.

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "key", Required = Required.Always)]
+        public string EntityKey { get; private set; } // do not remove set, is needed by Json Deserializer
+
+        /// <summary>
+        /// Returns the instance ID for a given entity ID.
+        /// </summary>
+        /// <param name="entityId">The entity ID.</param>
+        /// <returns>The corresponding instance ID.</returns>
+        public static string GetInstanceIdFromEntityId(EntityId entityId)
```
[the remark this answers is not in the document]
sebastianburckhardt: This is now redundant with the ToString() so I will remove it.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -624,6 +625,32 @@ public static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int task
...
+        /// <exception cref="ArgumentException"></exception>
+        internal static Guid CreateGuidFromHash(string stringToHash)
+        {
+            if (string.IsNullOrEmpty(stringToHash))
+            {
+                throw new ArgumentException("string to hash must not be null or empty", nameof(stringToHash));
+            }
+
+            byte[] hashByteArray;
+            using (HashAlgorithm hashAlgorithm = (HashAlgorithm)SHA1.Create())
```
cgillum, line 644: Is SHA1 consider FIPS-compliant? We had a problem using MD5 in the past for this exact purpose because of FIPS compliance violations: https://github.com/Azure/durabletask/issues/637.

sebastianburckhardt: Interesting. Never heard about FIPS, but I get your point. In this situation, there is no need for cryptographic quality of the hash. So yes, there should be no problem with replacing the cryptographically-insecure MD5 with an even less secure (and possibly less vetted in terms of statistic qualities) hash algorithm.  :)

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+        public IOrchestrationServiceClient ServiceClient { get; }
+
+        private void CheckEntitySupport(string name)
+        {
+            if (this.backendInformation == null)
+            {
+                throw new InvalidOperationException($"{nameof(TaskHubEntityClient)}.{name} is not supported because the chosen backend does not support entities.");
+            }
+        }
+        private void CheckQuerySupport(string name)
```
jviau, line 54: nit: add new line between functions.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+                throw new InvalidOperationException($"{nameof(TaskHubEntityClient)}.{name} is not supported because the chosen backend does not implement {nameof(IOrchestrationServicePurgeClient)}.");
+            }
+        }
+
+        /// <summary>
+        ///    Create a new TaskHubEntityClient from a given TaskHubClient.
+        /// </summary>
+        /// <param name="client">The taskhub client.</param>
+        /// <param name="stateDataConverter">The <see cref="DataConverter"/> to use for entity state deserialization, or null if the default converter should be used for that purpose.</param>
+        public TaskHubEntityClient(TaskHubClient client, DataConverter stateDataConverter = null)
```
jviau, line 74: nit: please order class:

1. Fields
2. Constructors
3. Properties
4. Methods

src/DurableTask.Core/Entities/ClientEntityContext.cs
```diff
@@ -0,0 +1,90 @@
...
+    using DurableTask.Core.Entities.OperationFormat;
+    using DurableTask.Core.Entities.StateFormat;
+    using Newtonsoft.Json.Linq;
+    using Newtonsoft.Json;
+    using System;
+
+    /// <summary>
+    /// Utility functions for clients that interact with entities, either by sending events or by accessing the entity state directly in storage
+    /// </summary>
+    public static class ClientEntityContext
```
jviau, line 25: Does this class need to be `public`? Are customers expected to call it directly?

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+        /// <param name="scheduledTimeUtc">A future time for which to schedule the start of this operation, or null if is should start as soon as possible.</param>
+        /// <returns>A task that completes when the message has been reliably enqueued.</returns>
+        public async Task SignalEntityAsync(EntityId entityId, string operationName, object operationInput = null, DateTime? scheduledTimeUtc = null)
+        {
+            this.CheckEntitySupport(nameof(SignalEntityAsync));
+
+            (DateTime original, DateTime capped)? scheduledTime = null;
+            if (scheduledTimeUtc.HasValue)
+            {
+                scheduledTime = this.backendInformation.GetCappedScheduledTime(DateTime.UtcNow, scheduledTimeUtc.Value.ToUniversalTime());
```
jviau, line 100: That fact that backends have this schedule time cap to it is really weird to me. This sounds like something the backend should be solving for you (as in AzureStorage should automatically break a future event beyond its threshold into these incremental 'forwarding' events).

I am not happy we need to expose this logic because of some backends limitation, but I also get why you need to solve this problem. If there is a way we can do without it, that would be awesome, but I understand if this is an unfortunate reality we have to live with.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+                EntityExists = false,
+                EntityState = default,
+            };
+        }
+
+        /// <summary>
+        /// The response returned by <see cref="ReadEntityStateAsync"/>.
+        /// </summary>
+        /// <typeparam name="T">The JSON-serializable type of the entity.</typeparam>
+        public struct StateResponse<T>
```
jviau, line 177: Please avoid nested types as part of public API.

Also I recommend making this type `readonly struct`

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+            this.CheckEntitySupport(nameof(ListEntitiesAsync));
+            this.CheckQuerySupport(nameof(ListEntitiesAsync));
+
+            OrchestrationQuery innerQuery = new OrchestrationQuery()
+            {
+                FetchInputsAndOutputs = query.FetchState,
+                ContinuationToken = query.ContinuationToken,
+                CreatedTimeFrom = query.LastOperationFrom,
+                CreatedTimeTo = query.LastOperationTo,
+                InstanceIdPrefix = "@",
```
jviau, line 208: What is the behavior going to be if someone inserts a non-entity orchestration with a prefix of `@` into their task hub?

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+
+            do
+            {
+                var result = await queryClient.GetOrchestrationWithQueryAsync(innerQuery, cancellationToken).ConfigureAwait(false);
+                entityResult.Entities.AddRange(result.OrchestrationState
+                    .Select(ConvertStatusResult)
+                    .Where(status => status != null));
+                entityResult.ContinuationToken = innerQuery.ContinuationToken = result.ContinuationToken;
+            }
+            while ( // run multiple queries if no records are found, but never in excess of 100ms
```
jviau, line 263: I am curious what the purpose of this is? The result is not guaranteed to be complete, so callers should ideally still check if `ContinuationToken` is not `null` and perform follow up queries anyways. If they are already doing that, what value does this add? Additionally, won't this violate the query's `pageSize` request?

I would prefer this to not try and perform successive queries and just return the first response. Let the caller determine how/when they want to perform continuations.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+                {
+                    return null;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Query condition for searching the status of entity instances.
+        /// </summary>
+        public class Query
```
jviau, line 306: Can this type extend `OrchestrationQuery`?

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+        /// Locks are considered orphaned, and are released, if the orchestration that holds them is not in state <see cref="OrchestrationStatus.Running"/>. This
+        /// should not happen under normal circumstances, but can occur if the orchestration instance holding the lock
+        /// exhibits replay nondeterminism failures, or if it is explicitly purged.</remarks>
+        /// <param name="removeEmptyEntities">Whether to remove empty entities.</param>
+        /// <param name="releaseOrphanedLocks">Whether to release orphaned locks.</param>
+        /// <param name="cancellationToken">Cancellation token that can be used to cancel the operation.</param>
+        /// <returns>A task that completes when the operation is finished.</returns>
+        public async Task<CleanEntityStorageResult> CleanEntityStorageAsync(bool removeEmptyEntities, bool releaseOrphanedLocks, CancellationToken cancellationToken)
+        {
+            this.CheckEntitySupport(nameof(CleanEntityStorageAsync));
```
jviau, line 418: Consider using https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-7.0, so the `nameof(CleanEntityStorageAsync)` will be inferred for you.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+                        if (isEmptyEntity && safeToRemoveWithoutBreakingMessageSorterLogic)
+                        {
+                            tasks.Add(DeleteIdleOrchestrationEntity(state));
+                        }
+                    }
+                }
+
+                async Task DeleteIdleOrchestrationEntity(OrchestrationState state)
+                {
+                    await this.purgeClient.PurgeInstanceStateAsync(state.OrchestrationInstance.InstanceId);
```
jviau, line 466: Between the query and purge, it is possible someone else purged this. In which case you _shouldn't_ increment `finalResult.NumberOfEmptyEntitiesRemoved`?

src/DurableTask.Core/Entities/EntitySchedulerException.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace DurableTask.Core.Entities
+{
+    using System;
+    using System.Runtime.Serialization;
+ 
+    /// <summary>
+    /// Exception used to describe various issues encountered by the entity scheduler.
+    /// </summary>
+    [Serializable]
+    public class EntitySchedulerException : Exception
```
jviau, line 23: Looking at the uses of this exception I am not sure I see the value in it. It does not expose any additional information and instead masks the real cause into `InnerException`. Unless we are going to get granular about our exception types and expose additional (and useful) information inside of them, I recommend we leave this out and just let the original exception bubble up.

src/DurableTask.Core/Entities/LocalSDK/TaskEntityContext.cs
```diff
@@ -0,0 +1,440 @@
...
+            }
+            else if (this.currentStateAccess == StateAccess.Accessed)
+            {
+                try
+                {
+                    string serializedState = this.taskEntity.StateDataConverter.Serialize(this.currentState);
+                    this.LastSerializedState = serializedState;
+                    this.currentStateAccess = StateAccess.Clean;          
+                }
+                catch (Exception serializationException)
```
jviau, line 176: add a check for non-fatal exception.

src/DurableTask.Core/Entities/LocalSDK/TaskEntityContext.cs
```diff
@@ -0,0 +1,440 @@
...
+                {
+                    // we cannot serialize the entity state - this is an application error. To help users diagnose this, 
+                    // we wrap it into a descriptive exception, and propagate this error result to any calling orchestrations.
+
+                    serializationException = new EntitySchedulerException(
```
jviau, line 181: This is the only place I see some importance to `EntityScheduleException`, primarily because this is outside of regular user control. But I still don't think it would be bad to just include the original exception and not wrap it in `EntitySchedulerException`

src/DurableTask.Core/Entities/EventToSend.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// <summary>
+    /// The data associated with sending an event to an orchestration.
+    /// </summary>
+    public readonly struct EventToSend
+    {
+        /// <summary>
+        /// The name of the event.
+        /// </summary>
+        public readonly string EventName;
```
jviau, line 30: nit: don't make fields `public`, use properties instead. This lets us make changes to the property getter in the future without it being breaking.

src/DurableTask.Core/Entities/EventToSend.cs
```diff
@@ -0,0 +1,55 @@
...
+    using DurableTask.Core.Entities.OperationFormat;
+    using DurableTask.Core.Entities.StateFormat;
+    using Newtonsoft.Json.Linq;
+    using Newtonsoft.Json;
+    using System;
+
+    /// <summary>
+    /// The data associated with sending an event to an orchestration.
+    /// </summary>
+    public readonly struct EventToSend
```
jviau, line 25: I take it this is the event sent _to_ entities? If so, can we name it like `EntityEvent` or `EntityOperationEvent`?

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -67,6 +70,11 @@ public abstract class OrchestrationContext
         /// </summary>
         internal ErrorPropagationMode ErrorPropagationMode { get; set; }
 
+        /// <summary>
+        /// Information about backend entity support, or null if the configured backend does not support entities.
+        /// </summary>
+        internal EntityBackendInformation EntityBackendInformation { get; set; }
```
jviau, line 76: Does this need to be on `OrchestrationContext`, or can it be on only `TaskOrchestrationContext`?

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -368,6 +376,70 @@ public virtual Task<T> CreateSubOrchestrationInstance<T>(Type orchestrationType,
...
+        /// </summary>
+        public abstract bool IsInsideCriticalSection { get; }
+
+        /// <summary>
+        /// Enumerates all the entities that can be called from within the current critical section. 
+        /// This set contains all the entities that were locked prior to entering the critical section,
+        /// and for which there is not currently an operation call pending.
+        /// </summary>
+        /// <returns>An enumeration of all the currently available entities.</returns>
+        public abstract IEnumerable<EntityId> GetAvailableEntities();
```
jviau, line 442: I am not sure if I am sold on the value of this. The caller knows what entities they passed to `LockEntitiesAsync`, so they theoretically have access to this already. If we must have this, can we make `LockEntitiesAsync` return a disposable struct which also contains the available entities on it?

src/DurableTask.Core/OrchestratorExecutionResult.cs
```diff
@@ -15,25 +15,27 @@ namespace DurableTask.Core
...
         /// </summary>
-        [JsonProperty("actions")]
+        [DataMember(Name = "actions")]
         public IEnumerable<OrchestratorAction> Actions { get; set; } = Array.Empty<OrchestratorAction>();
 
         /// <summary>
         /// The custom status, if any, of the orchestrator.
         /// </summary>
-        [JsonProperty("customStatus")]
+        [DataMember(Name = "customStatus")]
```
jviau, line 38: did not realize we had the Newtonsoft.Json properties here. I don't recommend changing this, who knows what it will impact.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
```diff
@@ -0,0 +1,538 @@
...
+    using DurableTask.Core.History;
+    using System.Linq;
+    using System.Diagnostics;
+    using DurableTask.Core.Query;
+    using Newtonsoft.Json;
+
+    /// <summary>
+    ///     Client used to manage and query entity instances
+    /// </summary>
+    public sealed class TaskHubEntityClient
```
jviau, line 33: I don't recommend sealing this. Infact go the opposite direction, make sure it is easily mockable for unit tests.

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -368,6 +376,70 @@ public virtual Task<T> CreateSubOrchestrationInstance<T>(Type orchestrationType,
...
+        /// <summary>
+        /// Calls an operation on an entity and returns the result asynchronously.
+        /// </summary>
+        /// <typeparam name="TResult">The result type of the operation.</typeparam>
+        /// <param name="entityId">The target entity.</param>
+        /// <param name="operationName">The name of the operation.</param>
+        /// <param name="operationInput">The input for the operation.</param>
+        /// <exception cref="EntityLockingRulesViolationException">if the orchestration is inside a critical section and the lock for this entity is not available.</exception>
+        /// <returns>A task representing the result of the operation.</returns>
+        public abstract Task<TResult> CallEntityAsync<TResult>(Entities.EntityId entityId, string operationName, object operationInput = null);
```
jviau, line 388: This is a breaking change adding `abstract` members to a class.

https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules

src/DurableTask.Core/Entities/EventFormat/RequestMessage.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Optionally, a scheduled time at which to start the operation.
+        /// </summary>
+        [DataMember(Name = "due", EmitDefaultValue = false)]
+        public DateTime? ScheduledTime { get; set; }
+
+        /// <summary>
+        /// A timestamp for this request.
+        /// Used for duplicate filtering and in-order delivery.
+        /// </summary>
+        [DataMember]
```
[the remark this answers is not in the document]
sebastianburckhardt: Changing this would break compatibility with already-serialized state in existing task hubs.

src/DurableTask.Core/Entities/EntityBackendInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    public class EntityBackendInformation
+    {
+        /// <summary>
+        /// Get the entity options specified by the orchestration service, or the default options if the service does not specify options.
+        /// </summary>
+        /// <param name="orchestrationService">The orchestration service.</param>
+        /// <param name="entityBackendInformation">The options that the provider specifies.</param>
+        /// <returns>The entity options</returns>
+        public static bool BackendSupportsEntities(IOrchestrationService orchestrationService, out EntityBackendInformation entityBackendInformation)
```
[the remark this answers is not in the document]
sebastianburckhardt: While revising the backend interface I have removed this method altogether. There is now an interface `IEntityOrchestrationService` which is implemented by the backends that support entities.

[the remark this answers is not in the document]
sebastianburckhardt: See above, method is no longer there.

src/DurableTask.Core/Entities/EntityBackendInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            if (orchestrationService is IInformationProvider optionsProvider)
+            {
+                entityBackendInformation = optionsProvider.GetEntityBackendInformation();
+                return true;
+            }
+            else
```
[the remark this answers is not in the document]
sebastianburckhardt: See above.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -274,5 +275,23 @@ internal LogHelper Logger
...
+        /// Gets or sets the maximum number of entity operations that are processed as a single batch.
+        /// </summary>
+        /// <remarks>
+        /// Reducing this number can help to avoid timeouts in execution environments that impose time limitations on work items.
+        /// If set to 1, batching is disabled, and each operation executes as a separate work item.
+        /// </remarks>
+        /// <value>
+        /// A positive integer configured by the host.
+        /// </value>
+        public int? MaxEntityOperationBatchSize { get; set; } = null;
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, good idea.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -274,5 +275,23 @@ internal LogHelper Logger
...
+
+        /// <summary>
+        /// Gets or sets the time window within which entity messages get deduplicated and reordered.
+        /// If set to zero, there is no sorting or deduplication, and all messages are just passed through.
+        /// </summary>
+        public int EntityMessageReorderWindowInMinutes { get; set; } = 30;
```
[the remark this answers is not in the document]
sebastianburckhardt: I agree that TimeSpan is better, but I think I still prefer to keep it as it is so the name stays consistent with the same setting currently being used in DF.

src/DurableTask.Core/Entities/EntityBackendInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        public TimeSpan MaximumSignalDelayTime { get; set; }
+
+        /// <summary>
+        /// Computes a cap on the scheduled time of an entity signal, based on the maximum signal delay time
+        /// </summary>
+        /// <param name="nowUtc"></param>
+        /// <param name="scheduledUtcTime"></param>
+        /// <returns></returns>
+        public (DateTime original, DateTime capped) GetCappedScheduledTime(DateTime nowUtc, DateTime scheduledUtcTime)
```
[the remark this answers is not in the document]
sebastianburckhardt: I can remove it. Yes, it is not needed. It just made the calling code look a little bit nicer.

src/DurableTask.Core/Entities/TaskEntity.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// serialization, and deserialization.</remarks>
+    public abstract class TaskEntity
+    {
+        /// <summary>
+        /// Executes the given operation batch and return the results.
+        /// </summary>
+        /// <param name="operations">The batch of operations to execute.</param>
+        /// <param name="options">Options to control entity execution.</param>
+        /// <returns></returns>
+        public abstract Task<OperationBatchResult> ExecuteOperationBatchAsync(OperationBatchRequest operations, EntityExecutionOptions options);
```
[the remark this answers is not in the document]
sebastianburckhardt: The batching is designed this way both for performance reasons and because it provides better separation between what is general and what is specific to an implementation. (The batched concept is the more general one - how to break the batch into operations is more implementation dependent and may vary between SDKs). Some aspects to consider:
- the entity state itself must be deserialized before the batch and serialized at the end of the batch, but we don't want to enforce serialization or deserialization between operations. Some implementations may do that, but it should not be enforced by the framework.
- the entity may be out of process in a different language (with its own serialization) and we don't want to go back and forth through the gRPC pipe for each individual request. Not supporting batching for out of process would kill performance.

src/DurableTask.Core/Entities/TaskEntity.cs
```diff
@@ -0,0 +1,122 @@
...
+        public virtual TState CreateInitialState(EntityContext<TState> context)
+        {
+            return default(TState);
+        }
+
+        /// <summary>
+        /// The data converter used for converting inputs and outputs for operations.
+        /// Implementations may override this setting.
+        /// </summary>
+        public virtual DataConverter MessageDataConverter => JsonDataConverter.Default;
```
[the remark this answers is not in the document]
sebastianburckhardt: I was copying what looked like the "DTFx way", since this is the DTFx SDK. And it does make a lot of sense to me to support a custom converter for the entity state. It makes it very easy to control how entity state is serialized/deserialized for a specific entity, as opposed to having to work with a single converter. It is very easy to support here and I don't see any harm.

Of course it does not mean you have to do the same for Isolated.

src/DurableTask.Core/Entities/TaskEntity.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// If true, all effects of an entity operation (all state changes and all actions) are rolled back
+        /// if the entity operation completes with an exception.
+        /// Implementations may override this setting.
+        /// </summary>
+        public virtual bool RollbackOnExceptions => true;
+
+        /// <summary>
+        /// Options for executing entities.
+        /// </summary>
+        public EntityExecutionOptions EntityExecutionOptions { get; set; }
```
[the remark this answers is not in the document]
sebastianburckhardt: I think we can remove it. I don't remember why it is there. I think at some point it was meant to provide a chance for the implementation to modify the options, but now all the modifiable options are actually part of TaskEntity.

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
```diff
@@ -0,0 +1,149 @@
...
+    /// <summary>
+    /// Context for an entity, which is available to the application code while it is executing entity operations.
+    /// </summary>
+    /// <typeparam name="TState">The JSON-serializable type of the entity state.</typeparam>
+    public abstract class EntityContext<TState>
+    {
+        /// <summary>
+        /// Gets the name of the currently executing entity.
+        /// </summary>
+        public abstract string EntityName { get; }
```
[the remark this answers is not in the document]
sebastianburckhardt: It is a bit redundant, but not harmful. Allows slightly shorter client code (`this.EntityName` instead of `this.EntityId.EntityName`).

That said, it definitely should not be abstract.

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            this.EntityName = entityName;
+            this.EntityKey = entityKey ?? throw new ArgumentNullException(nameof(entityKey), "Invalid entity id: entity key must not be null.");
+        }
+
+        /// <summary>
+        /// The name for this class of entities.
+        /// </summary>
+        [DataMember(Name = "name", IsRequired = true)]
+        public readonly string EntityName;
```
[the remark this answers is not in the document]
sebastianburckhardt: I think so. this will also tip the scale for the previous point in favor of removing the extra properties.

src/DurableTask.Core/Entities/EntityBackendInformation.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                entityBackendInformation = null;
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Interface for objects that provide entity backend information. 
+        /// </summary>
+        public interface IInformationProvider
```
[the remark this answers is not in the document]
sebastianburckhardt: Makes sense. I actually already changed this while revising the backend interface.

src/DurableTask.Core/Entities/TaskEntity.cs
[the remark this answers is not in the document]
sebastianburckhardt: I have changed the implementation so the batched execution is no longer part of the public interface. The batching loop is now part of the implementation of `EntityContext<TState>`.

src/DurableTask.Core/Entities/TaskEntity.cs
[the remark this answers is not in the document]
sebastianburckhardt: I like having separate options for converters and see no disadvantages other than cluttering the view. So I decided to hide them more. They have all been moved into `EntityContext<TState>` where they can be modified by users. This is exactly the same pattern as used by DTFx orchestrations.

src/DurableTask.Core/Entities/TaskEntity.cs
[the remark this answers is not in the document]
sebastianburckhardt: This has been moved to `EntityContext<TState>` where users can access it (and change the settings).

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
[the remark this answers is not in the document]
sebastianburckhardt: Done.

src/DurableTask.Core/Entities/EntityId.cs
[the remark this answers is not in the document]
sebastianburckhardt: Done.

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <summary>
+        /// Gets the input for this operation, as a deserialized value.
+        /// </summary>
+        /// <typeparam name="TInput">The JSON-serializable type used for the operation input.</typeparam>
+        /// <returns>The operation input, or default(<typeparamref name="TInput"/>) if none.</returns>
+        /// <remarks>
+        /// An operation invocation on an entity includes an operation name, which states what
+        /// operation to perform, and optionally an operation input.
+        /// </remarks>
+        public abstract TInput GetInput<TInput>();
```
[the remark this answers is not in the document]
sebastianburckhardt: Made it virtual.

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// An operation invocation on an entity includes an operation name, which states what
+        /// operation to perform, and optionally an operation input.
+        /// </remarks>
+        public abstract object GetInput(Type inputType);
+
+        /// <summary>
+        /// Returns the given result to the caller of this operation.
+        /// </summary>
+        /// <param name="result">the result to return.</param>
+        public abstract void Return(object result);
```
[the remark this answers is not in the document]
sebastianburckhardt: (As discussed yesterday)

I don't like the special objects because I don't want to assume a fixed relationship between how an operation affects the state (whether it creates state, deletes state, or is stateless) and what value it returns to the caller.

But I agree that using a `Return` statement is not so nice. I have therefore removed it. With the new signature of operations execution one can just return an object for the return value:

```csharp
ValueTask<object> ExecuteOperationAsync(EntityContext<TState> context);
```

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// operation to perform, and optionally an operation input.
+        /// </remarks>
+        public abstract string OperationName { get; }
+
+        /// <summary>
+        /// Whether this entity has state. 
+        /// </summary>
+        /// <remarks>The value of <see cref="HasState"/> changes as entity operations access or delete the state during their execution. It is set to true after <see cref="State"/> is accessed from within an operation, 
+        /// and it is set to false after <see cref="DeleteState"/> is called. This can happen repeatedly; neither creation nor deletion are permanent.</remarks>
+        public abstract bool HasState { get; }
```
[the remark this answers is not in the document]
sebastianburckhardt: I don't see the benefit of wrapping this into another class. Also, your suggestion seems to not be functionally equivalent (e.g. can't set the state directly), and `Clear` suggests that the state is returned to an intial state which is logically different from deletion of the entity.

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <summary>
+        /// Gets the input for this operation, as a deserialized value.
+        /// </summary>
+        /// <param name="inputType">The JSON-serializable type used for the operation input.</param>
+        /// <returns>The operation input, or default(<paramref name="inputType"/>) if none.</returns>
+        /// <remarks>
+        /// An operation invocation on an entity includes an operation name, which states what
+        /// operation to perform, and optionally an operation input.
+        /// </remarks>
+        public abstract object GetInput(Type inputType);
```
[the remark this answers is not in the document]
sebastianburckhardt: I believe GetInput is appropriate for the [function-based syntax for entities](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-entities?tabs=csharp#example-function-based-syntax---c). will become obsolete once we get around to implementing the [class-based syntax](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-entities?tabs=csharp#example-class-based-syntax---c). The latter implementation is built on top of the former.

src/DurableTask.Core/Entities/LocalSDK/EntityContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// Whether this entity has state. 
+        /// </summary>
+        /// <remarks>The value of <see cref="HasState"/> changes as entity operations access or delete the state during their execution. It is set to true after <see cref="State"/> is accessed from within an operation, 
+        /// and it is set to false after <see cref="DeleteState"/> is called. This can happen repeatedly; neither creation nor deletion are permanent.</remarks>
+        public abstract bool HasState { get; }
+
+        /// <summary>
+        /// The size of the current batch of operations.
+        /// </summary>
+        public abstract int BatchSize { get; }
```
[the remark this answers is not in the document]
sebastianburckhardt: We added this feature because we had customers explicitly ask for it.

I disagree a bit with the sentiment that exposing batching APIs is an implementation detail. It is often important for performance. APIs that don't have batching APIs often have to add them later because they realize that it is not possible to do things with the same level of efficiency when seeing only one operation at a time.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: ok.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: Why?

src/DurableTask.Core/Entities/ClientEntityContext.cs
sebastianburckhardt: It is called by the DF entity implementation, but not likely by customers directly.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: I agree that the backend should be responsible for this. But it is not doing that right now - so we have to do it here. 

Changing the backend APIs is not something I want add to this PR (it is already very big).

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: > Please avoid nested types as part of public API.

Why?

> Also I recommend making this type readonly struct

Sounds good.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: Things go wrong.

I suppose we could add some tests, but I am not sure we need that level of fool-proofness.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: The loop only repeats if there are *no* results, not just if there are *not enough* results. 

It is always correct and faster to retry in this body than to wait for the client to receive the result and issue another query.

We added this because we observed that it is very common to end up with a lot of empty pages given that we only return entities, not orchestrations. This slows down the querying a lot.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: I would prefer not to tie them in any way. Just adds extra mental overhead.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: Good catch. I will fix this by looking at the purge result.

src/DurableTask.Core/Entities/EntitySchedulerException.cs
sebastianburckhardt: > It does not expose any additional information and instead masks the real cause into InnerException.

Yes, it does. It describes precisely under what circumstances the inner exception was hit. That's its purpose.
I have found this to be quite useful when debugging my applications.
For example, it will tell you exactly where deserialization failed, not just that it failed:

```csharp
 throw new EntitySchedulerException($"Failed to serialize input for operation '{operationName}': {e.Message}", e);
```

src/DurableTask.Core/Entities/EventToSend.cs
sebastianburckhardt: This was a necessary consequence of making the struct readonly (one of your earlier suggestions).

I can undo that earlier change if you prefer. However, I don't really anticipate having to change the implementation of a property getter on a readonly struct in the future. It is readonly, after all. There is nothing hidden in there.

src/DurableTask.Core/Entities/EventToSend.cs
sebastianburckhardt: It is not actually sent. It is a tuple of information being gathered while preparing for sending the event. So, its name is purposefully different from something you would actually send or put on the wire.

src/DurableTask.Core/OrchestrationContext.cs
sebastianburckhardt: It is called on an `OrchestrationContext` in at least one place. I could add a cast, but it does not seem preferrable to me.

src/DurableTask.Core/OrchestrationContext.cs
sebastianburckhardt: The list is not just which entities you locked, but the list of entities you locked MINUS all the entities to which you currently have calls pending. This changes over time. Tracking those changes on your own is kind of complicated, so I made it possible to look at the internal list (which is very easy).

src/DurableTask.Core/OrchestratorExecutionResult.cs
sebastianburckhardt: yep, that was by accident. I got overzealous changing all the occurrences and did not realize it. Will change this back.

src/DurableTask.Core/Entities/LocalSDK/TaskEntityContext.cs
sebastianburckhardt: ok. Though I am no longer so sure that this "non-catching" of fatal exceptions really serves a useful purpose.

src/DurableTask.Core/Entities/LocalSDK/TaskEntityContext.cs
sebastianburckhardt: > primarily because this is outside of regular user control.

I don't understand this distinction.

All of the entity scheduler exceptions are situations where the thrown exception alone may not provide a lot of information about exactly where things went wrong, so I added them to provide these details. In my opinion it is much preferrable to have a text in the error somewhere that says what happened, as opposed to having to analyze stack traces.

src/DurableTask.Core/Entities/LocalSDK/TaskHubEntityClient.cs
sebastianburckhardt: This was copy-paste from `TaskHubClient`. I don't know what their reasoning is. That's why I kept it the same.

src/DurableTask.Core/OrchestrationContext.cs
sebastianburckhardt: yep, needs to have an implementation.

src/DurableTask.Core/Entities/LocalSDK/TaskEntityContext.cs
sebastianburckhardt: FYI, this screen snippets shows all the situations:

image

sebastianburckhardt:
I have pushed a commit which addresses all the feedback given so far.
Major changes include:
- the `TaskEntity<TState>.ExecuteOperationAsync ` now can return values directly  (and all tests were changed accordingly)
- TaskOrchestrationDispatcher and TaskEntityDispatcher are now cleanly separated (no more WorkItemProcessors)
- backends that support entities as a concept separate from orchestrations are recognized via `IEntityOrchestrationService` and those backends support separate LockNext...WorkItems calls for these two types of work items
- Now uses NameVersionManager in a special way for entities: name is the entity name as before, but version is the entity key. By default, uses a manager that ignores version. But it is also possible for implementations to supply a different manager that can cache TaskEntity objects - this makes it possible for applications to cache and reuse objects that represent entities.

src/DurableTask.Core/Entities/EntityBackendInformation.cs
```diff
@@ -0,0 +1,61 @@
...
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+using System;
+
+namespace DurableTask.Core.Entities
+{
+    /// <summary>
+    /// Entity processing characteristics that are controlled by the backend provider, i.e. the orchestration service.
+    /// </summary>
+    public class EntityBackendInformation
```
cgillum, line 20: For better consistency with existing .NET conventions, consider using `Options` or `Settings` rather than `Information` as the suffix.

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -0,0 +1,111 @@
...
+        public static EntityId FromString(string instanceId)
+        {
+            if (instanceId == null)
+            {
+                throw new ArgumentNullException(nameof(instanceId));
+            }
+            var pos = instanceId.IndexOf('@', 1);
+            if (pos <= 0)
+            {
+                throw new ArgumentException("instanceId is not a valid entityId", nameof(instanceId));
```
cgillum, line 78: Consider including the actual instance ID value in the exception to make it easier to debug if we ever encounter this.

```suggestion
                throw new ArgumentException($"Instance ID '{instanceId}' is not a valid entity ID.", nameof(instanceId));
```

src/DurableTask.Core/Entities/EntitySchedulerException.cs
cgillum: I think Jacob's comment might be more about whether we need a new `EntitySchedulerException` class or whether we can use an existing exception class. I don't personally have a preference either way, though I can't immediately think of an existing exception type that would be an obvious choice to use instead.

src/DurableTask.Core/Entities/EventToSend.cs
cgillum: I did a quick search on this and readonly structs are allowed to have properties. The caveat is that they must be read-only properties. More info here: https://www.educative.io/answers/what-are-readonly-structs-in-c-sharp. I would prefer we use read-only properties instead of read-only fields since that's the standard convention in C#.

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1177,6 +1180,192 @@ public DiscardingMessage(TaskOrchestrationWorkItem workItem, TaskMessage message
...
+            public string InstanceId { get; set; }
+
+            [StructuredLogField]
+            public string ExecutionId { get; set; }
+
+            [StructuredLogField]
+            public Guid CriticalSectionId { get; set; }
+
+            [StructuredLogField]
+            public EntityId[] LockSet { get; set; }
```
cgillum, line 1304: What does the output of this look like when written to logs, like Application Insights? I'm not sure what happens when we use arrays for these `[StructuredLogField]` properties.

src/DurableTask.Core/Logging/LogHelper.cs
```diff
@@ -16,8 +16,11 @@ namespace DurableTask.Core.Logging
     using System;
     using System.Collections.Generic;
     using System.Text;
+    using System.Web;
```
cgillum, line 19: Are we actually using `System.Web` somewhere in this code or was this an accident?

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -51,6 +51,17 @@ public FailureDetails(Exception e)
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FailureDetails"/> class from an exception object and
+        /// an explicitly specified inner exception.
+        /// </summary>
+        /// <param name="e">The exception used to generate the failure details.</param>
+        /// <param name="InnerException">The inner exception to use.</param>
+        public FailureDetails(Exception e, Exception InnerException)
```
cgillum, line 60: nit: constructor parameters should be lowercase.

```suggestion
        public FailureDetails(Exception e, Exception innerException)
```

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -51,6 +51,17 @@ public FailureDetails(Exception e)
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FailureDetails"/> class from an exception object and
+        /// an explicitly specified inner exception.
```
cgillum, line 56: Might be worth mentioning somewhere that any inner exception of the primary exception parameter will be ignored.

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -624,6 +625,40 @@ public static bool TryGetTaskScheduledId(HistoryEvent historyEvent, out int task
...
 
+        /// <summary>
+        /// Creates a determinstic Guid from a string using a hash function. This is a simple hash
+        /// meant to produce pseudo-random Guids, it is not meant to be cryptographically secure, 
+        /// and does not follow any formatting conventions for UUIDs (such as RFC 4122).
+        /// </summary>
+        /// <param name="stringToHash">The string to hash.</param>
+        /// <returns>A Guid constructed from the hash.</returns>
+        /// <exception cref="ArgumentException"></exception>
+        internal static Guid CreateGuidFromHash(string stringToHash)
```
cgillum, line 636: Nice - I'll probably want to borrow this at some point for some other projects. :)

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -57,6 +58,8 @@ class OrchestrationSessionManager : IDisposable
 
         internal IEnumerable<ControlQueue> Queues => this.ownedControlQueues.Values;
 
+        internal bool ProcessEntitiesSeparately { get; set; }
```
cgillum, line 61: Can you add comments explaining when this flag might be set to true and what the implications are?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -188,6 +221,92 @@ internal void ClearPendingActions()
             }
         }
 
+        public override Task<TResult> CallEntityAsync<TResult>(EntityId entityId, string operationName, object input = null)
```
cgillum, line 224: Do we need these new public APIs on `TaskOrchestrationContext` in order to support entities in .NET Isolated? As part of our public surface area discussion, I wonder if we can defer adding these, or hide them behind an interface until we feel ready to expose this functionality publicly in DTFx.

cgillum (commented):
I'm much happier with the recently refactoring to simplify the TaskOrchestrationDispatcher changes. Thank you!

I haven't done a full in-depth pass over this yet but took a look at a few areas of interest and left some comments.

We talked about whether we can avoid some of the public surface area debate by making some of the new public surface area internal, or move it into test code. Do you still think that this will be practical? I guess there are at least two categories of "public" APIs: the "TaskHub" APIs for managing orchestrations/entities and the orchestration context APIs for invoking entities from orchestrations.

src/DurableTask.Core/Entities/EntityBackendInformation.cs
sebastianburckhardt: I think the purpose of this class is a bit different from what `Options` and `Settings` are typically used for, so I would prefer to not call it that. It is used to describe properties of the backend, not to configure them. For example, whether a backend supports ordered message delivery is not really an option or a setting - it is true for Netherite and MSSQL but false for the Azure Storage provider. There are some properties that are also options, but not all of them, and they cannot be modified in this class. Maybe `EntityBackendProperties` would be a better name?

cgillum: I like Properties better. I reacted to Information partly because “information” sounds like something immutable, but this class contains mutable properties.

sebastianburckhardt: Thanks. I will make this change.

src/DurableTask.Core/Entities/EntityId.cs
sebastianburckhardt: Yes, makes sense. I will also add a check that the first character should be '@'.

src/DurableTask.Core/Logging/LogEvents.cs
sebastianburckhardt: I don't know who is serializing this data and how - seems not obvious based on the code. Probably safer to encode this as a string containing a Json array.

src/DurableTask.Core/Logging/LogHelper.cs
sebastianburckhardt: Must have been accidentally added by some automatic unwanted VS feature. Removed.

src/DurableTask.Core/FailureDetails.cs
sebastianburckhardt: yes.

src/DurableTask.Core/FailureDetails.cs
sebastianburckhardt: yes, added.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
sebastianburckhardt: Yes, done. 

Also, this reminded me that we wanted to have a `MaxConcurrentTaskEntityWorkItems` setting, analogous to the `MaxConcurrentTaskOrchestrationWorkItems`. I have added that as well.

src/DurableTask.Core/TaskOrchestrationContext.cs
sebastianburckhardt: see comment below. After addressing all the current feedback, I will create a separate PR that does not contain user-facing public SDK for entities.

src/DurableTask.Core/Entities/EventToSend.cs
sebastianburckhardt: Ok, sounds good. I did not know it was possible to have readonly properties.

src/DurableTask.Core/Entities/EntityId.cs
```diff
@@ -73,9 +73,9 @@ public static EntityId FromString(string instanceId)
                 throw new ArgumentNullException(nameof(instanceId));
             }
             var pos = instanceId.IndexOf('@', 1);
-            if (pos <= 0)
+            if ( pos <= 0 || instanceId[0] != '@')
```
cgillum, line 76: nit: we should probably update line 71 above to check for `string.IsNullOrEmpty(instanceId)` to avoid the potential of index-out-of-range exceptions.

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1284,8 +1286,17 @@ public EntityLockAcquired(string entityId, Core.Entities.EventFormat.RequestMess
                 this.InstanceId = message.ParentInstanceId;
                 this.ExecutionId = message.ParentExecutionId;
                 this.CriticalSectionId = message.Id;
-                this.LockSet = message.LockSet;
                 this.Position = message.Position;
+
+                if (message.LockSet != null)
```
cgillum, line 1291: Rather than taking a new dependency on a JSON library in this code, I would prefer we simply do the following:
```csharp
if (message.LockSet != null)
{
    this.LockSet = string.Join(",", message.LockSet.Select(id => id.ToString());
}
```
This should also be more efficient.

src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -134,8 +135,13 @@ public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory
...
             this.orchestrationService = orchestrationService ?? throw new ArgumentException("orchestrationService");
             this.logHelper = new LogHelper(loggerFactory?.CreateLogger("DurableTask.Core"));
-            this.entityOrchestrationService = orchestrationService as IEntityOrchestrationService;
-            this.entityOrchestrationService?.ProcessEntitiesSeparately(); // lets the backend know that this worker wants them separately
+
+            // if the backend supports entities, configure it to collect entity work items in a separate queue.
+            if (orchestrationService is IEntityOrchestrationService entityOrchestrationService)
+            {
+                this.entityOrchestrationService = entityOrchestrationService;
+                entityOrchestrationService.ProcessEntitiesSeparately();
```
cgillum, line 143: I need to look at this more closely, but I'm wondering why we can't simply infer the desire to process entities separately by just checking if `orchestrationService` is an `IEntityOrchestrationService` in other parts of the code. It would be great if that would lead to fewer flags to keep track of.

cgillum (commented):
A few more comments based on my reading of the most recent commit.

cgillum:
> would it help your reviewing process if I create a second PR by copying this one and removing the DTFx user-facing entity SDK and the tests altogether? So it would contain only the basic entity-mechanics.

This sounds like a great idea to me. The new public surface area changes look a bit intimidating, so separating out that part would make it easier to make progress on cementing the runtime changes.

src/DurableTask.Core/Logging/LogEvents.cs
sebastianburckhardt: I am fine with that. It may not always be parseable (e.g. if the entity ID contains commas) but that is not likely to be a major problem.

src/DurableTask.Core/TaskHubWorker.cs
sebastianburckhardt: FYI, the reason why I call `ProcessEntitiesSeparately` is because I do *not* want to process entities separately if the backend is being called by something that does not know about this difference (I did not want to break clients of IOrchestrationService that are unaware of entities and expect things to work as they did before). Maybe that is not a real concern.

sebastianburckhardt:
The reduced PR is now here: https://github.com/Azure/durabletask/pull/887

For further comments please use that PR. Also, I will address comments in this PR that have not been resolved yet there.

src/DurableTask.Core/Entities/EntityId.cs
sebastianburckhardt: Makes sense

sebastianburckhardt:
We are no longer actively pursuing this direction (adding entities to the DurableTask SDK).

PR #864, opened 2023-02-14
Update distributed tracing code based on spec

bachuv:
This PR updates the distributed tracing with OpenTelemetry code based on the spec in #847.

Scenarios from the spec that are not addressed:
- Span Status (e.g. OK, ERROR) (refer to comment in this PR for discussion)
- Fire and forget sub-orchestrations (not supporting for early preview)
- Event links (based on the conversation in the spec, it looks like we need to discuss the user experience more after the early preview)

### DIFF
--- a/src/DurableTask.Core/Tracing/Schema.cs
+++ b/src/DurableTask.Core/Tracing/Schema.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace DurableTask.Core.Tracing
+{
+    internal static class Schema
+    {
+        internal static class Task
+        {
+            internal const string Type = "durabletask.type";
+            internal const string Name = "durabletask.task.name";
+            internal const string Version = "durabletask.task.version";
+            internal const string InstanceId = "durabletask.task.instance_id";
+            internal const string ExecutionId = "durabletask.task.execution_id";
+            internal const string Status = "durabletask.task.status";
+            internal const string TaskId = "durabletask.task.task_id";
+            internal const string EventTargetInstanceId = "durabletask.event.target_instance_id";
+            internal const string FireAt = "durabletask.fire_at";
+        }
+
+        internal static class Status
+        {
+            internal const string Code = "otel.status_code";
+            internal const string Description = "otel.status_description";
+        }
+    }
+}
--- a/src/DurableTask.Core/Tracing/Schema.cs
+++ b/src/DurableTask.Core/Tracing/Schema.cs
@@ -1,6 +1,15 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
 
 namespace DurableTask.Core.Tracing
 {
--- a/src/DurableTask.Core/Tracing/TraceHelper.cs
+++ b/src/DurableTask.Core/Tracing/TraceHelper.cs
@@ -18,7 +18,6 @@ namespace DurableTask.Core.Tracing
     using System.Diagnostics;
     using System.Globalization;
     using System.Runtime.ExceptionServices;
-    using System.Reflection;
     using DurableTask.Core.Common;
     using DurableTask.Core.History;
     using DurableTask.Core.Serializing;
@@ -32,7 +31,14 @@ namespace DurableTask.Core.Tracing
 
         static readonly ActivitySource ActivityTraceSource = new ActivitySource(Source);
 
-        internal static Activity? CreateActivityForNewOrchestration(ExecutionStartedEvent startEvent)
+        /// <summary>
+        /// Starts a new trace activity for scheduling an orchestration from the client.
+        /// </summary>
+        /// <param name="startEvent">The orchestration's execution started event.</param>
+        /// <returns>
+        /// Returns a newly started <see cref="Activity"/> with orchestration-specific metadata.
+        /// </returns>
+        internal static Activity? StartActivityForNewOrchestration(ExecutionStartedEvent startEvent)
         {
             Activity? newActivity = ActivityTraceSource.StartActivity(
                 name: CreateSpanName("create_orchestration", startEvent.Name, startEvent.Version),
@@ -40,10 +46,10 @@ namespace DurableTask.Core.Tracing
 
             if (newActivity != null)
             {
-                newActivity.SetTag("durabletask.type", "orchestration");
-                newActivity.SetTag("durabletask.task.name", startEvent.Name);
-                newActivity.SetTag("durabletask.task.instance_id", startEvent.OrchestrationInstance.InstanceId);
-                newActivity.SetTag("durabletask.task.execution_id", startEvent.OrchestrationInstance.ExecutionId);
+                newActivity.SetTag(Schema.Task.Type, "orchestration");
+                newActivity.SetTag(Schema.Task.Name, startEvent.Name);
+                newActivity.SetTag(Schema.Task.InstanceId, startEvent.OrchestrationInstance.InstanceId);
+                newActivity.SetTag(Schema.Task.ExecutionId, startEvent.OrchestrationInstance.ExecutionId);
 
                 if (!string.IsNullOrEmpty(startEvent.Version))
                 {
@@ -63,7 +69,7 @@ namespace DurableTask.Core.Tracing
         /// <returns>
         /// Returns a newly started <see cref="Activity"/> with orchestration-specific metadata.
         /// </returns>
-        internal static Activity? StartTraceActivityForExecution(ExecutionStartedEvent? startEvent)
+        internal static Activity? StartTraceActivityForOrchestrationExecution(ExecutionStartedEvent? startEvent)
         {
             if (startEvent == null)
             {
@@ -90,13 +96,13 @@ namespace DurableTask.Core.Tracing
                 return null;
             }
 
-            activity.SetTag("durabletask.type", "orchestration");
-            activity.SetTag("durabletask.task.name", startEvent.Name);
-            activity.SetTag("durabletask.task.instance_id", startEvent.OrchestrationInstance.InstanceId);
+            activity.SetTag(Schema.Task.Type, "orchestration");
+            activity.SetTag(Schema.Task.Name, startEvent.Name);
+            activity.SetTag(Schema.Task.InstanceId, startEvent.OrchestrationInstance.InstanceId);
 
             if (!string.IsNullOrEmpty(startEvent.Version))
             {
-                activity.SetTag("durabletask.task.version", startEvent.Version);
+                activity.SetTag(Schema.Task.Version, startEvent.Version);
             }
 
             if (startEvent.ParentTraceContext.Id != null && startEvent.ParentTraceContext.SpanId != null)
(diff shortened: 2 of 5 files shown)


src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -35,14 +35,16 @@ public class TraceHelper
         internal static Activity? CreateActivityForNewOrchestration(ExecutionStartedEvent startEvent)
         {
             Activity? newActivity = ActivityTraceSource.StartActivity(
```
bachuv, line 37: For setting the OK or ERROR status for this span (described in [the spec here](https://github.com/Azure/durabletask/blob/vabachu/spec/docs/telemetry/traces/semantic-conventions.md?plain=1#L54)), should I add a try/catch around [TaskHubClient.cs line 614](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/TaskHubClient.cs#L614)? That would change the behavior right? I'm trying to figure out how to get that information without making any breaking changes. I wanted to get feedback for this span first before making the changes across all spans.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -208,33 +213,34 @@ public class TraceHelper
...
+                    new("durabletask.type", "event"),
+                    new("durabletask.task.name", eventRaised.Name),
+                    new("durabletask.event.target_instance_id", instance.InstanceId),
                 });
 
             return newActivity;
         }
 
-        internal static void EmitTraceActivityForTimer(OrchestrationInstance? instance, DateTime startTime, DateTime fireAt)
+        internal static void EmitTraceActivityForTimer(OrchestrationInstance? instance, DateTime startTime, TimerFiredEvent timerFiredEvent, string version)
```
cgillum, line 229: Timers don't have versions. Can we remove the `version` parameter?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -208,33 +213,34 @@ public class TraceHelper
...
             {
-                newActivity.AddTag("dtfx.type", "timer");
-                newActivity.AddTag("dtfx.instance_id", instance?.InstanceId);
-                newActivity.AddTag("dtfx.execution_id", instance?.ExecutionId);
-                newActivity.AddTag("dtfx.fire_at", fireAt.ToString("o"));
+                newActivity.AddTag("durabletask.type", "timer");
+                newActivity.AddTag("durabletask.fire_at", timerFiredEvent.FireAt.ToString("o"));
+                newActivity.AddTag("durabletask.task.version", version);
+                newActivity.AddTag("durabletask.task.instance_id", instance?.InstanceId);
+                newActivity.AddTag("durabletask.task.task_id", timerFiredEvent.EventId);
```
cgillum, line 243: I think this is supposed to be `timerFiredEvent.TimerId`. The `EventId` property will always be `-1`, which isn't useful.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -322,25 +330,41 @@ internal static void EmitTraceActivityForTimer(OrchestrationInstance? instance,
...
                 kind: ActivityKind.Producer,
                 parentContext: Activity.Current?.Context ?? default,
                 tags: new KeyValuePair<string, object?>[]
                 {
-                    new("dtfx.type", "externalevent"),
-                    new("dtfx.instance_id", instance?.InstanceId),
-                    new("dtfx.execution_id", instance?.ExecutionId)
+                    new("durabletask.type", "event"),
+                    new("durabletask.event.name", eventRaisedEvent.Name),
+                    new("durabletask.event.target_instance_id", targetInstanceId),
```
cgillum, line 350: I don't think I noticed this in the spec, but I anticipate `target_instance_id` will be confusing. Can we exclude this property if the value of `targetInstanceId` is `null` (which I expect will be true in 99% of cases)?

cgillum (commented):
Added some feedback.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: Yes, this was referring to the orchestration version. I removed it.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: I also added null checks for checking the version..

src/DurableTask.Core/Tracing/TraceHelper.cs
cgillum: A try/catch seems like the right thing to do. Something like the below pseudocode should work:
```csharp
try
{
    await this.ServiceClient.CreateTaskOrchestrationAsync(startMessage, dedupeStatuses);
}
catch (Exception e)
{
    TraceHelper.AddErrorDetailsToSpan(traceActivitySpan, e);
    throw;
}
```

What's the behavior change you were mentioning?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -35,14 +35,20 @@ public class TraceHelper
...
-                kind: ActivityKind.Internal);
+                name: CreateSpanName("create_orchestration", startEvent.Name, startEvent.Version),
+                kind: ActivityKind.Producer);
 
             if (newActivity != null)
             {
-                newActivity.SetTag("dtfx.type", "client");
-                newActivity.SetTag("dtfx.instance_id", startEvent.OrchestrationInstance.InstanceId);
-                newActivity.SetTag("dtfx.execution_id", startEvent.OrchestrationInstance.ExecutionId);
+                newActivity.SetTag("durabletask.type", "orchestration");
```
jviau, line 43: It might be helpful to collect all these tag names into a constants file. I like how OTel Geneva does this. Particularly the way of using nested classes to split up sections. This way we get something like `Schema.Task.InstanceId`.

 https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Internal/Schema.cs

The way they separate into versions is something we may consider in the future, but not needed right now.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -208,33 +239,33 @@ public class TraceHelper
...
                 startTime: startTime,
                 parentContext: Activity.Current?.Context ?? default);
 
             if (newActivity is not null)
             {
-                newActivity.AddTag("dtfx.type", "timer");
-                newActivity.AddTag("dtfx.instance_id", instance?.InstanceId);
-                newActivity.AddTag("dtfx.execution_id", instance?.ExecutionId);
-                newActivity.AddTag("dtfx.fire_at", fireAt.ToString("o"));
+                newActivity.AddTag("durabletask.type", "timer");
```
jviau, line 265: Can we include `durabletask.task.name` for the orchestration name here? May be valuable to know what orchestration this timer is for.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -255,8 +286,8 @@ internal static void EmitTraceActivityForTimer(OrchestrationInstance? instance,
             }
 
             Activity? newActivity = ActivityTraceSource.StartActivity(
-                name: taskScheduledEvent.Name ?? "",
-                kind: ActivityKind.Client,
+                name: CreateSpanName("activity", taskScheduledEvent.Name, taskScheduledEvent.Version),
+                kind: ActivityKind.Server,
```
jviau, line 290: This change caught me by surprise that we didn't have it as `Server` already. I think what may be beneficial is to do a pass on the name of these functions in this class to make sure they are descriptive of what exact span they are for. This name `CreateTraceActivityForTask` is not clear if it is for the orchestration calling an activity (Client), or for the worker processing an activity (Server). Can we do something to make it a bit more clear when reading just the function name?

Also some XML doc would be helpful for readers to know what this span is for.

src/DurableTask.Core/Tracing/TraceHelper.cs
bachuv: You can ignore the behavior change question. I misunderstood what was happening. 

For the error handling, do you have any pointers about where to set an ERROR status for the following spans?

1. [Client: Starting a sub-orchestration](https://github.com/Azure/durabletask/blob/vabachu/spec/docs/telemetry/traces/semantic-conventions.md#client-starting-a-sub-orchestration) - ERROR is set when the sub-orchestration fails to be enqueued. Right now, we create the span when the sub-orchestration completes or fails ([code link](https://github.com/Azure/durabletask/blob/cgillum/opentelemetry/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L750)). Currently, if the sub-orchestration fails to complete, we won't create a span. I thought we could add a try/catch around [this method call](https://github.com/Azure/durabletask/blob/cgillum/opentelemetry/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L565), but the ERROR message would show up as an attribute of the parent orchestration span.

2. [Worker: SendEvent](https://github.com/Azure/durabletask/blob/vabachu/spec/docs/telemetry/traces/semantic-conventions.md#sending-an-event) - ERROR is set when the event fails to be enqueued. This span is created [here](https://github.com/Azure/durabletask/blob/cgillum/opentelemetry/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L1063) and the `TaskMessage` is returned [here ](https://github.com/Azure/durabletask/blob/cgillum/opentelemetry/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L1069)(after the span is created and stopped). Similar to the sub-orchestration span above, I thought we could add a try/catch around [this method call](https://github.com/Azure/durabletask/blob/cgillum/opentelemetry/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L565), but the ERROR message would show up as an attribute of the parent orchestration span.

3. [Worker: Timer](durabletask) - ERROR is set when the timer is cancelled. I'm not sure where this would happen in `TaskOrchestrationDispatcher`. Amanda showed me that the cancellation is handled in `TaskOrchestrationContext` [here](https://github.com/Azure/durabletask/blob/cgillum/opentelemetry/src/DurableTask.Core/TaskOrchestrationContext.cs#L257), but I remember that we were avoiding adding code in `TaskOrchestrationContext` since it wouldn't be touched for apps using the dotnet-isolated model like Java?

cgillum: 1. This [this](https://github.com/Azure/durabletask/blob/5e52fde76ae8addc5f2e5b020eae54738fc6215a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L762) is where I'd expect we set the error status since this is where we get notified that the sub-orchestration failed. We wouldn't want to put a try/catch around `CompleteTaskOrchestrationWorkItemAsync` because that simply means that our attempt to save orchestration updates failed (i.e. some problem with Azure Storage) - we wouldn't throw there if an orchestration itself failed.
2. I think the try/catch would go [here, in `TaskHubClient.RaiseEventAsync()`](https://github.com/Azure/durabletask/blob/5e52fde76ae8addc5f2e5b020eae54738fc6215a/src/DurableTask.Core/TaskHubClient.cs#L707). I don't think we'd do anything in TaskOrchestrationDispatcher.cs - SendEvent in that path technically can't ever fail (at least not permanently).
3. Yeah, the way DTFx is implemented today, it's not possible for us to know if or when a timer is cancelled. We may have to defer on this for now.

bachuv: Thanks for the explanation!

1. From what I can understand, this is already handled then.
2. I added a try/catch in `TaskHubClient.RaiseEventAsync()` in my latest commit.
3. Ok, sounds good!

src/DurableTask.Core/Tracing/Schema.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
```
cgillum, line 1: We'll need a copyright header on this file.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -32,17 +31,30 @@ public class TraceHelper
...
-                newActivity.SetTag("dtfx.instance_id", startEvent.OrchestrationInstance.InstanceId);
-                newActivity.SetTag("dtfx.execution_id", startEvent.OrchestrationInstance.ExecutionId);
+                newActivity.SetTag(Schema.Task.Type, "orchestration");
+                newActivity.SetTag(Schema.Task.Name, startEvent.Name);
+                newActivity.SetTag(Schema.Task.InstanceId, startEvent.OrchestrationInstance.InstanceId);
+                newActivity.SetTag(Schema.Task.ExecutionId, startEvent.OrchestrationInstance.ExecutionId);
+
+                if (!string.IsNullOrEmpty(startEvent.Version))
+                {
+                    newActivity.SetTag("durabletask.task.version", startEvent.Version);
```
cgillum, line 56: Should we use `Schema.Task.Version` here?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -318,27 +364,117 @@ internal static void EmitTraceActivityForTimer(OrchestrationInstance? instance,
...
+            }
+
+            newActivity.AddTag(Schema.Task.Type, "event");
+            newActivity.AddTag(Schema.Task.Name, eventRaisedEvent.Name);
+            newActivity.AddTag(Schema.Task.InstanceId, instance?.InstanceId);
+            newActivity.AddTag(Schema.Task.ExecutionId, instance?.ExecutionId);
+
+            if (!string.IsNullOrEmpty(targetInstanceId))
+            {
+                newActivity.AddTag("durabletask.event.target_instance_id", targetInstanceId);
```
cgillum, line 397: Should we have a constant for this?

cgillum (approved):
I think this looks good to me! It has prompted a few other spec-related questions that I have, but I'll post those on the spec PR.

I did have one final suggestion but consider it non-blocking for merge.

PR #870, opened 2023-02-27
Update newtonsoft.json from 11.0.2 to 13.0.1

davidmrdavid:
This will remove certain static analysis warnings due to vulnerabilities. At runtime, our dependency is already being replaced with the newtonsoft.json version brought in by WebJobs, which is already `13.0.1`, so this change should be safe.

The change also required that we update the code in a few places to account for the possibility of the deserialization returning `null`. You will a few types being updated to include the `?` modifier as a result.

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -41,7 +41,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <!--Change the Storage SDK version with extreme caution, otherwise may break Durable Functions usage with Azure Storage Extension. -->
     <PackageReference Include="WindowsAzure.Storage" version="9.3.1" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -41,7 +41,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
cgillum, line 44: It's probably better to just remove this entirely since it's picked up transitively from DurableTask.Core.

cgillum (approved):
LGTM, though I wouldn't be surprised if we had some test projects that also needed to be updated.

cgillum:
> At runtime, our dependency is already being replaced with the newtonsoft.json version brought in by WebJobs, which is already 13.0.1, so this change should be safe.

Note that this comment only applies to _Durable Functions_ and not to any direct consumers of this framework.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
davidmrdavid: Ah, good idea. Will do

jviau:
Do we need to do any version rev for this?

davidmrdavid:
@jviau: Yeah, perhaps. I know for DF users this is a non-breaking change, but I struggle to determine if this is a major, minor, or patch release for DTFx users. After all, we're increasing a dependency by 2 major versions, and that seems significant. Thoughts, @jviau ?

PR #871, opened 2023-03-02
Add DurableTask.ApplicationInsights package and DurableTelemetryModule

bachuv:
This PR adds a new package `DurableTask.ApplicationInsights` which has a `DurableTelemetryModule`. This telemetry module listens to the `Activity`s that are emitted and converts them to Application Insights readable information (e.g. dependency telemetry). This is necessary since the Functions Host supports Application Insights and not OpenTelemetry yet.

Related Durable Functions PR - https://github.com/Azure/azure-functions-durable-extension/pull/2411

image


src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,104 @@
...
+                        }
+
+                        dependency.Success = activity.GetStatus() != ActivityStatusCode.Error;
+
+                        _telemetryClient.StopOperation(dependencyHolder);
+                        dependencyHolder.Dispose();
+                    }
+                },
+                Sample = (ref ActivityCreationOptions<ActivityContext> _) => ActivitySamplingResult.AllData,
+                SampleUsingParentId = (ref ActivityCreationOptions<string> _) => ActivitySamplingResult.AllData
```
davidmrdavid, line 74: would appreciate some comments here on end-effect this has on sampling :) . I'm not immediately sure what this is achieves.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    // Ignore for now; potentially log this in the future.
+                }
+                finally
+                {
+                    cts.Dispose();
+                }
+            }
+
+            GC.SuppressFinalize(this);
```
davidmrdavid, line 101: Interesting - did we find this to be necessary to avoid some perf hit or are we just being perf conscious from the start by adding this GC call?

davidmrdavid (commented):
Left some minor, non-urgent, comments from taking a quick look!

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        private static bool TryGetStatusCodeForTagValue(string statusCodeTagValue, out ActivityStatusCode statusCode)
+        {
+            ActivityStatusCode? tempStatusCode = GetStatusCodeForTagValue(statusCodeTagValue);
+
+            statusCode = tempStatusCode ?? default;
+
+            return tempStatusCode.HasValue;
+        }
+        private static ActivityStatusCode? GetStatusCodeForTagValue(string statusCodeTagValue)
```
jviau, line 82: nit: please add newline

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        static DiagnosticActivityExtensions()
+        {
+            s_getStatus = CreateGetStatus;
+        }
+
+        public static ActivityStatusCode GetStatus(this Activity activity)
+            => s_getStatus(activity);
+
+        public static bool TryGetStatus(this Activity activity, out ActivityStatusCode statusCode, out string statusDescription)
```
jviau, line 29: I think you will also need to use reflection for the System.Diagnostics.DiagnosticSources >= 6.0.0 scenario - checking if `Activity` has `.Status` and `.StatusDescription` properties on them and using those.

src/DurableTask.ApplicationInsights/DurableTask.ApplicationInsights.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
```
jviau, line 3: nit: align indentation with `<PropertyGroup>` below.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,104 @@
...
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            _telemetryClient = new TelemetryClient(configuration);
+
+            _listener = new ActivityListener
+            {
+                ShouldListenTo = source => source.Name.StartsWith("DurableTask"),
+                ActivityStarted = activity =>
+                {
+                    var dependency = _telemetryClient.StartOperation<DependencyTelemetry>(activity);
```
jviau, line 43: Our telemetry type will depend on `activity.Kind`.

Consumer or Server -> RequestTrelemetry
Client or Producer -> DependencyTelemetry. I am not sure what the telemetry type should be here. Should we make it our own string? Like "DurableTask"?
Internal -> DependencyTelemetry, "InProc" type.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,104 @@
...
+                    dependency.Telemetry.Type = DependencyTypeInProc; // Required for proper rendering in App Insights.
+                    activity.SetCustomProperty(DependencyTelemetryKey, dependency);
+                },
+                ActivityStopped = activity =>
+                {
+                    // Check for Exceptions events
+                    foreach (ActivityEvent activityEvent in activity.Events)
```
jviau, line 50: I don't think we want to keep this behavior - exceptions should be logged as `ExceptionTelemetry` independently of the span being logged.

This is because they are two different telemetry concepts: exception is an app-insights flavoring of the _log_ telemetry, and `Activity` here is a _trace_ telemetry. They have slightly different semantics around them, and we should not intertwine them like this. Particularly, traces may be sampled out - we wouldn't want exceptions logging to be ignored due to sampling.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
jviau: I am not certain the SuppressFinalize is needed? We have no finalizer.

jviau (changes requested):
@bachuv this would need to be a separate sample from `OpenTelemetrySample.csproj`, as this is using the _Application Insights_ SDK - which is different than the OTel SDK. These are independent from one another, and a customer would _typically_ have one or another and not both.

jviau:
Can we split out the fixes to our spans to a separate PR? The changes to client/server kind and parent/child relationships.

bachuv:
> Can we split out the fixes to our spans to a separate PR? The changes to client/server kind and parent/child relationships.

Yes, I moved those changes to #882

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
bachuv: Of course! I'll add a comment. These lines are just saying that the `ActivityListener` is going to collect all of the data for any `Activity` that's sent to the "DurableTask" source.

samples/DistributedTraceSample/ApplicationInsights/Processor.cs
```diff
@@ -0,0 +1,32 @@
...
+        private readonly ITelemetryProcessor _next;
+
+        public Processor(ITelemetryProcessor next)
+        {
+            _next = next;
+        }
+
+        public void Process(ITelemetry item)
+        {
+            if (item is DependencyTelemetry d && d.Target.StartsWith("127.0.0.1"))
```
cgillum, line 19: Can you add comments explaining what the code in this processor is doing? I don't immediately understand why you have these `if` statements.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        private static readonly string UnsetStatusCodeTagValue = "UNSET";
+        private static readonly string OkStatusCodeTagValue = "OK";
+        private static readonly string ErrorStatusCodeTagValue = "ERROR";
+
+        private static readonly Func<Activity, StatusTuple> s_getStatus;
+
+        static DiagnosticActivityExtensions()
+        {
+            s_getStatus = CreateGetStatus();
```
cgillum, line 26: nit: consider inlining this in line 22.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+                        {
+                            return ActivityStatusCode.Unset;
+                        }
+
+                        break;
+                    case "otel.status_description":
+                        statusDescription = tag.Value;
+                        break;
+                    default:
+                        continue;
```
cgillum, line 60: This `default: continue` is redundant and can be removed.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+            UnaryExpression convert = Expression.Convert(status, typeof(int));
+            convert = Expression.Convert(convert, typeof(ActivityStatusCode));
+            NewExpression result = Expression.New(ctor, convert, description);
+            return Expression.Lambda<Func<Activity, StatusTuple>>(result, activity).Compile();
+        }
+
+        private static bool TryGetStatusCodeForTagValue(string statusCodeTagValue, out ActivityStatusCode statusCode)
+        {
+            ActivityStatusCode? tempStatusCode = GetStatusCodeForTagValue(statusCodeTagValue);
+            statusCode = tempStatusCode ?? default;
```
cgillum, line 98: Is this `?? default` redundant?

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+namespace DurableTask.ApplicationInsights
+{
+    /// Telemetry Module to convert Activity events to App Insights API
```
cgillum, line 26: This needs to be a proper XML Doc command with `<summary>` tags.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,136 @@
...
+    public sealed class DurableTelemetryModule : ITelemetryModule, IAsyncDisposable
+    {
+        private TelemetryClient _telemetryClient;
```
cgillum, line 29: nit: this file doesn't follow the coding conventions for the rest of our durabletask code. For example, we don't prefix instance variables with `_`, we use `this.` (generally), and the `using` statements should be inside the `namespace` block. I suggest we keep this code consistent with existing style.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -248,6 +249,7 @@ public class TraceHelper
                 string statusDescription = "";
                 if (errorPropagationMode == ErrorPropagationMode.SerializeExceptions)
                 {
+                    // Null ref here if IncludeDetails is not enabled. Maybe failedEvent.Reason is good enough?
```
cgillum, line 252: I don't really understand what this comment is referring to. `IncludeDetails` is a property on `TaskOrchestrationDispatcher` that controls whether `TaskScheduledEvent`'s `Input` property gets assigned, IIRC. But the line of code below only deals with `TaskFailedEvent`, which shouldn't be impacted. Also, not sure why `FailedEvent.Reason` is being mentioned when we're not actually referencing it.

Also, we should code defensively. If we think that something can fail and result in an exception, we should handle it and do something reasonable.

samples/DistributedTraceSample/ApplicationInsights/Processor.cs
bachuv: Yes, this is more like a template to filter out/modify the telemetry. I actually see value in checking if `d.Target.Contains("core.windows.net")` to remove the storage access spans (example below). I updated the code and added a comment.

image

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+                }
+            }
+
+            return status;
+        }
+
+        private static Func<Activity, StatusTuple> CreateGetStatus()
+        {
+            PropertyInfo getStatus = typeof(Activity).GetProperty("Status");
+            if (getStatus is null)
```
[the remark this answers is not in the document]
bachuv: The default value for `Activity.Status` is `Unset` so I think we can actually remove the null check. @jviau, do you agree that it can be removed?

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
bachuv: No, I don't think so. `default` will resolve to `Unset` (first value in `ActivityStatusCode` enum).

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,136 @@
...
+        private ActivityListener _listener;
```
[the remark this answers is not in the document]
bachuv: No, they're assigned in `Initialize(TelemetryConfiguration configuration)`

samples/DistributedTraceSample/ApplicationInsights/Processor.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                return;
+            }
+
+            if (item is RequestTelemetry request && request.Name.StartsWith("orchestration"))
```
cgillum: Why do we have this `if` condition? If it's here for a reason, can we add code comments explaining why?

samples/DistributedTraceSample/ApplicationInsights/Processor.cs
```diff
@@ -0,0 +1,32 @@
+using Microsoft.ApplicationInsights;
```
cgillum, line 1: Need a copyright header on this file to be consistent with the other files in this project.

samples/DistributedTraceSample/ApplicationInsights/Processor.cs
cgillum: Oh sorry, you already did this. I was looking at the wrong commit/update.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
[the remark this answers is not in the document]
cgillum: My question was actually about when `typeof(Activity).GetProperty("Status")` would be null. Can we add a comment explaining this?

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
cgillum: I see. I misunderstood how the compiler was going to treat `default` in this particular expression.

.gitignore
```diff
@@ -13,6 +13,7 @@ obj/
 pkg/
 packages/
 TestResults/
+Debug/
```
jviau, line 16: This is fine to add, but I am curious what is outputting to this folder?

samples/DistributedTraceSample/ApplicationInsights/ApplicationInsightsSample.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <UserSecretsId>d4d9b2e3-fb2a-4de6-9747-3d6d3b639d1a</UserSecretsId>
+    <ApplicationInsightsResourceId>dummy-value</ApplicationInsightsResourceId>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
+    <PackageReference Include="Vio.DurableTask.Hosting" Version="2.2.1" />
```
jviau, line 13: I wrote this sample with my hosting package, mostly because I wanted to show using the module with DI and this was the quickest way to get DTFx with DI. 

But do we want to keep that this way? Do we want to start including samples that rely on external extension packages?

samples/DistributedTraceSample/ApplicationInsights/Processor.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public Processor(ITelemetryProcessor next)
+        {
+            _next = next;
+        }
+
+        public void Process(ITelemetry item)
+        {
+            // Filters out storage access spans so Gantt chart is easier to read
+            if (item is DependencyTelemetry d && d.Target.Contains("core.windows.net"))
```
jviau, line 32: Consider adding a filter for azurite / emulator, which is 127.0.0.1.

But also add a comment that it is not recommended to use this processor directly, and is just for demonstrative purposes. This processor will filter out **a lot** more than what the average customer would like

samples/DistributedTraceSample/ApplicationInsights/appsettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "TaskHub": {
+    "Name": "appinsightssample",
+    "ErrorPropagationMode": "UseFailureDetails",
```
jviau, line 4: note for testing / bug bash: we need to test with different error propagation modes and include details values.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,147 @@
...
+        public static ActivityStatusCode GetStatus(this Activity activity, out string statusDescription)
+        {
+            ActivityStatusCode status;
+            (status, statusDescription) = s_getStatus(activity);
+            return status;
+        }
+
+        private static ActivityStatusCode GetOtelStatus(this Activity activity, out string statusDescription)
+        {
+            Debug.Assert(activity != null, "Activity should not be null");
```
jviau, line 48: `Debug.Assert` only applies during `Debug` mode (and not in `Release`). As such, you want to still handle the release error case. There are two paths for this:

1. Change this to a null check + `ArgumentNullException` if null.
2. Return `ActivityStatusCode.Unset` and `null` statusDescription when `activity == null` (_after_ the `Debug.Assert`)
    - This way we are warned up front if something is awry in debug / tests, but we won't impact something using this in production.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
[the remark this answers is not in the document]
jviau: `Activity.Status` and `Activity.StatusDescription` was only added in `System.Diagnostics.DiagnosticSource/6.0.0`. We only target `5.0.0`. This code is designed to test if `6.0.0` (by virtue of the properties existing) was loaded and use the properties if they are available (via reflection - well, we compile a delegate which has direct property access performance levels). Otherwise, it will fallback to using tags to store these values.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        static OperationTelemetry CreateTelemetry(Activity activity)
+        {
+            OperationTelemetry telemetry;
+            ActivityStatusCode status = activity.GetStatus(out string description);
+            switch (activity.Kind)
+            {
+                case ActivityKind.Consumer or ActivityKind.Server:
+                    RequestTelemetry request = CreateTelemetryCore<RequestTelemetry>(activity);
+                    request.Success = status != ActivityStatusCode.Error;
```
jviau, line 71: For an added layer of resiliency, can you add a check for `request.StatusCode ??= (request.Success ? "success" : "fail");`. This is to guard against `durabletask.task.status` not being preset and we never set `StatusCode` (which if not set app insights silently fails to send this up).

Also instead of the raw strings `success` and `fail` - set it to the equivalent of `durabletask.task.status` in success/fail cases.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+
+            return telemetry;
+        }
+
+        async ValueTask IAsyncDisposable.DisposeAsync()
+        {
+            this.listener?.Dispose();
+            if (this.telemetryClient != null)
```
jviau, line 124: nit: this will function fine, but if we want to cross our t's and dot our i's we want to do one of the following for a proper disposal pattern:

1. Either make the `telemetryClient` and `listener` fields `readonly` and remove the two null-checks
2. Keep them non-readonly, and `null` them out during dispose, then you need to go update all other methods with a `VerifyNotDisposed` pattern to ensure they don't null-ref.
   - you also need to either perform locking or lock in the fields at the start of each method.
   - as I talk about this, it is much more complex than option 1 for not much of a benefit. You only need to do this pattern when you **must** release the references to let the GC clean them up. I'd just go with option 1.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -249,20 +249,7 @@ public class TraceHelper
...
-                else if (errorPropagationMode == ErrorPropagationMode.UseFailureDetails)
-                {
-                    FailureDetails? failureDetails = failedEvent.FailureDetails;
-                    if (failureDetails != null)
-                    {
-                        statusDescription = failureDetails.ErrorMessage;
-                    }
-                }
-
+                string statusDescription = failedEvent.Reason ?? "Unknown task failure";
```
jviau, line 252: Not sure I like `Unknown task failure` much. Worried customers will think it is the frameworks fault for not knowing. Can we have a generic message like "TaskActivity failed"?

Open to other suggestions as well.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -349,20 +337,7 @@ public class TraceHelper
...
-                else if (errorPropagationMode == ErrorPropagationMode.UseFailureDetails)
-                {
-                    FailureDetails? failureDetails = failedEvent.FailureDetails;
-                    if (failureDetails != null)
-                    {
-                        statusDescription = failureDetails.ErrorMessage;
-                    }
-                }
-
+                string statusDescription = failedEvent.Reason ?? "Unkown sub-orchestration failure";
```
jviau, line 340: Same here, can we have `Sub-orchestration failed`? Open to other suggestions.

src/DurableTask.ApplicationInsights/DurableTask.ApplicationInsights.csproj
```diff
@@ -0,0 +1,35 @@
...
+        <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevision)</FileVersion>
+        <!-- The assembly version is only the major/minor pair, making it easier to do in-place upgrades -->
+        <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
+        <!-- This version is used as the nuget package version -->
+        <Version>$(VersionPrefix)</Version>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
```
jviau, line 28: Is this package needed?

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau: This was from before the refactor to use `failedEvent.Reason`, when this code was trying to extract extra details from the exception. It was null-ref'ing in my testing, which led me to add the comment to investigate it later. The comment was more a TODO for my own reminder, doesn't need to be included in this PR - especially since I believe we are using `failedEvent.Reason` now.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            foreach (KeyValuePair<string, string> item in activity.Baggage)
+            {
+                telemetry.Properties[item.Key] = item.Value;
+            }
+
+            foreach (KeyValuePair<string, object> item in activity.TagObjects)
+            {
+                telemetry.Properties[item.Key] = item.Value.ToString();
+                if (item.Key == "durabletask.task.status" && telemetry is RequestTelemetry request)
```
jviau, line 112: Thinking about this a bit more, can you remove this if-statement entirely. See my other comment for the alternative.

samples/DistributedTraceSample/ApplicationInsights/ApplicationInsightsSample.csproj
cgillum: I don't feel strongly, but my preference would be to avoid relying on external packages unless we think they are an important part of the sample. That way users can more easily adapt samples to their own projects.

src/DurableTask.Core/Tracing/TraceHelper.cs
cgillum: How about "Unspecified sub-orchestration failure"? The main thing I'm hoping to communicate is that we don't have the details available to us.

samples/DistributedTraceSample/ApplicationInsights/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+    internal class Program
+    {
+        static async Task Main(string[] args)
+        {
+            IHost host = Host.CreateDefaultBuilder(args)
+                .ConfigureServices(services =>
+                {
+                    services.TryAddEnumerable(ServiceDescriptor.Singleton<ITelemetryModule, DurableTelemetryModule>());
+                    services.AddSingleton(GetOrchestrationService());
+                    AddApplicationInsights(services);
```
cgillum, line 38: I feel like there's a lot of setup code that we're asking developers to add and I would feel better if we could abstract it all from them. This might be related to Jacob's point about whether we need to use his library, because it seems like this sample is taking a dependency on it to do this setup. Are there alternate ways to configure this that we could consider?

cgillum (commented):
A few more comments from me (I also replied to some of the previous comment threads).

samples/DistributedTraceSample/ApplicationInsights/Program.cs
jviau: We aren't the ones asking developers to add this - it is how app insights SDK works. The only one we are specifically asking is the line:

``` CSharp
services.TryAddEnumerable(ServiceDescriptor.Singleton<ITelemetryModule, DurableTelemetryModule>());
```

Everything else is setup code someone using app insights would have already done. If they are using DI, they would use app insights DI extensions like we do here. If they are not using DI they would be doing whatever other way there is to construct and use app insights.

But yeah, if we do want to remove the DI library here we can - it won't make it any _less_ code (probably more), but just different setup code.

.gitignore
bachuv: There were some binaries and other files that were outputted. [This commit](https://github.com/Azure/durabletask/pull/871/commits/74f3048a1d75b8aab072fb2eea26d71a63147ca0) shows the files that were deleted.

src/DurableTask.ApplicationInsights/DiagnosticActivityExtensions.cs
bachuv: Thanks! I implemented the second option.

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
bachuv: I ended up adding the following check. I replaced `??=` with a null check because `request.ResponseCode` was never `null` and evaluated to an empty string - `""`. The `bool` cast is there for `request.Success` because it's nullable.

```
if (string.IsNullOrEmpty(request.ResponseCode))
{
    request.ResponseCode = (bool)request.Success ? "200" : "500";
}
```

src/DurableTask.ApplicationInsights/DurableTask.ApplicationInsights.csproj
bachuv: Yes, it's needed for `IAsyncDisposable`.

samples/DistributedTraceSample/ApplicationInsights/Program.cs
bachuv: > But yeah, if we do want to remove the DI library here we can - it won't make it any less code (probably more), but just different setup code.

@jviau, do you have any pointers you can share about this approach?

src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs
jviau: Ah right, forgot about that. Option 1 is out :).

This is fine for now, we can worry about improving dispose later.

samples/DistributedTraceSample/ApplicationInsights/Program.cs
jviau: This page has an example of adding AppInsights via a manual service collection; https://learn.microsoft.com/en-us/azure/azure-monitor/app/worker-service#net-corenet-framework-console-application

bachuv: Thanks! I'll take a look.

cgillum (approved):
Just a couple small things.

PR #876, opened 2023-03-23
Wait for AppLease lease acquisition to complete during shutdown

davidmrdavid:
**Context:**  we have recently observed that, during shutdown, the AppLease acquisition task is not guaranteed to return. This happens due to a race condition in the AppLease Manager's lease acquisition [task](https://github.com/Azure/durabletask/blob/3dd624ea067b6a3b2da9c1ae5210e5348b2da22b/src/DurableTask.AzureStorage/Partitioning/AppLeaseManager.cs#L129-L157).

The relevant bit is this:

```C#
while (!cancellationToken.IsCancellationRequested)
{
  try
  {
      while (!await this.TryAquireAppLeaseAsync())
      {
          await Task.Delay(this.settings.AppLeaseOptions.AcquireInterval, cancellationToken);
       }
       await this.StartAppLeaseAsync();
       await this.shutdownCompletedEvent.WaitAsync(Timeout.InfiniteTimeSpan, cancellationToken);

   // omitted

```

The line `await this.shutdownCompletedEvent.WaitAsync(Timeout.InfiniteTimeSpan, cancellationToken);` unblocks when either the `shutdownCompletedEvent` is `Set` or if the cancellation token is set.

Currently, during shutdown, we `Set` the `shutdownCompletedEvent` before the cancellationToken is Cancel'ed. As a result, we create a window of time where we unblock from WaitAsync and cancellationToken is not set. This allows the while loop to iterate one more time, pass it's conditional check (that cancellationToken is not set) and acquire the AppLease. In return, the AppLeaseManager is started anew, and so are the Partition Managers.

**This PR:**

This PR removes this race condition in two ways:
(1) During shutdown, it cancels the cancellation token before `Set`'ing the `shutdownCompletedEvent`. This already removes the race condition.
(2) The shutdown task now also explicitly awaits for the lease-acquisition task to complete. This is to properly model that the shutdown method cannot return unless we guarantee that the applease manager won't restart.

Finally, this PR also removes a typo where "Acquisition" was written as "Aquistion" :)

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/MessageManagerTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/MessageManagerTests.cs
@@ -1,0 +1,137 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+namespace DurableTask.AzureStorage.Tests
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Reflection;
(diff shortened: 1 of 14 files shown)


amdeel:
We don't actually want the AppLease component to stop the starter task, we want the outer layer to control the starter task, so this cancellation should stay in the public Stop method. StopAppLease may be called because we lost the AppLease, not only on shutdown events. In these situations we do not want to cancel the starter task.

I think it should be sufficient to simply reorder the Stop method to cancel the starter task token before stopping the AppLease.

amdeel (approved):
Looks good! Make sure to test this against your repro and share this package with the customers affected.

davidmrdavid:
Thanks @amdeel: I'll keep this open until I get confirmation from the customer that this worked for them.

davidmrdavid:
merging as original customer became inactive

Issue #877, opened 2023-03-24
DT.AzureStorage: netstandard2.0 message format is not backwards compatible with net462


cgillum:
### Description
Upgrading an existing app from .NET Framework to .NET Core or higher may cause existing orchestrations to get stuck due to message serialization compatibility issues.

Related: https://stackoverflow.com/questions/50190568/net-standard-4-7-1-could-not-load-system-private-corelib-during-serialization

New extensibility may be required to help mitigate this problem for users that need to migrate from .NET Framework to modern .NET versions.

### Expected behavior
Switching from .NET Framework to a newer version, like .NET 6, ideally should be safe and not require any special interventions.

### Actual behavior
Changes in the assembly name for primitive types like `Int32` and `String` from .NET Framework to .NET Core unfortunately result in deserialization problems because of the JSON settings used by DT.AzureStorage.

Example:
![image](https://user-images.githubusercontent.com/2704139/227653406-8a456537-3302-4ac0-9cec-43b3c48a8d7c.png)

### Known workarounds
None currently

PR #878, opened 2023-03-25
Allow customizing type binding for dequeued Azure Storage messages

madevoge:
Allow custom type binding for system types, Fixing #877

We are solving the deserialization problem by allowing configuration based bindings.
We cannot get rid of the Type handling in general due to the inheritance based logic used after deserialization.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -55,9 +55,9 @@ namespace DurableTask.AzureStorage
             {
                 TypeNameHandling = TypeNameHandling.Objects,
 #if NETSTANDARD2_0
-                SerializationBinder = new TypeNameSerializationBinder(),
+                SerializationBinder = new TypeNameSerializationBinder(settings.CustomMessageTypeBinder),
 #else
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -341,13 +353,20 @@ static class TypeNameSerializationHelper
         static readonly Assembly DurableTaskCore = typeof(DurableTask.Core.TaskMessage).Assembly;
         static readonly Assembly DurableTaskAzureStorage = typeof(AzureStorageOrchestrationService).Assembly;
 
-        public static void BindToName(Type serializedType, out string assemblyName, out string typeName)
+        public static void BindToName(ICustomTypeBinder customBinder, Type serializedType, out string assemblyName, out string typeName)
         {
-            assemblyName = null;
-            typeName = serializedType.FullName;
+            if (customBinder == null)
```
cgillum, line 358: Looks like this condition is flipped.
```suggestion
            if (customBinder != null)
```

cgillum (changes requested):

cgillum (approved):

PR #880, opened 2023-03-28
Patch tests post Newtonsoft.Json upgrade to 13.0.1

davidmrdavid:
Follow up to: https://github.com/Azure/durabletask/pull/870

It appears that CI did not run on the PR above, leading to broken tests post-merge. This PR fixes that by updating our Newtonsoft references across the test projects

### DIFF
--- a/Test/DurableTask.Core.Tests/ExceptionHandlingIntegrationTests.cs
+++ b/Test/DurableTask.Core.Tests/ExceptionHandlingIntegrationTests.cs
@@ -61,26 +61,26 @@ namespace DurableTask.Core.Tests
             if (mode == ErrorPropagationMode.SerializeExceptions)
             {
                 // The exception should be deserializable
-                InvalidOperationException e = JsonConvert.DeserializeObject<InvalidOperationException>(state.Output);
+                InvalidOperationException? e = JsonConvert.DeserializeObject<InvalidOperationException>(state.Output);
                 Assert.IsNotNull(e);
-                Assert.AreEqual("This is a test exception", e.Message);
+                Assert.AreEqual("This is a test exception", e?.Message);
             }
             else if (mode == ErrorPropagationMode.UseFailureDetails)
             {
                 // The failure details should contain the relevant exception metadata
-                FailureDetails details = JsonConvert.DeserializeObject<FailureDetails>(state.Output);
+                FailureDetails? details = JsonConvert.DeserializeObject<FailureDetails>(state.Output);
                 Assert.IsNotNull(details);
-                Assert.AreEqual(typeof(InvalidOperationException).FullName, details.ErrorType);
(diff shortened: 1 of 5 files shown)


Test/DurableTask.Core.Tests/ExceptionHandlingIntegrationTests.cs
```diff
@@ -61,26 +61,26 @@ await this.worker
             if (mode == ErrorPropagationMode.SerializeExceptions)
             {
                 // The exception should be deserializable
-                InvalidOperationException e = JsonConvert.DeserializeObject<InvalidOperationException>(state.Output);
+                InvalidOperationException? e = JsonConvert.DeserializeObject<InvalidOperationException>(state.Output);
```
cgillum, line 64: I'm assuming this is a change in the version of Newtonsoft we're using?

Test/DurableTask.Core.Tests/ExceptionHandlingIntegrationTests.cs
```diff
@@ -61,26 +61,26 @@ await this.worker
...
                 // The failure details should contain the relevant exception metadata
-                FailureDetails details = JsonConvert.DeserializeObject<FailureDetails>(state.Output);
+                FailureDetails? details = JsonConvert.DeserializeObject<FailureDetails>(state.Output);
                 Assert.IsNotNull(details);
-                Assert.AreEqual(typeof(InvalidOperationException).FullName, details.ErrorType);
-                Assert.IsTrue(details.IsCausedBy<InvalidOperationException>());
-                Assert.IsTrue(details.IsCausedBy<Exception>()); // check that base types work too
-                Assert.AreEqual("This is a test exception", details.ErrorMessage);
-                Assert.IsNotNull(details.StackTrace);
+                Assert.AreEqual(typeof(InvalidOperationException).FullName, details?.ErrorType);
```
cgillum, line 73: It's technically not correct for us to add the `?` here (and below) because we already did the null check above in `Assert.IsNotNull(details)`. If the compiler is still making you check for null, it might be because the version of `Assert.IsNotNull` we're calling doesn't have the right attributes on it. In that case, we should use `!` instead of `?` to suppress subsequent warnings.

cgillum (changes requested):

Test/DurableTask.Core.Tests/ExceptionHandlingIntegrationTests.cs
davidmrdavid: Correct. `DeserializeObject` may now return `null`

Test/DurableTask.Core.Tests/ExceptionHandlingIntegrationTests.cs
davidmrdavid: Fancy. Will do

cgillum (approved):
LGTM!

PR #882, opened 2023-03-29
Update task activity ActivityKind values and parents

bachuv:
This PR updates the `ActivityKind` values for task activity spans and also makes the task activity client span the parent of the task activity server span.

image

### DIFF
--- a/src/DurableTask.Core/History/TaskScheduledEvent.cs
+++ b/src/DurableTask.Core/History/TaskScheduledEvent.cs
@@ -84,11 +84,5 @@ namespace DurableTask.Core.History
         /// </summary>
         [DataMember]
         public DistributedTraceContext? ParentTraceContext { get; set; }
-
-        /// <summary>
-        /// The W3C span id associated with this task's client span.
-        /// </summary>
-        [DataMember]
-        public string? ClientSpanId { get; set; }
     }
 }
--- a/src/DurableTask.Core/History/TaskScheduledEvent.cs
+++ b/src/DurableTask.Core/History/TaskScheduledEvent.cs
@@ -84,5 +84,11 @@ namespace DurableTask.Core.History
         /// </summary>
         [DataMember]
         public DistributedTraceContext? ParentTraceContext { get; set; }
+
+        /// <summary>
+        /// The W3C span id associated with this task's client span.
+        /// </summary>
+        [DataMember]
+        public string? ClientSpanId { get; set; }
     }
 }
--- a/src/DurableTask.Core/ISupportsDurableTraceContext.cs
+++ b/src/DurableTask.Core/ISupportsDurableTraceContext.cs
@@ -61,8 +61,8 @@ namespace DurableTask.Core
             if (activityContext != null)
             {
                 wrapper.ParentTraceContext = new DistributedTraceContext(
-                    "00"+"-"+activityContext.TraceId.ToString()+"-"+activityContext.SpanId.ToString()+"-01",
-                    activityContext.TraceState?.ToString());
+                    "00-" + activityContext.TraceId.ToString() + "-" + activityContext.SpanId.ToString() + "-0" + ((int)activityContext.TraceFlags),
+                    activityContext.TraceState);
             }
         }
     }
--- a/src/DurableTask.Core/ISupportsDurableTraceContext.cs
+++ b/src/DurableTask.Core/ISupportsDurableTraceContext.cs
@@ -60,8 +60,9 @@ namespace DurableTask.Core
         {
             if (activityContext != null)
             {
+                // TODO: update trace flags casting to handle 2 digits
                 wrapper.ParentTraceContext = new DistributedTraceContext(
-                    "00-" + activityContext.TraceId.ToString() + "-" + activityContext.SpanId.ToString() + "-0" + ((int)activityContext.TraceFlags),
+                    $"00-{activityContext.TraceId}-{activityContext.SpanId}-0{activityContext.TraceFlags:d}",
                     activityContext.TraceState);
             }
         }
--- a/src/DurableTask.Core/ISupportsDurableTraceContext.cs
+++ b/src/DurableTask.Core/ISupportsDurableTraceContext.cs
@@ -55,5 +55,15 @@ namespace DurableTask.Core
                     activity.TraceStateString);
             }
         }
+
+        internal static void SetParentTraceContext(this ISupportsDurableTraceContext wrapper, ActivityContext activityContext)
+        {
+            if (activityContext != null)
+            {
+                wrapper.ParentTraceContext = new DistributedTraceContext(
+                    "00"+"-"+activityContext.TraceId.ToString()+"-"+activityContext.SpanId.ToString()+"-01",
+                    activityContext.TraceState?.ToString());
+            }
+        }
     }
 }
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -933,14 +933,15 @@ namespace DurableTask.Core
                 version: scheduleTaskOrchestratorAction.Version,
                 input: scheduleTaskOrchestratorAction.Input);
 
-            ActivitySpanId clientSpanId;
+            ActivitySpanId clientSpanId = ActivitySpanId.CreateRandom();
+
             if (parentTraceActivity != null)
             {
-                clientSpanId = ActivitySpanId.CreateRandom();
-                scheduledEvent.ClientSpanId = clientSpanId.ToString();
+                ActivityContext activityContext = new(parentTraceActivity.TraceId, clientSpanId, parentTraceActivity.ActivityTraceFlags, parentTraceActivity.TraceStateString);
+                scheduledEvent.SetParentTraceContext(activityContext);
             }
 
-            scheduledEvent.SetParentTraceContext(parentTraceActivity);
+
 
             taskMessage.Event = scheduledEvent;
             taskMessage.OrchestrationInstance = runtimeState.OrchestrationInstance;
@@ -955,10 +956,9 @@ namespace DurableTask.Core
 
                 if (parentTraceActivity != null)
                 {
-                    scheduledEvent.ClientSpanId = clientSpanId.ToString();
+                    ActivityContext activityContext = new(parentTraceActivity.TraceId, clientSpanId, parentTraceActivity.ActivityTraceFlags, parentTraceActivity.TraceStateString);
+                    scheduledEvent.SetParentTraceContext(activityContext);
                 }
-
-                scheduledEvent.SetParentTraceContext(parentTraceActivity);
             }
 
             this.logHelper.SchedulingActivity(
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -1018,6 +1018,9 @@ namespace DurableTask.Core
                 historyEvent.Input = createSubOrchestrationAction.Input;
             }
 
+            ActivitySpanId clientSpanId = ActivitySpanId.CreateRandom();
+            historyEvent.ClientSpanId = clientSpanId.ToString();
+
             runtimeState.AddEvent(historyEvent);
 
             var taskMessage = new TaskMessage();
@@ -1041,7 +1044,11 @@ namespace DurableTask.Core
                 Version = createSubOrchestrationAction.Version
             };
 
-            startedEvent.SetParentTraceContext(parentTraceActivity);
+            if (parentTraceActivity != null)
+            {
+                ActivityContext activityContext = new ActivityContext(parentTraceActivity.TraceId, clientSpanId, parentTraceActivity.ActivityTraceFlags, parentTraceActivity.TraceStateString);
+                startedEvent.SetParentTraceContext(activityContext);
+            }
 
             this.logHelper.SchedulingOrchestration(startedEvent);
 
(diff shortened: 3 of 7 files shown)


src/DurableTask.Core/History/TaskScheduledEvent.cs
```diff
@@ -84,5 +84,11 @@ public TaskScheduledEvent(int eventId)
         /// </summary>
         [DataMember]
         public DistributedTraceContext? ParentTraceContext { get; set; }
+
+        /// <summary>
+        /// The W3C span id associated with this task's client span.
+        /// </summary>
+        [DataMember]
+        public string? ClientSpanId { get; set; }
```
jviau, line 92: I don't think we need a separate spanId here. I was thinking instead `ParentTraceContext` would be set to contain the client span ID and _not_ the calling orchestrations server span ID.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -140,10 +139,13 @@ public class TraceHelper
                 return null;
             }
 
+            ActivitySpanId clientSpanId = ActivitySpanId.CreateFromString(scheduledEvent.ClientSpanId?.ToCharArray());
```
jviau, line 142: If we update `scheduledEvent.ParentTraceContext` to be the client trace context itself, then we won't need these extra steps.

src/DurableTask.Core/History/TaskScheduledEvent.cs
bachuv: `TaskScheduledEvent.ParentTraceContext` is used by both the client span and server span to get the parent. If we update the `ParentTraceContext` to store the client span trace context, then the client would be the parent of itself. I added some code links below. Let me know if you mean something else though!

Just a note that these links are from the target branch and the `ActivityKind` values are swapped in that branch.

Client span: 
https://github.com/Azure/durabletask/blob/0f953e6b0f1457bd72b8c1e73a87ab41c4602a45/src/DurableTask.Core/Tracing/TraceHelper.cs#L183
Server span:
https://github.com/Azure/durabletask/blob/0f953e6b0f1457bd72b8c1e73a87ab41c4602a45/src/DurableTask.Core/Tracing/TraceHelper.cs#L138

jviau: Ah I thought they were two separate events. Thank you for explaining.

jviau, line 92: Different suggestion now lol:

We are going to have to make a similar change for sub-orchestrations, correct? If so, maybe we wrap this into `DistributedTraceContext` itself. We can then make that structure smart enough to have `.RestoreServerActivity()` and `.RestoreClientActivity()` or whatever naming you like.

jviau (commented):
Do we also need to make a change for sub-orchestrations? Will that be this PR or a different one?

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau, line 142:
```suggestion
            ActivitySpanId clientSpanId = ActivitySpanId.CreateFromString(scheduledEvent.ClientSpanId?.AsSpan());
```

jviau: To clarify:

1. use `AsSpan()` for this overload.
2. put the null check outside of the `CreateFromString` because if it is null, it would throw. I think there is actually more work if it is null - like don't supply a parent at all.

src/DurableTask.Core/History/TaskScheduledEvent.cs
jviau: Looking at this again, `DistributedTraceContext` does have what we need - `ClientSpanId` should be part of `DistributedTraceContext.TraceParent`, right now it is the calling orchestrations `Server` ID, which we do not need here.

bachuv: >   ClientSpanId should be part of DistributedTraceContext.TraceParent

I'm not sure I understand this part. Can you explain how the `ClientSpanId` is a part of the `DistributedTraceContext.TraceParent` if it's the orchestrations `Server` ID?

bachuv: Yeah, I think we should increase the scope of this change to sub-orchestrations too. I'll work on making that change in this PR just to keep it all together.

> We can then make that structure smart enough to have .RestoreServerActivity() and .RestoreClientActivity() or whatever naming you like.

That's a good idea! It will make the code a lot cleaner.

jviau: For a given sub orchestration or task activity, we want the following span hierarchy:

`[Parent orchestration server span] -> [Client span for calling the task activity] -> [Task activity server span]`

If we look at `[Task activity server span]`, what should the `TraceParent` be? It is the client span for calling the task activity, **not** the parent orchestration server span. So, what I am saying is that right now you are setting `DistributedTraceContext.TraceParent` as the parent orchestration server span, which is not needed. It should be the client span only - which is why we do not need the extra `ClientSpanId` property.

The task activity server span is already stored in `DistributedTraceContext.SpanId` - so it will not be affected.

jviau: You currently have 3 distinct spanIds being carried with a TaskActivity execution. We only need 2 - the parent of the execution (which is the client span) and the execution span ID itself (which is the server span). Remember, `TraceParent` is of the form `00-0af7651916cd43dd8448eb211c80319c-b9c7c989f97918e1-01`, where that breaks down to `{version}-{trace id}-{span id}-{flags}`, and that span ID within` TraceParent` is where I am saying the client span should be stored, as it _is the parent_.

bachuv: Leaving a note here to summarize the current approach (as discussed offline):

1. The `task activity client span` span id is created when `TaskScheduledEvent` is created. The `task activity client span` span id is stored as part of `TaskScheduledEvent.ParentTraceContext`. We create the client span id here since the `task activity server span` is created first and needs to know its parent.
2. The `task activity server span`'s parent is `TaskScheduledEvent.ParentTraceContext`. This is the task activity client span.
3. The `task activity client span`'s parent is `Activity.Current`. This will be the `orchestration execution server span`.
4. We manually set the `task activity client span id` as the span id in `TaskScheduledEvent.ParentTraceContext`.

bachuv: I've been trying to make the same change for sub-orchestrations and would like some feedback/suggestions.

There are two parts to sub-orchestration spans:
1. `SubOrchestrationInstanceCreatedEvent` - used to create the `sub-orchestration client span`
2. `ExecutionStartedEvent` - used to create the `sub-orchestration server span`

I originally thought that the client span id could be stored in `SubOrchestrationInstanceCreatedEvent.ParentTraceContext` (we would add a new field) but realized that we wouldn't have access to that while creating the `sub-orchestration server span`.

I also thought we could add another field to `ExecutionStartedEvent` to store the client span information, but decided against in since it would be specific to the sub-orchestration scenario and not related to `ExecutionStartedEvent`.

@jviau, do you have any thoughts/ideas about this?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -142,14 +141,14 @@ public class TraceHelper
...
-                kind: ActivityKind.Client,
+                kind: ActivityKind.Server,
                 parentContext: activityContext);
 
             if (newActivity == null)
             {
                 return null;
             }
-
+
```
jviau, line 151: nit: whitespace

src/DurableTask.Core/History/TaskScheduledEvent.cs
jviau: I think you will need to store the same client spanId on both events:

1. `SubOrchestrationInstanceCreatedEvent` - add a `ClientSpanId` property
1. `ExecutionStartedEvent` - do the same as with task activity changes in this PR, that is to say the client span ID should be part of `TraceParent`

You should be able to do all of this within `ProcessCreateSubOrchestrationInstanceDecision`

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -946,7 +954,11 @@ private void AddFailureDetailstoActivity(OrchestrationRuntimeState runtimeState,
                     name: scheduleTaskOrchestratorAction.Name,
                     version: scheduleTaskOrchestratorAction.Version);
 
-                scheduledEvent.SetParentTraceContext(parentTraceActivity);
+                if (parentTraceActivity != null)
```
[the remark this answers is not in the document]
bachuv: `scheduledEvent` is passed as an event twice in this method so I set it twice - before line 946 and then again for the case where parameters are not included.

1. https://github.com/Azure/durabletask/blob/f11c51fee09054e3c6072e739fa60ebef8e9733b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L946
2. https://github.com/Azure/durabletask/blob/f11c51fee09054e3c6072e739fa60ebef8e9733b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L968

src/DurableTask.Core/ISupportsDurableTraceContext.cs
```diff
@@ -55,5 +55,15 @@ internal static void SetParentTraceContext(this ISupportsDurableTraceContext wra
                     activity.TraceStateString);
             }
         }
+
+        internal static void SetParentTraceContext(this ISupportsDurableTraceContext wrapper, ActivityContext activityContext)
+        {
+            if (activityContext != null)
+            {
+                wrapper.ParentTraceContext = new DistributedTraceContext(
+                    "00"+"-"+activityContext.TraceId.ToString()+"-"+activityContext.SpanId.ToString()+"-01",
```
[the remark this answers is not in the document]
bachuv: I ended up casting activityContext.TraceFlags to an int and it seems to work. 

`"00-" + activityContext.TraceId.ToString() + "-" + activityContext.SpanId.ToString() + "-0" + **((int)activityContext.TraceFlags)**`

src/DurableTask.Core/ISupportsDurableTraceContext.cs
```diff
@@ -55,5 +55,15 @@ internal static void SetParentTraceContext(this ISupportsDurableTraceContext wra
                     activity.TraceStateString);
             }
         }
+
+        internal static void SetParentTraceContext(this ISupportsDurableTraceContext wrapper, ActivityContext activityContext)
+        {
+            if (activityContext != null)
+            {
+                wrapper.ParentTraceContext = new DistributedTraceContext(
+                    "00-" + activityContext.TraceId.ToString() + "-" + activityContext.SpanId.ToString() + "-0" + ((int)activityContext.TraceFlags),
```
jviau, line 64: `"-0" + ((int)activityContext.TraceFlags)`

this will work for now, but it will break if `TraceFlags` ever has 2 digits to it. Can you add a TODO to fix that?

jviau (approved):

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1006,6 +1016,9 @@ private void AddFailureDetailstoActivity(OrchestrationRuntimeState runtimeState,
                 historyEvent.Input = createSubOrchestrationAction.Input;
             }
 
+            ActivitySpanId clientSpanId = ActivitySpanId.CreateRandom();
+            historyEvent.ClientSpanId = clientSpanId.ToString();
```
cgillum, line 1020: I haven't looked closely enough to understand why we have a `ClientSpanId` property for `SubOrchestrationInstanceCreatedEvent`, but is there a reason we don't set a value like this for `TaskScheduledEvent`? Or do we and I just missed it? I was expecting that we'd treat these two event types basically the same.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -180,22 +180,23 @@ public class TraceHelper
...
+                parentContext: Activity.Current?.Context ?? default);
 
             if (newActivity == null)
             {
                 return null;
             }
 
+            if (taskScheduledEvent.ParentTraceContext != null)
+            {
+                ActivityContext.TryParse(taskScheduledEvent.ParentTraceContext.TraceParent, taskScheduledEvent.ParentTraceContext?.TraceState, out ActivityContext parentContext);
```
cgillum, line 196: Do we need to check if `TryParse` returns `true` or `false`? I'm worried about whether the next line of code could throw a null-ref exception.

cgillum (commented):
Adding a few comments of my own.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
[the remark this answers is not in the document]
bachuv: We need it to exist in both places with the current implementation. Here is an example of using the `TaskScheduledEvent` stored in the message and the history.

### Message
https://github.com/Azure/durabletask/blob/f11c51fee09054e3c6072e739fa60ebef8e9733b/src/DurableTask.Core/TaskActivityDispatcher.cs#L128
https://github.com/Azure/durabletask/blob/f11c51fee09054e3c6072e739fa60ebef8e9733b/src/DurableTask.Core/TaskActivityDispatcher.cs#L131

### History
https://github.com/Azure/durabletask/blob/f11c51fee09054e3c6072e739fa60ebef8e9733b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L768

cgillum (approved):
LGTM! Thanks for the explanations!

PR #883, opened 2023-03-31
DT.AzureStorage: fix blob account for TrackingStore

### LINKED ISSUE #879: DT.AzureStorage: TrackingStore is ignored for large fields that spill to blobs
### Description

The TrackingStore setting appears to be ignored when storing blobs for large fields.

This means that the blobs used to represent large fields in the history or instance tables ends up in the default storage account, not the storage account specified for the TrackingStore, and are thus "separated" from the table they belong to.

### Known impact

In Azure/azure-functions-durable-extension#2420, a user is reporting this issue. They are constructing an external client that points to the tracking store and want to use this client to download the entity state. But this download fails because the blob is not in the specified account, but in a different account, for which the specified credentials are invalid.

### Expected Behavior

All the data for instance states and histories, including blobs representing large fields, are stored in the specified TrackingStore storage account.

### Actual Behavior

Any fields that spill to large blobs (e.g. large inputs or outputs) are not stored in the specified TrackingStore account, but in the default storage account.

### Workarounds

None

sebastianburckhardt:
When using a TrackingStore that is in a different account, blobs should also be stored in that account so they stay together with the instance and history tables.  

fixes #879.

### DIFF
--- a/src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
+++ b/src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
@@ -47,20 +47,21 @@ namespace DurableTask.AzureStorage.Storage
             this.Stats = new AzureStorageOrchestrationServiceStats();
             this.queueClient = account.CreateCloudQueueClient();
             this.queueClient.BufferManager = SimpleBufferManager.Shared;
-            this.blobClient = account.CreateCloudBlobClient();
-            this.blobClient.BufferManager = SimpleBufferManager.Shared;
-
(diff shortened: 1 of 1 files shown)


cgillum (approved):
Oh, I totally forgot about this feature! Now I better understand the issue.

I feel like this is the right change. One oddity is that now blobs for large queue messages will potentially live in a separate storage account, but I can't immediately think of any scenarios where that would be a problem since the runtime should always have access to both accounts (unlike clients).

PR #888, opened 2023-04-14
Fix exception when Continue-as-New and Termination happens at the same time

fuifuiyong:
Stop Continue-as-New when runtime state already in terminated status

Related issue: https://github.com/Azure/azure-functions-durable-extension/issues/2264


davidmrdavid (approved):
Thank  you so much for your contribution. This looks good to me, although we'll need to do a bit more testing before merging it to `main` proper. I've changed the target branch to a feature branch that we'll use for testing. Thanks!

Issue #890, opened 2023-04-20
DT.AzureStorage: wrong blob may win when split-brain instances race for the e-tag


sebastianburckhardt:
In split-brain situations, two hosts may try to write the same instance history simultaneously. 

To protect against state corruption, the `AzureTableTrackingStore` uses an e-Tag, stored in a sentinel row, which ensures that only one of the racing accesses can actually update the table. The losing access will then abandon the orchestration work item, thus resolving the split-brain situation (only the winner remains active).

Unfortunately, this mechanism is not sufficient to also protect against *races on blobs* that can happen when table properties are too large and are stored in a blob. In that case, the losing access (in terms of e-tag) may still be able to overwrite the blob of the winning access, leading to incorrect results as reported in https://github.com/Azure/azure-functions-durable-extension/issues/2437.

PR #891, opened 2023-04-20
DT.AzureStorage: add randomization to history blob names to avoid races

sebastianburckhardt:
As observed in #890 and Azure/azure-functions-durable-extension#2437, the current e-tag-based arbitration mechanism does not account for possible races on the blobs. The losing host may overwrite the blob of the winning host, causing inconsistent states.

This PR proposes a fix that adds randomization to blob names used for storing large history fields. 
This randomization means that competing hosts will use different blob names to store the property, preventing the race.


davidmrdavid (commented):
Thanks for the prompt patch. Just some questions!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1241,7 +1241,10 @@ static string GetBlobName(DynamicTableEntity entity, string property)
                 throw new InvalidOperationException($"Could not compute the blob name for property {property}");
             }
 
-            string blobName = $"{sanitizedInstanceId}/history-{sequenceNumber}-{eventType}-{property}.json.gz";
+            // randomize the blob name to prevent accidental races in split-brain situations (#890)
+            uint random = (uint)(new Random()).Next();
```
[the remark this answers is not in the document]
cgillum: > Also - do we need to be concerned about creating orphaned blobs as a result of this?

I don't have it in front of me, but it would be good to check our purge logic to see if it might end up missing orphaned blobs.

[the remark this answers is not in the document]
sebastianburckhardt: A GUID of 128 bit is indeed more random than the 32bit used here. But for this particular purpose I think 32bit is more than enough - split brain does not happen frequently enough to make collisions likely. I didn't want to waste so much space in the blob name.

The purge operation seems to wipe out the entire blob directory belonging to the instance id, so no problem there.

One thing I am not sure about is the case where a history is replaced (e.g. after ContinueAsNew). There should be some code somewhere that removes blobs for the old history (even without the proposed randomization) but I haven't quite spotted it yet.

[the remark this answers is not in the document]
sebastianburckhardt: I looked at the current code some more. It simply does not remove any old blobs when updating histories - so those blobs may accumulate until purging the instance. However, importantly, in some select cases, it can overwrite old blobs with a new blob of the same name. Tor example, when calling continue as new and the new blob happens to have the same event type, and the same large field, and be in the same row of the history, then the blob is overwritten.

This latter case is the important one for entities, since they often have large inputs (that is where the entity state is stored) so that state can be overwritten every time the entity does continue-as-new, which is all the time.

This does mean this PR currently has a problem with entities (and eternal orchestrators) since the blobs will accumulate forever.  It does not seem very easy to fix since the TrackingStore implementation does not know whether the old history contained any large blobs that need to be removed. We can either issue a query, or track that information somehow.

sebastianburckhardt:
I have implemented what I think is an o.k. solution, after some discussions. We still need add some tests though.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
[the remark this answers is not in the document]
sebastianburckhardt: All the issues were addressed.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -843,7 +845,7 @@ public override async Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string
 
             // It is possible that the queue message was small enough to be written directly to a queue message,
             // not a blob, but is too large to be written to a table property.
-            await this.CompressLargeMessageAsync(entity);
+            await this.CompressLargeMessageAsync(entity, null);
```
davidmrdavid, line 848: nit - mind specifying the keyword argument for the `null` parameter? Helps with readability outside the IDE

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -923,16 +925,29 @@ public override Task StartAsync()
...
-        public override async Task<string> UpdateStateAsync(
+        public override Task<string> UpdateStateAsync(
             OrchestrationRuntimeState newRuntimeState,
             OrchestrationRuntimeState oldRuntimeState,
             string instanceId,
             string executionId,
             string eTagValue)
+        {
+            throw new NotSupportedException("Use the overload that includes trackingStoreData");
+        }
```
davidmrdavid, line 936: I find this a bit surprising. Why are we keeping this implementation around if it just throws an exception? Is it not possible to update all call sites?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -923,16 +925,29 @@ public override Task StartAsync()
...
+            OrchestrationRuntimeState oldRuntimeState,
+            string instanceId,
+            string executionId,
+            string eTagValue,
+            object trackingStoreData)
         {
             int estimatedBytes = 0;
             IList<HistoryEvent> newEvents = newRuntimeState.NewEvents;
             IList<HistoryEvent> allEvents = newRuntimeState.Events;
+            TrackingStoreData data = (TrackingStoreData) trackingStoreData;
```
davidmrdavid, line 950: Why can't the parameter type be `TrackingStoreData`? That way, we'd skip the casting in line 950.
I would prefer the callers cast their arguments at the callsites

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -953,7 +968,15 @@ public override Task StartAsync()
...
-           
+
+            // check if we are replacing a previous execution with blobs; those will be deleted from the store after the update
+            List<string> blobsToDelete = null;
+            if (oldRuntimeState != newRuntimeState && data.Blobs.Count > 0)
+            {
+                blobsToDelete = data.Blobs;
+                data.Blobs = new List<string>();
+            }
+
```
davidmrdavid, line 979: To confirm, does `oldRuntimeState != newRuntimeState` confirm we're in a continueAsNew scenario? I'm trying to figure out exactly this allows us to conclude.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1177,7 +1212,7 @@ Type GetTypeForTableEntity(DynamicTableEntity tableEntity)
             }
         }
 
-        async Task CompressLargeMessageAsync(DynamicTableEntity entity)
+        async Task CompressLargeMessageAsync(DynamicTableEntity entity, List<string> listOfBlobs)
```
davidmrdavid, line 1215: Should the parameter type have a `?` in the type declaration?

davidmrdavid (commented):
Left more preliminary questions

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: sounds good.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: The mechanism is supposed to be independent of the implementation of TrackingStore. So it can't really depend on a type specifically designed for AzureTableTrackingStore.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: Yes, it does. This same test is used in `InstanceStoreBackedTrackingStore`.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: This file is not compiled with `# nullable enable`

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: I can't remove the method from the public interface `ITrackingStore`, so it needs to be implemented in some way. I thought throwing an exception was the best way to catch future mistakes where people call the wrong one (because it is still there).

cgillum (commented):
I'm quite uncomfortable with this change as it's currently implemented, mainly for two reasons:

1. We're adding new public surface area that doesn't add any value to the abstraction
2. The code is becoming more complicated since we need to cast interfaces and add new conditionals

As far as I can tell, `ITrackingStore` is a public interface but we don't expose any way for 3rd parties to introduce their own tracking stores. We own all the implementations. For that reason, I'd much prefer we make a breaking change to `ITrackingStore` rather than creating a new `ITrackingStoreXXX` interface.

sebastianburckhardt:
> For that reason, I'd much prefer we make a breaking change to ITrackingStore rather than creating a new ITrackingStoreXXX interface.

I am all for it - would be my preference also. Definitely cleaner and simpler. Will update the PR accordingly.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -953,7 +957,15 @@ public override Task StartAsync()
...
             };
-           
+
+            // check if we are replacing a previous execution with blobs; those will be deleted from the store after the update
+            List<string> blobsToDelete = null;
+            if (oldRuntimeState != newRuntimeState && data.Blobs.Count > 0)
+            {
+                blobsToDelete = data.Blobs;
+                data.Blobs = new List<string>();
+            }
```
davidmrdavid, line 967: nit: let's call out explicitly this is a ContinueAsNew scenario. I think it's helpful to convey the high-level API that would drive this.
```suggestion
            // check if we are replacing a previous execution with blobs; those will be deleted from the store after the update. This could occur in a ContinueAsNew scenario
            List<string> blobsToDelete = null;
            if (oldRuntimeState != newRuntimeState && data.Blobs.Count > 0)
            {
                blobsToDelete = data.Blobs;
                data.Blobs = new List<string>();
            }
```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
davidmrdavid: In that case, should we do a check here in case the type casting fails?

davidmrdavid (commented):
Just a few more questions

davidmrdavid (commented):
Just a few more questions

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: I am not sure what benefit an explicit check would add over the exception that is already being thrown if this should ever happen. There is not much risk here that the cast would ever fail IMO.

davidmrdavid: Now that I think about it, you're right - the cast should fail if it's not valid. I was getting confused with the behavior of casting in PLs with gradual typing, where sometimes the cast fails silently. Ok then, please disregard my comment here.

davidmrdavid (approved):
LGTM but I suspect we may want to wait on @cgillum's approval too

davidmrdavid (approved):
LGTM but I suspect we may want to wait on @cgillum's approval too

davidmrdavid (approved):
LGTM but I suspect we may want to wait on @cgillum's approval too

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -216,6 +216,12 @@ public Task<string> DownloadAndDecompressAsBytesAsync(Uri blobUri)
             return DownloadAndDecompressAsBytesAsync(blob);
         }
 
+        public Task<bool> DeleteOrphanedBlobAsync(string blobName)
```
cgillum, line 219: nit: Can we call this method `DeleteBlobAsync`? I don't see any logic here related to orphaned data.

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
```diff
@@ -69,7 +69,8 @@ interface ITrackingStore
         /// <param name="instanceId">InstanceId for the Orchestration Update</param>
         /// <param name="executionId">ExecutionId for the Orchestration Update</param>
         /// <param name="eTag">The ETag value to use for safe updates</param>
-        Task<string> UpdateStateAsync(OrchestrationRuntimeState newRuntimeState, OrchestrationRuntimeState oldRuntimeState, string instanceId, string executionId, string eTag);
+        /// <param name="trackingStoreData">The additional data that is maintained for this execution.</param>
+        Task<string> UpdateStateAsync(OrchestrationRuntimeState newRuntimeState, OrchestrationRuntimeState oldRuntimeState, string instanceId, string executionId, string eTag, object trackingStoreData);
```
cgillum, line 73: nit: `trackingStoreContext` might be a better word since "data" is heavily overloaded in this context (it would almost read as if this "data" is part of what we're updating).

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -32,7 +33,7 @@ namespace DurableTask.AzureStorage.Tracking
     /// <summary>
     /// Tracking store for use with <see cref="AzureStorageOrchestrationService"/>. Uses Azure Tables and Azure Blobs to store runtime state.
     /// </summary>
-    class AzureTableTrackingStore : TrackingStoreBase
+    class AzureTableTrackingStore : TrackingStoreBase, ITrackingStore
```
cgillum, line 36: `TrackingStoreBase` already implements `ITrackingStore` so you can remove this.

src/DurableTask.AzureStorage/Tracking/OrchestrationHistory.cs
```diff
@@ -35,11 +35,20 @@ public OrchestrationHistory(IList<HistoryEvent> historyEvents, DateTime lastChec
             this.LastCheckpointTime = lastCheckpointTime;
             this.ETag = eTag;
```
cgillum, line 36: Let's update this overload to call your new overload.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1108,6 +1120,18 @@ public override Task StartAsync()
                 episodeNumber,
                 orchestrationInstanceUpdateStopwatch.ElapsedMilliseconds);
 
+            // finally, delete orphaned blobs from the previous execution history.
+            // We had to wait until the new history has committed to make sure the blobs are no longer necessary.
+            if (blobsToDelete != null)
+            {
+                var tasks = new List<Task>();
```
cgillum, line 1127:
```suggestion
                var tasks = new List<Task>(blobsToDelete.Count);
```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1241,7 +1271,10 @@ static string GetBlobName(DynamicTableEntity entity, string property)
                 throw new InvalidOperationException($"Could not compute the blob name for property {property}");
             }
 
-            string blobName = $"{sanitizedInstanceId}/history-{sequenceNumber}-{eventType}-{property}.json.gz";
+            // randomize the blob name to prevent accidental races in split-brain situations (#890)
+            uint random = (uint)(new Random()).Next();
+
```
cgillum, line 1276: Do we have to allocate a new `Random()` each time or can we use a static `Random` instance?

cgillum (approved):
A few more comments, though I don't think any of them are necessarily blocking.

src/DurableTask.AzureStorage/MessageManager.cs
sebastianburckhardt: Good idea.

src/DurableTask.AzureStorage/Tracking/ITrackingStore.cs
sebastianburckhardt: I like it.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: Absolutely.

src/DurableTask.AzureStorage/Tracking/OrchestrationHistory.cs
sebastianburckhardt: Makes sense.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: It's a little tricky because the `Random` class is unfortunately not thread-safe, so I would have to do something complicated or ugly. I think it is better to keep it simple to read and maintain, and the performance is probably fine considering that this is only called when we need a large blob, which is by itself going to use significant CPU and memory (so this little allocation is a tiny fraction of the cost).

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sebastianburckhardt: Sure.

collins-benj:
We have installed a preview version of this lib (`1.13.7-preview`) given to us through our support ticket to hopefully fix our issue https://github.com/Azure/azure-functions-durable-extension/issues/2437. We have seen the following log:
> @holdreleasecommandprocessor@7399a8cd-878d-48a2-a37c-684210bcb902: Split brain detected: Another worker already updated the history for this instance - the 3 history event result(s) will be discarded

immediately after this, we see this error: 
> DurableTask.AzureStorage.Storage.DurableTaskStorageException: Element 3 in the batch returned an unexpected response code.

At this point, the orchestration that calls our entity halts in the "running" state, and never continues as new as we would expect. To unstick we have had to purge orchestration history and start new. This has happened on 3 occasions over the past week.

PR #892, opened 2023-04-24
[AzureServiceFabric] Handle missing ExecutionStartedEvent orchestrations

shankarsama:
This PR is handling the following cases:
1. When an orchestration is terminated, sometimes it's adding a TerminateEvent message to a non-existent instance id or execution id
2. If we clean up the session store then the in-memory session queue may still contain an entry which tries to find an already cleaned up session.

### DIFF
--- a/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
+++ b/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
@@ -159,7 +159,7 @@ namespace DurableTask.AzureServiceFabric
 
                 // Do not return the orchestration workitem if 'ExecutionStartedEvent' is missing.
                 // This can happen when an orchestration message like TerminateEvent is sent to an already finished orchestration
-                // TODO: Fix race this condition that introduced the bad orchestration
+                // TODO: Fix the race condition that introduced the bad orchestration
                 if (currentRuntimeState.ExecutionStartedEvent == null)
                 {
                     ServiceFabricProviderEventSource.Tracing.UnexpectedCodeCondition($"Orchestration with no execution started event found: {currentSession.SessionId}");
--- a/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
+++ b/src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
@@ -157,9 +157,9 @@ namespace DurableTask.AzureServiceFabric
                     OrchestrationRuntimeState = currentRuntimeState
                 };
 
-                // Do not return the orchestraiton if 'ExecutionStartedEvent' is missing.
-                // This can happen when an when an orchestration message like TerminateEvent is sent to an already finished orchestration
-                // TODO: Fix race this condition
+                // Do not return the orchestration workitem if 'ExecutionStartedEvent' is missing.
+                // This can happen an orchestration message like TerminateEvent is sent to an already finished orchestration
+                // TODO: Fix race this condition that introduced the bad orchestration
                 if (currentRuntimeState.ExecutionStartedEvent == null)
                 {
                     ServiceFabricProviderEventSource.Tracing.UnexpectedCodeCondition($"Orchestration with no execution started event found: {currentSession.SessionId}");
@@ -357,7 +357,6 @@ namespace DurableTask.AzureServiceFabric
             }
         }
 
-        // Caller should ensure the workItem has reached terminal state.
         private async Task DropOrchestrationAsync(TaskOrchestrationWorkItem workItem)
         {
             await RetryHelper.ExecuteWithRetryOnTransient(async () =>
@@ -372,7 +371,7 @@ namespace DurableTask.AzureServiceFabric
                     await this.orchestrationProvider.DropSession(txn, workItem.OrchestrationRuntimeState.OrchestrationInstance);
                     await txn.CommitAsync();
                 }
-            }, uniqueActionIdentifier: $"OrchestrationId = '{workItem.InstanceId}', Action = '{nameof(HandleCompletedOrchestration)}'");
+            }, uniqueActionIdentifier: $"OrchestrationId = '{workItem.InstanceId}', Action = '{nameof(DropOrchestrationAsync)}'");
 
             this.instanceStore.OnOrchestrationCompleted(workItem.OrchestrationRuntimeState.OrchestrationInstance);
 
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -346,6 +357,33 @@ public Task RenewTaskOrchestrationWorkItemLockAsync(TaskOrchestrationWorkItem wo
             }
         }
 
+        // Caller should ensure the workItem has reached terminal state.
```
lchew7, line 360: Is this comment correct?

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -346,6 +357,33 @@ public Task RenewTaskOrchestrationWorkItemLockAsync(TaskOrchestrationWorkItem wo
...
+                    // DropSession does 2 things (like mentioned in the comments above) - remove the row from sessions dictionary
+                    // and delete the session messages dictionary. The second step is in a background thread and not part of transaction.
+                    // However even if this transaction failed but we ended up deleting session messages dictionary, that's ok - at
+                    // that time, it should be an empty dictionary and we would have updated the runtime session state to full completed
+                    // state in the transaction from Complete method. So the subsequent attempt would be able to complete the session.
+                    await this.orchestrationProvider.DropSession(txn, workItem.OrchestrationRuntimeState.OrchestrationInstance);
+                    await txn.CommitAsync();
+                }
+            }, uniqueActionIdentifier: $"OrchestrationId = '{workItem.InstanceId}', Action = '{nameof(HandleCompletedOrchestration)}'");
+
```
lchew7, line 376: Should this be 'DropOrchestrationAsync' instead?

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -346,6 +357,33 @@ public Task RenewTaskOrchestrationWorkItemLockAsync(TaskOrchestrationWorkItem wo
...
+            this.instanceStore.OnOrchestrationCompleted(workItem.OrchestrationRuntimeState.OrchestrationInstance);
+
+            string message = string.Format("Orchestration with instanceId : '{0}' and executionId : '{1}' dropped due to state corruption (missing ExecutionStartedEvent)",
```
lchew7, line 379: Minor: This seems like a generic method.  Other people could call it for other reasons.  Maybe we can change this method to accept a string 'reason' parameter?

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -149,13 +149,24 @@ public async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAs
...
                     OrchestrationRuntimeState = currentRuntimeState
                 };
 
+                // Do not return the orchestraiton if 'ExecutionStartedEvent' is missing.
+                // This can happen when an when an orchestration message like TerminateEvent is sent to an already finished orchestration
+                // TODO: Fix race this condition
+                if (currentRuntimeState.ExecutionStartedEvent == null)
+                {
+                    ServiceFabricProviderEventSource.Tracing.UnexpectedCodeCondition($"Orchestration with no execution started event found: {currentSession.SessionId}");
+                    await this.DropOrchestrationAsync(workItem);
```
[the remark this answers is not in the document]
shankarsama: If there are no new messages, then means it could be a completed orchestration.

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
shankarsama: fixed

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
shankarsama: Not until we fix the race condition that caused this bad orchestration.

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -149,13 +149,24 @@ public async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAs
...
+
                 var workItem = new TaskOrchestrationWorkItem()
                 {
                     NewMessages = newMessages.Select(m => m.Value.TaskMessage).ToList(),
                     InstanceId = currentSession.SessionId.InstanceId,
                     OrchestrationRuntimeState = currentRuntimeState
                 };
 
+                // Do not return the orchestration workitem if 'ExecutionStartedEvent' is missing.
+                // This can happen an orchestration message like TerminateEvent is sent to an already finished orchestration
```
lchew7, line 161: Minor: This can happen *when* an

src/DurableTask.AzureServiceFabric/Stores/SessionProvider.cs
```diff
@@ -108,23 +109,28 @@ public async Task<PersistentSession> AcceptSessionAsync(TimeSpan receiveTimeout)
...
                                         {
-                                            var errorMessage = $"Internal Server Error : Unexpected to dequeue the session {returnInstanceId} which was already locked before";
+                                            var errorMessage = $"Internal Server Error : Unexpected to dequeue the session {returnInstanceId} which was already locked before. Do not re-enture the item-without-session";
                                             ServiceFabricProviderEventSource.Tracing.UnexpectedCodeCondition(errorMessage);
                                             throw new Exception(errorMessage);
                                         }
                                     }
                                     else
                                     {
+                                        // If the session state is cleared using ForceTerminateOrchestration + reason (cleanupstore) then we may endup with a stale entry in the queue
```
lchew7, line 119: Minor: endup -> end up

src/DurableTask.AzureServiceFabric/FabricOrchestrationService.cs
```diff
@@ -149,13 +149,24 @@ public async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAs
...
+                // TODO: Fix race this condition that introduced the bad orchestration
```
lchew7, line 162: Minor: "Fix the race condition that"

PR #901, opened 2023-05-05
[WIP] Fix exception when Continue-as-New and Termination happens at the same time

davidmrdavid:
A feature branch, for testing, containing the community contribution here: https://github.com/Azure/durabletask/pull/888
This is a fix contributed by @fuifuiyong.

Creating this branch for creating a private package with which to test this, and to perform edits as necessary.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -805,7 +805,7 @@ namespace DurableTask.Core
             out bool continuedAsNew)
         {
             ExecutionCompletedEvent executionCompletedEvent;
-            continuedAsNew = (completeOrchestratorAction.OrchestrationStatus == OrchestrationStatus.ContinuedAsNew);
+            continuedAsNew = (completeOrchestratorAction.OrchestrationStatus == OrchestrationStatus.ContinuedAsNew) && (runtimeState.OrchestrationStatus != OrchestrationStatus.Terminated);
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -805,7 +805,7 @@ bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem)
             out bool continuedAsNew)
         {
             ExecutionCompletedEvent executionCompletedEvent;
-            continuedAsNew = (completeOrchestratorAction.OrchestrationStatus == OrchestrationStatus.ContinuedAsNew);
+            continuedAsNew = (completeOrchestratorAction.OrchestrationStatus == OrchestrationStatus.ContinuedAsNew) && (runtimeState.OrchestrationStatus != OrchestrationStatus.Terminated);
```
cgillum, line 808: Instead of making a change here, I think it would be better to make a change further upstream in `OrchestrationRuntimeState.SetMarkerEvents`. Instead of always throwing if we see that `ExecutionCompletedEvent == null`, we could add an additional check that allows transitioning from a `ContinuedAsNew` state to the `Terminated` state. I feel like this is a reasonable thing to allow since termination should be considered forceful.

I'd be fine going even further than this too. I don't see a strong reason to throw an exception if multiple completion events are found as long as we can make the expected outcome predictable. Ideally we'd simply log a warning if we encounter multiple completion events. I feel like failing with an exception is only useful when we don't have a reasonable way to procede.

PR #910, opened 2023-05-26
Validate that timeout is positive in WaitForOrchestrationAsync, and enable nullable checks

davidmrdavid:
This PR validates that the timeout in WaitForOrchestrationAsync is positive. Otherwise, we risk accidentally returning `null` without doing any meaningful work.

This PR also enables nullable analysis for part of the `AzureStorageOrchestrationService` file

I encountered this in the process of repro'ing this issue: https://github.com/microsoft/durabletask-dotnet/issues/143

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1843,7 +1843,7 @@ namespace DurableTask.AzureStorage
         /// </summary>
         /// <param name="instanceId">The orchestration instance to wait for.</param>
         /// <param name="executionId">The execution ID (generation) of the specified instance.</param>
-        /// <param name="timeout">>Max timeout to wait. Only positive <see cref="TimeSpan"/> values or <see cref="Timeout.InfiniteTimeSpan"/> are allowed.</param>
+        /// <param name="timeout">Max timeout to wait. Only positive <see cref="TimeSpan"/> values or <see cref="Timeout.InfiniteTimeSpan"/> are allowed.</param>
         /// <param name="cancellationToken">Task cancellation token.</param>
         public async Task<OrchestrationState?> WaitForOrchestrationAsync(
             string instanceId,
@@ -1865,7 +1865,7 @@ namespace DurableTask.AzureStorage
             }
 
             TimeSpan statusPollingInterval = TimeSpan.FromSeconds(2);
-            while (!cancellationToken.IsCancellationRequested && (isInfiniteTimeSpan || (timeout >= TimeSpan.Zero)))
+            while (!cancellationToken.IsCancellationRequested)
             {
                 OrchestrationState state = await this.GetOrchestrationStateAsync(instanceId, executionId);
                 if (state == null ||
@@ -1885,6 +1885,14 @@ namespace DurableTask.AzureStorage
                         state.Output = await this.messageManager.FetchLargeMessageIfNecessary(state.Output);
                     }
                     return state;
+                }
+
+                // For a user-provided timeout of `TimeSpan.Zero`,
+                // we want to check the status of the orchestration once and then return.
+                // Therefore, we check the timeout condition after the first status check.
+                if (!isInfiniteTimeSpan && (timeout <= TimeSpan.Zero))
+                {
+                    break;
                 }
             }
 
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1843,7 +1843,7 @@ namespace DurableTask.AzureStorage
         /// </summary>
         /// <param name="instanceId">The orchestration instance to wait for.</param>
         /// <param name="executionId">The execution ID (generation) of the specified instance.</param>
-        /// <param name="timeout">Max timeout to wait. Only positive `TimeSpan` values, or `Timeout.InfiniteTimeSpan`, are allowed.</param>
+        /// <param name="timeout">>Max timeout to wait. Only positive <see cref="TimeSpan"/> values or <see cref="Timeout.InfiniteTimeSpan"/> are allowed.</param>
         /// <param name="cancellationToken">Task cancellation token.</param>
         public async Task<OrchestrationState?> WaitForOrchestrationAsync(
             string instanceId,
@@ -1861,11 +1861,11 @@ namespace DurableTask.AzureStorage
             {
                 throw new ArgumentException($"The parameter {nameof(timeout)} cannot be negative." +
                     $" The value for {nameof(timeout)} was '{timeout}'." +
-                    $" Please provide either positive timeout value, or `Timeout.InfiniteTimeSpan`.");
+                    $" Please provide either a positive timeout value or Timeout.InfiniteTimeSpan.");
             }
 
             TimeSpan statusPollingInterval = TimeSpan.FromSeconds(2);
-            while (!cancellationToken.IsCancellationRequested && (isInfiniteTimeSpan || (timeout > TimeSpan.Zero)))
+            while (!cancellationToken.IsCancellationRequested && (isInfiniteTimeSpan || (timeout >= TimeSpan.Zero)))
             {
                 OrchestrationState state = await this.GetOrchestrationStateAsync(instanceId, executionId);
                 if (state == null ||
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1857,6 +1856,13 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
                 throw new ArgumentException(nameof(instanceId));
             }
 
+            if (timeout < TimeSpan.Zero)
```
jviau, line 1859: Should we accept negative values? There seems to be a precedence that negative implies infinite for timeouts. Especially since `Timeout.InfiniteTimeSpan` is negative.

davidmrdavid: Maybe, I think it's up for debate, and I no strong feelings on this. I simply noticed that the code _today_ assumes the timeout will be positive, and so I made that assumption explicit by validating the input.

Using a combination of `Datetime.now` and the provided timespan, we can modify the currrent implementation to work with infinite timespans as well.

Would you prefer if we accepted negative timespans?

jviau: I think it would be a better experience to accept negative as an infinite timeout.

cgillum: I'd say let's be specific and allow for `Timeout.InfiniteTimeSpan` to mean infinite, whereas we throw for any other negative value. I believe this is mostly consistent with other .NET APIs and helps us differentiate an explicit desire for infinite vs. an accidental negative value provided by the user.

So here, I think we'd say `if (timeout < TimeSpan.Zero && timeout != Timeout.InfiniteTimeSpan)` and below, we change the while loop condition to be `while (!cancellationToken.IsCancellationRequested)` (removing the `timeout` check).

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1857,8 +1856,10 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
                 throw new ArgumentException(nameof(instanceId));
             }
 
+            // only check timeout only if it is a positive value
+            var ignoreTimeout = timeout < TimeSpan.Zero;
```
cgillum, line 1860: nit: please use `bool` instead of `var`. We should generally avoid `var` in this codebase unless using the variable type name is redundant, which is not the case here.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1838,15 +1837,15 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
...
             return storagePurgeHistoryResult.ToCorePurgeHistoryResult();
         }
-
+#nullable enable
         /// <summary>
         /// Wait for an orchestration to reach any terminal state within the given timeout
         /// </summary>
         /// <param name="instanceId">The orchestration instance to wait for.</param>
         /// <param name="executionId">The execution ID (generation) of the specified instance.</param>
         /// <param name="timeout">Max timeout to wait.</param>
```
cgillum, line 1846: Please update the documentation to indicate how to wait indefinitely.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: I like the compromise of explicitly checking for `Timeout.InfiniteTimeSpan`, and erroring out in other negative `TimeSpan` cases. My latest commit tries to do just that.

My if statement is as you suggested, @cgillum. However, the while loop does preserve the timeout check because I think we need it for the "happy case" where the timeout is positive. I suppose I could also move it to an if-statement check inside the while loop though.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: Good to know. Using `bool` now.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: Sounds good, I added a note on `InfiniteTimeSpan` on my latest commit

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1843,7 +1843,7 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
         /// </summary>
         /// <param name="instanceId">The orchestration instance to wait for.</param>
         /// <param name="executionId">The execution ID (generation) of the specified instance.</param>
-        /// <param name="timeout">Max timeout to wait. Only positive values are allowed</param>
+        /// <param name="timeout">Max timeout to wait. Only positive `TimeSpan` values, or `Timeout.InfiniteTimeSpan`, are allowed.</param>
```
cgillum, line 1846: XML doc comments for C# have a specific convention for referencing other types.
```suggestion
        /// <param name="timeout">>Max timeout to wait. Only positive <see cref="TimeSpan"/> values or <see cref="Timeout.InfiniteTimeSpan"/> are allowed.</param>
```
If you simply want the equivalent of markdown's inline code block (`code`), then you can use `<c>code</c>`.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1856,15 +1856,16 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
...
             }
 
-            if (timeout < TimeSpan.Zero)
+            bool isInfiniteTimeSpan = timeout == Timeout.InfiniteTimeSpan;
+            if (timeout < TimeSpan.Zero && !isInfiniteTimeSpan)
             {
                 throw new ArgumentException($"The parameter {nameof(timeout)} cannot be negative." +
                     $" The value for {nameof(timeout)} was '{timeout}'." +
-                    $" Please provide a positive timeout value.");
+                    $" Please provide either positive timeout value, or `Timeout.InfiniteTimeSpan`.");
```
cgillum, line 1864:
```suggestion
                    $" Please provide either a positive timeout value or Timeout.InfiniteTimeSpan.");
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1857,8 +1856,16 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
...
+            if (timeout < TimeSpan.Zero && !isInfiniteTimeSpan)
+            {
+                throw new ArgumentException($"The parameter {nameof(timeout)} cannot be negative." +
+                    $" The value for {nameof(timeout)} was '{timeout}'." +
+                    $" Please provide either positive timeout value, or `Timeout.InfiniteTimeSpan`.");
+            }
+
             TimeSpan statusPollingInterval = TimeSpan.FromSeconds(2);
-            while (!cancellationToken.IsCancellationRequested && timeout > TimeSpan.Zero)
+            while (!cancellationToken.IsCancellationRequested && (isInfiniteTimeSpan || (timeout > TimeSpan.Zero)))
```
cgillum, line 1868: I wonder if we should change this to support `TimeSpan.Zero`.
```suggestion
            while (!cancellationToken.IsCancellationRequested && (isInfiniteTimeSpan || (timeout >= TimeSpan.Zero)))
```
Otherwise, specifying `TimeSpan.Zero` will always result in `null`, which would be unexpected. Normally, specifying `TimeSpan.Zero` means that you want to "check the thing but don't wait for it if it's not ready.".

cgillum (approved):
Happy to see you decided to support `Timeout.InfiniteTimeSpan`. :)

Added some non-blocking suggestions. I realize the one regarding `TimeSpan.Zero` would require some refactoring for a "perfect" implementation, so I'm fine if you want to skip that.

davidmrdavid:
@cgillum: let's try to get this right all the way. Does this latest commit look good to you?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1879,6 +1886,14 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
                     }
                     return state;
                 }
+
+                // For a user-provided timeout of `TimeSpan.Zero`,
+                // we want to check the status of the orchestration once and then return.
+                // Therefore, we check the timeout condition after the first status check.
+                if (!isInfiniteTimeSpan && (timeout <= TimeSpan.Zero))
```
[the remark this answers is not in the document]
davidmrdavid: Ah, you're right, totally missed that. Let me fix that and actually refactor this code a bit further

cgillum (approved):
LGTM!

PR #913, opened 2023-06-12
Update Log Info for GetMessagesAsync

nytiannn:
To enhance the logging clarity in the `GetMessageAsync` method, update the log to include both lease loss and shutdown as cancellation reasons. This will provide more detailed information in the Kusto logs.

**Old implementation**:
image
When least lost happened, the method still logged cancellation reason as `shutdown`. 

**Update**:
when queues are released due to lease loss, the log specifies the reason as `lease lost` rather than `shutdown`.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
@@ -184,7 +184,7 @@ namespace DurableTask.AzureStorage.Messaging
                     }
                 }
 
-                this.Release(CloseReason.Shutdown, $"ControlQueue GetMessagesAsync cancelled by: {(this.releaseCancellationToken.IsCancellationRequested ? "control queue released token cancelled" : "")} {(cancellationToken.IsCancellationRequested ? "shutdown token cancelled" : "")}");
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -184,7 +184,7 @@ await batch.ParallelForEachAsync(async delegate (QueueMessage queueMessage)
                     }
                 }
 
-                this.Release(CloseReason.Shutdown, $"ControlQueue GetMessagesAsync cancelled by: {(this.releaseCancellationToken.IsCancellationRequested ? "control queue released token cancelled" : "")} {(cancellationToken.IsCancellationRequested ? "shutdown token cancelled" : "")}");
+                this.Release(cancellationToken.IsCancellationRequested ? CloseReason.Shutdown : CloseReason.LeaseLost, $"ControlQueue GetMessagesAsync cancelled by: {(this.releaseCancellationToken.IsCancellationRequested ? "control queue released token cancelled" : "")} {(cancellationToken.IsCancellationRequested ? "shutdown token cancelled" : "")}");
```
davidmrdavid, line 187: Just for my own understanding: can you please explain how we know that if `cancellationToken.IsCancellationRequested` is false, that the closing reason is that the lease was lost? Is there anywhere in the code that demonstrates this?

davidmrdavid (commented):
Just a quick clarifying question

davidmrdavid (approved):

PR #917, opened 2023-06-22
[WIP] Report Azure Storage metrics

davidmrdavid:


davidmrdavid:
@cgillum: this is still WIP - but I'm curious if you know whether periodically peeking at the storage metrics, like we do in this PR, will increase storage costs? Since it's not actually adding more data to the storage itself, I'm not sure if this is charged.

cgillum:
@davidmrdavid yes, all storage transactions will incur costs. These charges are separate from the actual storage costs that customers get charged.

Here's some example info on queue storage and transaction costs: https://azure.microsoft.com/pricing/details/storage/queues/

torosent:

PR #918, opened 2023-06-22
Automatically generate version suffixes for private releases of DTFx.AS and DTFx.Core

davidmrdavid:
When making a private extension release, we usually add a version suffix to the package so that it can be distinguished from an official release.

This PR automates that process so that the CI automatically adds a version suffix when the release pipeline is triggered in a feature branch. The version suffix will be -preview.<buildVersion> where <buildVersion is automatically generated by the CI.

Since multiple groups work in this repo, I've only made this change for DTFx.AS and DTFx.Core. In order to achieve that, I created a copy of the release pipeline yaml and filtered it to only apply to those two packages.

### DIFF
--- a/azure-pipelines-release-azure-storage.yml
+++ b/azure-pipelines-release-azure-storage.yml
@@ -1,0 +1,136 @@
+trigger: none
+pr: none
+
+pool:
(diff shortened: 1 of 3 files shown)


torosent:

PR #921, opened 2023-07-04
Suggested edits for table partition manager

cgillum:
@nytiannn These are mostly stylistic edits for your current partition manager PR (naming, whitespace, removing redundancies, etc.). Please review feel free to merge if you like all of them or start discussions if there are some that you don't understand or agree with. I figured batching my nits in a PR would be more efficient than adding more comments to your PR. :)

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
+++ b/Test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
@@ -27,16 +27,13 @@ namespace DurableTask.AzureStorage.Tests
     [TestClass]
     public class TestTablePartitionManager
     {
-
         readonly string connection = TestHelpers.GetTestStorageAccountConnectionString();
 
-        [TestMethod]
         // Start with one worker and four partitions.
         // Test the worker could claim all the partitions in 5 seconds.
+        [TestMethod]
         public async Task TestOneWorkerWithFourPartition()
         {
-            //string storageConnectionString = TestHelpers.GetTestStorageAccountConnectionString();
-            //var storageAccount = CloudStorageAccount.Parse(storageConnectionString);
             string testName = nameof(TestOneWorkerWithFourPartition);
             var settings = new AzureStorageOrchestrationServiceSettings
             {
@@ -54,20 +51,20 @@ namespace DurableTask.AzureStorage.Tests
                 timeout: TimeSpan.FromSeconds(5),
                 condition: () =>
                 {
-                var partitions = service.ListTableLeases();
-                Assert.AreEqual(4, partitions.Count());
-                return partitions.All(p => p.CurrentOwner == "0");
+                    var partitions = service.ListTableLeases();
+                    Assert.AreEqual(4, partitions.Count());
+                    return partitions.All(p => p.CurrentOwner == "0");
                 });
 
             await service.StopAsync();
             await service.DeleteAsync();
         }
 
-        [TestMethod]
         //Starts with two workers and four partitions.
         //Test that one worker can acquire two partitions. 
         //Since two worker can not start at the same time, and one will start earlier than the another one.
         //There would be a steal process, and test that the lease tranfer will take no longer than 30 sec.
+        [TestMethod]
         public async Task TestTwoWorkerWithFourPartitions()
         {
             var services = new AzureStorageOrchestrationService[2];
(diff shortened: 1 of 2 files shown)


Test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
```diff
@@ -710,14 +706,18 @@ static bool CheckAllArraysUnique(params string[][] arrays)
 
         static void WaitForCondition(TimeSpan timeout, Func<bool> condition)
         {
+            if (Debugger.IsAttached)
+            {
+                // Give more time for debugging so we can step through the code
+                timeout = TimeSpan.FromMinutes(3);
```
cgillum, line 712: This is a trick I use to make debugging tests a little easier.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -23,12 +23,14 @@ namespace DurableTask.AzureStorage.Partitioning
...
     using DurableTask.AzureStorage.Storage;
 
-
     /// <summary>
     /// Partition Manager V3 based on Azure Storage V2.  
     /// </summary>
     sealed class TablePartitionManager : IPartitionManager, IDisposable
     {
+        // Used for logging purposes only, indicating that a particular parameter (usually a partition ID) doesn't apply.
+        const string NotApplicable = "";
```
cgillum, line 32: I created this to avoid having too many redundant comments. Hopefully it's clear enough.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -196,36 +199,41 @@ async Task IPartitionManager.StopAsync()
...
                 }
-                
-                if(failureCount != 0 && failureCount < maxFailureCount)
-                {
-                    timeToSleep = TimeSpan.FromSeconds(0);
-                }
-                //If failed to update the table too many times, extend the wait time to reeduce excessive loggings.
-                else if(failureCount > maxFailureCount)
+
+                if (failureCount > 0)
```
cgillum, line 214: I wrapped this logic in a separate if-check to make it more clear that it only applies when failures happen.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -397,51 +409,53 @@ bool TryClaimLease(TableLease partition)
...
-                            if (partition.NextOwner == null)
-                            {
-                                ownershipLeaseCount++;
-                            }
-                            else// Lease is stolen by others.
-                            {
-                                response.IsDrainingPartition = true;
-                                // In case operations on dictionary and isDraining are not sync, always check first if we have any drain tasks in the dictionary.
-                                bool isDrainTaskExists = this.drainTask.TryGetValue(partition.RowKey!, out Task? task);
+                            return;
```
cgillum, line 414: Returning early from a method like this is a common practice to reduce code complexity.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -482,28 +496,14 @@ void CheckOtherWorkerLease(TableLease partition, Dictionary<string, List<TableLe
...
-                }
-                // Exception for failed table update due to ETag expired. 
-                // Re-read the table immediately to get a new ETag.
-                catch (RequestFailedException ex) when (ex.Status == 412)
-                {
-                    throw;
                 }
-                
 
+                await this.BalanceLeases(partitionDistribution, partitions, ownershipLeaseCount, response);
```
cgillum, line 505: I noticed that the code previously did try/catch/throw, which seemed unnecessary, so I just removed it.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -584,29 +585,28 @@ public async Task BalanceLeases(Dictionary<string, List<TableLease>> partitionDi
...
+                                this.settings.WorkerId,
+                                partition.RowKey,
+                                $"Failed to update table entry due to an Etag mismatch. Failed ETag value: '{etag}'");
                             throw;
                         }
                         // Eat any exceptions during lease stealing.
                         catch (Exception exception)
                         {
-                            this.settings.Logger.PartitionManagerError(
+                            this.settings.Logger.PartitionManagerWarning(
```
cgillum, line 598: I don't feel strongly about this, but I usually try to use Error only when something really bad actually happens. That's usually not the case when we swallow exceptions like this.

cgillum (commented):
Added a few short explanations.

Test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
nytiannn: I am sorry I am not very familiar with this... What is this `Debugger.IsAttached` and why it's a trick?

cgillum: Here is the documentation: [Debugger.IsAttached](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debugger.isattached?view=net-7.0). It lets you know whether Visual Studio is attached to the process. It's a convenient way to increase the timeout when you're slowly stepping through code while not increasing the timeout for regular unit test execution, which happens without any debuggers attached.

PR #922, opened 2023-07-05
More suggested edits

cgillum:
Here are a few more suggested updates based on my review of the code. They are mostly small things related to naming and comments.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
@@ -24,22 +24,30 @@ namespace DurableTask.AzureStorage.Partitioning
     using DurableTask.AzureStorage.Storage;
 
     /// <summary>
-    /// Partition Manager V3 based on Azure Storage V2.  
+    /// Partition Manager V3 based on Azure Storage Tables.
     /// </summary>
+    /// <remarks>
+    /// Previous partition managers were based on Azure Storage blobs, which are more complex to manage
+    /// and have more expensive per-transaction costs, particularly in Azure Storage V2 accounts. This
+    /// table storage-based partition manager aims to be both simpler, cheaper, easier to debug, and 
+    /// faster when it comes to rebalancing.
+    /// </remarks>
     sealed class TablePartitionManager : IPartitionManager, IDisposable
     {
         // Used for logging purposes only, indicating that a particular parameter (usually a partition ID) doesn't apply.
         const string NotApplicable = "";
-        Task partitionManagerTask;
 
         readonly AzureStorageClient azureStorageClient;
         readonly AzureStorageOrchestrationService service;
         readonly AzureStorageOrchestrationServiceSettings settings;
-        readonly CancellationTokenSource partitionManagerCancellationSource;
+        readonly CancellationTokenSource gracefulShutdownToken;
+        readonly CancellationTokenSource forcefulShutdownToken;
         readonly string storageAccountName;
         readonly TableClient partitionTable;
         readonly TableLeaseManager tableLeaseManager;
         readonly LeaseCollectionBalancerOptions options;
+
+        Task partitionManagerTask;
 
         /// <summary>
         /// Constructor to initiate new instances of TablePartitionManager.
@@ -53,19 +61,22 @@ namespace DurableTask.AzureStorage.Partitioning
             this.azureStorageClient = azureStorageClient;
             this.service = service;
             this.settings = this.azureStorageClient.Settings;
-            string connectionString = this.settings.StorageConnectionString ?? this.settings.StorageAccountDetails.ConnectionString;
             this.storageAccountName = this.azureStorageClient.TableAccountName;
+
+            string connectionString = this.settings.StorageConnectionString ?? this.settings.StorageAccountDetails.ConnectionString;
             if (string.IsNullOrEmpty(connectionString))
             {
                 throw new InvalidOperationException("A connection string is required to use the table partition manager. Managed identity is not yet supported when using the table partition manager. Please provide a connection string to your storage account in your app settings.");
             }
+
             this.options = new LeaseCollectionBalancerOptions
             {
                 AcquireInterval = this.settings.LeaseAcquireInterval,
                 LeaseInterval = this.settings.LeaseInterval,
                 ShouldStealLeases = true
             };
-            this.partitionManagerCancellationSource = new CancellationTokenSource();
+            this.gracefulShutdownToken = new CancellationTokenSource();
+            this.forcefulShutdownToken = new CancellationTokenSource();
             this.partitionTable = new TableClient(connectionString, this.settings.PartitionTableName);
             this.tableLeaseManager = new TableLeaseManager(this.partitionTable, this.service, this.settings, this.storageAccountName, this.options);
             this.partitionManagerTask = Task.CompletedTask;
@@ -77,7 +88,10 @@ namespace DurableTask.AzureStorage.Partitioning
         /// </summary>
         Task IPartitionManager.StartAsync()
         {
-            this.partitionManagerTask = Task.Run(() => this.PartitionManagerLoop(this.partitionManagerCancellationSource.Token));
+            // Run the partition manager loop in the background
+            this.partitionManagerTask = this.PartitionManagerLoop(
+                this.gracefulShutdownToken.Token,
+                this.forcefulShutdownToken.Token);
             this.settings.Logger.PartitionManagerInfo(
                 this.storageAccountName,
                 this.settings.TaskHubName,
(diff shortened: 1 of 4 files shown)


nytiannn:
[like]  Naiyuan Tian reacted to your message:
________________________________
From: Chris Gillum ***@***.***>
Sent: Wednesday, July 5, 2023 5:41:22 PM
To: Azure/durabletask ***@***.***>
Cc: Naiyuan Tian ***@***.***>; Review requested ***@***.***>
Subject: [Azure/durabletask] More suggested edits (PR #922)


Here are a few more suggested updates based on my review of the code. They are mostly small things related to naming and comments.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/Azure/durabletask/pull/922

Commit Summary

  *   befb693<https://github.com/Azure/durabletask/pull/922/commits/befb6932f6914c61ecea3eb020b24b7b25d89948> More suggested edits

File Changes

(4 files<https://github.com/Azure/durabletask/pull/922/files>)

  *   M src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs<https://github.com/Azure/durabletask/pull/922/files#diff-d3dd829fbd3263bc8bc50ca289389c1ac1b9c671975a10cdf1a5c36059ba6cc5> (4)
  *   M src/DurableTask.AzureStorage/Partitioning/LeaseCollectionBalancerOptions.cs<https://github.com/Azure/durabletask/pull/922/files#diff-8f9abc802065646c6a8d81b0d3e8d7be9d043fc9899967622002a71df971afec> (14)
  *   M src/DurableTask.AzureStorage/Partitioning/TableLease.cs<https://github.com/Azure/durabletask/pull/922/files#diff-b98f982f76183d32af822501e13832c66abc2a9e8d92190887d28ab33d749d58> (8)
  *   M src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs<https://github.com/Azure/durabletask/pull/922/files#diff-39098a4550ee86c62805d3045d0e83f53b60cb9ab4c0c30cd7a2e45623f8cdd0> (57)

Patch Links:

  *   https://github.com/Azure/durabletask/pull/922.patch
  *   https://github.com/Azure/durabletask/pull/922.diff

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -196,40 +215,55 @@ async Task PartitionManagerLoop(CancellationToken cancellationToken)
...
                 this.settings.TaskHubName,
                 this.settings.WorkerId,
                 partitionId: NotApplicable,
                 "Started draining the in-memory messages of all owned control queues for shutdown.");
 
+            // Wait 5 minutes for the partition manager to shutdown gracefully. Otherwise force a shutdown.
+            var timeout = TimeSpan.FromMinutes(5);
+            var timeoutTask = Task.Delay(Timeout.Infinite, this.forcefulShutdownToken.Token);
+            this.forcefulShutdownToken.CancelAfter(timeout);
+            await Task.WhenAny(this.partitionManagerTask, timeoutTask);
```
cgillum, line 234: This additional logic is how we ensure that we don't want any longer than N minutes for the shutdown to complete.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -294,13 +328,15 @@ sealed class TableLeaseManager
...
+            /// <param name="forcefulShutdownToken">Cancellation of this token forcefully aborts the partition manager loop.</param>
             /// <returns> The <see cref="ReadTableReponse"/> incidates whether the worker is waiting to claim a stolen lease from other workers or working on releasing any ownership leases.</returns>
             /// <exception cref="RequestFailedException">will be thrown if failed to update the partition table. Partition Manager loop will catch it and re-read the table to get the latest information.</exception>
-            public async Task<ReadTableReponse> ReadAndWriteTable(CancellationToken cancellationToken)
+            public async Task<ReadTableReponse> ReadAndWriteTableAsync(bool isShuttingDown, CancellationToken forcefulShutdownToken)
             {
                 var response = new ReadTableReponse();
-                Pageable<TableLease> partitions = this.partitionTable.Query<TableLease>();
+
+                IReadOnlyList<TableLease> partitions = this.partitionTable.Query<TableLease>(cancellationToken: forcefulShutdownToken).ToList();
```
cgillum, line 343: Two important things here:
1. Added a cancellation token to ensure we don't get stuck here if storage is slow
2. We call `.ToList()` on the result to make sure that the results are evaluated immediately. If we don't do this, we might have redundant iteration of the result list depending on the internal implementation details of `Pageable<T>`. This is a common and sometimes severe problem that many developers run into when they don't understand how the underlying collections received from various SDKs work. Calling `.ToList()` ensures that we're always dealing with an in-memory list.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -317,37 +353,49 @@ public async Task<ReadTableReponse> ReadAndWriteTable(CancellationToken cancella
...
-                        await CheckOwnershipLease(partition, response);
+                        claimedLease = this.TryClaimLease(partition);
+
+                        this.CheckOtherWorkersLeases(
+                            partition,
+                            partitionDistribution,
+                            response,
+                            ref ownershipLeaseCount,
+                            ref previousOwner,
+                            ref stoleLease);
```
cgillum, line 374: The previous use of local methods was making it really hard to review the code because:
1. They were making the body of `ReadAndWriteTable` really large and
2. They were modifying externally defined variables in a way that is hard to detect and verify

I switched to private instance methods with `ref` parameters to eliminate the complexity mentioned above.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -358,185 +406,219 @@ public async Task<ReadTableReponse> ReadAndWriteTable(CancellationToken cancella
...
-                        }
-                        // As long as the lease is not released, keep renewing the lease in case lease expired since it's still draining. 
-                        if (partition.CurrentOwner != null)
-                        {
-                            this.RenewLease(partition);
-                            renewedLease = true;
+                            this.backgroundDrainTasks.Remove(partition.RowKey!);
+                            this.ReleaseLease(partition);
+                            releasedLease = true;
+                            drainTask.GetAwaiter().GetResult(); // surface any exceptions that may have occurred during the drain
```
cgillum, line 504: I'm doing this instead of `await drainTask` so that I could remove `async` from the method signature, which was a bit misleading. We didn't need the method to be async since we don't actually block and wait for drain tasks to complete. This small change should help simplify the code and make the intent more obvious.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -358,185 +406,219 @@ public async Task<ReadTableReponse> ReadAndWriteTable(CancellationToken cancella
...
+                ref bool releasedLease,
+                ref bool drainedLease,
+                ref bool renewedLease)
+            {
+                response.IsDrainingPartition = true;
 
+                if (partition.CurrentOwner != this.workerName)
+                {
+                    // If the lease is not owned by this worker, we don't need to drain it.
+                    return;
```
cgillum, line 587: To reduce code complexity, I'm exiting early if the current worker is not the partition owner. It should be equivalent to the previous implementation, but please confirm that it matches your intent.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -358,185 +406,219 @@ public async Task<ReadTableReponse> ReadAndWriteTable(CancellationToken cancella
...
                 }
-                await this.BalanceLeases(partitionDistribution, partitions, ownershipLeaseCount, response);
 
-                // If shutdown is requested and the worker releases all ownership leases, then set the response.IsReleasesAllLease to true to notify the partitionManagerLoop to stop.
-                if (cancellationToken.IsCancellationRequested)
+                ownershipLeaseCount++;
+
+                // NOTE: This logic is largely redundant with logic in RenewOrReleaseMyLease. Changes to one
+                //       method may need to be copied to the other method.
+                if (this.backgroundDrainTasks.TryGetValue(partition.RowKey!, out Task? drainTask))
```
cgillum, line 594: The logic here is redundant as I noted in the comments. However, I wanted to avoid trying to consolidate it since there are some very subtle differences that might make it easy for me to accidentally introduce bugs.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -548,43 +630,47 @@ async Task TryDrainAndReleaseAllPartitions(TableLease partition, ReadTableRepons
...
                 {
-                    // Already have enough leases, and thus no need to steal, just stop the balancing.
+                    // Partitions are already balanced across all workers.
                     return;
                 }
 
-                //Check all the other active workers when my ownership lease number is not larger than the average number.
-                foreach (KeyValuePair<string, List<TableLease>> pair in partitionDistribution)
+                // Check all the other active workers when my ownership lease number is not larger than the average number.
+                foreach (IReadOnlyList<TableLease> ownedPartitions in partitionDistribution.Values)
```
cgillum, line 658: We were never accessing the key, so I simplified this code to loop over the values directly using easy-to-understand variable names.

cgillum (commented):
Adding some comments for additional context.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -548,43 +634,47 @@ async Task TryDrainAndReleaseAllPartitions(TableLease partition, ReadTableRepons
...
+                    // No partitions to be balanced.
                     return;
                 }
 
-                int averageLeasesCount = (partitions.Count()) / (partitionDistribution.Count + 1);
+                int averageLeasesCount = partitions.Count / (partitionDistribution.Count + 1);
                 if (averageLeasesCount < ownershipLeaseCount)
                 {
-                    // Already have enough leases, and thus no need to steal, just stop the balancing.
+                    // Partitions are already balanced across all workers.
```
nytiannn, line 653: I would update this commit with the original one. Cause the condition here is: If this worker owns more leases than the average count, then it shouldn't steal any other partitions because it already have enough.  But it doesn't indicate the partition distribution among all workers is balanced.  I think the new commit will be a little misleading?

cgillum: You're right. My suggested comment isn't accurate - it's more that the current worker is "balanced" (using the term loosely), though others may not be. Feel free to change it back.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -125,51 +139,60 @@ async Task PartitionManagerLoop(CancellationToken cancellationToken)
...
+
+                try
                 {
-                    try
+                    if (isShuttingDown || forcefulShutdownToken.IsCancellationRequested)
                     {
-                        // Introduce a token to enable immediate shutdown initiation for the worker, even if it's at rest when receiving the shutdown request.
-                        await Task.Delay(timeToSleep, cancellationToken);
+                        // If shutdown is required, we sleep for a short period to ensure a relatively fast shutdown process
+                        await Task.Delay(TimeSpan.FromSeconds(1), forcefulShutdownToken);
```
nytiannn, line 182: So here we don't have the old design that if a update failed due to concurrency, re-read the table immediately to get the latest E-tag. Do you think we need to add it? 

I am not quite sure. But I think 1 sec is already frequent enough.

cgillum: Oh, I actually hadn't considered this - so this change is a mistake on my part. With my change we always do 1 second instead of doing 0 seconds after a failure, if I understand your concern correctly. I'm okay with changing this back to `timeToSleep` to preserve the original intended behavior.

PR #947, opened 2023-08-27
Merge with Main and Remove SimpleBufferManager

wsugarman:
This PR merges with the `main` branch some changes that have been added since the last time. It also removes the existing `SimpleBufferManager` and replaces it with a more standard call pattern.

See [these commits](https://github.com/Azure/durabletask/pull/947/files/4618b4ee32e778f31d189cd0a457f64b5f3e275f..eccc9f62aa2c9ecddacc130cbe76603f8ffdd7fb) for more details around what is actually changed.


src/DurableTask.AzureStorage/Storage/Blob.cs
```diff
@@ -90,9 +89,21 @@ public async Task UploadFromByteArrayAsync(byte[] buffer, int index, int byteCou
             await this.blockBlobClient.UploadAsync(stream, cancellationToken: cancellationToken).DecorateFailure();
         }
 
+        public async Task<Stream> OpenWriteAsync(CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                return await this.blockBlobClient.OpenWriteAsync(overwrite: true, cancellationToken: cancellationToken);
```
wsugarman, line 96: This method doesn't return `Response<T>` and I'm reluctant to make an extension method for all `Task<T>` that performs this `try`/`catch`.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -175,39 +176,25 @@ internal MessageData DeserializeMessageData(string json)
...
-                        }
-                        finally
-                        {
-                            SimpleBufferManager.Shared.ReturnBuffer(buffer);
-                        }
-                    }
-
-                    return new ArraySegment<byte>(memory.GetBuffer(), 0, (int)memory.Length);
-                }
+                // TODO: Change default pipe options if necessary
```
wsugarman, line 190: Oops! I previously used pipes, but that wasn't necessary. Still -- my comment survived. I'll remove.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -175,39 +176,25 @@ internal MessageData DeserializeMessageData(string json)
...
+                await payloadStream.CopyToAsync(compressedStream, bufferSize: 81920, cancellationToken: cancellationToken);
+                await compressedStream.FlushAsync(cancellationToken);
+                await blobStream.FlushAsync(cancellationToken);
+            }
+            catch (RequestFailedException rfe)
```
wsugarman, line 195: I still wrap in a `try`/`catch` in case of stream failures while writing to the internal `BlockBlobWriteStream`

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -24,9 +24,9 @@
   <!-- Common package dependencies -->
   <ItemGroup>
     <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.12.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
-    <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
-    <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.1" />
```
wsugarman, line 27: These libraries needed to be updated to run in the latest Visual Studio version. While that is probably a VS bug, it is a good reminder to update these.

nytian (approved):
Thanks for the PR!

test/DurableTask.AzureServiceFabric.Integration.Tests/DurableTask.AzureServiceFabric.Integration.Tests.csproj
```diff
@@ -8,9 +8,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
-    <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
-    <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.1" />
+    <PackageReference Include="MSTest.TestAdapter" Version="3.1.1" />
+    <PackageReference Include="MSTest.TestFramework" Version="3.1.1" />
```
davidmrdavid, line 13: Is it 'ok' for us to change these bits in the ServiceFabric backend, @cgillum?

src/DurableTask.AzureStorage/Storage/Blob.cs
davidmrdavid: Hmm, not sure I follow the significance of this comment. Can you please elaborate, @wsugarman? How does this affect callers of this method?

davidmrdavid (commented):
Just a few small questions.

test/DurableTask.AzureServiceFabric.Integration.Tests/DurableTask.AzureServiceFabric.Integration.Tests.csproj
cgillum: I assumed Will's comment further below explained why he made this change? I'm personally okay with it. @shankarsama can maybe comment if this might be a problem for development of the Service Fabric backend?

cgillum (approved):
I'm good with these changes.

Just so that I fully understand the implications, does this new method of compressing and decompressing remove the need for `SimpleBufferManager` or are we just removing it because it might be causing more problems than it solves? I guess the answer to that depends on what `blob.CopyToAsync` is doing internally? I don't mind either way, but curious to know if this changes the memory usage profile in any way.

src/DurableTask.AzureStorage/Storage/Blob.cs
wsugarman: I just meant that typically in this changeset I have been using a new extension method called `DecorateFailure()` to catch `RequestFailedException` and return `DurableTaskStorageException`. However, those extension methods are usually for type `Task<Response<T>>`, and here the method `OpenWriteAsync` method simply returns `Task<Stream>`. So instead of writing a new extension method for `Task<T>` or specifically `Task<Stream>`, I in-lined what would be the extension method here.

wsugarman:
> I'm good with these changes.
> 
> Just so that I fully understand the implications, does this new method of compressing and decompressing remove the need for `SimpleBufferManager` or are we just removing it because it might be causing more problems than it solves? I guess the answer to that depends on what `blob.CopyToAsync` is doing internally? I don't mind either way, but curious to know if this changes the memory usage profile in any way.

I mostly removed it to simplify the logic, but it does change the allocations a bit. However, I think we're pretty well off at the end of the day. For example in `CompressAndUploadAsBytesAsync`, previously we would allocate an array (as part of `MemoryStream`) to store the entire compressed message blob while using a smaller pooled buffer to incrementally pass data from the source `Stream` to the `GZipStream`. However, now we stream the compressed bytes from the source `Stream` directly to the `GZipStream` that is wrapping a network stream (technically a [`BlockBlobWriteStream`](https://github.com/Azure/azure-sdk-for-net/blob/613bba76f42c24062139c19fd4382ac87cd40388/sdk/storage/Azure.Storage.Blobs/src/BlockBlobWriteStream.cs#L15)). Internally. Inside of the [`MemoryStream.CopyToAsync`](https://github.com/dotnet/runtime/blob/v6.0.22/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L483), we're really just calling `GzipStream.WriteAsync` for the entire `MemoryStream` buffer.

Likewise for `DownloadAndDecompressAsBytesAsync`, previously we would allocate a large array via `MemoryStream` to hold the contents of the blob and incrementally copy decompressed bytes from the `GZipStream` wrapping the blob's stream to the destination `MemoryStream` using a small pooled buffer. Then that array would be converted into a UTF-8 `string`. However, now we skip the middle-man and directly use [`StreamReader.ReadToEndAsync`](https://github.com/dotnet/runtime/blob/4bb6dc195c0a3bc4c7e24ff54a8925b98db4fecd/src/libraries/System.Private.CoreLib/src/System/IO/StreamReader.cs#L949) to write the bytes into a `StringBuilder` and in turn a `string`. Internally, this method does something similar to the logic before by incrementally filling up its own buffer (whose size can be specified in the `StreamReader` ctor) and writing the bytes to the `StringBuilder`. While `StringBuilder` looks to be able to [quickly create the string with an efficient `memmove`](https://github.com/dotnet/runtime/blob/4bb6dc195c0a3bc4c7e24ff54a8925b98db4fecd/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs#L372), it still is copying the bytes.

So, in actuality, `DownloadAndDecompressAsBytesAsync` is probably less memory efficient as it no longer pools the intermediate buffer -- although it may be slightly faster. But we do have big memory efficiency gains in `CompressAndUploadAsBytesAsync` by streaming the entire operation.

test/DurableTask.AzureServiceFabric.Integration.Tests/DurableTask.AzureServiceFabric.Integration.Tests.csproj
wsugarman: Yeah, this was related to what I assumed was a VS bug where I couldn't run the tests properly on this older version of MSTest. I wouldn't think this would have an impact specific to Service Fabric.

PR #949, opened 2023-09-05
Log warning on multiple execution completed events and allow termination to override previous execution completed events

davidmrdavid:
Replaces: https://github.com/Azure/durabletask/pull/901/files
From this feedback: https://github.com/Azure/durabletask/pull/901/files#r1316325503

Currently, we throw an exception if we receive multiple `ExecutionCompletedEvent`s. That behavior is a bit too strict because it is possible for multiple `ExecutionCompletedEvent`s to be received at once. For example: when a termination signal arrives at the same time as an orchestrator is requested to continue-as-new.

This PR introduces two changes to this behavior:
(1) Instead of throwing an exception like we do today, we now log a warning notifying the reader that we've found multiple execution completed events and how DTFx is resolving this conflict.

(2) If orchestrator is ~not in the terminated state and the a new event aims to terminate it, the new event overrides the previously encountered `ExecutionCompletedEvent`~ planning to continue-as-new but a termination request is received, we transition to the "Terminated" state; this is because termination should be forceful. In all other cases, we discard the new execution completed event.

### DIFF
--- a/src/DurableTask.Core/Logging/EventIds.cs
+++ b/src/DurableTask.Core/Logging/EventIds.cs
@@ -63,5 +63,7 @@ namespace DurableTask.Core.Logging
         public const int RenewOrchestrationWorkItemStarting = 70;
         public const int RenewOrchestrationWorkItemCompleted = 71;
         public const int RenewOrchestrationWorkItemFailed = 72;
+
+        public const int OrchestrationDebugTrace = 73;
     }
 }
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -262,12 +262,17 @@ namespace DurableTask.Core
             {
                 if (ExecutionCompletedEvent != null)
                 {
-                    throw new InvalidOperationException(
-                        "Multiple ExecutionCompletedEvent found, potential corruption in state storage");
+                    TraceHelper.Trace(TraceEventType.Warning,
+                        "OrchestrationRuntimeState-DuplicateEvent",
+                        "The orchestration '{0}' already received an 'ExecutionCompletedEvent' with id {1} but is now receiving a new one with id {2}. The latter event will be discarded.",
+                        this.OrchestrationInstance?.InstanceId ?? "",
+                        ExecutionCompletedEvent.EventId,
+                        completedEvent.EventId);
                 }
-
-                ExecutionCompletedEvent = completedEvent;
-                orchestrationStatus = completedEvent.OrchestrationStatus;
+                else { 
+                    ExecutionCompletedEvent = completedEvent;
+                    orchestrationStatus = completedEvent.OrchestrationStatus;
+                }
             }
             else if (historyEvent is ExecutionSuspendedEvent)
             {
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -262,17 +262,27 @@ namespace DurableTask.Core
             {
                 if (ExecutionCompletedEvent != null)
                 {
-                    TraceHelper.Trace(TraceEventType.Warning,
-                        "OrchestrationRuntimeState-DuplicateEvent",
-                        "The orchestration '{0}' already received an 'ExecutionCompletedEvent' with id {1} but is now receiving a new one with id {2}. The latter event will be discarded.",
-                        this.OrchestrationInstance?.InstanceId ?? "",
-                        ExecutionCompletedEvent.EventId,
-                        completedEvent.EventId);
+                    // It's not generally expected to receive multiple execution completed events for a given orchestrator, but it's possible under certain race conditions.
+                    // For example: when an orchestrator is signaled to terminate at the same time as it attempts to continue-as-new.
+
+                    var log = $"The orchestration '{this.OrchestrationInstance?.InstanceId ?? ""}' " +
+                        $"had already received an 'ExecutionCompletedEvent' with EventId={ExecutionCompletedEvent.EventId} " +
+                        $"and orchestrationStatus={orchestrationStatus} but it now receiving a new one with id={completedEvent.EventId} and orchestrationStatus={completedEvent.OrchestrationStatus}. ";
+                    
+                    if (orchestrationStatus != OrchestrationStatus.Terminated && completedEvent.OrchestrationStatus == OrchestrationStatus.Terminated)
+                    {
+                        // If the new event requests termination, we override the pre-existing execution completed event. This is because termination should be considered to be forceful.
+                        log += "Discarding previous 'ExecutionCompletedEvent' as termination is forceful.";
+                    }
+                    else
+                    {
+                        log += "Discarding new 'ExecutionCompletedEvent'.";
+                    }
+
+                    TraceHelper.Trace(TraceEventType.Warning, "OrchestrationRuntimeState-DuplicateEvent", log);
                 }
-                else { 
-                    ExecutionCompletedEvent = completedEvent;
-                    orchestrationStatus = completedEvent.OrchestrationStatus;
-                }
+                ExecutionCompletedEvent = completedEvent;
+                orchestrationStatus = completedEvent.OrchestrationStatus;
             }
             else if (historyEvent is ExecutionSuspendedEvent)
             {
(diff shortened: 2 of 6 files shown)


src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -262,12 +262,28 @@ void SetMarkerEvents(HistoryEvent historyEvent)
...
+                        ExecutionCompletedEvent = completedEvent;
+                        orchestrationStatus = completedEvent.OrchestrationStatus;
+                    }
+                    else
+                    {
+                        // otherwise, we ignore the new event.
+                        log += "Discarding new 'ExecutionCompletedEvent'.";
+                    }
+
+                    TraceHelper.Trace(TraceEventType.Warning, "OrchestrationRuntimeState-DuplicateEvent", log);
```
cgillum, line 284: This uses the old unstructured logging mechanism, which most users will never see. I think we need to instead find a way to use the `LogHelper` class here, perhaps by adding overloads to the methods that call into here which accept a `LogHelper` object.

Also, we'll want to be sure to use a log event that include an `InstanceId` field so that people filtering for events based on the instance ID can find it.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -262,12 +262,28 @@ void SetMarkerEvents(HistoryEvent historyEvent)
...
+                    // It's not generally expected to receive multiple execution completed events for a given orchestrator, but it's possible under certain race conditions.
+                    // For example: when an orchestrator is signaled to terminate at the same time as it attempts to continue-as-new.
+                    var log = $"The orchestration '{this.OrchestrationInstance?.InstanceId ?? ""}' " +
+                        $"had already received an 'ExecutionCompletedEvent' with EventId={ExecutionCompletedEvent.EventId} " +
+                        $"and orchestrationStatus={orchestrationStatus} but is now receiving a new one with id={completedEvent.EventId} and orchestrationStatus={completedEvent.OrchestrationStatus}. ";
+                    
+                    if (orchestrationStatus != OrchestrationStatus.Terminated && completedEvent.OrchestrationStatus == OrchestrationStatus.Terminated)
+                    {
+                        // If the new event requests termination, we override the pre-existing execution completed event. This is because termination should be considered to be forceful.
```
cgillum, line 273: I'm not sure this is the right thing to do. If an orchestration has already completed, then we should let it remain in the completed state. There's no reason to change it and I think it's harmless to drop the terminated event. The only case for replacing the status would be if it's in a `ContinuedAsNew` state and then received a terminate event.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -262,12 +262,28 @@ void SetMarkerEvents(HistoryEvent historyEvent)
             {
                 if (ExecutionCompletedEvent != null)
                 {
-                    throw new InvalidOperationException(
-                        "Multiple ExecutionCompletedEvent found, potential corruption in state storage");
+                    // It's not generally expected to receive multiple execution completed events for a given orchestrator, but it's possible under certain race conditions.
+                    // For example: when an orchestrator is signaled to terminate at the same time as it attempts to continue-as-new.
+                    var log = $"The orchestration '{this.OrchestrationInstance?.InstanceId ?? ""}' " +
```
cgillum, line 267: This log message feels a little too verbose. Can we simplify it to something like `$"Ignoring {completedEvent.GetType().Name} event since the orchestration is already in the {orchestrationStatus} state."`.

cgillum (changes requested):

src/DurableTask.Core/OrchestrationRuntimeState.cs
davidmrdavid: Fair enough, I've restricted this logic to only applies to the `ContinueAsNew` case.

src/DurableTask.Core/OrchestrationRuntimeState.cs
davidmrdavid: Hmm, I see what you mean. 
My only concern with this request is that the refactoring to bring `LogHelper` here isn't trivial from what I can tell (e.g I can't _just_ add a `LogHelper` parameter `OrchestrationRuntimeState.AddEvent()` because `AddEvent` is public whereas `LogHelper` is internal), and so I worry about growing the scope of this PR.

Would you be willing to track modernizing the logging in this codepath that as a separate work item, or is this a blocker for you? If it's a blocker, I'll go ahead and starting working on it, but first I'm trying to determine my options.

cgillum: A simple, if slightly hackly, solution would be to add an internal method or constructor overload that sets a private `LogHelper` field. if that field is non-null, then we can write the log message. It's probably not worth blocking on this, but at a minimum we should use a `TraceHelper` method that accepts an orchestration instance ID parameter. Otherwise, the logging is going to be pretty unhelpful.

src/DurableTask.Core/OrchestrationRuntimeState.cs
davidmrdavid: Sure, incorporated this in my next commit.

src/DurableTask.Core/Logging/EventIds.cs
```diff
@@ -63,5 +63,7 @@ static class EventIds
         public const int RenewOrchestrationWorkItemStarting = 70;
         public const int RenewOrchestrationWorkItemCompleted = 71;
         public const int RenewOrchestrationWorkItemFailed = 72;
+
+        public const int OrchestrationDebugTrace = 73;
```
davidmrdavid, line 67: It seemed to me that we were missing a general "debug trace" for logs like this one, so I decided to create one. It did not seem appropriate to create a very specific new category just for this log, so instead I created a more general one.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -262,12 +271,27 @@ void SetMarkerEvents(HistoryEvent historyEvent)
...
+                        ExecutionCompletedEvent = completedEvent;
+                        orchestrationStatus = completedEvent.OrchestrationStatus;
+                    }
+                    else
+                    {
+                        // otherwise, we ignore the new event.
+                        log += "Discarding new 'ExecutionCompletedEvent'.";
+                    }
+
+                    LogHelper?.OrchestrationDebugTrace(this.OrchestrationInstance?.InstanceId ?? "", this.OrchestrationInstance?.ExecutionId ?? "", log);
```
davidmrdavid, line 292: I did not find a pre-existing loghelper method that fit this case.

src/DurableTask.Core/OrchestrationRuntimeState.cs
davidmrdavid: I took your suggestion to create an internal method that sets a `LogHelper` field.

cgillum (approved):
:shipit:

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -269,11 +269,16 @@ void SetMarkerEvents(HistoryEvent historyEvent)
...
+                {
+                    ExecutionCompletedEvent = completedEvent;
+                    orchestrationStatus = completedEvent.OrchestrationStatus;
+                }
+                else
                 {
                     // It's not generally expected to receive multiple execution completed events for a given orchestrator, but it's possible under certain race conditions.
                     // For example: when an orchestrator is signaled to terminate at the same time as it attempts to continue-as-new.
-                    var log = $"Ignoring {completedEvent.GetType().Name} event since the orchestration is already in the {orchestrationStatus} state.";
+                    var log = $"Received new {completedEvent.GetType().Name} event despite the orchestration being already in the {orchestrationStatus} state.";
```
cgillum, line 281: I'm not a fan of this log change. You're saying we received an event but you're not saying what we're doing with it. The previous log made it clear that we're ignoring the new event.

davidmrdavid: @cgillum: I think there might be some confusion here. The log grows to include whether we decided to ignore the event later on. We don't always ignore the new event, as seen in the snippet below:

https://github.com/Azure/durabletask/blob/5a068aac2bdd2a6dc751be6e57ad5ae2fb939f37/src/DurableTask.Core/OrchestrationRuntimeState.cs#L287-L295

So it's a log message that's created in two parts. First we capture that we're in a situation where we have to make a decision of which ExecutionCompletedEvent to preserve, and we include which one we decide to preserve (the new one or the previous one).

cgillum: I see. In that case it's fine, but I think you need to add a space between the period in the first half of the log and the start of the next sentence.

davidmrdavid: addressed in latest commit: https://github.com/Azure/durabletask/pull/949/commits/c23d3cc18da93e241b94839a69bedfae52292dc1

superLionn:
Hi team, may I ask the official release time?

PR #950, opened 2023-09-06
Enable Tags to be surfaced via TaskHubClient

### LINKED ISSUE #840: DT.AzureStorage to support querying orchestration tags

twsouthwick:
This change adds a property to the Azure tracking table ("Tags") that will store the tags for an orchestration. If it is too large, it is offloaded to a blob rather than being stored as a property.

As part of this change, `OrchestrationInstanceStatus` from the table is no longer created with reflection; instead, properties are searched for and then set on the status DTO.

Fixes #840

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationInstanceStatus.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.AzureStorage
 {
     using System;
+    using System.Collections.Generic;
     using Microsoft.WindowsAzure.Storage.Table;
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


cgillum (commented):
Thanks for this contribution! Overall, the change looks good, and I think this is a feature that other users will appreciate!

PR #953, opened 2023-09-06
Rename OperationBatchRequest and OperationBatchResponse

sebastianburckhardt:
For better consistency across repositories and easier understanding, these are now named EntityBatchRequest and EntityBatchResponse.

### DIFF
--- a/src/DurableTask.Core/Entities/OperationFormat/EntityBatchRequest.cs
+++ b/src/DurableTask.Core/Entities/OperationFormat/EntityBatchRequest.cs
@@ -18,7 +18,7 @@ namespace DurableTask.Core.Entities.OperationFormat
     /// <summary>
     /// A request for execution of a batch of operations on an entity.
     /// </summary>
-    public class OperationBatchRequest
+    public class EntityBatchRequest
     {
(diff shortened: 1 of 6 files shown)


src/DurableTask.Core/Entities/OperationFormat/EntityBatchRequest.cs
```diff
@@ -18,7 +18,7 @@ namespace DurableTask.Core.Entities.OperationFormat
     /// <summary>
     /// A request for execution of a batch of operations on an entity.
     /// </summary>
-    public class OperationBatchRequest
+    public class EntityBatchRequest
```
davidmrdavid, line 21: changing these names will necessitate a change in the extension, right?

sebastianburckhardt: The new naming is already reflected in the PRs that depend on this.
So, I only have to change anything if you object to this name change!

sebastianburckhardt: Also, note that these classes were newly introduced in the feature branch. So, we don't have any released code that depends on it.

jviau (approved):

PR #954, opened 2023-09-06
Revise how entity batches are executed and handle failures

sebastianburckhardt:
There are two changes here:

1. Removes the unnecessary `EntityExecutionOptions` from `TaskEntity`
2. Revises failure handling of entity batches. The entire entity batch can now return a failure result. Previously, only individual operations could return failure results.


src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -246,6 +246,23 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                     {
                         // execute the user-defined operations on this entity, via the middleware
                         var result = await this.ExecuteViaMiddlewareAsync(workToDoNow, runtimeState.OrchestrationInstance, schedulerState.EntityState);
+                        var operationResults = result.Results!;
+
+                        // if we encountered an, record it as the result of the operations
+                        // so that callers are notified that the operation did not succeed.
```
davidmrdavid, line 252: did you mean this?
```suggestion
                        // if we encountered an error, record it as the result of the operations
                        // so that callers are notified that the operation did not succeed.
```

src/DurableTask.Core/Entities/OperationFormat/OperationBatchResult.cs
```diff
@@ -24,10 +24,9 @@ public class OperationBatchResult
...
         /// <summary>
-        /// The results of executing the operations in the batch. The length of this list must match
-        /// the size of the batch if all messages were processed; In particular, all execution errors must be reported as a result.
-        /// However, this list of results can be shorter than the list of operations if
-        /// some suffix of the operation list was skipped, e.g. due to shutdown, send throttling, or timeouts.  
+        /// The results of executing the operations in the batch. If there were (non-application-level) errors, the length of this list may
+        /// be shorter than the number of requests. In that case, <see cref="FailureDetails"/> contains the reason why not all requests 
+        /// were processed.
         /// </summary>
         public List<OperationResult>? Results { get; set; }
```
davidmrdavid, line 31: if there's less results and input operations, how do we know what results came from which operations? Or do we not need to associate them?

davidmrdavid (commented):
Some questions

src/DurableTask.Core/Entities/OperationFormat/OperationBatchResult.cs
sebastianburckhardt: We always execute operations in order, so the returned results correspond to a prefix of the operations.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -246,6 +246,23 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
+                        var operationResults = result.Results!;
+
+                        // if we encountered an error, record it as the result of the operations
+                        // so that callers are notified that the operation did not succeed.
+                        if (result.FailureDetails != default(FailureDetails))
+                        {
+                            OperationResult errorResult = new OperationResult()
+                            {
+                                Result = null,
+                                ErrorMessage = "entity dispatch failed",
```
jviau, line 258: I am still unsure how this is supposed to be used. Does the presence of `FailureDetails` not already indicate an error occurred? Is `ErrorMessage` offering any _new_ information?

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -246,6 +246,23 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
+                                FailureDetails = result.FailureDetails,
+                            };
+                            for (int i = operationResults.Count; i < workToDoNow.OperationCount; i++)
```
jviau, line 261: nit: 
```suggestion

                            for (int i = operationResults.Count; i < workToDoNow.OperationCount; i++)
```

jviau (approved):

src/DurableTask.Core/TaskEntityDispatcher.cs
sebastianburckhardt: I got the impression that FailureMessage is not sufficient as an indication because it is not used when using `ErrorPropagationMode.SerializeExceptions`. Also, this was designed to allow us to switch seamlessly from legacy DF in which case there are no FailureDetails present.

PR #955, opened 2023-09-06
Revise entity state and status format and access

sebastianburckhardt:
Three changes here:
- on the entity scheduler state: since we are maintaining the invariant (entity exists) <=> (serialized state is not null) there is no reason for defining (and serializing/deserializing) an `entityExists` field. Instead, just use a property.
- in client entity helpers, provide access to helper functions so external code can deserialize strings that represent `EntityStatus` and `SchedulerState` without us having to make the serialization format public.  This is needed by backends that provide entity query functionality directly. 
- also provide a shortcut (optimized) way to determine if an entity exists when looking at the entity status. Again, this is useful for backends implementing queries.

### DIFF
--- a/src/DurableTask.Core/Entities/StateFormat/SchedulerState.cs
+++ b/src/DurableTask.Core/Entities/StateFormat/SchedulerState.cs
@@ -24,11 +24,8 @@ namespace DurableTask.Core.Entities
     [DataContract]
     internal class SchedulerState
     {
-        /// <summary>
-        /// Whether this entity exists or not.
-        /// </summary>
-        [DataMember(Name = "exists", EmitDefaultValue = false)]
-        public bool EntityExists { get; set; }
+        [IgnoreDataMember]
+        public bool EntityExists => this.EntityState != null;
 
         /// <summary>
         /// The last serialized entity state.
--- a/src/DurableTask.Core/Entities/ClientEntityHelpers.cs
+++ b/src/DurableTask.Core/Entities/ClientEntityHelpers.cs
@@ -69,6 +69,7 @@ namespace DurableTask.Core.Entities
                 Id = "fix-orphaned-lock", // we don't know the original id but it does not matter
             };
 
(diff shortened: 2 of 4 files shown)


src/DurableTask.Core/Entities/StateFormat/SchedulerState.cs
```diff
@@ -24,11 +24,8 @@ namespace DurableTask.Core.Entities
     [DataContract]
     internal class SchedulerState
     {
-        /// <summary>
-        /// Whether this entity exists or not.
-        /// </summary>
-        [DataMember(Name = "exists", EmitDefaultValue = false)]
-        public bool EntityExists { get; set; }
+        [IgnoreDataMember]
+        public bool EntityExists => this.EntityState != null;
```
davidmrdavid, line 28: Is this change backwards compatible? Not immediately sure who's consuming this.

cgillum: This change should be backwards compatible since it should always (in theory) return the same value as before. But entities haven't been shipped/announced as a public feature of DTFx, so I assume we don't need to worry about backwards compatibility yet?

davidmrdavid (approved):

src/DurableTask.Core/Entities/StateFormat/SchedulerState.cs
davidmrdavid: > entities haven't been shipped/announced as a public feature of DTFx, 
ah, good call out

sebastianburckhardt: > so I assume we don't need to worry about backwards compatibility yet?

I actually did design the core-entities with compatibility in mind - I was planning to switch the old entity implementation in DF to the new implementation of entities in DT core, so non-isolated users can benefit from it as well. However, that is not currently our priority - for now we just care about getting Isolated working.

PR #957, opened 2023-09-06
revise entity backend properties and implement entity backend queries

sebastianburckhardt:
This PR does
1. simplify how features are supported via `IEntityOrchestrationService`. 
2. implements direct support for entity queries in the `AzureStorageOrchestrationService`.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -295,8 +295,7 @@ namespace DurableTask.AzureStorage
         EntityBackendQueries IEntityOrchestrationService.EntityBackendQueries
            => new EntityTrackingStoreQueries(
                 this.messageManager,
-                (this.trackingStore as AzureTableTrackingStore)
-                    ?? throw new NotSupportedException("entity queries not supported for custom tracking stores"),
+                this.trackingStore,
                 this.EnsureTaskHubAsync,
                 ((IEntityOrchestrationService)this).EntityBackendProperties,
                 this.SendTaskOrchestrationMessageAsync);
@@ -307,7 +306,7 @@ namespace DurableTask.AzureStorage
         {
             if (!this.settings.UseSeparateQueueForEntityWorkItems)
             {
-                throw new InvalidOperationException("backend was not configured for separate entity processing");
+                throw new InvalidOperationException("Internal configuration is inconsistent. Backend is using single queue for orchestration/entity dispatch, but frontend is pulling from individual queues.");
             }
             return this.LockNextTaskOrchestrationWorkItemAsync(false, cancellationToken);
         }
@@ -318,7 +317,7 @@ namespace DurableTask.AzureStorage
         {
             if (!this.settings.UseSeparateQueueForEntityWorkItems)
             {
-                throw new InvalidOperationException("backend was not configured for separate entity processing");
+                throw new InvalidOperationException("Internal configuration is inconsistent. Backend is using single queue for orchestration/entity dispatch, but frontend is pulling from individual queues.");
             }
             return this.LockNextTaskOrchestrationWorkItemAsync(entitiesOnly: true, cancellationToken);
         }
@@ -679,7 +678,7 @@ namespace DurableTask.AzureStorage
         {
             if (this.settings.UseSeparateQueueForEntityWorkItems)
             {
-                throw new InvalidOperationException("backend was configured for separate orchestration/entity processing, must use specialized methods to get work items");
+                throw new InvalidOperationException("Internal configuration is inconsistent. Backend is using separate queues for orchestration/entity dispatch, but frontend is pulling from single queue.");
             }
             return LockNextTaskOrchestrationWorkItemAsync(entitiesOnly: false, cancellationToken);
         }
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -281,7 +281,7 @@ namespace DurableTask.AzureStorage
 
         #region IEntityOrchestrationService
 
-        EntityBackendProperties IEntityOrchestrationService.GetEntityBackendProperties()
+        EntityBackendProperties IEntityOrchestrationService.EntityBackendProperties
            => new EntityBackendProperties()
            {
                EntityMessageReorderWindow = TimeSpan.FromMinutes(this.settings.EntityMessageReorderWindowInMinutes),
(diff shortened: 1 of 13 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -281,34 +281,31 @@ public BehaviorOnContinueAsNew EventBehaviourForContinueAsNew
...
-        EntityBackendProperties IEntityOrchestrationService.GetEntityBackendProperties()
+        EntityBackendProperties IEntityOrchestrationService.EntityBackendProperties
            => new EntityBackendProperties()
            {
                EntityMessageReorderWindow = TimeSpan.FromMinutes(this.settings.EntityMessageReorderWindowInMinutes),
                MaxEntityOperationBatchSize = this.settings.MaxEntityOperationBatchSize,
                MaxConcurrentTaskEntityWorkItems = this.settings.MaxConcurrentTaskEntityWorkItems,
                SupportsImplicitEntityDeletion = false, // not supported by this backend
                MaximumSignalDelayTime = TimeSpan.FromDays(6),
+               UseSeparateQueueForEntityWorkItems = this.settings.UseSeparateQueueForEntityWorkItems,
```
davidmrdavid, line 292: What is this "use separate queue" feature?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -281,34 +281,31 @@ public BehaviorOnContinueAsNew EventBehaviourForContinueAsNew
...
-            else
-            {
-                return false;
-            }
-        }
+        EntityBackendQueries IEntityOrchestrationService.EntityBackendQueries
+           => new EntityTrackingStoreQueries(
+                this.messageManager,
+                (this.trackingStore as AzureTableTrackingStore)
+                    ?? throw new NotSupportedException("entity queries not supported for custom tracking stores"),
```
davidmrdavid, line 299: nit:
```suggestion
                    ?? throw new NotSupportedException("entity queries are not supported for custom tracking stores"),
```

Also, why separate tracking stores not supported?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -680,6 +677,10 @@ static TaskHubInfo GetTaskHubInfo(string taskHub, int partitionCount)
             TimeSpan receiveTimeout,
             CancellationToken cancellationToken)
         {
+            if (this.settings.UseSeparateQueueForEntityWorkItems)
+            {
+                throw new InvalidOperationException("backend was configured for separate orchestration/entity processing, must use specialized methods to get work items");
+            }
```
davidmrdavid, line 683: This error message is not particularly customer actionable as far as I can tell. What can the customer do to deal with this error?

src/DurableTask.Core/Entities/EntityBackendProperties.cs
```diff
@@ -14,6 +14,7 @@
 namespace DurableTask.Core.Entities
 {
     using System;
+    using System.Threading;
```
davidmrdavid, line 17: do we really need this here?

src/DurableTask.Core/Entities/IEntityOrchestrationService.cs
```diff
@@ -23,18 +23,17 @@ namespace DurableTask.Core.Entities
     public interface IEntityOrchestrationService : IOrchestrationService
     {
         /// <summary>
-        /// The entity orchestration service.
+        /// Properties of the backend.
         /// </summary>
-        /// <returns>An object containing properties of the entity backend.</returns>
-        EntityBackendProperties GetEntityBackendProperties();
+        /// <returns>An object containing properties of the entity backend, or null if the backend does not natively support entities.</returns>
+        EntityBackendProperties? EntityBackendProperties { get; }
```
davidmrdavid, line 29: Is this about supporting entities in general, or entity **queries** in particular?

src/DurableTask.AzureStorage/EntityTrackingStoreQueries.cs
```diff
@@ -0,0 +1,258 @@
...
+            do
+            {
+                DurableStatusQueryResult result = await this.trackingStore.GetStateAsync(condition, filter.PageSize ?? 100, continuationToken, cancellation);
+                entityResult = await ConvertResultsAsync(result.OrchestrationState);
+                continuationToken = result.ContinuationToken;
+            }
+            while ( // continue query right away if the page is completely empty, but never in excess of 100ms
+                continuationToken != null
+                && entityResult.Count == 0
+                && stopwatch.ElapsedMilliseconds <= 100);
```
davidmrdavid, line 92: have we always had this 100ms limit?

davidmrdavid (commented):
A few questions from a quick pass

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: This means the backend maintains two separate queues for orchestration work items and entity work items. Among other things, this makes it possible to specify separate concurrency limits for the two.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: Accessing the internal methods of AzureTableTrackingStore is a bit more efficient. However, now that I look at it again, it seems not very hard to add a little bit of code to make it work with tracking stores also.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: If the customer sees this we messed up. So the message is really for us. I can modify the text to make this clearer.

src/DurableTask.Core/Entities/IEntityOrchestrationService.cs
sebastianburckhardt: It's about direct support for Durable.Task.Core entities in the backend. Relevant mostly for allowing separate dispatch. For out-of-proc v2 SDKs, this is a requirement for supporting entities. For the old SDKs, they can continue to work without this support (but we may allow them to take advantage).

src/DurableTask.AzureStorage/EntityTrackingStoreQueries.cs
sebastianburckhardt: yes, this was already there (but it was in the DF extension since that's where we implemented this query).

src/DurableTask.Core/Entities/EntityBackendProperties.cs
sebastianburckhardt: It seems needed by one of the references in the XML comments.

src/DurableTask.Core/Entities/IEntityOrchestrationService.cs
sebastianburckhardt: I have updated the comment to explain a bit more.

src/DurableTask.Core/Entities/EntityBackendQueries.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <param name="request">The request which describes what to clean.</param>
+        /// <param name="cancellation">The cancellation token to cancel the operation.</param>
+        /// <returns>A task that completes when the operation is finished.</returns>
+        public abstract Task<CleanEntityStorageResult> CleanEntityStorageAsync(
+            CleanEntityStorageRequest request = default, CancellationToken cancellation = default);
+
+        /// <summary>
+        /// Metadata about an entity, as returned by queries.
+        /// </summary>
+        public struct EntityMetadata
```
jviau, line 57: Recommend moving all these out of a nested class. I only use `public` nested class types very sparingly.

jviau (approved):

jviau:
Do you know why tests are failing? Is that going to be addressed afterwards?

src/DurableTask.Core/Entities/EntityBackendQueries.cs
sebastianburckhardt: I prefer the nesting since it is much easier to keep everything that belongs together in one place.

sebastianburckhardt:
I am going to take a look at the failing tests after I merge this (because it is difficult to maintain different branches for all the PRs in my local build).

PR #960, opened 2023-09-07
[WIP] feature/core-entities

sebastianburckhardt:
Feature branch for the new entity support in DT.Core and DT.AzureStorage.


src/DurableTask.AzureStorage/EntityTrackingStoreQueries.cs
```diff
@@ -0,0 +1,270 @@
...
+            this.trackingStore = trackingStore;
+            this.ensureTaskHub = ensureTaskHub;
+            this.properties = properties;
+            this.sendEvent = sendEvent;
+        }
+
+        public async override Task<EntityMetadata?> GetEntityAsync(
+            EntityId id, 
+            bool includeState = false, 
+            bool includeStateless = false,
```
jviau, line 55: Think this should change to `includeTransient` before GA

jviau (approved):

src/DurableTask.AzureStorage/EntityTrackingStoreQueries.cs
sebastianburckhardt: oh right

PR #965, opened 2023-09-11
revise operation result encoding and add more comments.

sebastianburckhardt:
Revises the operation result representation and adds more comment.

### DIFF
--- a/src/DurableTask.Core/TaskEntityDispatcher.cs
+++ b/src/DurableTask.Core/TaskEntityDispatcher.cs
@@ -250,12 +250,15 @@ namespace DurableTask.Core
 
                         // if we encountered an error, record it as the result of the operations
                         // so that callers are notified that the operation did not succeed.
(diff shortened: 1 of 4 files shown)


src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -250,12 +250,15 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
 
                         // if we encountered an error, record it as the result of the operations
                         // so that callers are notified that the operation did not succeed.
-                        if (result.FailureDetails != default(FailureDetails))
+                        if (result.FailureDetails != null)
```
davidmrdavid, line 253: isn't `default(FailureDetails)`  the same as `null`?

jviau: Yes, but you often use `null` for reference types for consistency.

jviau (approved):

PR #966, opened 2023-09-13
Add RawInput to workaround double-serialization issue

jviau:
This PR is to address a conflicting need in Durable Functions extension. We want to use `TaskHubClient` from within `LocalGrpcListener` so that we leverage all the logic there, especially the distributed tracing work. However, our input is already serialized and `TaskHubClient` would double serialize it. To work around this an "internal" API called `RawInput` has been added. When this type is used as the input for `TaskHubClient`, we will directly use `RawInput.Value` instead of serializing the input.

`RawInput` is "internal" in the sense that it is `public`, but in an `.Internal` namespace, with an `[Obsolete]` attribute, and xmldoc advising customers to use the type at their own risk. This is a pattern used in a few .NET libraries (efcore, webjobs, functions worker, etc) to address the internal-yet-public functionality.

### DIFF
--- a/src/DurableTask.Core/Serializing/DataConverter.cs
+++ b/src/DurableTask.Core/Serializing/DataConverter.cs
@@ -14,6 +14,7 @@
 namespace DurableTask.Core.Serializing
 {
     using System;
+    using DurableTask.Core.Serializing.Internal;
 
     /// <summary>
     /// Abstract class for serializing and deserializing data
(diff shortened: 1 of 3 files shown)


davidmrdavid (approved):
This makes sense to me. Tagging @cgillum for a final approval

cgillum (approved):
LGTM - I assume part of the reason why we're marking this `[Obsolete]` is because it's not really a fully fleshed-out feature (can't use `RawValue` in any other contexts currently)?

jviau:
> I assume part of the reason why we're marking this `[Obsolete]` is because it's not really a fully fleshed-out feature (can't use `RawValue` in any other contexts currently)?

Essentially - it is used to raise a flag to customers that while this type is public, we don't want them to use it. So, every usage of it comes with a pseudo "I agree" via suppressing the warning.

PR #969, opened 2023-09-20
Add private setter for TraceParent

bachuv:
This PR adds a private setter for `TraceParent` to allow it to get serialized. Without this addition, we weren't able to run Netherite apps with the distributed tracing changes because we were seeing this exception: `InvalidDataContractException: No set method for property 'TraceParent' in type 'DurableTask.Core.Tracing.DistributedTraceContext'.`

I'm open to other approaches that solve this issue instead of adding a private setter.

### DIFF
--- a/src/DurableTask.Core/Tracing/DistributedTraceContext.cs
+++ b/src/DurableTask.Core/Tracing/DistributedTraceContext.cs
@@ -44,7 +44,7 @@ namespace DurableTask.Core.Tracing
         /// The W3C traceparent data: https://www.w3.org/TR/trace-context/#traceparent-header
         /// </summary>
         [DataMember]
-        public string TraceParent { get; }
+        public string TraceParent { get; private set; }
 
         /// <summary>
         /// The optional W3C tracestate parameter: https://www.w3.org/TR/trace-context/#tracestate-header


src/DurableTask.Core/Tracing/DistributedTraceContext.cs
```diff
@@ -44,7 +44,7 @@ public DistributedTraceContext(string traceParent, string? traceState = null)
         /// The W3C traceparent data: https://www.w3.org/TR/trace-context/#traceparent-header
         /// </summary>
         [DataMember]
-        public string TraceParent { get; }
+        public string TraceParent { get; private set; }
```
jviau, line 47: Might as well just make this public, all other fields are public. Also can you verify if we need a parameterless constructor?

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
jviau, line 1: I noticed another change needed here:

Line 40 has special logic for truncating `TraceState`. We should move that logic into the setter for `TraceState`.

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
```diff
@@ -31,26 +31,35 @@ public class DistributedTraceContext
...
         /// <summary>
         /// The optional W3C tracestate parameter: https://www.w3.org/TR/trace-context/#tracestate-header
         /// </summary>
         [DataMember]
-        public string? TraceState { get; set; }
+        public string? TraceState
+        {
+            get
+            {
+                return this.TraceState;
```
jviau, line 51: You need a field `string? traceState;` and get/set that field.

What you have no is a stack overflow.

jviau (changes requested):

jviau (approved):

PR #972, opened 2023-09-21
Revise serialization of entitymessages

sebastianburckhardt:
Revises how to avoid double serialization, and exposure to application-defined serialization settings, for entity message serialization. This is a challenge because orchestrators that send messages to entities have to do so by calling `SendEvent`, which serializes the content using the application-defined serializer.

Previously, double serialization was avoided by serializing to `JObject`. But that is hacky and does not actually work with STJ.

Instead we now revise the `SendEvent` implementation in DurableTask.Core directly, so it can detect EntityMessageEvents and use the internally defined serialization mechanism.


src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -199,7 +199,17 @@ public override void SendEvent(OrchestrationInstance orchestrationInstance, stri
             }
 
             int id = this.idCounter++;
-            string serializedEventData = this.MessageDataConverter.Serialize(eventData);
+
+            string serializedEventData;
+            if (eventData is EntityMessageEvent entityMessageEvent)
+            {
+                serializedEventData = entityMessageEvent.AsSerializedString(); // bypass application-defined serializer
```
jviau, line 206: If you merge main, I have introduced a `RawInput` object - maybe you can leverage that instead of special casing this type.

sebastianburckhardt: That might work. I will give it a try.

jviau: `RawInput` is marked `Obsolete` on purpose - just suppress the message. It is done that way to have an "I acknowledge" before using it.

src/DurableTask.Core/Entities/EntityMessageEvent.cs
```diff
@@ -78,13 +68,24 @@ public TaskMessage AsTaskMessage()
...
                 }
             };
         }
 
+#pragma warning disable CS0618 // Type or member is obsolete. Intentional internal usage.
+        /// <summary>
+        /// Returns the content as an already-serialized string. Can be used to bypass the application-defined serializer.
+        /// </summary>
+        /// <returns></returns>
+        public RawInput AsRawInput()
```
jviau, line 83: Where is `AsRawInput()` expected to be called? By the dotnet isolated SDK shim layer?

jviau (approved):

src/DurableTask.Core/Entities/EntityMessageEvent.cs
sebastianburckhardt: Yes. This is the corresponding PR: https://github.com/microsoft/durabletask-dotnet/pull/207

PR #977, opened 2023-09-26
Update ActivitySource in OpenTelemetry sample

bachuv:
This PR updates the `ActivitySource` from `DurableTask` to `DurableTask.Core`.


cgillum:
I think we also need to update the README.md file in this directory.

cgillum (approved):

PR #981, opened 2023-10-04
Revise entity queries

sebastianburckhardt:
Several changes here:

- rename 'includeDeleted' property to `includeStateless` because it is more accurate (there are other reasons than having been deleted why an entity could be stateless). Added a comment on this to explain a bit more.

- added `BackLogQueueSize` and `LockedBy` properties to the entity metadata returned by queries. This information was already being stored in storage but not returned by queries.

### DIFF
--- a/src/DurableTask.Core/Entities/StateFormat/EntityStatus.cs
+++ b/src/DurableTask.Core/Entities/StateFormat/EntityStatus.cs
@@ -38,7 +38,7 @@ namespace DurableTask.Core.Entities
         }
 
         /// <summary>
-        /// Whether this entity exists or not.
+        /// Whether this entity currently has a user-defined state or not.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/DurableTask.Core/Entities/StateFormat/EntityStatus.cs
```diff
@@ -47,7 +47,7 @@ public static bool TestEntityExists(string serializedJson)
         /// The size of the queue, i.e. the number of operations that are waiting for the current operation to complete.
         /// </summary>
         [DataMember(Name = "queueSize", EmitDefaultValue = false)]
-        public int QueueSize { get; set; }
+        public int BacklogQueueSize { get; set; }
```
sebastianburckhardt, line 50: I kept the JSON member name the same as before so this remains compatible with existing storage layouts, allowing us to migrate non-isolated DF apps at some point.

jviau (approved):

PR #983, opened 2023-10-05
Rename DT.AS package version suffix for df v3 release 

nytian:
[Issue#2615](https://github.com/Azure/azure-functions-durable-extension/issues/2615), we will release the preview `DurableTask.AzureStorage` package named as `DurableTask.AzureStorage-rc`. 

This PR is for creating the official preview package that will be published in NuGet. So this PR don't need to be merged. We already have internal ADO package named as -preview. To distinguish between them we name the official preview package as rc(release candidate).


cgillum:
To be clear, this is not renaming the package. This is just adding a suffix to the _version number_, right @nytian? If so, can you please retitle this issue to avoid confusion?

nytian:
> To be clear, this is not renaming the package. This is just adding a suffix to the _version number_, right @nytian? If so, can you please retitle this issue to avoid confusion?

@cgillum This is to rename the package not adding suffix number. We renamed it as DTFx.AS-rc. Since we already have DTFx.AS-preview.5 in our ADO feeds. And I think it will be a little confused for customers if we released a public preview version named as `preview.5`. So, with David's suggestions, I renamed the new AS track2 pkg as DTFx.AS-rc.

cgillum:
> This is to rename the package not adding suffix number. We renamed it as DTFx.AS-rc for our public preview release. Since we already have DTFx.AS-preview.5 in our ADO feeds. And I think it will be a little confused for customers if we released a public preview version named as `preview.5`. So, with David's suggestions, I renamed the new AS track2 pkg as DTFx.AS-rc.

In that case, I recommend you not use `-rc` as a suffix because that's a convention which is only used for version numbers (which is why I misunderstood the intention of this PR). Instead, use the normal nuget package naming convention, like `DurableTask.AzureStorage.V3Preview`, or something like that.

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -24,7 +24,7 @@
     <PatchVersion>0</PatchVersion>
 
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix>preview.5</VersionSuffix>
+    <VersionSuffix>rc</VersionSuffix>
```
cgillum, line 27: This does not change the package name. This only changes the package version.

PR #995, opened 2023-10-13
Replace STJ package with Newtonsoft

nytian:
The System.Text.Json package refers to Systems.Buffer which is not supported in DurableFunctions V1 host. Replace this package with Newtonsoft.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/TagsSerializer.cs
+++ b/src/DurableTask.AzureStorage/Tracking/TagsSerializer.cs
@@ -12,16 +12,16 @@
 //  ----------------------------------------------------------------------------------
 
 using System.Collections.Generic;
-using System.Text.Json;
+using Newtonsoft.Json;
 
 namespace DurableTask.AzureStorage.Tracking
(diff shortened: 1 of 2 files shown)


nytian:
Please ignore the servicebus pipeline. That's the pipeline I added for df v3 tests. I canceled that one and then it becomes failure :(

src/DurableTask.AzureStorage/Tracking/TagsSerializer.cs
bachuv, line 1: Just to double check, did you validate this PR with the functions v1 smoke test?

nytian: Thanks for reminding! I will. but I would like to do after the CI passed.

cgillum (approved):
The change looks good to me. As Varshi already mentioned, let's validate using the Functions V1 smoke test before merging.

nytian:
Tested with smoketestsv1 and it works at my end.

PR #1016, opened 2023-12-13
Fix bug where input is not passed along when signaling an entity

sebastianburckhardt:
As reported in https://github.com/Azure/azure-functions-durable-extension/issues/2690, the `TaskEntityDispatcher` contains a bug where the input field is lost when creating a entity request message representing a signal.

### DIFF
--- a/src/DurableTask.Core/TaskEntityDispatcher.cs
+++ b/src/DurableTask.Core/TaskEntityDispatcher.cs
@@ -712,6 +712,7 @@ namespace DurableTask.Core
                 Id = Guid.NewGuid(),
                 IsSignal = true,
                 Operation = action.Name,
+                Input = action.Input,
                 ScheduledTime = action.ScheduledTime,
             };
             string eventName;


cgillum (approved):
Is there a test update we could do to help confirm that this code path is working?

thomaseyde:
When can we expect an updated NuGet package?

davidmrdavid:
Should be coming up in the next release, or the one after that (there's a release in progress, not sure it it'll pick this up), as it's just been merged!

thomaseyde:
Fix is not included in https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Extensions.DurableTask/1.1.1

Is there a different NuGet package I should reference?

Twiddeldum:
We have the same issue and we are waiting for nuget package containing the fix too.

davidmrdavid:
@thomaseyde, @Twiddeldum: seems a corresponding `Microsoft.Azure.Worker.Extension.DurableTask` package with this fix was not released. My apologies for the delay. We're starting another release of our packages this week, and the aformentioned package is in scope, so I expect it should be released with it. ETA: 10 days.

PR #1018, opened 2023-12-16
Fix InvalidCastException for history event type mismatch

cgillum:
While investigating https://github.com/microsoft/durabletask-mssql/issues/198, I noticed that some of the new distributed tracing code casts history events based on certain assumptions, which aren't always true in production. The issue can result in a poison message with the following error:

```
System.InvalidCastException: Unable to cast object of type 'DurableTask.Core.History.TaskCompletedEvent' to type 'DurableTask.Core.History.TaskScheduledEvent'.
at DurableTask.Core.TaskOrchestrationDispatcher.ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem, String dispatcher, ErrorPropagationMode errorPropagationMode, LogHelper logHelper) in //src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 860
at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in //src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 345
at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem)
at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem) in //src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 211
at DurableTask.Core.WorkItemDispatcher`1.ProcessWorkItemAsync(WorkItemDispatcherContext context, Object workItemObj) in //src/DurableTask.Core/WorkItemDispatcher.cs:line 373
```

This PR replaces the cast operation with a type-safe check, removing the possibility of `InvalidCastException`.

Note that the root cause of the bug observed by the user is probably specific to the MSSQL backend: it might not be deserializing history events correctly, resulting in unexpected Event ID matches. Regardless, I felt it made sense to make a fix here for defense in depth.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -842,14 +842,14 @@ namespace DurableTask.Core
                     }
                     else if (historyEvent is SubOrchestrationInstanceCompletedEvent subOrchestrationInstanceCompletedEvent)
                     {
-                        SubOrchestrationInstanceCreatedEvent subOrchestrationCreatedEvent = (SubOrchestrationInstanceCreatedEvent)workItem.OrchestrationRuntimeState.Events.FirstOrDefault(x => x.EventId == subOrchestrationInstanceCompletedEvent.TaskScheduledId);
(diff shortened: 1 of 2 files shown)


bachuv (approved):
Thanks for making this change!

PR #1020, opened 2023-12-21
Upgrade various dependencies with advisory warnings

davidmrdavid:
As of recently, building the DTFx project failed with errors of the following kind:

> "Warning as Error: Package <package name> has a known high severity vulnerability" and it points to this advisory: <link>"

The warnings were for:
* Newtonsoft.Json < 13.0.1, which linked to: https://github.com/advisories/GHSA-5crp-9r3c-p9vr
* System.Data.SqlClient < 4.8.5, which linked to: https://github.com/advisories/GHSA-8g2p-5pqh-5jmc

In this PR, I upgrade these dependencies to remove the warnings. In particular,
* We upgrade `System.Data.SqlClient` to `4.8.5`. This is only used in our test project, so it's safe.

* We upgrade `Newtonsoft.Json` to `13.0.1` across the board. This is potentially less safe as it's a major version upgrade, and it's used in of the projects contained within the `DurableTask.sln` solution.

While making this upgrade, I also noticed that we reference `Newtonsoft.Json` in almost every `.csproj` despite it being a transitive dependency. To minimize our management burden in the future, I removed this transitive dependency from those `.csproj`s.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -51,6 +51,7 @@ namespace DurableTask.AzureStorage
             this.settings = settings;
             this.azureStorageClient = azureStorageClient;
             this.blobContainer = this.azureStorageClient.GetBlobContainerReference(blobContainerName);
+#pragma warning disable CS0618 // Type or member is obsolete
             this.taskMessageSerializerSettings = new JsonSerializerSettings
             {
                 TypeNameHandling = TypeNameHandling.Objects,
@@ -60,6 +61,7 @@ namespace DurableTask.AzureStorage
                 Binder = new TypeNameSerializationBinder(settings.CustomMessageTypeBinder),
 #endif
             };
+#pragma warning restore CS0618 // Type or member is obsolete
             this.serializer = JsonSerializer.Create(taskMessageSerializerSettings);
 
             if (this.settings.UseDataContractSerialization)
--- a/src/DurableTask.Core/Common/Utils.cs
+++ b/src/DurableTask.Core/Common/Utils.cs
@@ -58,7 +58,7 @@ namespace DurableTask.Core.Common
 #if NETSTANDARD2_0
             SerializationBinder = new PackageUpgradeSerializationBinder()
 #else
-            Binder = new PackageUpgradeSerializationBinder()
+            SerializationBinder = new PackageUpgradeSerializationBinder()
 #endif
         };
         private static readonly JsonSerializer DefaultObjectJsonSerializer = JsonSerializer.Create(ObjectJsonSettings);
--- a/src/DurableTask.Core/DurableTask.Core.csproj
+++ b/src/DurableTask.Core/DurableTask.Core.csproj
@@ -32,7 +32,7 @@
   
   <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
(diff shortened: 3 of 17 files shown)


src/DurableTask.Core/Serializing/JsonDataConverter.cs
```diff
@@ -42,7 +42,7 @@ public JsonDataConverter()
 #if NETSTANDARD2_0
                 SerializationBinder = new PackageUpgradeSerializationBinder()
 #else
-                Binder = new PackageUpgradeSerializationBinder()
+                SerializationBinder = new PackageUpgradeSerializationBinder()
```
davidmrdavid, line 45: necessary change as part of the upgrade

src/DurableTask.Core/Common/Utils.cs
```diff
@@ -58,7 +58,7 @@ public static class Utils
 #if NETSTANDARD2_0
             SerializationBinder = new PackageUpgradeSerializationBinder()
 #else
-            Binder = new PackageUpgradeSerializationBinder()
+            SerializationBinder = new PackageUpgradeSerializationBinder()
```
davidmrdavid, line 61: `Binder` is obsolete after upgrading.

src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
```diff
@@ -29,7 +29,6 @@
     <PackageReference Include="Microsoft.Data.OData" version="5.8.5" />
     <PackageReference Include="Microsoft.Data.Services.Client" version="5.8.5" />
     <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" version="3.2.1" />
-    <PackageReference Include="Newtonsoft.Json" version="7.0.1" />
```
davidmrdavid, line 32: We should get sign-off from the ServiceBus owners about this change.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -60,6 +61,7 @@ class MessageManager
                 Binder = new TypeNameSerializationBinder(settings.CustomMessageTypeBinder),
 #endif
             };
+#pragma warning restore CS0618 // Type or member is obsolete
```
davidmrdavid, line 64: `Binder` is obsolete after upgrading our Newtonsoft dependency. However, I couldn't get `TypeNameSerializationBinder` assigned to `SerializationBinder` in the `#else` pragma for some strange reason. That's why I added the `#pragma` here.

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -32,15 +32,14 @@
...
-    <PackageReference Include="Newtonsoft.Json" Version="7.0.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
   </ItemGroup>
 
   <ItemGroup>
+	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
davidmrdavid, line 42: Not sure why the indentation is showing up like this, but on my end this seems correctly indented. Doesn't seem like a tabs vs spaces issue afaik either. I think it's the GH UI

davidmrdavid:
Replaced by: https://github.com/Azure/durabletask/pull/1023

PR #1024, opened 2024-01-05
Add orchestrator scheduled actions check when suspended

nytian:
This PR fixes issue [#2519](https://github.com/Azure/azure-functions-durable-extension/issues/2519)

When calling SuspendAsync URL, condition can happen that the orchestrator is suspended when there is already an action scheduled and added to the `orchestratorActionsMap`. However, our design will add any later event into a temporary queue during suspended time. Thus, this scheduled action will never be removed or completed and will be executed endless times until orchestrator is resumed. And this execution history will also be added to the History Table which will cause the non-deterministic error when orchestrator is resumed. 

The race condition occurs when an `ExecutionSuspended` message is processed after `TaskCompleted`. Additionally, there is a subsequent activity task pending scheduling after the completion of this task. During the processing`TaskCompleted`, the result is sent back to the extension level, and thus, durable extension will go to the next line and then schedule a new task, adds it to the action map which will cause the issue. 

For debug, we can verify if a task has been executed multiple times **by checking if column`TaskEventId` in Kusto or `TaskScheduledId` in History Table.** Each task should have a unique identifier in these fields.

The fix of this PR is to guarantee `orchestratorActionsMap` is empty when orchestrator is suspended.  Basically, we will check the `orchestratorActionsMap` when suspending an orchestrator. If `orchestratorActionsMap` is not null, the action will be saved to a temporary dictionary and will be added back to the `orchestratorActionsMap` when the orchestrator is resumed.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -17,6 +17,7 @@ namespace DurableTask.Core
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Globalization;
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
     using DurableTask.Core.Command;
@@ -35,6 +36,7 @@ namespace DurableTask.Core
         private bool executionCompletedOrTerminated;
         private int idCounter;
         private readonly Queue<HistoryEvent> eventsWhileSuspended;
+        private readonly IDictionary<int, OrchestratorAction> orchestratorActionsMapSuspended;
 
         public bool IsSuspended { get; private set; }
 
@@ -63,6 +65,7 @@ namespace DurableTask.Core
             this.EntityParameters = entityParameters;
             ErrorPropagationMode = errorPropagationMode;
             this.eventsWhileSuspended = new Queue<HistoryEvent>();
+            this.orchestratorActionsMapSuspended = new SortedDictionary<int, OrchestratorAction>();
         }
 
         public IEnumerable<OrchestratorAction> OrchestratorActions => this.orchestratorActionsMap.Values;
@@ -568,11 +571,30 @@ namespace DurableTask.Core
         public void HandleExecutionSuspendedEvent(ExecutionSuspendedEvent suspendedEvent)
         {
             this.IsSuspended = true;
+            
+            if (this.orchestratorActionsMap.Any())
+            {
+                foreach(var pair in this.orchestratorActionsMap)
+                {
+                    this.orchestratorActionsMapSuspended.Add(pair.Key, pair.Value);
+                }
+                this.orchestratorActionsMap.Clear();
+            }
         }
 
         public void HandleExecutionResumedEvent(ExecutionResumedEvent resumedEvent, Action<HistoryEvent> eventProcessor)
         {
             this.IsSuspended = false;
+            
+            if (this.orchestratorActionsMapSuspended.Any())
+            {
+                foreach(var pair in this.orchestratorActionsMapSuspended)
+                {
+                    this.orchestratorActionsMap.Add(pair.Key, pair.Value);
+                }
+                this.orchestratorActionsMapSuspended.Clear();
+            }
+
             while (eventsWhileSuspended.Count > 0)
             {
                 eventProcessor(eventsWhileSuspended.Dequeue());
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -36,7 +36,7 @@ namespace DurableTask.Core
         private bool executionCompletedOrTerminated;
         private int idCounter;
         private readonly Queue<HistoryEvent> eventsWhileSuspended;
-        private readonly IDictionary<int, OrchestratorAction> orchestratorActionsMapSuspended;
+        private readonly IDictionary<int, OrchestratorAction> suspendedActionsMap;
 
         public bool IsSuspended { get; private set; }
 
@@ -65,7 +65,7 @@ namespace DurableTask.Core
             this.EntityParameters = entityParameters;
             ErrorPropagationMode = errorPropagationMode;
             this.eventsWhileSuspended = new Queue<HistoryEvent>();
-            this.orchestratorActionsMapSuspended = new SortedDictionary<int, OrchestratorAction>();
+            this.suspendedActionsMap = new SortedDictionary<int, OrchestratorAction>();
         }
 
         public IEnumerable<OrchestratorAction> OrchestratorActions => this.orchestratorActionsMap.Values;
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -35,6 +36,7 @@ internal class TaskOrchestrationContext : OrchestrationContext
         private bool executionCompletedOrTerminated;
         private int idCounter;
         private readonly Queue<HistoryEvent> eventsWhileSuspended;
+        private readonly IDictionary<int, OrchestratorAction> orchestratorActionsMapSuspended;
```
davidmrdavid, line 39: nit - I think the word `orchestrator` here can be removed without affecting clarity, so how about we remove it and rename this to: `suspendedActionsMap`?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -568,11 +571,30 @@ public void HandleEventWhileSuspended(HistoryEvent historyEvent)
         public void HandleExecutionSuspendedEvent(ExecutionSuspendedEvent suspendedEvent)
         {
             this.IsSuspended = true;
+            
+            if (this.orchestratorActionsMap.Any())
```
davidmrdavid, line 575: can you please add a comment explaining why this is necessary

davidmrdavid (commented):
Some nits, but mostly looks good!

davidmrdavid:
I know there's a "perf tests" project [here](https://github.com/Azure/durabletask/tree/main/test/DurableTask.Stress.Tests) that I believe is meant to catch race conditions like the bug you're fixing. I wonder if you can re-use that infrastructure to set up a test for suspend/resume. Do you might giving that a look?

src/DurableTask.Core/TaskOrchestrationContext.cs
nytian: thanks for the naming nit!

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -571,12 +571,16 @@ public void HandleEventWhileSuspended(HistoryEvent historyEvent)
         public void HandleExecutionSuspendedEvent(ExecutionSuspendedEvent suspendedEvent)
         {
             this.IsSuspended = true;
-            
+
+            //When the orchestrator is suspended, a task could potentially be added to the orchestratorActionsMap.
+            //This could lead to the task being executed repeatedly without completion until the orchestrator is resumed.
+            //To prevent this scenario, check if orchestratorActionsMap is empty before proceeding.
```
davidmrdavid, line 577: stylistic nit: let's always add a space between "//" and the text :) . This can be automated with an IDE style linter, would love to invest on that eventually.

```suggestion

            // When the orchestrator is suspended, a task could potentially be added to the orchestratorActionsMap.
            // This could lead to the task being executed repeatedly without completion until the orchestrator is resumed.
            // To prevent this scenario, check if orchestratorActionsMap is empty before proceeding.
```

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -571,12 +571,16 @@ public void HandleEventWhileSuspended(HistoryEvent historyEvent)
...
             if (this.orchestratorActionsMap.Any())
             {
-                foreach(var pair in this.orchestratorActionsMap)
+                //If not, store its contents to a temporary dictionary to allow processing of the task later when orchestrator resumes.
```
davidmrdavid, line 580: nit - as above, let's add spaces between "//" and the text
```suggestion
                // If not, store its contents to a temporary dictionary to allow processing of the task later when orchestrator resumes.
```

davidmrdavid (commented):
@nytian: Do you have ideas for any telemetry that would have helped identify this bug faster? If so, I wonder if we could add that to this PR.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -568,11 +571,34 @@ public void HandleEventWhileSuspended(HistoryEvent historyEvent)
...
+            
+            if (this.suspendedActionsMap.Any())
+            {
+                foreach(var pair in this.suspendedActionsMap)
+                {
+                    this.orchestratorActionsMap.Add(pair.Key, pair.Value);
+                }
+                this.suspendedActionsMap.Clear();
+            }
+
```
bachuv, line 601: I agree with David's comment about adding telemetry. Maybe we could add a log saying how many actions were stored in the `suspendedActionsMap` and then how many were added back to the `orchestratorActionsMap`? Or a log that says that the `orchestratorActionsMap` was cleared when the orchestration was suspended.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -568,11 +571,34 @@ public void HandleEventWhileSuspended(HistoryEvent historyEvent)
...
+                }
+                this.orchestratorActionsMap.Clear();
+            }
         }
 
         public void HandleExecutionResumedEvent(ExecutionResumedEvent resumedEvent, Action<HistoryEvent> eventProcessor)
         {
             this.IsSuspended = false;
+            
+            if (this.suspendedActionsMap.Any())
```
bachuv, line 593: Can we add a comment here just saying that we're adding the events from before the orchestration was suspended back to the `orchestratorActionsMap`?

bachuv (commented):
Thanks for fixing this issue!

src/DurableTask.Core/TaskOrchestrationContext.cs
davidmrdavid: I'm thinking we may be able to come up with a more general log here (not specific to suspend/resume, but still able to identify this bug).

In this case, it seems to have been key to realize that we were re-scheduling the **same** task over and over and that this task was never marked as completed due to us buffering events while suspended.

I think this means that a task with the same TaskId was being scheduled infinitely. When we see this error on Azure today, do we log the TaskId being scheduled? If not - then maybe that's the missing piece

nytian: Thanks for the review, @bachuv ! 

We already have the action count log in column `ActionCount` in TaskName `orchestrationExecuted` provided from `DurableTask.Core` in [here](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L388) . 

I think we can use this one to check if the action count is 0 when orchestration is suspended.

nytian: As we talked about, @davidmrdavid , there it is one column named as `TaskEventId` in kusto logs which will be incremented with the sequence in orchestrator. We can use this column to check if a task has been executed repeatedly. Like the screenshot provided below. 


image

davidmrdavid (approved):
I sync'ed w/ Naiyuan and, after the fact, we determined that we could have identified this bug by paying attention to the `TaskScheduledId` column of our telemetry. So I don't strictly think we need new logs anymore, just better team-wide awareness to pay attention to that source of information. Still, please get @bachuv's approval as well before merging.

Finally, @nytian, you wrote the following:
> During the processingTaskCompleted, the result is sent back to the extension level, where SuspendAsync call is impossible to prevent the triggering of the next line. Thus, the extension will schedule a new activity and adds it to the action map which will cause the issue

Assuming I'm reading this correctly, I'm not this statement is 100% accurate. In particular, DTFx.Core and the Extension _run in the same thread_ so it's not quite right to say that "it's impossible [for DTFx] to prevent the triggering of the next line [of DF Extension user-code]." This is definitely possible :) . Let's discuss this offline later.

bachuv (approved):

PR #1025, opened 2024-01-09
Add Automatic suffix generation for DTFx.Core and DTFx.AzureStorage

davidmrdavid:
As in title: automation to generate preview packages of DTFx.Core and DTFx.AzureStorage on ADO. This is the DTFX equivalent of: https://github.com/Azure/azure-functions-durable-extension/pull/2709

This PR does two things:
(1) It makes the pipeline run automatically after every commit to the `main` branch
(2) On any automatic run, it appends a suffix to the generated package version.

In the long run, we can take this package version and automatically push it to our ADO feed so that it may be consumed by our pre-existing nightly DF Extension testing feed, so that we have nightly testing coverage across repos.


azure-pipelines-release.yml
```diff
@@ -1,3 +1,10 @@
+schedules:
+- cron: '0 5 * * *'
+  displayName: Daily 10pm (PDT) build
```
davidmrdavid, line 3: runs at 10pm because DF Extension nightly build is at midnight, so we have a 2 hour buffer.

azure-pipelines-release.yml
```diff
@@ -1,11 +1,22 @@
-trigger: none
+trigger:
+- main
+- dajusto/patch-nightly-build
```
davidmrdavid, line 3: for testing only, will be removed before merge

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -22,15 +22,21 @@
...
     <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevision)</FileVersion>
     <!-- The assembly version is only the major/minor pair, making it easier to do in-place upgrades -->
     <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
-    <!-- This version is used as the nuget package version -->
-    <Version>$(VersionPrefix)</Version>
+  </PropertyGroup>
+
+   <!-- This version is used as the nuget package version -->
+  <PropertyGroup Condition="$(VersionSuffix) == ''">
+    <Version>$(MajorVersion).$(MinorVersion).$(PatchVersion)</Version>
```
cgillum, line 36: Should we continue to use `$(VersionPrefix)` like we were before, or did you have a reason to remove it?

azure-pipelines-release.yml
```diff
@@ -1,11 +1,21 @@
...
 pr: none
 
 pool:
   name: '1ES-Hosted-DurableTaskFramework'
   demands:
     - ImageOverride -equals MMS2022TLS
 
+variables:
+
+  - name: PackageSuffix
```
cgillum, line 13: It seems `PackageSuffix` serves the same purpose as `VersionSuffix`. Was there a reason not to simply assign `VersionSuffix` directly as an alternative to creating a new build variable?

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
davidmrdavid: I've re-added it. I can't recall exactly why I removed it (a local version of this PR has been sitting on my devbox for a while).

azure-pipelines-release.yml
davidmrdavid: I've unified them in `VersionSuffix`. I think I had made them differently because I wasn't sure to what extend I could just continue reading env variables in the `.csproj`. So far, it seems like I can do that with no restrictions, so I've removed the `VersionSuffix` xml variable I introduced and replaced it with this env var.

davidmrdavid:
@cgillum: could you give this another pass?

azure-pipelines-release.yml
```diff
@@ -1,11 +1,21 @@
...
 
 pool:
   name: '1ES-Hosted-DurableTaskFramework'
   demands:
     - ImageOverride -equals MMS2022TLS
 
+variables:
+
+  - name: VersionSuffix
+    ${{ if eq(variables['Build.Reason'], 'IndividualCI') }}:
```
cgillum, line 14: In what cases will this expression be true? Can we add some comments explaining how this check works for the benefit of others? I'm not familiar with `Build.Reason` or how/why it's related to `Build.BuildNumber`.

davidmrdavid: Yeah you're right, it was a bit mysterious.
I added detailed comments in my latest commit, please let me know if they make sense.

cgillum: Thanks!

cgillum (approved):

PR #1028, opened 2024-01-12
Remove AzureServiceFabric build temporarily

nytian:
This PR don't need to be merged.

Remove    `src/DurableTask.AzureServiceFabric/DurableTask.AzureServiceFabric.csproj` from restore and build.  Since it includes `Newtonsoft.Json 7.0.1`  which will cause the release pipeline failed to restore. Remove this one temporarily to unblock the release process.


nytian:
Close this PR as finish building package.

PR #1047, opened 2024-02-27
Add checks for entity support

sebastianburckhardt:
As reported by microsoft/durabletask-dotnet#246, there is currently not always a descriptive error message when using a backend that does not support entities.

This PR adds checks to throw an exception if entity-features are used from within an orchestrator while using a backend that does not support entities. Such exceptions are thrown on the orchestrator thread, which should then cause the orchestrator to fail and log or otherwise propagate the error message.


cgillum (approved):
Much appreciated!

PR #1053, opened 2024-03-18
Increase DTFx.Core and DTFx.AS versions

davidmrdavid:
As in title, routine PR for upcoming release

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <MajorVersion>1</MajorVersion>
     <MinorVersion>17</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>2</PatchVersion>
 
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <MajorVersion>1</MajorVersion>
     <MinorVersion>17</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>2</PatchVersion>
```
cgillum, line 24: I recommend we exclude this change from the release since 1) it appears we didn't make any changes to DurableTask.AzureStorage and 2) the change to DurableTask.Core doesn't apply to DurableTask.AzureStorage (AFAICT, the entity-specific error handling only applies to backends like MSSQL that don't yet implement out-of-proc entity support).

davidmrdavid: yeah I felt similarly that this was a rather light release so it makes sense not to release DTFx.AS at all. Incorporated your feedback.

cgillum (approved):
:shipit:

Issue #785, opened 2022-08-17
Support metrics using System.Diagnostics.Metrics 


jviau:
With the release of .NET 6 last year, a new metrics API was introduced. This is available in System.Diagnostics.DiagnosticSource 6.0 package, which is backwards compatible with older .net runtimes.

We should use this API to emit metrics for select DTFx scenarios. Customers can then listen to these metrics themselves and export them out of process appropriately, or use an existing SDK like OpenTelemetry to export them.

We can start with building a list of metrics we want to collect, their names, value significance, and any dimensions.

cgillum:
One thing we've received asks for by customers is metrics for Azure Storage, like queue length. It's obviously specific to the DurableTask.AzureStorage backend but it would be useful if the different backends could add their own metrics as part of this work.

jviau:
# Metrics

| Name | Instrument Type |  Unit | Unit ([ucum](https://ucum.org/ucum.html)) | Description |
| - | - | - | - | - |
| `dtfx.task.limit` | Counter | default unit |  `{concurrent_task_limit}` | The configured limit of concurrent tasks for this worker. Attributes will define orchestration vs activity. |
| `dtfx.task.current` | Counter | default unit | `{conccurent_task_current}` | The current concurrent tasks (activity or orchestration) running on the worker. Attributes will define orchestration vs activity. |
| `dtfx.queue.delay` | Histogram | milliseconds | `ms` | Measures the delay of the messages as they are dequeued. |
| `dtfx.errors` | Counter | default unit | `{errors}` | Number of task invocation errors. |
| `dtfx.task.duration` | Histogram | milliseconds | `{ms}` | Measures the duration of a task. |

# Attributes

| Name | Requirement level | Notes and examples |
| - | - | - |
| `dtfx.task.type` | Required | The type of task being ran. SHOULD be one of: `activity`, `orchestration` |
| `dtfx.task.name` | Required | The name of the task being ran. Example `MyOrchestration` |
| `dtfx.task.version` | Optional | The version of the task being ran. Omitted when version is `null` |

jviau:
> One thing we've received asks for by customers is metrics for Azure Storage, like queue length. It's obviously specific to the DurableTask.AzureStorage backend but it would be useful if the different backends could add their own metrics as part of this work.

Yeah that is definitely important. But I do wonder if that is something DTFx should implement? Or should Azure Storage be responsible for that? I guess DTFx could add one for now, but have it opt-in only via some startup value.

[...]

PR #1058, opened 2024-04-03
[Proposal] Adding control-queue monitor with dedicated orchestrator to each control-queue.

pasaini-microsoft:
# Motivation: 
Issue: Unpredictable time to detect (TTD) for any orchestration being stuck in control-queue.
- We have been facing issues where DTF orchestration used to get stuck at random. Given that customer load is not very regular in our service, it was challenging to understand upfront if the orchestration would be processed or will be stuck. 
- More often customers used to reach out with incidents complaining their request not completing for long time.
Motivation: motivation was to reduce the TTD for finding if orchestration can be stuck/waiting-forever in a control-queue irrespective of the cause.
- This is where heartbeat orchestration comes into picture, which evaluates the control-queue siting outside the system (let's say system is completely down and cannot emit any metric or emitting false/partial metrics).
- We built this system locally in our service, and it has been detecting the issue with 100% accuracy. We build monitors around it to raise ICM and mitigate the situation manually using TSGs.

Issue: Manual efforts in identifying the impacted worker and mitigation steps.
- Post having heartbeat orchestrations, next issue was it used to take manual steps and active engineering cycles to identify the impacted worker and trigger mitigation.
Motivation: motivation was to make detection of impacted worker and perform mitigation steps automatically.
- This is where monitor for these heartbeat orchestration and owner of control-queue arrives.
- We built this system locally in our service, and it has been identifying the impacted worker predictably and auto-mitigates the issue.

> With both these systems in place, we are able to run DTF with no orchestrations stuck, without any manual steps for detection and mitigation. To bring this capability to all users of DTF users, proposing below change.

# Proposals:
Adding control-queue monitor with dedicated orchestrator to each control-queue.

This change consists of 2 main portions:
1. Addition of orchestration and its instances exactly one for each control-queue. 
- This orchestrator (aka ControlQueueHeartbeatTaskOrchestrator) is very simple and quick to run.
- It just waits for some interval (configurable) and logs a heartbeat message and continue itself as new.
- It validates if the context is correct, just in case partition count changes (it assumes partition count doesn't change, otherwise it may result in closing off the orchestrators). 
- It provides a way for user to provide a delegate (callback) to run with each heartbeat.
- This one gets the partition-count information and creates instance-ids such that one is allotted to each control-queue.

2. Addition of a monitoring of these orchestrations from last processed time. 
- Now, with the instance id for each control-queue, it checks for last updated time of orchestration instance and uses it measuring against threshold.
- If orchestration found stuck, appropriate message is logged. 
- This one does a bit more to find point in time owners of each control-queue, to pin the impacted taskhubworker owning the control-queue. 
- It also provides a way for user to provide delegate (callback) to run when it detect any anomaly, like stuck orchestration, fetching owner fails, fetching orchestration instance fails, or any of these times out).

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
@@ -32,6 +32,12 @@ namespace DurableTask.AzureStorage
 
         internal static readonly TimeSpan DefaultMaxQueuePollingInterval = TimeSpan.FromSeconds(30);
 
+        internal static readonly TimeSpan DefaultControlQueueHearbeatOrchestrationInterval = TimeSpan.FromSeconds(60);
+
+        internal static readonly TimeSpan DefaultControlQueueOrchHeartbeatDetectionInterval = TimeSpan.FromSeconds(600);
+
+        internal static readonly TimeSpan DefaultControlQueueOrchHeartbeatDetectionThreshold = TimeSpan.FromSeconds(600);
+
         LogHelper logHelper;
 
         /// <summary>
@@ -176,6 +182,21 @@ namespace DurableTask.AzureStorage
         /// Maximum interval for polling control and work-item queues.
         /// </summary>
         public TimeSpan MaxQueuePollingInterval { get; set; } = DefaultMaxQueuePollingInterval;
+
+        /// <summary>
+        /// Time interval between control queue heartbeat orchestration.
+        /// </summary>
+        public TimeSpan ControlQueueHearbeatOrchestrationInterval { get; set; } = DefaultControlQueueHearbeatOrchestrationInterval;
+
+        /// <summary>
+        /// Time interval between control queue heartbeat orchestration.
+        /// </summary>
+        public TimeSpan ControlQueueOrchHeartbeatDetectionInterval { get; set; } = DefaultControlQueueOrchHeartbeatDetectionInterval;
+
+        /// <summary>
+        /// Time interval between control queue heartbeat orchestration.
+        /// </summary>
+        public TimeSpan ControlQueueOrchHeartbeatDetectionThreshold { get; set; } = DefaultControlQueueOrchHeartbeatDetectionThreshold;
 
         /// <summary>
         /// If true, takes a lease on the task hub container, allowing for only one app to process messages in a task hub at a time.
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
@@ -36,7 +36,7 @@ namespace DurableTask.AzureStorage
 
         internal static readonly TimeSpan DefaultControlQueueOrchHeartbeatDetectionInterval = TimeSpan.FromSeconds(600);
 
-        internal static readonly TimeSpan DefaultControlQueueOrchHeartbeatDetectionThreshold = TimeSpan.FromSeconds(600);
+        internal static readonly TimeSpan DefaultControlQueueOrchHeartbeatLatencyThreshold = TimeSpan.FromSeconds(600);
 
         LogHelper logHelper;
 
@@ -189,14 +189,14 @@ namespace DurableTask.AzureStorage
         public TimeSpan ControlQueueHearbeatOrchestrationInterval { get; set; } = DefaultControlQueueHearbeatOrchestrationInterval;
 
         /// <summary>
-        /// Time interval between control queue heartbeat orchestration.
+        /// Time interval between control queue heartbeat detection.
         /// </summary>
         public TimeSpan ControlQueueOrchHeartbeatDetectionInterval { get; set; } = DefaultControlQueueOrchHeartbeatDetectionInterval;
 
         /// <summary>
-        /// Time interval between control queue heartbeat orchestration.
+        /// Time threshold for latency in control queue heartbeat.
         /// </summary>
-        public TimeSpan ControlQueueOrchHeartbeatDetectionThreshold { get; set; } = DefaultControlQueueOrchHeartbeatDetectionThreshold;
+        public TimeSpan ControlQueueOrchHeartbeatLatencyThreshold { get; set; } = DefaultControlQueueOrchHeartbeatLatencyThreshold;
 
         /// <summary>
         /// If true, takes a lease on the task hub container, allowing for only one app to process messages in a task hub at a time.
--- a/src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
+++ b/src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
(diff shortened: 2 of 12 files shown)


src/DurableTask.AzureStorage/ControlQueueHeartbeat/FileWriter.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace DurableTask.AzureStorage.ControlQueueHeartbeat
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class FileWriter
```
pasaini-microsoft, line 11: This will be removed and structed logging will be added.

davidmrdavid:
Hi @pasaini-microsoft: Thanks for opening this PR. As we discussed internally, I'll loop in the other DTFx/DF maintainers for discussion regarding the scope of the PR, as well as the design. I'm personally most excited about the ability to create an instanceID targetting a specific partition, so at the very least I'd like to get that in.

It would help **greatly** if you could update the PR description to contain not just a list of the changes, but also a small motivation and background behind this change. For example, you can explain that you're using this utility in your own app to help you detect stuck orchestrations, and so on. Thanks!

src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
```diff
@@ -0,0 +1,118 @@
...
+            // This queues the user provided callback without waiting for it to finish.
+            // This is to keep heartbeat orchestrator thin and fast.
+            if (!context.IsReplaying)
+            {
+                // No wait to complete provided delegate. The current orchestrator need to be very thin and quick to run. 
+                bool isQueued = ThreadPool.QueueUserWorkItem(async (_) =>
+                {
+                    await this.callBack(context.OrchestrationInstance, controlQueueHeartbeatTaskContextInput, controlQueueHeartbeatTaskContextInit, this.cancellationTokenSrc.Token);
+                });
+            }
```
davidmrdavid, line 100: I realize this was on @cgillum's suggestion, but I still think it's worth calling out in a comment where why we're not following the DF orchestrator coding constraints of not having threading in orchestrators.

src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace DurableTask.AzureStorage.ControlQueueHeartbeat
+{
+    /// <summary>
+    /// Control-queue heartbeat orchestrator.
+    /// This is supposed to be initialized with ControlQueueHeartbeatTaskContext informing orchestrator about configuration of taskhubworker and heartbeat interval.
+    /// </summary>
+    internal class ControlQueueHeartbeatTaskOrchestrator : TaskOrchestration<ControlQueueHeartbeatTaskResult, ControlQueueHeartbeatTaskInputContext>
+    {
+        public const string OrchestrationName = "ControlQueueHeartbeatTaskOrchestrator";
+
```
davidmrdavid, line 15: Given that this orchestrator is rather simple to write, my intuition is that customers wouldn't mind writing these themselves so long as they have the right utilities to let them schedule it every partition. 

With that in mind, my recommendation would be to make this orchestrator part of our samples, and to instead prioritize merging the work done to help users schedule an orchestrator in a given partitionID. I think this latter utility  function has value beyond this queue-monitoring scenario, so it's easier to justify building that into the framework.

davidmrdavid (commented):
Finished a partial pass. Please see my comment here for my thoughts on how I'd like to move this forward: https://github.com/Azure/durabletask/pull/1058#pullrequestreview-1987067747

jviau:
I love the idea of monitoring partition queue processing, but I don't think via an orchestration is the right way to do this. This is a health check after all, and there is a health check ecosystem in .NET. We should evaluate leveraging that. Metrics (#785) may also be a good avenue for this feature.

https://www.nuget.org/packages/Microsoft.Extensions.Diagnostics.HealthChecks/9.0.0-preview.2.24128.4

src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
pasaini-microsoft: The reason for keeping it with the DurableTask.AzureStorage was:
- Not giving user the choice of choosing their orchestrators as they might have gaps like when to continue as new vs complete.
- The instance-ids are dedicated in such a way it helps reverse engineer the control-queue and instance id to identify which control-queue is stuck.
- The owner-id is information is not publicly available which is required for detection.

Though it is possible to allow orchestration of arbitrary type or as just samples, but it will increase the possibility of false positives if user creates faulty orchestrators.

src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
pasaini-microsoft: updated.

src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskResult.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DurableTask.AzureStorage.ControlQueueHeartbeat
+{
+    internal enum ControlQueueHeartbeatTaskResult
+    {
+        Unknown,
+        Succeeded,
+        InvalidInput,
+        InputContextMismatch
```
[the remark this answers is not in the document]
pasaini-microsoft: added comments.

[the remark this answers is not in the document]
pasaini-microsoft: usually it happens if someone updates the partition-count.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -177,6 +183,21 @@ public class AzureStorageOrchestrationServiceSettings
...
+        /// Time interval between control queue heartbeat orchestration.
+        /// </summary>
+        public TimeSpan ControlQueueOrchHeartbeatDetectionInterval { get; set; } = DefaultControlQueueOrchHeartbeatDetectionInterval;
+
+        /// <summary>
+        /// Time interval between control queue heartbeat orchestration.
+        /// </summary>
+        public TimeSpan ControlQueueOrchHeartbeatDetectionThreshold { get; set; } = DefaultControlQueueOrchHeartbeatDetectionThreshold;
+
         /// <summary>
```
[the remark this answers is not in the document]
pasaini-microsoft: updated.

pasaini-microsoft:
> I love the idea of monitoring partition queue processing, but I don't think via an orchestration is the right way to do this. This is a health check after all, and there is a health check ecosystem in .NET. We should evaluate leveraging that. Metrics (#785) may also be a good avenue for this feature.
> 
> https://www.nuget.org/packages/Microsoft.Extensions.Diagnostics.HealthChecks/9.0.0-preview.2.24128.4

Thanks jviau. The reason for using orchestration was to keep the detection as independent from control-queue processing system as possible. This helps detect if control-queue is stuck for any reason including if taskhubworker is just absent or some control-queue just missed to be owned by any worker. 
This basically helps avoid false negatives.

src/DurableTask.AzureStorage/ControlQueueHeartbeat/ControlQueueHeartbeatTaskOrchestrator.cs
jviau: I agree with @davidmrdavid here. Focusing this work to just a feature of manual control of what partition an orchestration goes to is something I am open to accepting.

As for the heartbeat orchestration, I'd have to think more on if we even want it as a sample. Even in samples we could have issues opened on us regarding it. Is that something we want to take ownership of? Or could you host the sample yourself on github? Or even package it up into your own nuget?

PR #1060, opened 2024-04-05
Change CreateTaskOrchestrationAsync tp throw OrchestrationAlreadyExistsException

cgillum:
This is to allow exception handling code to catch a more specific exception and differentiate between unexpected errors and expected errors.

This is largely non-breaking because `OrchestrationAlreadyExistsException` already derives from `InvalidOperationException`.

Note that the `IOrchestrationServiceClient` interface already documents that the `CreateTaskOrchestrationAsync` methods should be throwing `OrchestrationAlreadyExistsException`, so this change fixes the AzureStorageOrchestrationService's implementation to follow the existing spec.


davidmrdavid (approved):
let's wait for CI

cgillum:
Looks like the failing CI is new and not part of our normal CIs. I think we can safely ignore its results.

davidmrdavid:
@cgillum: agreed. Surprised to see it running automatically on every PR, but that's fine since it's just tests.

PR #1061, opened 2024-04-05
Add automatic push to ADO step when signing

davidmrdavid:
Another step in the effort to minimize our release burden. This PR makes it so that every time a package is signed, it is also pushed to ADO. In most scenarios, if we sign a package, we want to share them with another user, in which case ADO is the preferred mechanism. This prevents us from having to manually kick off the "release to ADO" feed, it will run automatically instead.

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -219,10 +219,6 @@ steps:
 
 # Next, we push directly to ADO
 steps:
-- task: NuGetToolInstaller@1
-  displayName: 'Use NuGet '
-
-steps:
 - task: NuGetCommand@2
   displayName: 'Push to durabletask ADO feed'
   inputs:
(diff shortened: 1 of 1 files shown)


davidmrdavid:
back to draft mode to review some issues

davidmrdavid:
I have confirmed this works, so back to "ready for review"

azure-pipelines-release.yml
```diff
@@ -215,4 +215,17 @@ steps:
...
+  displayName: 'Use NuGet '
+
+- task: NuGetCommand@2
+  displayName: 'Push to durabletask ADO feed'
+  inputs:
+    command: push
+    feedsToUse: config
+    packagesToPush: '$(build.artifactStagingDirectory)/*.nupkg'
+    publishVstsFeed: '734e7913-2fab-4624-a174-bc57fe96f95d/d55248c1-5b53-411f-bfe7-73efc9e540d1'
+    allowPackageConflicts: true
```
bachuv, line 231: What does this mean? Why are we allowing package conflicts?

bachuv (commented):
Thanks for adding this step in the pipeline!

jviau:
Why is `[nalehman] DTFx Build & Run Tests Core` a PR trigger now?

PR #1063, opened 2024-04-12
Add simple poison message handling for Azure Storage

davidmrdavid:
Poison messages are a rare but destructive scenario where DTFx attempts to process a message infinitely and is somehow unable to make progress. This can create application instability and grow the control queue backlogs.

In those cases, we want to identify these "poison" messages and take them out of circulation, by putting the message on a "poison container" where the message can be manually reviewed and handled by the user, while also stopping further processing attempts.

This PR adds a simple poison message handling solution for orchestrator and activity messages. When an orchestrator or activity poison message is encountered (defined by having a DequeCount larger than 20, or some user-configured value), we place it on a **new** Azure Storage table called `<taskhubName>-poison`, which is used to hold poison messages and immediately deleted from the queue. This table is only created on demand, when a poison message is encountered.

From there, the consumer of that message is notified of the poison message.
In the case of an orchestrator poison message, the orchestrator is terminated.
In the case of an activity poison message, the activity is marked as failed, which in turn throws a catch-able exception at calling the orchestrator.
The case for a poison message in Entities is unhandled - I'd appreciate guidance on how we think that should be handled, if at all.


src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -102,6 +105,30 @@ public async Task<IReadOnlyList<MessageData>> GetMessagesAsync(CancellationToken
...
+                                // add to poison table
+                                var poisonMessagesTable = this.azureStorageClient.GetTableReference("PoisonMessagesTable");
+                                await poisonMessagesTable.CreateIfNotExistsAsync();
+                                await poisonMessagesTable.InsertAsync(poisonMessage);
+
+                                // delete from queue so it doesn't get processed again.
+                                await this.storageQueue.DeleteMessageAsync(queueMessage);
+
+                                // TODO: here we should do a forceful update in the instance table. The code doesn't seem to make that easy today :(
+                                return; // ignore this message completely
```
cgillum, line 128: Just a random idea, but maybe you could replace the poison message with a generated orchestration termination message or even an orchestration completed (with failure) message? This way we could transition the orchestration into a terminated or failed state, and even include some data explaining why it was terminated/failed.

davidmrdavid: Yeah I did consider that, and haven't fully made up my mind. I'd like to avoid having poison messages generate more messages, even if that message is for termination. It makes sense to me that a poison message handler should circumvent the standard messaging system, for robustness.

But then again, I'm just supposed to be hacking this together, so maybe I shouldn't worry about these things :)

davidmrdavid: **oh wait I misunderstood your comment:** you're saying the message _in place_ not to send a new message to the queue.

_what a great idea!_ Let me hack it real quick

davidmrdavid: このような？
https://github.com/Azure/durabletask/pull/1063/commits/748b27960b6bb26d271a64d6a61f71192c1fdbab

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -139,10 +140,19 @@ internal static bool TryGetLargeMessageReference(string messagePayload, out Uri
...
+        public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMessage, string queueName, bool isPoison = false)
         {
+            // TODO: if it's poison, we should consider the possibility that deserialization fails. In that case, it may be impossible
+            // to properly transition an orchestrator to the failed state :( . But we can move the poison message to a poison queue.
             MessageData envelope = this.DeserializeMessageData(queueMessage.Message);
 
+            // override HistoryEvent w/ a suspend event if it's a poisonMessage
+            if (isPoison)
+            {
+                ExecutionSuspendedEvent suspendEvent = new ExecutionSuspendedEvent(-1, "poison message detected");
```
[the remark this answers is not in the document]
davidmrdavid: Good points. Will change to terminate on the next pass, though for a Monday experiment I'm hoping suspend will do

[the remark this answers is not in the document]
davidmrdavid: incorporated

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1046,6 +1046,14 @@ bool IsExecutableInstance(OrchestrationRuntimeState runtimeState, IList<TaskMess
                     message = null;
                     return true;
                 }
+                // It's possible for the the original `ExecutionStarted` event to be poisonous, so here we account for that case.
+                // Though it is a misnomber to claim this is an executable instance in that case. I wonder if instead we should add a branching condition
+                // at the calling site of this method.
+                else if (newMessages.Count == 1 && newMessages[0].Event is ExecutionTerminatedEvent poisonEventCandidate && poisonEventCandidate.Input.Contains("poison"))
```
davidmrdavid, line 1052: this is rather hacky, I know. Just trying to identify the right code paths to tweak first

davidmrdavid:
As of the latest commit, poison activities are handled as well.

src/DurableTask.AzureStorage/Messaging/WorkItemQueue.cs
```diff
@@ -30,6 +31,37 @@ class WorkItemQueue : TaskHubQueue
 
         protected override TimeSpan MessageVisibilityTimeout => this.settings.WorkItemQueueVisibilityTimeout;
 
+        private async Task HandleIfPoisonMessageAsync(MessageData messageData)
```
davidmrdavid, line 34: I'll probably move this to the queue base class, since it's shared across both control and workitem queues

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -46,6 +50,36 @@ class ControlQueue : TaskHubQueue, IDisposable
...
 
+        private async Task HandleIfPoisonMessageAsync(MessageData messageData)
+        {
+            var isPoison = false;
+            var queueMessage = messageData.OriginalQueueMessage;
+
+            // if deuque count is large, just flag it as poison. Don't even deserialize it!
+            if (queueMessage.DequeueCount > 5) // TODO: make configurable
+            {
+                var poisonMessage = new DynamicTableEntity(queueMessage.Id, this.Name)
```
davidmrdavid, line 61: I assume the messageId is unique across queues?

davidmrdavid:
to figure out: what does an Entity poison message look like?

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -192,6 +192,16 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
 
                         try
                         {
+                            if (scheduledEvent.IsPoison)
+                            {
+                                // if the activity is "poison", then we should not executed again. Instead, we'll manually fail the activity
+                                // by throwing an exception on behalf of the user-code. In the exception, we provide the storage-provider's guidance
+                                // on how to deal with the poison message.
+                                // TODO: this exception seems to not be getting de-serialized correctly on the orchestrator end. Why?
```
davidmrdavid, line 200: this seems to be due to the mixed exception serialization modes we have - I need to account for the serialization strategy. This already exists within the `TaskActivity` interface's RunAsync method, so I may try to refactor that into a shared utility of some kind.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -105,27 +105,27 @@ await batch.ParallelForEachAsync(async delegate (QueueMessage queueMessage)
...
+                                await this.HandleIfPoisonMessageAsync(messageData);
                             }
-                            catch (Exception e)
+                            catch (Exception exception)
                             {
-                                // We have limited information about the details of the message
-                                // since we failed to deserialize it.
-                                this.settings.Logger.MessageFailure(
-                                    this.storageAccountName,
-                                    this.settings.TaskHubName,
```
davidmrdavid, line 118: this was moved to `this.AbandonMessageAsync`

src/DurableTask.AzureStorage/EntityTrackingStoreQueries.cs
```diff
@@ -235,9 +235,10 @@ bool OrchestrationIsRunning(OrchestrationStatus? status)
             {
                 // first, retrieve the entity scheduler state (= input of the orchestration state), possibly from blob storage.
                 string serializedSchedulerState;
-                if (MessageManager.TryGetLargeMessageReference(state.Input, out Uri blobUrl))
+                if (MessageManager.TryGetLargeMessageReference(state.Input, out Uri? blobUrl))
                 {
-                    serializedSchedulerState = await this.messageManager.DownloadAndDecompressAsBytesAsync(blobUrl);
+                    // we know blobUrl is not null because TryGetLargeMessageReference returned true
+                    serializedSchedulerState = await this.messageManager.DownloadAndDecompressAsBytesAsync(blobUrl!);
```
davidmrdavid, line 241: this is here because I added nullable analysis

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -61,9 +61,13 @@ class MessageManager
 #endif
             };
 
-            if (this.settings.UseDataContractSerialization)
+            JsonSerializer newtonSoftSerializer = JsonSerializer.Create(taskMessageSerializerSettings);
+
+            if (this.settings.UseDataContractSerialization) // for hotfix to work, set setting to `true`
```
jviau, line 66: To help lab services move forward and avoid this serialization issue again, lets just remove the `if` check and always have this workaround enabled. This way we can have them set this to `false` (rather, remove the setting of it to `true`) right away and when they do move back to GA train they don't need to worry about this setting.

davidmrdavid:
abandoning in favor of: https://github.com/Azure/durabletask/pull/1130

PR #1065, opened 2024-04-17
[WIP] Ensure popReceipt is updated between RenewMessage operations

davidmrdavid:
In order to update messages on Azure Storage, we need to provide an up-to-date "popReceipt" to the Azure Storage SDK. Otherwise, the request will fail with "404"-type errors. After every message update operation, the popReceipt for a message is updated.

From recent reports, we learned that in the new Azure Storage SDK, the popReceipt of an `queueMessage` object is not automatically updated in-place when the update API is called. Instead, the update API now returns an `updateReceipt` object containing the new popRecepit, and leaves the `popReceipt` in the `queueMessage` object intact, making it stale.
 
This suggests that we cannot simply access `queueMessage.popReceipt` to obtain the latest receipt, instead we'll need to maintain an extra mapping between messageIds and their latest popReceipts. This PR implements a rather naive mapping, but further refinement may be needed.

_work in progress_

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -17,10 +17,6 @@ steps:
     verbosityRestore: Minimal
     projects: |
       src/DurableTask.AzureStorage/DurableTask.AzureStorage.sln
-      src/DurableTask.Emulator/DurableTask.Emulator.csproj
-      src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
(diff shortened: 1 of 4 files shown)


davidmrdavid:
note: let's refactor as per this PR (https://github.com/Azure/durabletask/pull/1066 so that we actually update the popReceipt. Unclear yet if the PR should be merged as-is, or incorporated here. Needs some thinking.

PR #1067, opened 2024-04-18
New APIs and updates related to failure details propagation

cgillum:
First step towards properly fixing https://github.com/Azure/azure-functions-durable-extension/issues/2476.

This PR fixes a couple things:

1. The `IsCausedBy<T>()` API doesn't correctly handle exceptions defined outside of mscorlib. This PR fixes that.
2. Exception details aren't propagating correctly from a sub-orchestration to the parent orchestration. This PR also fixes the propagation.
3. Minor public API changes to make some of the failure details primitives usable by other SDKs

I also went ahead and bumped the minor version number of DurableTask.Core since these changes impact the public API surface area.

### DIFF
--- a/src/DurableTask.Core/FailureDetails.cs
+++ b/src/DurableTask.Core/FailureDetails.cs
@@ -21,6 +21,9 @@ namespace DurableTask.Core
     using DurableTask.Core.Exceptions;
     using Newtonsoft.Json;
 
+    // NOTE: This class is very similar to https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskFailureDetails.cs.
+    //       Any functional changes to this class should be mirrored in that class and vice versa.
+
     /// <summary>
     /// Details of an activity, orchestration, or entity operation failure.
     /// </summary>
@@ -129,11 +132,6 @@ namespace DurableTask.Core
         /// <returns>Returns <c>true</c> if the <see cref="ErrorType"/> value matches <typeparamref name="T"/>; <c>false</c> otherwise.</returns>
         public bool IsCausedBy<T>() where T : Exception
         {
-            if (string.IsNullOrEmpty(this.ErrorType))
-            {
-                return false;
-            }
-
             // This check works for .NET exception types defined in System.Core.PrivateLib (aka mscorelib.dll)
             Type? exceptionType = Type.GetType(this.ErrorType, throwOnError: false);
 
--- a/src/DurableTask.Core/FailureDetails.cs
+++ b/src/DurableTask.Core/FailureDetails.cs
@@ -14,6 +14,9 @@
 namespace DurableTask.Core
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
     using System.Runtime.Serialization;
     using DurableTask.Core.Exceptions;
     using Newtonsoft.Json;
@@ -40,6 +43,16 @@ namespace DurableTask.Core
             this.StackTrace = stackTrace;
             this.InnerFailure = innerFailure;
             this.IsNonRetriable = isNonRetriable;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FailureDetails"/> class from an exception object.
+        /// </summary>
+        /// <param name="e">The exception used to generate the failure details.</param>
+        /// <param name="innerFailure">The inner cause of the failure.</param>
+        public FailureDetails(Exception e, FailureDetails innerFailure)
+            : this(e.GetType().FullName, GetErrorMessage(e), e.StackTrace, innerFailure, false)
+        {
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/DurableTask.Core/FailureDetails.cs
```diff
@@ -116,7 +129,38 @@ public override string ToString()
         /// <returns>Returns <c>true</c> if the <see cref="ErrorType"/> value matches <typeparamref name="T"/>; <c>false</c> otherwise.</returns>
         public bool IsCausedBy<T>() where T : Exception
         {
+            if (string.IsNullOrEmpty(this.ErrorType))
+            {
+                return false;
+            }
```
davidmrdavid, line 135: is this even valid? For the ErrorType to be null or empty? Should we consider a more aggressive failure here?

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -638,6 +638,11 @@ public void FailOrchestration(Exception failure)
                     details = orchestrationFailureException.Details;
                 }
             }
+            else if (failure is TaskFailedException tfe && this.ErrorPropagationMode == ErrorPropagationMode.UseFailureDetails)
+            {
+                // Propagate the original FailureDetails
+                failureDetails = tfe.FailureDetails;
+            }
```
davidmrdavid, line 645: nit - I think we should avoid single letter 3 word abbreviations :-)
```suggestion
            else if (failure is TaskFailedException taskFailedException && this.ErrorPropagationMode == ErrorPropagationMode.UseFailureDetails)
            {
                // Propagate the original FailureDetails
                failureDetails = taskFailedException .FailureDetails;
            }
```

src/DurableTask.Core/FailureDetails.cs
davidmrdavid: would appreciate a comment around this if statement explaining when we'd expect this to be null or empty, I find it surprising.

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -116,7 +129,38 @@ public override string ToString()
...
+                    .ToList();
+                if (matchingExceptionTypes.Count == 1)
+                {
+                    exceptionType = matchingExceptionTypes[0];
+                }
+                else if (matchingExceptionTypes.Count > 1)
+                {
+                    throw new AmbiguousMatchException($"Multiple exception types with the name '{this.ErrorType}' were found.");
+                }
+            }
```
davidmrdavid, line 162: I knew I had seen this somewhere before! https://github.com/microsoft/durabletask-dotnet/blob/a275358fcfb97dd513af8d803d1b3b1ecf91b724/src/Abstractions/TaskFailureDetails.cs#L66-L98

Might be worthwhile to add a comment somewhere saying this and the durabletask-dotnet file should probably be kept in sync / be rather similar? I'd like to avoid bugs fixed here to remain in the other codebase

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -116,7 +129,38 @@ public override string ToString()
         /// <returns>Returns <c>true</c> if the <see cref="ErrorType"/> value matches <typeparamref name="T"/>; <c>false</c> otherwise.</returns>
         public bool IsCausedBy<T>() where T : Exception
         {
+            if (string.IsNullOrEmpty(this.ErrorType))
```
davidmrdavid, line 132: seems this code needs to be rather similar to the one in durabletask-dotnet, so I want to propose that we add a note about this

```suggestion
        public bool IsCausedBy<T>() where T : Exception
        {
            // This code is mostly a duplicate of this file: https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskFailureDetails.cs
            // when making changes here, especially bug fixes, please review the file above in case the changes need to be ported there as well.

            if (string.IsNullOrEmpty(this.ErrorType))
```

davidmrdavid (approved):
Approved w/ suggestions

src/DurableTask.Core/FailureDetails.cs
cgillum: It's a good point - it's never expected to be null or empty. Static nullability checks should help ensure that it's at least never null. I'm okay with removing this check to avoid creating invalid expectations.

src/DurableTask.Core/TaskOrchestrationContext.cs
cgillum: This is common Go conventions (simple local variable names) sneaking into my C# code.

src/DurableTask.Core/FailureDetails.cs
davidmrdavid: I'm fine keeping it as well so long as we document that this is just being extra safe. I'll leave it up to you, to either remove it or document the expectation :)

cgillum: I removed it.

src/DurableTask.Core/FailureDetails.cs
cgillum: Added a comment at the top of the file.

davidmrdavid (approved):
approved once CI goes green

PR #1069, opened 2024-04-19
Get instance id for desired control-queue(s)

pasaini-microsoft:
# Motivation
#1079 
## Issue: No way of targeting an orchestrator instance to a desired control-queue.
- We have been facing issues where DTF orchestration used to get stuck at random. Given that customer load is not very regular in our service, it was challenging to understand upfront if the orchestration would be processed or will be stuck.
- More often customers used to reach out with incidents complaining their request not completing for long time.
- This is where we needed orchestration instances to observe health of each queue by targeting one instance for desired control-queue.
### Motivation: 
- motivation was to reduce the TTD for finding if orchestration can be stuck/waiting-forever in a control-queue irrespective of the cause.

## Issue: No way to load lightly loaded control-queues.
- We have face a few situations where some of control-queues are overwhelmed with orchestration instances while the others are happily processing almost nothing.
### Motivation: 
-  motivation was to target new instances of orchestration instances to set control-queue which are not heavily loaded  

# Proposal
## API to generate instance id for a set of control-queues.
- This API receives set of control-queues and prefix for instance id. 
- Implementation detail is: Allow special way of creating instance id with a suffix unsigned integer after delimiter '!' and explicitly use that value to route to control-queue (say suffixNumber % partitionCount). If this pattern is not used, it would goes back to default (current) which is hash(instance-id)%partition-count.


src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -318,5 +318,11 @@ internal LogHelper Logger
         /// Consumers that require separate dispatch (such as the new out-of-proc v2 SDKs) must set this to true.
         /// </summary>
         public bool UseSeparateQueueForEntityWorkItems { get; set; } = false;
+
+        /// <summary>
+        /// Enabled explicit placement of instance to parition id.
+        /// if the instance id ends with !nnn, where nnn is an unsigned number, it indicates explicit partition placement
+        /// </summary>
```
davidmrdavid, line 325:
```suggestion
        /// Whether to allow instanceIDs to use special syntax to land on a specific partition.
        /// If enabled, when an instanceID ends with suffix '!nnn', where 'nnn' is an unsigned number, the instance will land on the partition/queue for to that number.
        /// </summary>
```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2075,6 +2075,28 @@ public Task<string> DownloadBlobAsync(string blobUri)
...
+            int placementSeparatorPosition = instanceId.LastIndexOf('!');
+
+            // if the instance id ends with !nnn, where nnn is an unsigned number, it indicates explicit partition placement
+            if (
+                this.settings.EnableExplicitPartitionPlacement
+                && placementSeparatorPosition != -1
+                && uint.TryParse(instanceId.Substring(placementSeparatorPosition + 1), out uint index))
+            {
+                var partitionId = index % totalPartitions;
+                return (uint)partitionId;
```
davidmrdavid, line 2091: So this allows for the following scenario:
Assume we have 4 partitions,
And user creates instanceID "abc!7", they will still land on some queue, but it won't be the 7th queue, because that doesn't exist. 

From first principles, I would think we'd want to error out in this case. But it seems this behavior is consistent with Netherite. I would prefer to throw in this case, but curious to know what others think.

davidmrdavid (commented):
Some thoughts

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: @pasaini-microsoft - now that I think about it, it may be a good idea to keep the behavior you implemented. If in the future, we make the partitionCount something that users can change 'on the fly' (not possible today, but I'm working to make this happen), then this behavior would be resilient to changes in the number of partitions.

Let's keep this behavior for now but let's also try to emit a warning for when the total number of partitions is less than the customer's specified target number. That will help notify the customer that something possibly unintuitive is taking place. Thanks

Test/DurableTask.AzureStorage.Tests/TestPartitionIndex.cs
```diff
@@ -0,0 +1,101 @@
...
+            azureStorageOrchestrationService = new AzureStorageOrchestrationService(settings);
+
+            controlQueueNumberToNameMap = new Dictionary<string, int>();
+
+            for (int i = 0; i < partitionCount; i++)
+            {
+                var controlQueueName = AzureStorageOrchestrationService.GetControlQueueName(settings.TaskHubName, i);
+                controlQueueNumberToNameMap[controlQueueName] = i;
+            }
+        }
```
davidmrdavid, line 69: are we still using this in the new tests? No, right?

davidmrdavid (commented):
Have we tested this for external events as well?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sebastianburckhardt: I am actually in favor of not generating a warning. The reason is that the warning would actually fire almost constantly in all the applications where I have used this. 

The expected use is that applications want to distribute things over the queues but dont actually know or care how many queues there are (e.g. like partition keys in Azure Storage).

jviau: I agree with @sebastianburckhardt - I feel this will be more noisy than helpful

davidmrdavid: Ok, let's ignore the warning, I'm convinced. Agreed on the need to document it

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -318,5 +318,11 @@ internal LogHelper Logger
         /// Consumers that require separate dispatch (such as the new out-of-proc v2 SDKs) must set this to true.
         /// </summary>
         public bool UseSeparateQueueForEntityWorkItems { get; set; } = false;
+
+        /// <summary>
+        /// Whether to allow instanceIDs to use special syntax to land on a specific partition.
+        /// If enabled, when an instanceID ends with suffix '!nnn', where 'nnn' is an unsigned number, the instance will land on the partition/queue for to that number.
+        /// </summary>
+        public bool EnableExplicitPartitionPlacement { get; set; } = false;
```
davidmrdavid, line 326: Something to consider - it is not safe to change this from `false` to `true` (or vice-versa) while an orchestrator with the special syntax is in-flight. If we do that, any pre-existing messages for that orchestrator may now be considered to be "in the wrong queue".

Let's call this out in the intellisense

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2075,6 +2075,28 @@ public Task<string> DownloadBlobAsync(string blobUri)
             return cachedQueue;
         }
 
+        internal uint GetPartitionIndex(string instanceId)
+        {
+            uint totalPartitions = (uint)this.settings.PartitionCount;
+
+            int placementSeparatorPosition = instanceId.LastIndexOf('!');
+
+            // if the instance id ends with !nnn, where nnn is an unsigned number, it indicates explicit partition placement
```
davidmrdavid, line 2084: can we add a test that documents the behavior if the customer uses an instanceID with multiple `!` in there? Say instanceID "A!1!B!3` should probably map to partition "3", right?

davidmrdavid: Let's also add a test that checks that instanceID `myinstanceID!NotANumber` does not trigger any errors / that it correctly ignores the explicit placement logic.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -755,6 +755,7 @@ public Task TerminateInstanceAsync(OrchestrationInstance orchestrationInstance)
         }
 
         /// <summary>
+        /// Add test for this.
```
davidmrdavid, line 758: was this intended?

davidmrdavid (commented):
This looks almost ready to me, but please note there's some outstanding suggestions (and a missing test request from my last review). 

FYI @gillum, whose approval we should seek before merging this.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -318,5 +318,12 @@ internal LogHelper Logger
         /// Consumers that require separate dispatch (such as the new out-of-proc v2 SDKs) must set this to true.
         /// </summary>
         public bool UseSeparateQueueForEntityWorkItems { get; set; } = false;
+
+        /// <summary>
+        /// Whether to allow instanceIDs to use special syntax to land on a specific partition.
+        /// If enabled, when an instanceID ends with suffix '!nnn', where 'nnn' is an unsigned number, the instance will land on the partition/queue for to that number.
+        /// ** DO NOT CHANGE THIS FLAG FOR PRE-EXISTING MESSAGES AS IT MAY BE CONSIDERED IN THE WRONG QUEUE **
+        /// </summary>
```
davidmrdavid, line 326: nit - I believe `remark` is the right docs tag for this sort of thing. Let's also not use all-caps for comments, I realize it has a useful effect, but it's not in our usual style for customer-facing docs.

```suggestion
        /// <summary>
        /// Whether to allow instanceIDs to use special syntax to land on a specific partition.
        /// If enabled, when an instanceID ends with suffix '!nnn', where 'nnn' is an unsigned number, the instance will land on the partition/queue for to that number.
        /// ** DO NOT CHANGE THIS FLAG FOR PRE-EXISTING MESSAGES AS IT MAY BE CONSIDERED IN THE WRONG QUEUE **
        /// </summary>
        /// <remarks>
        /// It is not generally safe to change to this flag for pre-existing TaskHubs, as it may change the expected target queue for an instanceID. 
        /// </remarks>
```

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -755,6 +756,7 @@ public Task TerminateInstanceAsync(OrchestrationInstance orchestrationInstance)
...
```
davidmrdavid, line 758:
```suggestion
```

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -707,6 +707,7 @@ void CreateAndTrackDependencyTelemetry(TraceContextBase requestTraceContext)
...
```
davidmrdavid:
```suggestion
```

davidmrdavid (commented):
FYI - I'll take the liberty of deleting the deleting the comment about adding a test at the beginning of `TaskHubClient`, since this is public documentation so it does not belong _there_. After that, I'll leave a 'LGTM'

davidmrdavid (commented):
LGTM. Please get an approval from a current repo owner (@cgillum, or @jviau, for example) before merging. But this seems good to me, and that any remaining tests we could just add on your behalf with ease.

PR #1071, opened 2024-04-22
Add Versioning support in durabletask(phase1)

skyao:
Code updates:

- OrchestrationContext

  Add a new field called Version to indicate the version of the current orchestration instance. Version will be set when the orchestration instance is created.

- TaskOrchestrationExecutor

  Get the instance version from version field on the ExecutionStartedEvent and save it to TaskOrchestrationContext.Version so that it can be used in the orchestration code.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationExecutor.cs
+++ b/src/DurableTask.Core/TaskOrchestrationExecutor.cs
@@ -131,7 +131,17 @@ namespace DurableTask.Core
                                 this.context.CurrentUtcDateTime = decisionStartedEvent.Timestamp;
                                 continue;
                             }
-
+                            
+                            if (historyEvent.EventType == EventType.ExecutionStarted)
+                            {
+                                var executionStartedEvent = (ExecutionStartedEvent)historyEvent;
+                                // get the instance version from version field on the ExecutionStartedEvent
+                                var instanceVersion = executionStartedEvent.Version;
+                                // copy the instance version to OrchestrationInstance in TaskOrchestrationContext
+                                // so that it can be used in the orchestration code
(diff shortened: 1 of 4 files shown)


src/DurableTask.Core/TaskOrchestrationExecutor.cs
```diff
@@ -131,7 +131,17 @@ void ProcessEvents(IEnumerable<HistoryEvent> events)
...
-
+                            
+                            if (historyEvent.EventType == EventType.ExecutionStarted)
+                            {
+                                var executionStartedEvent = (ExecutionStartedEvent)historyEvent;
+                                // get the instance version from version field on the ExecutionStartedEvent
+                                var instanceVersion = executionStartedEvent.Version;
+                                // copy the instance version to OrchestrationInstance in TaskOrchestrationContext
+                                // so that it can be used in the orchestration code
+                                this.context.OrchestrationInstance.InstanceVersion = instanceVersion;
```
cgillum, line 142: A couple questions:

* Why does the version need to be made part of `OrchestrationInstance`, and why does it need to be serializable? It's already part of `ExecutionStartedEvent`, which is already serialized into the state store.
* For making the version available to the orchestrator code, couldn't we just expose a new property directly onto `OrchestrationContext` base class and set that new property instead of going through `OrchestrationInstance`?

skyao: OK, code updated, rollback changes of OrchestrationInstance, adding version on OrchestrationContext.

cgillum (approved):
This looks good to me as a first step. @jviau do you have any objections to this change?

jviau (approved):
This is great! Getting version was annoying before - I always had to extract it in the middleware and save it somewhere (typically a property on the orchestration)

PR #1077, opened 2024-04-25
Merge main into azure-storage-v12

nytian:
As titled. Routine PR to merges the recent changes from the main branch into azure-storage-v12 for durable v3 GA. 

Changes in main branch is ensured to be workable at the  target branch with PR #1081.


bachuv (approved):

davidmrdavid (approved):

davidmrdavid (changes requested):
My apologies, I would actually like to inspect the commit history a bit closer - I notice the merge commit is rather large (https://github.com/Azure/durabletask/pull/1077/commits/d8863e1aa40b78c83088bc6f530cca4131ac6598) but I'd expect it to be only the conflict resolution diff. I'd like to triple check why that is, so let's please hold off on this for a bit.

davidmrdavid (approved):
After reading up a bit more on how merge commits represent diffs, I'm no longer concerned. Looks good, although in the future it would be good to keep branches up to date with each often more often so we don't have large PRs like this one, but that's everyone's responsibility :). Thank you for handling the difficult merges!

PR #1080, opened 2024-04-30
Fix Argument Order in MergeTags() when Scheduling Orchestration in Entity Dispatcher

nytian:
This PR fixes a bug where tags were incorrectly passed when entities scheduling orchestrations, causing entities to wrongly receive a `SubOrchestrationInstanceCompleted` message. By correcting the tag passing, this update ensures the intended fire-and-forget behavior of orchestration scheduling.

Add error handling in TaskEntityDispatcher. If there is any unexpected events received by the dispatcher, throw error indicating that this is not expected and should be an issue. 

Add unit test for TaskEntityDispatcher.


nytian:
I would like to add a unit test for the change I made but I didn't find a appropriate file or test class for EntityDispatcher. So I only did the E2E test. Let me know if you guys have any idea about the testing. Thanks!

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -801,8 +801,8 @@ void ProcessSendStartMessage(WorkItemEffects effects, OrchestrationRuntimeState
             var executionStartedEvent = new ExecutionStartedEvent(-1, action.Input)
             {
                 Tags = OrchestrationTags.MergeTags(
-                    runtimeState.Tags,
-                    new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } }),
+                    new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } },
+                    runtimeState.Tags),
                 OrchestrationInstance = destination,
```
davidmrdavid, line 806: can you please clarify why the order of tags matters? Ideally we'd document that in the code as well, to prevent regressions :)

davidmrdavid (commented):
Left a question:

Agreed that it would be good to have a test. Not immediately sure if there's entity tests at this level, but if they don't exist, then we should create them, though I wouldn't block this PR for that (since we have an active case)

src/DurableTask.Core/TaskEntityDispatcher.cs
nytian: Thanks for reviewing!

It's because the logic in [MergeTags()](https://github.com/Azure/durabletask/blob/b2a8a56ace3233895070135a2a36601c5eb3443d/src/DurableTask.Core/OrchestrationTags.cs#L49) In summary, the FireAndForget value from newTags is intended to overwrite the one in existingTags. However, in our case, newTags include a key-value pair of `FireAndForget: true`, while the existing tag is null(which is the default). Due to incorrect ordering, the existing tag was mistakenly treated as new tags, causing the FireAndForget value to be overwritten as null. That's how we caused the issue.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -801,8 +801,8 @@ void ProcessSendStartMessage(WorkItemEffects effects, OrchestrationRuntimeState
             var executionStartedEvent = new ExecutionStartedEvent(-1, action.Input)
             {
                 Tags = OrchestrationTags.MergeTags(
-                    runtimeState.Tags,
-                    new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } }),
+                    new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } },
+                    runtimeState.Tags),
```
davidmrdavid, line 805: To prevent future error, let's add explicit the parameter names here, so that we know which tags are supposed to be the new ones, and which are the old ones:

```suggestion
                    newTags: new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } },
                    existingTags: runtimeState.Tags),
```

src/DurableTask.Core/TaskEntityDispatcher.cs
davidmrdavid: Got it. Good catch, thanks! Left an edit request based on that here.

davidmrdavid (commented):
Good catch, this was a subtle bug.
Can you please also add a handler in the Entity event processor so that, if a SubOrchestration response  message (or any message outside the expected set) is received, we throw an exception? That would have also helped us catch this error faster!

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -563,6 +563,11 @@ void DetermineWork(OrchestrationRuntimeState runtimeState, out SchedulerState sc
                         }
 
                         break;
+
+                    default:
+                        {
+                            throw new EntitySchedulerException($"Unexpected event type {e.EventType} in entity scheduler history.");
```
davidmrdavid: nit: I like to provide customers with an actionable error where possible. In this case, this would be a bug in our end, so I'd encourage them to file a bug:

``suggestion
                            throw new EntitySchedulerException($"The entity with instanceId '{instanceId}' received an unexpected event type of type '{e.EventType}'. This is not a valid entity message. This is a framework-internal error, please report this issue in the GitHub repo: 'https://github.com/Azure/durabletask/'");
```

File: src/DurableTask.Core/Entities/EntityTestHelper.cs [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+using DurableTask.Core.Logging;
+using DurableTask.Core.Middleware;
+using Microsoft.Extensions.Logging;
+
+namespace DurableTask.Core.Entities
+{
+    /// <summary>
+    /// Class to help test TaskEntityDispatcher.
+    /// </summary>
+    public class EntityTestHelper
```
nytian, line 11: Have to add this file due to some class's accessibility level. Let me know if you have any questions about it!

Test/DurableTask.AzureStorage.Tests/TestEntityDispatcher.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using DurableTask.Core.Logging;
+using DurableTask.Core.Entities.OperationFormat;
+using System.Text;
+using Microsoft.Extensions.Azure;
+using DurableTask.Core.History;
+
+namespace DurableTask.AzureStorage.Tests
```
davidmrdavid, line 13: it is a bit odd to me that this is an AzureStorage test, and not something in DTFx.Core's tests, but that's ok :)

Test/DurableTask.AzureStorage.Tests/TestEntityDispatcher.cs
```diff
@@ -0,0 +1,40 @@
...
+            };
+            var service = new AzureStorageOrchestrationService(settings);
+            var entityTestHelper = new EntityTestHelper(service);
+            var testEvent = entityTestHelper.TestScheduleOrchestrationTags();
+            
+            Assert.IsInstanceOfType(testEvent, typeof(ExecutionStartedEvent));
+            var testEventTags = (ExecutionStartedEvent)testEvent;
+            bool containsKey = testEventTags.Tags.ContainsKey(OrchestrationTags.FireAndForget);
+            Assert.IsTrue(containsKey);
+        }
```
davidmrdavid, line 37: some nitpicks:

```suggestion
        readonly string connectionString = TestHelpers.GetTestStorageAccountConnectionString();
        
        [TestMethod]
        public void TestScheduleOrchestration()
        {
            var settings = new AzureStorageOrchestrationServiceSettings()
            {
                StorageConnectionString = this.connectionString ,
                TaskHubName = "scheduleOrchestratorFromEntity",
            };
            var service = new AzureStorageOrchestrationService(settings);
            var entityTestHelper = new EntityTestHelper(service);
            var testEvent = entityTestHelper.TestScheduleOrchestrationTags();
            
            Assert.IsInstanceOfType(testEvent, typeof(ExecutionStartedEvent));
            var testEventTags = (ExecutionStartedEvent)testEvent;
            bool containsKey = testEventTags.Tags.ContainsKey(OrchestrationTags.FireAndForget);
            Assert.IsTrue(containsKey);
        }
```

src/DurableTask.Core/Entities/EntityTestHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+        TaskEntityDispatcher dispatcher;
+
+        /// <summary>
+        /// Constructor for EntityTestHelper
+        /// </summary>
+        /// <param name="service">The <see cref="IOrchestrationService"/> for initialize TaskEntityDispatcher instance.</param>
+        public EntityTestHelper(IOrchestrationService service)
+        {
+            this.orchestrationService = service;
+            this.dispatcher ??= CreateTestEntityDispatcherInstance();
```
davidmrdavid, line 23: Why are we using `??=` here instead of always creating the dispatcher (i.e using `=`)?

src/DurableTask.Core/Entities/EntityTestHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        /// <summary>
+        /// Create a  instance of the TaskEntityDispatcher class for testing.
```
davidmrdavid, line 27:
```suggestion
        /// Create an instance of the TaskEntityDispatcher class for testing.
```

src/DurableTask.Core/Entities/EntityTestHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+                entityManager,
+                entityMiddleware,
+                logger,
+                ErrorPropagationMode.SerializeExceptions);
+        }
+
+        /// <summary>
+        /// Method to test orchestration settings when entities schedule it.
+        /// </summary>
+        /// <returns> HistoryEvent contains the schedule message.</returns>
```
davidmrdavid, line 49:
```suggestion
        /// <returns> HistoryEventing contains the scheduled message.</returns>
```

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -880,5 +883,35 @@ async Task<EntityBatchResult> ExecuteViaMiddlewareAsync(Work workToDoNow, Orches
 
             return result;
         }
+
+        /// <summary>
+        /// Test only. 
+        /// </summary>
+        public HistoryEvent TestScheduleSendEvent()
+        {
+            // Initialize arguments to pass to ProcessSendStartMessage().
```
davidmrdavid, line 892: hmm, I don't think we should be exposing a public method like this just for a test, this will creation confusion for our users. Can't we make this `internal` instead and define the test as part of the DurableTask.Core's tests?

davidmrdavid (commented):
Left some thoughts

Test/DurableTask.AzureStorage.Tests/TestEntityDispatcher.cs
nytian: I agree with you. My original thought was that I can use AzureStorageOrchestrationService, but in this case we can replace it with another LocalMockService. I will move this one into core tests.

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -25,11 +25,12 @@ namespace DurableTask.Emulator
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
+    using DurableTask.Core.Entities;
 
     /// <summary>
     /// Fully functional in-proc orchestration service for testing
     /// </summary>
-    public class LocalOrchestrationService : IOrchestrationService, IOrchestrationServiceClient, IDisposable
+    public class LocalOrchestrationService : IOrchestrationService, IOrchestrationServiceClient, IEntityOrchestrationService, IDisposable
```
nytian, line 33: Add new inheritor so that LocalOrchestrationService can be transferred to IentityOrchestrationService in TaskEntityDispatcher;

src/DurableTask.Core/Entities/EntityTestHelper.cs
nytian: I wonder what is HistoryEventing?

davidmrdavid: whoops, that's a typo - I only mean to change `schedule` for `scheduled` :P .

Test/DurableTask.Core.Tests/TestTaskEntityDispatcher.cs
```diff
@@ -0,0 +1,88 @@
...
+            // Invoke the dispatcher and obtain resulting event
+            dispatcher.ProcessSendStartMessage(effects, runtimeState, action);
+            HistoryEvent resultEvent = effects.InstanceMessages[0].Event;
+
+            Assert.IsInstanceOfType(resultEvent, typeof(ExecutionStartedEvent));
+            var executionStartedEvent = (ExecutionStartedEvent)resultEvent;
+
+            // The resulting event should contain a fire and forget tag
+            bool hasFireAndForgetTag = executionStartedEvent.Tags.ContainsKey(OrchestrationTags.FireAndForget);
+            Assert.IsTrue(hasFireAndForgetTag);
```
nytian, line 85: Thus there should be IsTrue

src/DurableTask.Emulator/LocalOrchestrationService.cs
```diff
@@ -669,5 +670,36 @@ void Dispose(bool disposing)
                 this.cancellationTokenSource.Dispose();
             }
         }
+
+        /// <inheritdoc />
+        /// Test only for core entities. The value is set as default.
```
davidmrdavid, line 675:
```suggestion
        /// Used for testing entities. The values correspond to the defaults.
```

davidmrdavid (approved):
LGTM, thanks for your hard work here.

src/DurableTask.Core/TaskEntityDispatcher.cs
kweebtronic: https://github.com/Azure/durabletask/issues/1095

PR #1085, opened 2024-05-07
Rev SqlClient dependency in SqlServer tests

davidmrdavid:
As in title, due to CVE warning making builds difficult.


cgillum (approved):
Let's go ahead and take this, but we should just remove the DurableTask.SqlServer content in this repo. We don't use it for anything (it was a contribution from a long time ago) and isn't related to the backend in microsoft/durabletask-sqlserver repo.

PR #1087, opened 2024-05-08
Increase DTFx.AzureStorage and DTFx.ApplicationInsights

nytian:
As titled. 
DTFx.AzureStorage  v1.17.1 -> v1.17.2
DTFx.ApplicationInsights v0.1.4 ->v0.1.5 

DTFx.Core has been updated as v2.17.0 from v2.16.1 in PR #1067

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <MajorVersion>1</MajorVersion>
     <MinorVersion>17</MinorVersion>
-    <PatchVersion>1</PatchVersion>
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <MajorVersion>1</MajorVersion>
     <MinorVersion>17</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>2</PatchVersion>
```
nytian, line 24: DTFx.Core increased a minor version, but since we don't have new feature added in DTFx.AS, so I guess at DTFx.AS there should still be just a patch version increasement.

bachuv (approved):

PR #1100, opened 2024-05-23
Remove user-provided details from various logs

davidmrdavid:
In various log events, our telemetry captures user-provided details that may contain sensitive information, particularly in the "Details" column. For safety, this PR does two things:
(1) We stop logging the reason for client operations like terminate, suspend, and resume. These are user-provided strings, and they have minimal use during investigations.

(2) When it comes to logging exceptions, we avoid logging the exception _message_, which may contain sensitive data. To do this, the orchestratorCompleted action now always carries a FailureDetails object, which can be inspected to obtain only the error trace and error type of an exception. This doesn't mean we do not honor the user's choice of error propagation mode, it just means that the FailureDetails property is always populated. Obviously, this may have a memory perf impact, but I found all alternatives to have tricky tradeoffs. Given that FailureDetails is positioned to be the _preferred_ error serialization mode of future SDKs, I think it should be 'ok' to leverage it here.


davidmrdavid:
Just remembered that capturing stack traces and exception _types_ is really useful, and getting just that should be safe. So I'll modify this PR to do that.

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1507,9 +1508,11 @@ internal class TaskActivityFailure : StructuredLogEvent, IEventSourceEvent
                 this.Name = name;
                 this.TaskEventId = taskEvent.EventId;
                 this.Details = exception.ToString();
```
jviau, line 1510: Exception ToString is still here, which I believe will become a parameter for `ILogger` calls. Do you know if `ILogger` calls make it to DF kusto tables? Or is it only the event source which is sent?

But passing exception contents as structured log parameter is something we really should not be doing. ILogger has an explicit parameter for exception which we should be using. Is there any way we can redesign this to use that? Or is that too large of a change for this PR?

davidmrdavid: @jviau - I am under the impression that, in the context of Azure Functions, the `ILogger` calls here go only to the customer's own Application Insights. Perhaps @cgillum can confirm.

>  ILogger has an explicit parameter for exception which we should be using. Is there any way we can redesign this to use that? Or is that too large of a change for this PR?

I'm interested, but I'd like to make it a part of a follow up PR so that I can tackle the immediate issue of removing exception **messages** from our telemetry. 

Beyond the `ILogger`, and assuming @cgillum can confirm it's use is primarily to redirect to Application Insights, does the change in this PR seem sound to you?

cgillum: `ILogger` calls definitely go to Application Insights, but I don't remember clearly how it works for Kusto. The Durable Functions-specific Kusto table is populated only from ETW events, but the general Functions table *might* contain these `ILogger` logs.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -626,16 +626,13 @@ public void FailOrchestration(Exception failure)
                     CorrelationTraceClient.TrackException(failure);
                 });
 
+            // In this block, we always create a failureDetails variable, which is needed to produce sanitized telemetry
             if (failure is OrchestrationFailureException orchestrationFailureException)
             {
-                if (this.ErrorPropagationMode == ErrorPropagationMode.UseFailureDetails)
+                failureDetails = orchestrationFailureException.FailureDetails;
```
cgillum, line 632: Is this refactoring an important part of this PR? I worry that there could be unintended side effects of populating failure details when not explicitly enabled. There's a lot of code out there (not just in DurableTask.Core) that may behave differently depending on whether failure details are present, without considering the value of `ErrorPropagationMode`, since that setting isn't globally accessible.

src/DurableTask.Core/Logging/LogEvents.cs
davidmrdavid: Ah, I think you're right - that the FunctionsLogs may get populated by this ILogger. That makes things more complicated. I wanted to avoid breaking the behavior of Application Insights logging by leaving the ILogger behavior untouched.

src/DurableTask.Core/TaskOrchestrationContext.cs
davidmrdavid: Yes - I was worried about that same possibility, @cgillum. And yes - this was an intended refactoring, but I'm open to alternatives.

My goal here was to be able to log something about the failure without changing how exceptions are serialized to Applicaiton Insights. So I wanted to sanitize the logs for our kusto clusters, but keep the user's Application Insights instance as-is.
 
However - since you mentioned that the `ILogger` here may also be used to log the the FunctionsLogs kusto table, then perhaps there's no alternative to sanitizing the exception in the ILogger and in ETW. If that's the case, then I can fix this issue without needing to leverage `FailureDetails`.

cgillum: I'm not sure if my question was understood, so I'm not sure I know how to respond. I probably should have asked my question on line 646 instead of here. I understand the goals, but my question is whether we need to create a `FailureDetails` object even if `ErrorPropagationMode` is set to `SerializeExceptions`. You responded with "yes", but I didn't get a sense of why it's required in your current proposal. Can you elaborate on what this specific code change is doing?

davidmrdavid: Of course. So rely on FailureDetails for the new behavior I'm introducing in `LogEvents.cs`, here: https://github.com/Azure/durabletask/blob/3de7889db52eeacb8d1fb8a26a39341f8f9f9106/src/DurableTask.Core/Logging/LogEvents.cs#L1057.

See below for the relevant two snippets:

(1) In the constructor of `OrchestrationCompleted`, I'm creating a "sanitizedDetails" string which contains the error type and error trace.

https://github.com/Azure/durabletask/blob/3de7889db52eeacb8d1fb8a26a39341f8f9f9106/src/DurableTask.Core/Logging/LogEvents.cs#L1044-L1061

(2) Then I log that via ETW to the `Details` column. This replaces logging the `toString()` representation of the exception in the `Details` column.

https://github.com/Azure/durabletask/blob/3de7889db52eeacb8d1fb8a26a39341f8f9f9106/src/DurableTask.Core/Logging/LogEvents.cs#L1086-L1095

Does that make sense? So this PR relied on always constructing a FailureDetails object so that I could always log this sanitized string. Does that make sense or am I still missing your question?

cgillum: Thanks for the explanation. I think I get it now. I won't block on this further but be aware that there is a small risk that we might accidentally regress something in the (default) `SerializeExceptions` case. Hopefully our existing test cases will help ensure that we're safe. I don't have alternative suggestions at this time.

cgillum (approved):
No blockers for me, but I recommend doing a bit of due diligence on the two points raised so far in this PR to ensure we're not regressing anything major.

We'll definitely want to call out this change clearly in the release notes.

src/DurableTask.Core/Logging/LogEvents.cs
davidmrdavid: I actually didn't find any DTFx logs in FunctionsLogs, so I think we're safe on this front. If we forward any logs to FunctionsLogs via ILogger, I think those would be coming from the WebJobs extension, but not at this level.

jviau: Good enough for me. I also believe latest host versions will track that these logs are emitted from within a function invocation scope, consider it 'user' code, and ignore it.

jviau (approved):

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1506,10 +1512,13 @@ internal class TaskActivityFailure : StructuredLogEvent, IEventSourceEvent
                 this.ExecutionId = instance.ExecutionId;
                 this.Name = name;
                 this.TaskEventId = taskEvent.EventId;
-                this.Details = exception.ToString();
+#nullable enable
                 this.exception = exception;
+                this.Details = exception.ToString();
             }
 
+            private Exception exception;
```
cgillum, line 1520: Why was this moved (and no longer `readonly`)? In C#, the normal convention is to put fields at the top of the class definition.

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1049,13 +1049,20 @@ internal class OrchestrationCompleted : StructuredLogEvent, IEventSourceEvent
...
-                this.Details = action.Details;
                 this.SizeInBytes = Encoding.UTF8.GetByteCount(action.Result ?? string.Empty);
+
+#nullable enable
+                Exception? exception = runtimeState.Exception;
+                this.exception = exception;
+                this.Details = action.Details ?? string.Empty;
             }
 
+            private Exception? exception { get; }
```
cgillum, line 1064: nit: better to use a `readonly` field instead of a property (and move it to the top of the class definition). ALso, properties don't start with a lowercase letter.

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -632,7 +632,7 @@ public TerminatingInstance(OrchestrationInstance instance, string reason)
                 StructuredEventSource.Log.TerminatingInstance(
                     this.InstanceId,
                     this.ExecutionId,
-                    this.Details,
+                    Details: string.Empty, // User-provided details may contain sensitive data, so we don't log it.
```
cgillum, line 635: Totally optional, but if you want to provide a bit more information, you could perhaps log the number of characters of data that was associated with this. For example:
```suggestion
                    Details: $"(Redacted {this.Details?.Length ?? 0} characters}), // User-provided details may contain sensitive data, so we don't log it.
```
The reason to *consider* this is if you want to try and distinguish between empty and non-empty, or perhaps to get a hint about whether there might be some large messages at play. It also could help make it clear that something was explicitly redacted. I don't have a concrete scenario for this, but it could be useful.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -134,6 +134,14 @@ public OrchestrationRuntimeState(IList<HistoryEvent>? events)
         /// </remarks>
         public FailureDetails? FailureDetails => ExecutionCompletedEvent?.FailureDetails;
 
+        /// <summary>
+        /// Failure that caused an orchestrator to complete, if any.
+        /// <remarks>
```
cgillum, line 139: nit: `<remarks>` is a sibling to `<summary>`, not a nested child. See the `FailureDetails` property above for an example of how to correctly use `<remarks>`.

cgillum (approved):
This approach looks sound to me.

src/DurableTask.Core/OrchestrationRuntimeState.cs
davidmrdavid: Thanks for pointing this out. I wish Visual Studio had flagged this somehow. Im sure there's a way to set that up, but I'll leave that to another day due to time :)

src/DurableTask.Core/Logging/LogEvents.cs
davidmrdavid: It was moved without realizing (I tried undoing a bunch of changes and failed to move this to the original position). Will restore the `readonly` as well. Thanks for the convention note.

src/DurableTask.Core/Logging/LogEvents.cs
davidmrdavid: Thanks. I'll be following the format as the other `exception` variable in `TaskActivityFailure`, where this is `readonly` but not a property. Will move it to the top of the class as well.

src/DurableTask.Core/Logging/LogEvents.cs
davidmrdavid: Good idea, will incorporate.

src/DurableTask.Core/OrchestrationRuntimeState.cs
davidmrdavid: Will be fixed in next commit

davidmrdavid:
Incorporated feedback, merging as such. Thanks for the prompt reviews!

PR #1107, opened 2024-05-28
Allow TaskHubClient's CreateOrchestrationInstanceAsync to schedule orchestrators

davidmrdavid:
Related: https://github.com/Azure/azure-functions-durable-extension/pull/2805

In the PR above, we learned if a user wants to _schedule_ an orchestrator to start a particular time, they need to use the `CreateScheduledOrchestrationInstanceAsync` instead of using the more commonly used `CreateOrchestrationInstanceAsync` API. 

Allowing `CreateOrchestrationInstanceAsync` to also schedule orchestrators by accepting a new `fireAt` would simplify things downstream (in durabletask-dotnet, it would simplify the PR linked above^) so this PR adds that "missing" parameter. This does create some repetition (there are now two APIs that allow users to schedule orchestrators) but I think it makes the orchestrator-scheduling feature more discoverable while also helping simplify downstream code.

This PR also include 2 additional minor contributions:
1) It enables nullable analysis in TaskHubClient, and performs the minimal edits needed to get the analysis to pass.
2) It adds a simple TaskHubClient test validating that the class is able to schedule orchestrators to fire at a given time.

### DIFF
--- a/Test/DurableTask.Core.Tests/TaskHubClientTests.cs
+++ b/Test/DurableTask.Core.Tests/TaskHubClientTests.cs
@@ -1,0 +1,73 @@
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+#nullable enable
+namespace DurableTask.Core.Tests
+{
+    using DurableTask.Core.History;
+    using DurableTask.Emulator;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A test implementation of <see cref="LocalOrchestrationService"/> that captures messages for later inspection.
+    /// </summary>
+    public class TestLocalOrchestrationService : LocalOrchestrationService
+    {
+        public List<TaskMessage> Messages { get; } = new List<TaskMessage>();
+
+        public override Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[] dedupeStatuses)
+        {
+            // capture message for later inspection
+            this.Messages.Add(creationMessage);
+
+            // call base implementation
+            return base.CreateTaskOrchestrationAsync(creationMessage, dedupeStatuses);
+        }
+    }
+
+    [TestClass]
+    public class TaskHubClientTests
+    {
+        /// <summary>
+        /// Tests that scheduled orchestrations can be created.
+        /// </summary>
+        [TestMethod]
+        public async Task CanCreateScheduledOrchestrations()
+        {
+            // create test orchestration service that allows us to inspect the generated HistoryEvents
+            TestLocalOrchestrationService service = new TestLocalOrchestrationService();
+            TaskHubClient client = new TaskHubClient(service);
+
+            // orchestrator should fire in 5 minutes
+            DateTime fireAt = DateTime.UtcNow.AddMinutes(5);
+
+            // invoke client method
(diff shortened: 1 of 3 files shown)


cgillum:
If another method already exists that implements the scheduling behavior, why is this PR needed? In other words, why can’t we have the consuming SDK use the existing method? That wasn’t clear in this PR description and I’m hesitant to agree to adding a redundant method overload.

davidmrdavid:
Yeah I didn't make this clear. I see two main reasons to make this change, but please do push back if needed:

First - The overloads for `CreateOrchestrationInstanceAsync` accept a strictly larger set of parameters than `CreateScheduledOrchestrationInstanceAsync`.  Therefore, `durabletask-dotnet` can't simply call the `CreateScheduledOrchestrationInstanceAsync` without possibly **dropping** parameters like `dedupStatuses` and `tags`.

For evidence:  if we do a CTRL+F for "Task<OrchestrationInstance> CreateScheduledOrchestrationInstanceAsync" in [`TaskHubClient`](https://github.com/Azure/durabletask/blob/7a778c3bca8bb4e33e9f19e3ffae2cd003fe0cce/src/DurableTask.Core/TaskHubClient.cs) you'll see that there's only 3 matches:

(A.1)
https://github.com/Azure/durabletask/blob/7a778c3bca8bb4e33e9f19e3ffae2cd003fe0cce/src/DurableTask.Core/TaskHubClient.cs#L83-L86

(A.2) 
https://github.com/Azure/durabletask/blob/7a778c3bca8bb4e33e9f19e3ffae2cd003fe0cce/src/DurableTask.Core/TaskHubClient.cs#L108-L113

(A.3) 
https://github.com/Azure/durabletask/blob/7a778c3bca8bb4e33e9f19e3ffae2cd003fe0cce/src/DurableTask.Core/TaskHubClient.cs#L135

Meanwhile, a CTRL+F for `Task<OrchestrationInstance> CreateOrchestrationInstanceAsync` yields 9 results, too many to list here. At it's largest parameter set, this method accepts up to 7 parameters, compared to the 5 parameters listed in (A.3), see here:

https://github.com/Azure/durabletask/blob/7a778c3bca8bb4e33e9f19e3ffae2cd003fe0cce/src/DurableTask.Core/TaskHubClient.cs#L344-L351

So, it seems to me that there's a parameter-parity gap between the two APIs that favors `CreateOrchestrationInstanceAsync`. That makes me think it would be better, maintenance-wise, for `CreateOrchestrationInstanceAsync` to fully subsume the behavior of `CreateScheduledOrchestrationInstanceAsync` by giving it that missing (fireAt parameter). The alternative is giving `CreateScheduledOrchestrationInstanceAsync` more overloads.

Second - `CreateOrchestrationInstanceAsync` can actually already schedule orchestrations, see the last parameter of this overload for evidence: https://github.com/Azure/durabletask/blob/7a778c3bca8bb4e33e9f19e3ffae2cd003fe0cce/src/DurableTask.Core/TaskHubClient.cs#L175

So it seems to me that `CreateOrchestrationInstanceAsync` was already intended to subsume `CreateScheduledOrchestrationInstanceAsync`, but that we're missing an overload that allows us to set  `fireAt`  in addition to all the other parameters. This PR adds that missing overload.

Hope that makes sense, sorry for the wall of text. In short - I think the two APIs already have overlapping responsibilities, and I'm just trying to give the API with more parameters another overload to have it fully subsume the other one.

davidmrdavid:
Ultimately though, I don't feel strongly. Just think this makes sense in the long run.

cgillum:
@davidmrdavid got it, the two reasons you called out have eased my concerns. I'm good with these changes. Thanks for explaining!

Test/DurableTask.Core.Tests/TaskHubClientTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    using DurableTask.Emulator;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A test implementation of <see cref="LocalOrchestrationService"/> that captures messages for later inspection.
+    /// </summary>
+    public class TestLocalOrchestrationService : LocalOrchestrationService
```
cgillum, line 27: Since this file is the only place where this new class is being used, can it instead be an inner class of `TaskHubClientTests` instead of being a top-level class?

cgillum (approved):

jviau (approved):

Test/DurableTask.Core.Tests/TaskHubClientTests.cs
davidmrdavid: Sorry, I completely missed this and I already _just merged_ the PR. I'll add this to my ToDo list for a future iteration

PR #1112, opened 2024-06-06
Migrate DurableTask.ServiceBus Table and Blob store SDK from WindowsAzure.Storage to Azure.Data.Tables and Azure.Storage.Blobs

arlynch:
### Notable Changes

- The class CloudStorageAccount no longer exists in the new libraries. Therefore, any constructors that leverage it were replaced with new constructors that take an endpoint to the store and a TokenCredentials object
- TableStore entities were able to implement a method that allowed the code to "unpack" complex types to save and load from tables. These methods are no longer available. Instead, a series of public accessors with custom getters and setters are now used to get the same behavior.

### Validation

- Running DurableTask.ServiceBus.Tests project: Test runs are green via both the pipeline and running tests against a live storage account

### DIFF
--- a/src/DurableTask.ServiceBus/Tracking/AzureStorageBlobStore.cs
+++ b/src/DurableTask.ServiceBus/Tracking/AzureStorageBlobStore.cs
@@ -41,7 +41,7 @@ namespace DurableTask.ServiceBus.Tracking
         }
 
         /// <summary>
-        /// Construct a blob storage client instance with hub name and cloud storage account
+        /// Creates a new AzureStorageBlobStore using the supplied hub name, endpoint and token credential
         /// </summary>
         /// <param name="hubName">The hub name</param>
         /// <param name="endpoint">Uri Endpoint for the blob store</param>
--- a/src/DurableTask.ServiceBus/Tracking/AzureStorageBlobStore.cs
+++ b/src/DurableTask.ServiceBus/Tracking/AzureStorageBlobStore.cs
@@ -16,9 +16,9 @@ namespace DurableTask.ServiceBus.Tracking
     using System;
     using System.IO;
     using System.Threading.Tasks;
+    using Azure.Core;
     using DurableTask.Core;
     using DurableTask.Core.Tracking;
-    using Microsoft.WindowsAzure.Storage;
 
     /// <summary>
     /// Azure blob storage to allow save and load large blobs, such as message and session, as a stream using Azure blob container.
@@ -41,13 +41,14 @@ namespace DurableTask.ServiceBus.Tracking
         }
 
         /// <summary>
-        /// Creates a new AzureStorageBlobStore using the supplied hub name and cloud storage account
+        /// Construct a blob storage client instance with hub name and cloud storage account
         /// </summary>
-        /// <param name="hubName">The hub name for this store</param>
-        /// <param name="cloudStorageAccount">Azure Cloud Storage Account</param>
-        public AzureStorageBlobStore(string hubName, CloudStorageAccount cloudStorageAccount)
+        /// <param name="hubName">The hub name</param>
+        /// <param name="endpoint">Uri Endpoint for the blob store</param>
+        /// <param name="credential">Token Credentials required for accessing the blob store</param>
+        public AzureStorageBlobStore(string hubName, Uri endpoint, TokenCredential credential)
         {
-            this.blobClient = new BlobStorageClient(hubName, cloudStorageAccount);
+            this.blobClient = new BlobStorageClient(hubName, endpoint, credential);
         }
 
         /// <summary>
--- a/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
+++ b/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
@@ -59,6 +59,8 @@ namespace DurableTask.ServiceBus.Tracking
 
             this.tableClient = new TableServiceClient(tableConnectionString);
             this.hubName = hubName;
+            this.jumpStartTableClient = tableClient.GetTableClient(JumpStartTableName);
+            this.historyTableClient = tableClient.GetTableClient(TableName);
         }
 
         public AzureTableClient(string hubName, Uri endpoint, TokenCredential credential)
@@ -80,6 +82,8 @@ namespace DurableTask.ServiceBus.Tracking
 
             this.tableClient = new TableServiceClient(endpoint, credential);
             this.hubName = hubName;
+            this.jumpStartTableClient = tableClient.GetTableClient(JumpStartTableName);
+            this.historyTableClient = tableClient.GetTableClient(TableName);
         }
 
         public string TableName => AzureTableConstants.InstanceHistoryTableNamePrefix + "00" + this.hubName;
@@ -88,14 +92,14 @@ namespace DurableTask.ServiceBus.Tracking
 
         internal async Task CreateTableIfNotExistsAsync()
         {
-            await this.tableClient.CreateTableIfNotExistsAsync(TableName);
             this.historyTableClient = tableClient.GetTableClient(TableName);
+            await this.tableClient.CreateTableIfNotExistsAsync(TableName);
         }
 
         internal async Task CreateJumpStartTableIfNotExistsAsync()
         {
-            await this.tableClient.CreateTableIfNotExistsAsync(JumpStartTableName);
             this.jumpStartTableClient = tableClient.GetTableClient(JumpStartTableName);
+            await this.tableClient.CreateTableIfNotExistsAsync(JumpStartTableName);
         }
 
         internal async Task DeleteTableIfExistsAsync()
@@ -392,7 +396,7 @@ namespace DurableTask.ServiceBus.Tracking
             where T : class, ITableEntity
         {
             var pageableResults = tableClient.QueryAsync<T>(filter: filter, 
-                maxPerPage: count > 0? count: default);
+                maxPerPage: count > 0? count: null);
 
             var results = new List<T>();
 
--- a/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
+++ b/src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
@@ -121,7 +121,7 @@ namespace DurableTask.ServiceBus.Tracking
         public Task<IEnumerable<AzureTableOrchestrationStateEntity>> QueryOrchestrationStatesAsync(
             OrchestrationStateQuery stateQuery)
         {
-            var query = CreateQueryInternal(stateQuery,false);
+            var query = CreateQueryInternal(stateQuery, false);
             return ReadAllEntitiesAsync<AzureTableOrchestrationStateEntity>(query, this.historyTableClient);
         }
 
@@ -267,13 +267,13 @@ namespace DurableTask.ServiceBus.Tracking
 
         string GetSecondaryFilterExpression(OrchestrationStateQueryFilter filter)
         {
-            string filterExpression;
+            string query;
 
             if (filter is OrchestrationStateInstanceFilter orchestrationStateInstanceFilter)
             {
                 if (orchestrationStateInstanceFilter.StartsWith)
                 {
-                    filterExpression = string.Format(CultureInfo.InvariantCulture,
+                    query = string.Format(CultureInfo.InvariantCulture,
                         AzureTableConstants.InstanceQuerySecondaryFilterRangeTemplate,
                         orchestrationStateInstanceFilter.InstanceId, ComputeNextKeyInRange(orchestrationStateInstanceFilter.InstanceId));
                 }
@@ -281,12 +281,12 @@ namespace DurableTask.ServiceBus.Tracking
                 {
                     if (string.IsNullOrWhiteSpace(orchestrationStateInstanceFilter.ExecutionId))
                     {
-                        filterExpression = string.Format(CultureInfo.InvariantCulture,
+                        query = string.Format(CultureInfo.InvariantCulture,
                             AzureTableConstants.InstanceQuerySecondaryFilterTemplate, orchestrationStateInstanceFilter.InstanceId);
                     }
                     else
                     {
-                        filterExpression = string.Format(CultureInfo.InvariantCulture,
+                        query = string.Format(CultureInfo.InvariantCulture,
                             AzureTableConstants.InstanceQuerySecondaryFilterExactTemplate, orchestrationStateInstanceFilter.InstanceId,
                             orchestrationStateInstanceFilter.ExecutionId);
                     }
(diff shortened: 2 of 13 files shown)


davidmrdavid:
green run: https://durabletaskframework.visualstudio.com/Durable%20Task%20Framework%20CI/_build/results?buildId=6300&view=ms.vss-test-web.build-test-results-tab

src/DurableTask.ServiceBus/Tracking/AzureStorageBlobStore.cs
```diff
@@ -41,13 +41,14 @@ public AzureStorageBlobStore(string hubName, string connectionString)
         }
 
         /// <summary>
-        /// Creates a new AzureStorageBlobStore using the supplied hub name and cloud storage account
+        /// Creates a new AzureStorageBlobStore using the supplied hub name, endpoint and token credential
```
davidmrdavid, line 44: Just confirming - does this mean that _only_ managed identity (i.e token credentials, right?) would be supported now?

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -21,62 +21,69 @@ namespace DurableTask.ServiceBus.Tracking
...
-        volatile CloudTable jumpStartTable;
+        volatile TableClient historyTableClient;
+        volatile TableClient jumpStartTableClient;
 
         public AzureTableClient(string hubName, string tableConnectionString)
-            : this(hubName, CloudStorageAccount.Parse(tableConnectionString))
         {
+            if(string.IsNullOrEmpty(hubName))
+            {
+                throw new ArgumentException("Invalid hub name", nameof(hubName));
```
davidmrdavid, line 52: non-blocking nitpick: technically speaking, we could be more precise an say that the hubName is not just invalid, but it's either null or empty. That might provide a better clue of what went wrong. But I recognize this matches old behavior, so I think it's up to you whether you want to make this improvement across this and similar exceptions showing up in the diff (I know we're time constrained)

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -85,85 +92,84 @@ private static CloudTableClient CreateAzureTableClient(CloudStorageAccount cloud
...
+        public async Task<Page<AzureTableOrchestrationStateEntity>> QueryOrchestrationStatesSegmentedAsync(
+            OrchestrationStateQuery stateQuery, string continuationToken, int count)
         {
-            TableQuery<AzureTableOrchestrationStateEntity> query = CreateQueryInternal(stateQuery, count, false);
-            return this.historyTable.ExecuteQuerySegmentedAsync(query, continuationToken);
+            int? nullIfNegativeCount = count < 0 ? null : count;
+            var query = CreateQueryInternal(stateQuery, false);
+            var pageableResults = this.historyTableClient.QueryAsync<AzureTableOrchestrationStateEntity>(filter: query, maxPerPage: nullIfNegativeCount);
+            await using var enumerator = pageableResults.AsPages(continuationToken, nullIfNegativeCount).GetAsyncEnumerator();
+            return await enumerator.MoveNextAsync() ? enumerator.Current : default;
```
davidmrdavid, line 135: nit - this seems to be a pattern used across other query methods. Should we refactor it into a common helper method?

src/DurableTask.ServiceBus/Tracking/AzureTableCompositeTableEntity.cs
```diff
@@ -41,40 +41,13 @@ public abstract class AzureTableCompositeTableEntity : ITableEntity
         /// <summary>
         /// Gets or sets the row timestamp
         /// </summary>
-        public DateTimeOffset Timestamp { get; set; } = DateTimeOffset.Now;
+        public DateTimeOffset? Timestamp { get; set; } = DateTimeOffset.Now;
 
         /// <summary>
         /// Gets or sets the entity etag
         /// </summary>
-        public string ETag { get; set; }
```
davidmrdavid, line 49: dropping these (and other public fields) is technically a breaking change. Just jotting this down for down, still unsure what replaces this (or if functionality was dropped)

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
```diff
@@ -93,8 +93,31 @@ public AzureTableOrchestrationHistoryEventEntity()
         /// <summary>
         /// Gets or set the history event detail for the tracking entity
         /// </summary>
+        [IgnoreDataMember]
         public HistoryEvent HistoryEvent { get; set; }
 
+#pragma warning disable 1591
```
davidmrdavid, line 99: what particular warning is `1591`?

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
```diff
@@ -93,8 +93,31 @@ public AzureTableOrchestrationHistoryEventEntity()
         /// <summary>
         /// Gets or set the history event detail for the tracking entity
         /// </summary>
+        [IgnoreDataMember]
```
davidmrdavid, line 96: nit - let's add an inline comment documenting where this data will come from
```suggestion
        [IgnoreDataMember] // see HistoryEventJson
```

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
```diff
@@ -93,8 +93,31 @@ public AzureTableOrchestrationHistoryEventEntity()
...
         public HistoryEvent HistoryEvent { get; set; }
 
+#pragma warning disable 1591
+        // In order to maintain table schema with the new Azure SDK, we need the following accessor.
+        // As a result, this HistoryEventJson does not need to be documented.
+        [DataMember(Name = "HistoryEvent")]
+        public string HistoryEventJson
```
davidmrdavid, line 103: I'm a bit confused - why can't these getter and setter be part of the `HistoryEvent` declared above?

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationJumpStartEntity.cs
```diff
@@ -50,6 +50,7 @@ public AzureTableOrchestrationJumpStartEntity(OrchestrationJumpStartInstanceEnti
         /// <summary>
         /// Gets a OrchestrationJumpStartInstanceEntity
         /// </summary>
+        [IgnoreDataMember]
         public OrchestrationJumpStartInstanceEntity OrchestrationJumpStartInstanceEntity => new OrchestrationJumpStartInstanceEntity
```
davidmrdavid, line 54: why is this ignored now?

davidmrdavid (commented):
A few preliminary questions. The changes make sense at first glance, but I see many instances where properties are marked with "IgnoreDataMember" only to be replaced by slightly differently-named properties below; I'm unsure why we cannot just change the implementation of the pre-existing properties. 

Thank you for this PR!

src/DurableTask.ServiceBus/Tracking/AzureTableCompositeTableEntity.cs
arlynch: This is being replaced with an ETag class provided by Azure Core, and a required part of [the new Azure SDK](https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.itableentity?view=azure-dotnet). While unfortunate this is a breaking change, this should not change functionality in a substantial manner.

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
arlynch: This is a warning regarding requiring xml documentation for public properties. The code comment below explains the motivation for suppressing the warning, although I'm open to justifications to have the xml comments.

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
arlynch: To preserve functionality, I thought it best to allow the HistoryEvent object be accessed via default getters and setters. This HistoryEventJson property only exists to allow the entity to cooperate with the new Azure SDK, as table read and writes for entities now behave differently.

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationJumpStartEntity.cs
arlynch: During a table read or write, the new Azure sdk attempts to serialize all public properties of a given entity. However, it can only do so for simple types (string, int, double, etc). As a result. we need to add IgnoreDataMember to complex properties. The information related to this particular property is already safely saved and read from the table, so this prevents an error occurring with the new Azure SDK. If you'd like, I can put a less verbose comment on this property explaining why we need to Ignore it.

src/DurableTask.ServiceBus/Tracking/AzureStorageBlobStore.cs
arlynch: To preserve previous functionality, we are maintaining support for connection strings. The breaking change is to introduce support for token credentials and removing support for the "cloud storage account" as that class no longer exists in the new Azure SDK.

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
arlynch: I agree that a more thorough check of not only hubname, but other parameters like URI or connection string would be beneficial. I do unfortunately agree that we are time constrained. Perhaps this would be a good addition to a minor version update?

NateLehman (approved):
LGTM 🚢 it

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -85,85 +92,87 @@ private static CloudTableClient CreateAzureTableClient(CloudStorageAccount cloud
...
+                await this.tableClient.DeleteTableAsync(JumpStartTableName);
             }
         }
 
         public Task<IEnumerable<AzureTableOrchestrationStateEntity>> QueryOrchestrationStatesAsync(
             OrchestrationStateQuery stateQuery)
         {
-            TableQuery<AzureTableOrchestrationStateEntity> query = CreateQueryInternal(stateQuery, -JumpStartTableScanIntervalInDays, false);
-            return ReadAllEntitiesAsync(query, this.historyTable);
+            var query = CreateQueryInternal(stateQuery,false);
```
davidmrdavid, line 124: nit -
```suggestion
            var query = CreateQueryInternal(stateQuery, false);
```

src/DurableTask.ServiceBus/Tracking/AzureTableClient.cs
```diff
@@ -197,15 +206,7 @@ internal TableQuery<AzureTableOrchestrationStateEntity> CreateQueryInternal(Orch
...
-            TableQuery<AzureTableOrchestrationStateEntity> query =
-                new TableQuery<AzureTableOrchestrationStateEntity>().Where(filterExpression);
-
-            if (count != -1)
-            {
-                query.TakeCount = count;
-            }
-
-            return query;
+            return filterExpression;
```
davidmrdavid, line 209: nit - a better name here might be 'query' but won't insist.

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
```diff
@@ -169,5 +121,27 @@ public override string ToString()
         {
             return $"Instance Id: {InstanceId} Execution Id: {ExecutionId} Seq: {SequenceNumber.ToString()} Time: {TaskTimeStamp} HistoryEvent: {HistoryEvent.EventType.ToString()}";
         }
+
+        #region
```
davidmrdavid, line 125: I think regions can be given a _name_ (as in `#region AzureStorageHelpers`). I recommend giving this a name to make the region more self-documenting

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationJumpStartEntity.cs
davidmrdavid: Yes, if you could please add a comment explaining where this is being taken care of, that'd be great. Thanks

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationStateEntity.cs
```diff
@@ -66,131 +70,254 @@ internal override IEnumerable<ITableEntity> BuildDenormalizedEntities()
...
+        }
 
-            returnValues.Add("InstanceId", new EntityProperty(State.OrchestrationInstance.InstanceId));
-            returnValues.Add("ExecutionId", new EntityProperty(State.OrchestrationInstance.ExecutionId));
+        #region
+        // Public Accessors with the Sate prefix are equivalent to those in the already existing State Property.
+        // These are used only to safely interact with the Azure Table Storage SDK which reads and writes using reflection.
+        // This is an artifact of updating from and SDK that did not use reflection.
+        [DataMember(Name = "InstanceId")]
+        public string StateInstanceId
```
davidmrdavid, line 95: Thank you for this comment. Please add a newline between it and in property below so it's clear that this comment applies to the region, not to `StateInstanceId` specifically.

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationStateEntity.cs
```diff
@@ -66,131 +70,254 @@ internal override IEnumerable<ITableEntity> BuildDenormalizedEntities()
...
+                "Instance Id: {0} Execution Id: {1} Name: {2} Version: {3} CreatedTime: {4} CompletedTime: {5} LastUpdated: {6} Status: {7} User Status: {8} Input: {9} Output: {10} Size: {11} CompressedSize: {12}",
+                State.OrchestrationInstance.InstanceId, State.OrchestrationInstance.ExecutionId, State.Name,
+                State.Version, State.CreatedTime, State.CompletedTime,
+                State.LastUpdatedTime, State.OrchestrationStatus, State.Status, State.Input, State.Output, State.Size,
+                State.CompressedSize);
+        }
 
-            returnValues.Add("InstanceId", new EntityProperty(State.OrchestrationInstance.InstanceId));
-            returnValues.Add("ExecutionId", new EntityProperty(State.OrchestrationInstance.ExecutionId));
+        #region
```
davidmrdavid, line 90: same comment on giving regions names so they can be self-documenting. otherwise, it's unclear why there's a region in the first place.

davidmrdavid (commented):
Mostly LGTM - just a few requests in terms of adding comments and naming regions. Afterwards, I'll be ready to approve.

davidmrdavid (approved):

PR #1116, opened 2024-06-14
HotFixes: Enable Support for Forward Compatibility  between DTFx.AzureStorage 1.x to 2.x

nytian:
DurableTask.AzureStorage v1.x and v2.x use different version of Azure Storage SDK which use different encoding methods. Consequently, when upgrading from v1 to v2, the app may be stuck with queue processing because workers loaded with DTFx.AS v2 cannot successfully read messages encoded with DTFx.AS v1. This PR introduces a hotfix to ensure forward compatibility.

End-to-end test has been made.

### DIFF
--- a/src/DurableTask.AzureStorage/MessageManager.cs
+++ b/src/DurableTask.AzureStorage/MessageManager.cs
@@ -161,9 +161,9 @@ namespace DurableTask.AzureStorage
             {
                 envelope = this.DeserializeMessageData(Encoding.UTF8.GetString(body));
             }
-            catch
+            catch(JsonReaderException)
             {
-                // This catch block is a hotfix for sev1 and better implementation might be needed in future. 
+                // This catch block is a hotfix and better implementation might be needed in future. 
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -156,7 +156,17 @@ public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMe
         {
             // TODO: Deserialize with Stream?
             byte[] body = queueMessage.Body.ToArray();
-            MessageData envelope = this.DeserializeMessageData(Encoding.UTF8.GetString(body));
+            MessageData envelope;
+            try
+            {
+                envelope = this.DeserializeMessageData(Encoding.UTF8.GetString(body));
+            }
+            catch
```
nytian, line 164: Do we want to add specific encoding exception here?

bachuv: Yes, let's add a specific exception.

davidmrdavid (approved):

bachuv (approved):

PR #1117, opened 2024-06-18
Run DTFx.Core tests in GitHub action

davidmrdavid:
As part of 1ES migration - replaces part of the "DTFx Build & Run tests" pipeline with a GitHub action, in particular, the DTFx.Core tests are replaced here. DTFx.AzureStorage cannot be replaced with a GH action until DTFx.AS v2 is released, where Azurite will be compatible with the tests - the current tests depend on the old Azure Storage emulator and that is not compatible with GitHub pipelines (because they run on ARM64)

### DIFF
--- a/.github/workflows/validate-build.yml
+++ b/.github/workflows/validate-build.yml
@@ -31,13 +31,13 @@ jobs:
       run: dotnet restore $solution
 
     - name: Build
-      run: dotnet build $solution --configuration $config #--no-restore -p:FileVersionRevision=$GITHUB_RUN_NUMBER -p:ContinuousIntegrationBuild=true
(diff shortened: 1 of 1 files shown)


nytian (approved):

.github/workflows/validate-build.yml
```diff
@@ -0,0 +1,60 @@
+name: Validate Build (DTFx.Core)
+
+on:
+  push:
+    branches:
+      - main
+      - 'dajusto/add-tests-as-gh-action'
```
bachuv, line 7: Should we remove this line before merging?

bachuv (approved):
LGTM! Just left one comment.

PR #1120, opened 2024-06-24
Enable DTFx.AzureStorage v1.x backward compatibility support

nytian:
DTFx.AzureStorage v1.x and v2.x use different Azure Storage SDK versions, each with its own encoding method. This PR adds backward compatibility to DTFx.AzureStorage v1.x, ensuring that queue messages created with DTFx.AS v2 remain readable after downgrading to DTFx.AS v1.

The `RawProperty` is an internal attribute of `CloudQueueMessage`, and we use System.Reflection to access it. The included catch block should only be triggered in the scenario described above.

End-to-End tests have been done.

### DIFF
--- a/src/DurableTask.AzureStorage/Storage/QueueMessage.cs
+++ b/src/DurableTask.AzureStorage/Storage/QueueMessage.cs
@@ -14,7 +14,6 @@
 namespace DurableTask.AzureStorage.Storage
 {
     using System;
-    using System.Text;
     using Microsoft.WindowsAzure.Storage.Queue;
 
     class QueueMessage
@@ -28,12 +27,15 @@ namespace DurableTask.AzureStorage.Storage
             }
             catch (FormatException)
             {
+                // This try-catch block ensures backward compatibility with DTFx.AzureStorage v1.x.
+                // It should only be executed in scenarios requiring this compatibility.
                 try
                 {
+                    // RawString is an internal property of CloudQueueMessage.
                     System.Reflection.PropertyInfo rawStringProperty = typeof(CloudQueueMessage).GetProperty("RawString", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
                     this.Message = (string)rawStringProperty.GetValue(this.CloudQueueMessage);
                 }
-                catch (Exception)
+                catch (Exception) // If failed to read the RawString property, throw an exception.
                 {
                     throw new InvalidOperationException("Failed to read the cloud queue message.");
                 }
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Storage/QueueMessage.cs
```diff
@@ -21,7 +21,25 @@ class QueueMessage
...
+                {
+                    // RawString is an internal property of CloudQueueMessage.
+                    System.Reflection.PropertyInfo rawStringProperty = typeof(CloudQueueMessage).GetProperty("RawString", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+                    this.Message = (string)rawStringProperty.GetValue(this.CloudQueueMessage);
+                }
+                catch (Exception) // If failed to read the RawString property, throw an exception.
+                {
+                    throw new InvalidOperationException("Failed to read the cloud queue message.");
+                }
+            }
```
davidmrdavid, line 42: some nits on wording. Also, if you're just going to re-throw the caught exception, you don't need to catch it in the first place :)

```suggestion
            catch (FormatException)
            {
                // This try-catch block ensures forwards compatibility with DTFx.AzureStorage v2.x, which does not guarantee base64 encoding of messages (messages not encoded at all).
                // Therefore, if we try to decode those messages as base64, we will have a format exception that will yield a poison message
                // RawString is an internal property of CloudQueueMessage, so we need to obtain it via reflection.
                System.Reflection.PropertyInfo rawStringProperty = typeof(CloudQueueMessage).GetProperty("RawString", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
                this.Message = (string)rawStringProperty.GetValue(this.CloudQueueMessage);
            }
```

davidmrdavid (commented):
Some minor feedback.
@nytian - Given this created an outage, I think we should have an automated test for this, either a unit test, or an E2E test. I think an E2E test, written as a github action, might be simplest. We just need to populate a DTFx. v1 queue with a DTFx v2 message, and see that it is able to be processed by DTFx.AS v1

nytian:
> Some minor feedback. @nytian - Given this created an outage, I think we should have an automated test for this, either a unit test, or an E2E test. I think an E2E test, written as a github action, might be simplest. We just need to populate a DTFx. v1 queue with a DTFx v2 message, and see that it is able to be processed by DTFx.AS v1

@davidmrdavid  what do you mean by a E2E test within GitHub actions? You want me to add a new folder/file that include the e2e test and add it to our workflow?

davidmrdavid:
@nytian - actually, let me take that back - I just realized that writing this E2E test will be difficult for DTFx.AS V1 since DTFx.AS V1 doesn't work well with Azurite. 

For now - don't worry about this. I trust your manual test, let's just make a note to perform another one right before release.

davidmrdavid (approved):

PR #1123, opened 2024-06-28
Fix Azure Storage HNS Support

wsugarman:
This PR fixes a bug when enumerating the lease blobs. When hierarchical namespaces (HNS) are enabled for a storage account, the "List Blobs" API returns the folder itself as a "blob" in the response (in addition to the blobs themselves), which results in a `NullReferenceException`. This PR skips those blobs by checking for the internal `hdi_isfolder` metadata property.

This PR also includes a refactor of how projects/filters are done on top of paginated data.


src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,137 @@
...
+                }
+
+                public async IAsyncEnumerator<Page<TResult>> GetAsyncEnumerator(CancellationToken cancellationToken = default(CancellationToken))
+                {
+                    await foreach (Page<TSource> page in this.source.WithCancellation(cancellationToken))
+                    {
+                        yield return Page<TResult>.FromValues(
+                            this.transform(page).ToList(),
+                            page.ContinuationToken,
+                            page.GetRawResponse());
```
cgillum, line 91: I'm not familiar with the AsyncPageable APIs. Can you provide a link to some docs explaining how `Page<TResult>.FromValues(...)` is intended to be used? I'm not quite sure I follow what's happening here.

(Though I now see that this is actually just a refactoring of existing code, so I assume there's no behavior change)

cgillum (approved):

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
wsugarman: No problem! This is an API introduced with Azure's new client SDKs that provides an asynchronous way to enumerate paginated data from Azure services (e.g. fetching the results from an Azure Table Storage query). [`AsyncPageable<T>`](https://learn.microsoft.com/en-us/dotnet/api/azure.asyncpageable-1?view=azure-dotnet) provides a method for fetching each "chunk" of data via `AsPages(string?, int?)` where each page is logically one call to a REST API with an optional continuation token (e.g. [List Blobs](https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?tabs=microsoft-entra-id)). For consumers that do not care that the result set is composed of multiple remote network calls, `AsyncPageable<T>` also implements `IAsyncEnumerable<T>` to abstract it all away.

For our use case, the DTFx code often records the number of storage API calls for the sake of telemetry, and so the preservation of pages is important for certain code paths.
E.g.
```csharp
await foreach (Page<T> p in pages)
{
    apiCalls++;
    // Process page
}
```

 However, there are instances where we also want to transform the results of the paginated data, such as projecting a different type or filtering their values. So, I came up with this workaround by applying some arbitrary transformation per page. This way we can both preserve the `AsyncPageable<T>` APIs while providing our own data types.

As for `Page<TResult>.FromValues`, this API is honestly used mostly for [testing](https://learn.microsoft.com/en-us/dotnet/azure/sdk/unit-testing-mocking?tabs=csharp), but it is also the easiest way to create arbitrary pages. I would have perhaps written a smarter [`Page<T>`](https://learn.microsoft.com/en-us/dotnet/api/azure.page-1?view=azure-dotnet) class that lazily transforms the results of each `Page<T>` instead of allocating a new `List<T>`, but the API doesn't really allow for it. `Page<T>` exposes its values as an `IReadOnlyList<T>` so this methodology seemed like the most straightforward.

src/DurableTask.AzureStorage/Storage/BlobContainer.cs
```diff
@@ -68,9 +70,12 @@ public async Task<bool> DeleteIfExistsAsync(string? appLeaseId = null, Cancellat
...
         {
-            return new AsyncPageableProjection<BlobItem, Blob>(
-                this.blobContainerClient.GetBlobsAsync(BlobTraits.Metadata, BlobStates.None, prefix, cancellationToken),
-                x => this.GetBlobReference(x.Name)).DecorateFailure();
+            return this.blobContainerClient
+                .GetBlobsAsync(BlobTraits.Metadata, BlobStates.None, prefix, cancellationToken)
+                .DecorateFailure()
+                .TransformPages(p => p.Values
+                    .Where(b => !IsHnsFolder(b))
+                    .Select(b => this.GetBlobReference(b.Name)));
```
davidmrdavid, line 78: nit - let's add an inline comment here documenting why we have the `IsHnsFolder` filter.
```suggestion
            //  When hierarchical namespaces (HNS) are enabled for a storage account, the "List Blobs" API returns the
            // folder itself as a "blob" in the response (in addition to the blobs themselves), 
            // which can result in a NullReferenceException if not accounted for. 
            // Therefore, we filter out all HNS folders in the result
            return this.blobContainerClient
                .GetBlobsAsync(BlobTraits.Metadata, BlobStates.None, prefix, cancellationToken)
                .DecorateFailure()
                .TransformPages(p => p.Values
                    .Where(b => !IsHnsFolder(b))
                    .Select(b => this.GetBlobReference(b.Name)));
```

src/DurableTask.AzureStorage/Storage/BlobContainer.cs
```diff
@@ -100,5 +105,14 @@ public Task RenewLeaseAsync(string leaseId, CancellationToken cancellationToken
...
         }
+
+        static bool IsHnsFolder(BlobItem item)
+        {
+            // See https://github.com/Azure/azure-sdk-for-python/issues/24814
+            return item.Metadata != null
+                && item.Metadata.TryGetValue("hdi_isfolder", out string value)
+                && bool.TryParse(value, out bool isFolder)
+                && isFolder;
+        }
```
davidmrdavid, line 116:
```suggestion
        static bool IsHnsFolder(BlobItem item)
        {
            // See https://github.com/Azure/azure-sdk-for-python/issues/24814
            // Checks if blob is a hierarchical namespace folder (HNS)
            return item.Metadata != null
                && item.Metadata.TryGetValue("hdi_isfolder", out string value)
                && bool.TryParse(value, out bool isFolder)
                && isFolder;
        }
```

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,137 @@
...
+                    }
+                }
+            }
+        }
+
+        sealed class AsyncPageableAsyncTransformation<TSource, TResult> : AsyncPageable<TResult>
```
davidmrdavid, line 97: I think this name is starting to get a bit too verbose, especially with the double "Async" there. Could we just call this something like `AsyncPageableProjection`?

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,137 @@
...
+            where TSource : notnull
+            where TResult : notnull
+        {
+            readonly AsyncPageable<TSource> source;
+            readonly Func<Page<TSource>, CancellationToken, IAsyncEnumerable<TResult>> transformAsync;
```
davidmrdavid, line 102: nit - here we use the term "transform" to describe the function, but earlier in this file we used "projection". I don't mind one or the other, but I think it would be better to keep the terms consistent. I imagine "transformAsync" is more intuitive than "projectAsync", which is more math-y. What do you think about renaming some variables in this file to use either "transform" or "project" but not both?

davidmrdavid (commented):
Thanks. I really like this change as well, but I have some suggestions here for readability.
I also suspect it would be good have caught this during testing. I'm curious - is Azurite even able to simulate HNS storage configurations? If so - it might be worth adding a test for that (we can take care of that if you don't have the cycles)

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
wsugarman: I actually did try to be intentional with my terminology here, but it may be confusing. Please let me know. For `SelectAsync`, this is really a _projection_ of types A to B, so I tried to convey that as much by using the term "project." The transformations on the other hand are a combination of arbitrary LINQ operations including projections, filters, aggregations, etc. So I tried to use a more generic term like _transformation_.

Let me change the word "projection" to like "selector."

wsugarman:
> Thanks. I really like this change as well, but I have some suggestions here for readability. I also suspect it would be good have caught this during testing. I'm curious - is Azurite even able to simulate HNS storage configurations? If so - it might be worth adding a test for that (we can take care of that if you don't have the cycles)

Unfortunately, Azurite does not support ADLS Gen 2, so we would need a proper integration test connecting to a real Azure Storage account with HNS enabled ☹️

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
davidmrdavid: For reference to other readers, that change can be seen here: https://github.com/Azure/durabletask/pull/1123/commits/d1623a9d26d3cdd6dd066a83898b2c8f3d908351

I like it. Does that mean we no longer need this class though?

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
davidmrdavid: I see now the difference you're making. Ok, I won't insist, it makes sense to me. My only recommendations is to add a comment next to the variable calling out the naming choice.

src/DurableTask.AzureStorage/Linq/AsyncEnumerableExtensions.cs
wsugarman: I just renamed the classes to be the same. They're `private` inner classes, so they can have name collisions without issue.

davidmrdavid (approved):

davidmrdavid:
Let's wait until GitHub action is executed on this branch

PR #1125, opened 2024-07-03
Remove net462 TFM

bachuv:
Removed .NET Framework 4.6.2 TFM and upgraded DurableTask.Core version to 3.0.0.

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -46,9 +46,6 @@
     <PackageReference Include="Azure.Storage.Queues" Version="12.18.0" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
   </ItemGroup>
 
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -2,7 +2,7 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
     <Description>Azure Storage provider extension for the Durable Task Framework.</Description>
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -48,11 +48,6 @@
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
-    <PackageReference Include="WindowsAzure.Storage" version="7.2.1" />
-  </ItemGroup>
-
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
```
cgillum, line 52: The expression `Condition="'$(TargetFramework)' == 'netstandard2.0'"` is always going to be true now, so let's consolidate all package references to be in the same `<ItemGroup>` above.

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -34,12 +34,6 @@
...
     <Version>$(VersionPrefix)-$(VersionSuffix)</Version>
   </PropertyGroup>
-  
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
-    <!--Newtonsoft.Json 7.0.1 is kept for backwards compatibility with Functions runtime V1, so we supress warning NU1903. -->
-    <PackageReference Include="Newtonsoft.Json" Version="7.0.1" NoWarn="NU1903" />
-  </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
cgillum, line 38: Similar comment as in the other csproj file, let's consolidate all the `<PackageReference .../>` nodes to be in the same `<ItemGroup>`

davidmrdavid (changes requested):
LGTM but let's make sure the GitHub actions are running on all PRs, as this has not had the validation executed on it :)

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
```diff
@@ -35,11 +35,7 @@ internal class AzureTableOrchestrationHistoryEventEntity : AzureTableCompositeTa
         private static readonly JsonSerializerSettings ReadJsonSettings = new JsonSerializerSettings
         {
             TypeNameHandling = TypeNameHandling.Objects,
-#if NETSTANDARD2_0
```
nytian, line 38: Just wondering can we update the code in ServiceBus? Doesn't this belong to other teams?

nytian (approved):
Thanks for the PR! Left a small question :)

src/DurableTask.ServiceBus/Tracking/AzureTableOrchestrationHistoryEventEntity.cs
davidmrdavid: ServiceBus is not actively maintained. It should be ok. I think you're thinking of ServiceFabric, which is maintained by another team.

davidmrdavid (approved):
I'll approve since the GitHub actions should run as soon as we try to merge to main. Let's open that PR asap

PR #1130, opened 2024-07-09
Poison message handling for DT.AS V2

bachuv:
This PR adds poison message handling for DurableTask.AzureStorage V2. There's a new `PoisonMessageDeuqueCountThreshold` setting in `AzureStorageOrchestrationServiceSettings` where the default is set to 20. If a message is dequeued and fails to deserialize more than the number that's set for `PoisonMessageDeuqueCountThreshold`, then it gets added to a new `<TaskHubName>Poison` table in Azure Storage. Customers can go to this table in their storage account and look at these poison messages.

### DIFF
--- a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
+++ b/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
@@ -106,28 +106,27 @@ namespace DurableTask.AzureStorage.Messaging
                             MessageData messageData;
                             try
                             {
+                                // Try to de-serialize the message
                                 messageData = await this.messageManager.DeserializeQueueMessageAsync(
                                     queueMessage,
                                     this.storageQueue.Name);
-                            }
-                            catch (Exception e)
-                            {
-                                // We have limited information about the details of the message
-                                // since we failed to deserialize it.
-                                this.settings.Logger.MessageFailure(
-                                    this.storageAccountName,
-                                    this.settings.TaskHubName,
-                                    queueMessage.MessageId /* MessageId */,
-                                    string.Empty /* InstanceId */,
-                                    string.Empty /* ExecutionId */,
-                                    this.storageQueue.Name,
-                                    string.Empty /* EventType */,
-                                    0 /* TaskEventId */,
-                                    e.ToString());
 
-                                // Abandon the message so we can try it again later.
-                                // Note: We will fetch the message again from the queue before retrying, so no need to read the receipt
-                                _ = await this.AbandonMessageAsync(queueMessage);
+                                // If successful, check if it's a poison message. If so, we handle it
+                                // and log metadata about it as the de-serialization succeeded.
+                                await this.HandleIfPoisonMessageAsync(messageData);
+                            }
+                            catch (Exception exception)
+                            {
+                                // Deserialization errors can be persistent, so we check if this is a poison message.
+                                bool isPoisonMessage = await this.TryHandlingDeserializationPoisonMessage(queueMessage, exception);
+                                if (isPoisonMessage)
+                                {
+                                    // we have already handled the poison message, so we move on.
+                                    return;
+                                }
+
+                                // This is not a poison message (at least not yet), so we abandon it to retry later.
+                                await this.AbandonMessageAsync(queueMessage, exception);
                                 return;
                             }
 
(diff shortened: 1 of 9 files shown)


src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -150,6 +152,96 @@ async Task<MessageData> AddMessageAsync(TaskMessage taskMessage, OrchestrationIn
...
+                    messageData.TaskMessage.Event.EventId,
+                    messageData.OriginalQueueMessage.MessageId,
+                    messageData.TaskMessage.OrchestrationInstance.InstanceId,
+                    messageData.TaskMessage.OrchestrationInstance.ExecutionId,
+                    this.Name,
+                    messageData.OriginalQueueMessage.DequeueCount);
+            }
+        }
+
+        public async Task<bool> TryHandlingDeserializationPoisonMessage(QueueMessage queueMessage, Exception deserializationException)
```
cgillum, line 202: It looks like there's a lot of redundancy between these two methods. Is it possible to refactor them to share some logic?

src/DurableTask.Core/History/HistoryEvent.cs
```diff
@@ -93,5 +93,16 @@ protected HistoryEvent(int eventId)
...
+        /// <summary>
+        /// Gets or sets whether this is a poison message.
+        /// </summary>
+        public bool IsPoison { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets user-facing details for why a message was labeled as poison.
+        /// This is to be set by each storage provider.
+        /// </summary>
+        public string PoisonGuidance { get; set; } = "";
```
cgillum, line 106: I'm not comfortable with adding these properties to every DTFx history event, especially if we don't have a plan or design for whether/how to implement poison message handling for other backend providers. For the purposes of this PR, I'd prefer we make changes only to DurableTask.AzureStorage unless we can get broad agreement from across the team that this is the right approach.

cgillum (changes requested):
Biggest blocker for me on this PR are the changes to DurableTask.Core. I don't think a case has been made for this yet, and it needs further discussion. At this point, my preference is to change only DurableTask.AzureStorage.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -192,8 +191,21 @@ await batch.ParallelForEachAsync(async delegate (QueueMessage queueMessage)
         }
 
         // This overload is intended for cases where we aren't able to deserialize an instance of MessageData.
-        public Task<UpdateReceipt?> AbandonMessageAsync(QueueMessage queueMessage)
+        public Task<UpdateReceipt?> AbandonMessageAsync(QueueMessage queueMessage, Exception exception)
         {
+            // We have limited information about the details of the message
+            // since we failed to deserialize it.
+            this.settings.Logger.MessageFailure(
```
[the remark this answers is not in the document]
davidmrdavid: I believe this overload is only called in cases where there's an exception, as you can see by the comment in line 193: https://github.com/Azure/durabletask/blob/934a5b673c52ee217df3de249498f982ef72652a/src/DurableTask.AzureStorage/Messaging/ControlQueue.cs#L192 . In that case, doesn't that mean the exception will never be null, @cgillum ?

src/DurableTask.Core/History/HistoryEvent.cs
davidmrdavid: Makes sense. Just to add more context here (may be obvious) - the reason for the DTFx.Core change here is solely that it allows us to _easily_ fail the function (activity or orchestrator) with the poison message. In the activity/orchestrator dispatchers, we are checking if `isPoison` is true and, if so, we're replacing the history event of the poison message with a failure event.

I assume that if instead of doing this, we replace the **message** itself with a failure event in the DTFx.AS level, we should be able to get away without this "isPoison" property. I recall trying to do this, and not finding it to be super obvious, but it's worth trying again. @cgillum if you have suggestions, I'm all ears!

Kencdk:
Instead of having a "one-size fits no-one" implementation; could this be extended to include an error handler which can be defined by developers instead, potentially with this as the default implementation?

That would allow developers to easily add new metrics, logs and store messages where they see fit, instead of having to rely on table storage and updates to the main library in case we would like more data readily available.

PR #1131, opened 2024-07-11
Add CodeQL support

davidmrdavid:
CodeQL (Code Query Language?) is a service that checks against CVEs and other compliance requirements using static analysis in our repos. It needs to be executed weekly.

Our 1ES pipeline already has this, but we need to make it run weekly, so our 1ES pipeline now has a weekly schedule.
We also need to add it directly on GitHub, which is done through the `codeQL.yml` GitHub workflow.

_In addition_, our open telemetry sample was throwing a warning because there was an `async` method nor await'ing anything. I've made the method synchronous to make the warning go away. See: https://github.com/Azure/durabletask/pull/1131/commits/8d2223a1ed472091a6bdae91041edf69a231b74d


nytian (approved):

.github/workflows/codeQL.yml
```diff
@@ -0,0 +1,79 @@
+# This workflow generates weekly CodeQL reports for this repo, a security requirements.
+# The workflow is adapted from the following reference: https://github.com/Azure-Samples/azure-functions-python-stream-openai/pull/2/files
+# Generic comments on how to modify these file are left intactfor future maintenance.
+
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main", "dajusto/code-ql", "*" ] # TODO: remove development branch after approval
```
cgillum, line 9: @davidmrdavid do the development branches still need to be removed?

davidmrdavid: Yes, I'll remove this in a follow up PR, totally blanked out, my mistake

PR #1136, opened 2024-07-15
Add Check for Lease Ownership Bug in Table Partition Manager

nytian:
This PR addresses a bug reported by customers involving the table partition manager. The issue occurs when a worker is the owner of a partition without claiming the lease. This can happen when a worker restarts or crashes and returns to a healthy state before the lease expires, thus it remains the owner of the partition. In our current logic, the worker only listens to a control queue when it has claimed a lease. Therefore, in the above scenario worker will continue to renew the lease without listening to queue messages.

To resolve this, the PR adds a check to ensure that when a worker reads the partition table and identifies itself as the current lease owner, it verifies if it is actively listening to the queue. If it is not, the worker should initiate the process to start listening to the queue. This ensures that the lease renewal logic only applies when the worker is actively managing the partition.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -574,12 +574,6 @@ namespace DurableTask.AzureStorage
             }
         }
 
-        internal bool CheckIfListeningToOwnedQueue(TablePartitionLease partition)
-        {
-            return this.allControlQueues.TryGetValue(partition.RowKey, out ControlQueue controlQueue) &&
-                this.OwnedControlQueues.Contains(controlQueue);
-        }
-
         internal Task OnOwnershipLeaseReleasedAsync(BlobPartitionLease lease, CloseReason reason)
         {
             this.orchestrationSessionManager.RemoveQueue(lease.PartitionId, reason, "Ownership LeaseCollectionBalancer");
@@ -588,6 +582,14 @@ namespace DurableTask.AzureStorage
 
         internal async Task OnTableLeaseAcquiredAsync(TablePartitionLease lease)
         {
+            // Check if the worker is already listening to the control queue.
+            if(this.allControlQueues.TryGetValue(lease.RowKey, out ControlQueue queue) &&
+                this.OwnedControlQueues.Contains(queue))
+            {
+                return;
+            }
+
+            // If the worker hasn't listened to the queue yet, then add it.
             var controlQueue = new ControlQueue(this.azureStorageClient, lease.RowKey, this.messageManager);
             await controlQueue.CreateIfNotExistsAsync();
             this.orchestrationSessionManager.AddQueue(lease.RowKey, controlQueue, this.shutdownSource.Token);
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -18,7 +18,6 @@ namespace DurableTask.AzureStorage
     using System.Diagnostics.Tracing;
     using System.Linq;
     using System.Net;
-    using System.Runtime.CompilerServices;
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -18,6 +18,7 @@ namespace DurableTask.AzureStorage
     using System.Diagnostics.Tracing;
     using System.Linq;
     using System.Net;
+    using System.Runtime.CompilerServices;
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -574,6 +574,12 @@ internal void DropLostControlQueue(TablePartitionLease partition)
             }
         }
 
+        internal bool CheckIfListeningToOwnedQueue(TablePartitionLease partition)
```
nytian, line 577: Let me know if you have thoughts about the method naming! :)

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -352,11 +352,18 @@ public async Task<ReadTableReponse> ReadAndWriteTableAsync(bool isShuttingDown,
...
 
                     // String previousOwner is for the steal process logs. Only used for stealing leases of any worker which is in shutdown process in this loop.
                     string previousOwner = partition.CurrentOwner ?? this.workerName;
 
+                    // If the partition is owned by this worker, check if the worker is listening to the owned control queue.
+                    if(partition.CurrentOwner == this.workerName)
+                    {
+                        isListeningToOwnedQueue = this.service.CheckIfListeningToOwnedQueue(partition);
+                    }
+
```
davidmrdavid, line 366: I'm a bit worried about adding more cases to this for-loop, so I'd prefer to re-use the existing `renewedLease` case/boolean to check if we're actively listening to the queue.

Can we use the `renewedLease` variable to enter the if-statement leading to `OnTableLeaseAcquiredAsync`. Then, inside `OnTableLeaseAcquiredAsync`, you could check, somehow, if the queue is already in our listened-to set?

davidmrdavid (commented):
Left a small comment. I'd also love to add some test here

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
nytian: Updated. Let me know if I understood it correctly.

test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
```diff
@@ -659,6 +659,78 @@ public async Task EnsureOwnedQueueExclusive()
...
+                // Worker[1] that shares name "0" should owned the partition without cliaming the lease. 
+                await Task.Delay(1000);
+                Assert.AreEqual(1, services[1].OwnedControlQueues.Count());
+
+                // Check all the instances could be processed successfully. 
+                OrchestrationState[] states = await Task.WhenAll(
+                    instances.Select(i => client.WaitForOrchestrationAsync(i, TimeSpan.FromSeconds(30))));
+                Assert.IsTrue(
+                    Array.TrueForAll(states, s => s?.OrchestrationStatus == OrchestrationStatus.Completed),
+                    "Not all orchestrations completed successfully!");
```
bachuv, line 730: Why does it say "Not all orchestration completed successfully"? Shouldn't they all complete successfully?

bachuv (commented):
Thanks for fixing this bug! I appreciate how you added comments in the code explaining the changes :)

test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
cgillum: This is the error message which will be written to the test logs if `Assert.IsTrue` fails (in other words, we're giving a clear indication of what the problem is: some orchestrations didn't complete).

cgillum (approved):
LGTM! I appreciate that you added a new test case to cover and validate this scenario.

davidmrdavid (changes requested):
Just want to make sure this works well for the case when we're supposed to be draining the queue. In that case - shouldn't we renew the lease but stop listening to new messages? I worry this PR may work against that.

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -407,7 +407,9 @@ public async Task<ReadTableReponse> ReadAndWriteTableAsync(bool isShuttingDown,
                             throw;
                         }
 
-                        if (claimedLease)
+                        // Worker should listen to the control queue if 1) worker just claimed the lease,
+                        // or 2) worker is already the lease owner but hasn't listened to the queue yet for some reason.
+                        if (claimedLease || renewedLease)
```
davidmrdavid, line 412: I _think_ this is what we want:

```suggestion
                        // Ensure worker is listening to the control queue iff either:
                        // 1) worker just claimed the lease,
                        // 2) worker was already the owner in the partitions table and is not actively draining the queue. Note that during draining, we renew the lease but do not want to listen to new messages. Otherwise, we'll never finish draining our in-memory messages.
                        bool isRenewingToDrainQueue = renewedLease & response.IsDrainingPartition
                        if (claimedLease || !isRenewingToDrainQueue)
```

davidmrdavid (commented):
I think we want to do the following. If this checks out to you, can you please help me with a test?

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
davidmrdavid: You did, thanks!

davidmrdavid:
Closing as it was merged here: https://github.com/Azure/durabletask/pull/1142

PR #1138, opened 2024-07-16
Convert timer target to UTC

### LINKED ISSUE #1127: System.NotSupportedException is thrown during task execution when a timer created with non UTC DateTime originally (v2.0.0-rcX).
We had a timer which was created with non UTC DateTime format:

await context.CreateTimer<object>(allocationResult.NextProjectedChange.Value.DateTime, null);

it used to be working with pre 2.0.0 DTF version but now it throws an exception, later in stack, after the timer is created and when a task is picked up for execution:

message": TaskOrchestrationDispatcher-b16f0d531284437aa1371a4493157b75-0: Unhandled exception with work item '22d6a199b42b4ab2ba41c4fc69ac7071:1': System.NotSupportedException: DateTime 06/26/2024 22:55:07 has a Kind of Unspecified. Azure SDK requires it to be UTC. You can call DateTime.SpecifyKind to change Kind property value to DateTimeKind.Utc.
   at Azure.Core.TypeFormatters.ToString(DateTime value, String format)
   at Azure.Core.Utf8JsonWriterExtensions.WriteObjectValue(Utf8JsonWriter writer, Object value)
   at Azure.Data.Tables.TableRestClient.CreateUpdateEntityRequest(String table, String partitionKey, String rowKey, Nullable`1 timeout, String ifMatch, IDictionary`2 tableEntityProperties, QueryOptions queryOptions)
   at Azure.Data.Tables.TableClient.CreateUpdateOrMergeRequest(TableRestClient batchOperations, ITableEntity entity, TableUpdateMode mode, ETag ifMatch)
   at Azure.Data.Tables.TableClient.BuildChangeSet(TableRestClient batchOperations, IEnumerable`1 batch, Dictionary`2 requestLookup, Guid batchId, Guid changesetId)
   at Azure.Data.Tables.TableClient.SubmitTransactionInternalAsync(IEnumerable`1 transactionalBatch, Guid batchId, Guid changesetId, Boolean async, CancellationToken cancellationToken)
   at Azure.Data.Tables.TableClient.SubmitTransactionAsync(IEnumerable`1 transactionActions, CancellationToken cancellationToken)
   at DurableTask.AzureStorage.Storage.ClientResponseExtensions.DecorateFailure[T](Task`1 responseTask) in /_/src/DurableTask.AzureStorage/Storage/ClientResponseExtensions.cs:line 48
   at DurableTask.AzureStorage.Storage.Table.ExecuteBatchAsync(IEnumerable`1 batchOperation, CancellationToken cancellationToken) in /_/src/DurableTask.AzureStorage/Storage/Table.cs:line 174
   at DurableTask.AzureStorage.Tracking.AzureTableTrackingStore.UploadHistoryBatch(String instanceId, String sanitizedInstanceId, String executionId, IList`1 historyEventBatch, StringBuilder historyEventNamesBuffer, Int32 numberOfTotalEvents, Int32 episodeNumber, Int32 estimatedBatchSizeInBytes, Nullable`1 eTagValue, Boolean isFinalBatch, CancellationToken cancellationToken) in /_/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs:line 1193
   at DurableTask.AzureStorage.Tracking.AzureTableTrackingStore.UpdateStateAsync(OrchestrationRuntimeState newRuntimeState, OrchestrationRuntimeState oldRuntimeState, String instanceId, String executionId, Nullable`1 eTagValue, Object trackingStoreContext, CancellationToken cancellationToken) in /_/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs:line 972
   at DurableTask.AzureStorage.AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync(TaskOrchestrationWorkItem workItem, OrchestrationRuntimeState newOrchestrationRuntimeState, IList`1 outboundMessages, IList`1 orchestratorMessages, IList`1 timerMessages, TaskMessage continuedAsNewMessage, OrchestrationState orchestrationState) in /_/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs:line 1218
   at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 612
   at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 211
   at DurableTask.Core.WorkItemDispatcher`1.ProcessWorkItemAsync(WorkItemDispatcherContext context, Object workItemObj) in /_/src/DurableTask.Core/WorkItemDispatcher.cs:line 459

While it might be considered not correct usage of the CreateTimer method it is still a breaking change, and the stack trace doesn't point to a correct place where the actual issue is. It would be helpful to add some check during a timer creation and throw an exception when a non-UTC date time provided.

AnatoliB:
Fixes #1127 

The old Azure SDK (e.g. WindowsAzure.Storage v9.3.1) silently converts non-UTC datetime values to UTC before persisting them in a table. As a result, passing non-UTC datetime to `CreateTimer` worked, even though this is not a recommended pattern.

The new Azure SDK (e.g. Azure.Data.Tables v12.8.3) throws an exception on attempts to persist non-UTC datetime values, which causes a hard-to-diagnose issue later, after a successful `CreateTimer` invocation.

This fix makes sure `CreateTimer` always converts the provided datetime value to UTC, to match the end-to-end behavior after the Azure SDK upgrade.

Since this is a potentially breaking change to both DurableTask.Core and DurableTask.AzureStorage, it should be released only with **DurableTask.Core v2** and **DurableTask.AzureStorage v2**.

**2024-07-18 update:** we decided to limit this fix to DurableTask.AzureStorage only, so no impact to DurableTask.Core


test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1285,6 +1285,28 @@ public async Task TimerExpiration(bool enableExtendedSessions)
...
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        [DataRow(false, false)]
+        public async Task TimerDelay(bool enableExtendedSessions, bool useUtc)
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
+            {
+                await host.StartAsync();
+
+                var now = useUtc ? DateTime.UtcNow : DateTime.Now;
```
davidmrdavid, line 1299:
```suggestion
                // by convention, DateTime objects are expected to be in UTC, but previous version of DTFx.AzureStorage
                // performed a implicit conversions to UTC when different timezones where used. This test ensures
                // that behavior is backwards compatible, despite not being recommended.
                var now = useUtc ? DateTime.UtcNow : DateTime.Now;
```

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -249,7 +249,7 @@ public override async Task<T> CreateTimer<T>(DateTime fireAt, T state, Cancellat
             var createTimerOrchestratorAction = new CreateTimerOrchestratorAction
             {
                 Id = id,
-                FireAt = fireAt,
+                FireAt = fireAt.ToUniversalTime(),
```
davidmrdavid, line 252: We have to be a bit careful here since we're changing behavior in DTFx.Core, which means it affects all backends, not just Azure Storage. For example - I'm not sure if Netherite performs an implicit conversion to universal time, or if it assumes all time is UTC.

This makes me think it may be safer to throw if the time is not in UTC. That way, it'll be simpler to predict the behavior. Not sure, would appreciate more thoughts

davidmrdavid (commented):
First - congrats on the contribution!

Left some preliminary comments

src/DurableTask.Core/TaskOrchestrationContext.cs
AnatoliB: Yes, I thought about it, and I'm still on the fence. I realize it will impact all backends, whether we throw or silently convert, and this is why it is intended to go to the next major version of DT.Core. I agree with the benefits of throwing an exception (after all, this is what the Azure SDK owners chose to do), but adhering to the old behavior would make the 
 migration from the previous version easier (for example, the original issue would not even be raised).

davidmrdavid: Another option would be not to do the conversion here (DTFx.Core), but to do it in DTFx.AzureStorage, at some point before we save state in the history/queue. By looking at the original trace, you may be able to spot where that conversion should be.

Test/DurableTask.Core.Tests/TaskOrchestrationContextTests.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace DurableTask.Core.Tests
+{
+    using DurableTask.Core.Command;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System;
+    using System.Linq;
+
+    [TestClass]
+    public class TaskOrchestrationContextTests
+    {
```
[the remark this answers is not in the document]
AnatoliB: This one is tricky. This cannot be a unit test for the `TaskOrchestrationContext` class, as this class just provides CurrentUtcDateTime as storage, but does not initialize this property. I'm trying to track this down... Alternatively, we can just do this check in a scenario test.

src/DurableTask.Core/TaskOrchestrationContext.cs
AnatoliB: > Another option would be not to do the conversion here (DTFx.Core), but to do it in DTFx.AzureStorage, at some point before we save state in the history/queue. By looking at the original trace, you may be able to spot where that conversion should be.

Yes, this is an option. What I didn't like about it is that it will require checking every event type and explicit type casting, probably around here: https://github.com/Azure/durabletask/blob/d4b98d46158e470c6b769601cd76a86d9bb6d096/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L1020
However, we already do type-specific processing there, so this patch will not make it significantly worse.

Ok, this will limit the impact to the AzureStorage backend. Now, the question is: do we _want_ to limit this to the Azure Storage backend? Regardless of the backend and the current behavior, if the customer is passing non-UTC time to `CreateTimer`, chances are they may be confused. Given that UTC is expected in many other parts, including DT.Core (such as context.CurrentUtcDateTime), do we want them to continue mixing-and-matching? Accidentally, we are bumping the DT.Core version now, so this is a good opportunity to make an intentional breaking change.

cgillum: Unless we have a concrete (non-theoretical) reason to change DurableTask.Core, I suggest we avoid it and only change DurableTask.AzureStorage. I personally think it's fine if users want to pass local times to our APIs. I consider the fact that we internally save those timestamps as UTC as an internal implementation detail. Since APIs like `context.CurrentUtcDateTime` already make it clear that the timestamp returned is in UTC, I don't think we need to be concerned about confusing them.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1285,6 +1285,30 @@ public async Task TimerExpiration(bool enableExtendedSessions)
             }
         }
 
+        [DataTestMethod]
+        [DataRow(true, true)]
+        [DataRow(true, false)]
+        [DataRow(false, true)]
+        [DataRow(false, false)]
+        public async Task TimerDelay(bool enableExtendedSessions, bool useUtc)
```
cgillum, line 1293: I suggest removing the `enableExtendedSessions` flag from this test. I don't expect it will have any impact, and removing this permutation will help save some time.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1285,6 +1285,30 @@ public async Task TimerExpiration(bool enableExtendedSessions)
...
+        {
+            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(enableExtendedSessions))
+            {
+                await host.StartAsync();
+                // by convention, DateTime objects are expected to be in UTC, but previous version of DTFx.AzureStorage
+                // performed a implicit conversions to UTC when different timezones where used. This test ensures
+                // that behavior is backwards compatible, despite not being recommended.
+                var now = useUtc ? DateTime.UtcNow : DateTime.Now;
```
cgillum, line 1301: The problem with using `DateTime.Now` in this test is that CI machines are likely going to be configured with the local time as being UTC, making this test somewhat useless. My recommendation would be to test with two specific timezones: one that's ahead of UTC and one that's behind. There are some StackOverflow answers that provide some tips on how you can manufacture `DateTime` objects in specific time zones, which may be useful: https://stackoverflow.com/q/246498/2069. We should then conform that the orchestration completes roughly in the amount of time expected. You're currently asserting that it completes in less than 30 seconds, which covers one case, but we should also assert that the orchestration completes in *no less than 3 seconds*, covering the other case.

src/DurableTask.Core/TaskOrchestrationContext.cs
AnatoliB: Reverted DurableTask.Core changes, so this fix is limited to DurableTask.AzureStorage now.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
AnatoliB: Done

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
AnatoliB: @cgillum I have no problem manufacturing `DateTime` objects in specific time zones _in the test code_, but how do we make the _production_ code respect the intended timezone, if it's not the one configured on the machine? We would need to pass it to the framework somehow (for example, by letting the user pass `DataTimeOffset` to `CreateTimer`). Any ideas?

Adding the "no less than" check.

AnatoliB: > Adding the "no less than" check.

I'm glad I did. The non-UTC case finishes almost immediately, no delay. When I return to the original fix (conversion in DT.Core), the delay works properly. However, the new DT.AS fix makes sure the table update succeeds, but there is no delay now. We are missing something else, perhaps yet another place to convert.

cgillum: > how do we make the production code respect the intended timezone

I'm not sure if I know the answer to this question. This might reflect my lack of understanding of how `DateTime` handles foreign time zones. I assumed from that StackOverflow link that it was possible without using `DateTimeOffset`, but I could be wrong. I was hoping we could write a test that creates a `DateTime` with `DateTime.UtcNow.AddSeconds(3)`, convert that into some foreign time zone, call `CreateTime(...)` with that non-UTC `DateTime`, and then validate that the orchestration correctly completes in ~3 seconds. Feel free to tell me if that's not possible.

cgillum: > We are missing something else, perhaps yet another place to convert.

Looking at the code changes again, it seems you're only updating the history events that get saved to the History table. However, there is another copy of these history events that gets serialized into queue messages, and it's the queue message version which actually controls the delay. We might need to make the code change in AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync, which is where we commit new history events both to table storage and to queue storage.

AnatoliB: > > how do we make the production code respect the intended timezone
> 
> I'm not sure if I know the answer to this question. This might reflect my lack of understanding of how `DateTime` handles foreign time zones. I assumed from that StackOverflow link that it was possible without using `DateTimeOffset`, but I could be wrong. I was hoping we could write a test that creates a `DateTime` with `DateTime.UtcNow.AddSeconds(3)`, convert that into some foreign time zone, call `CreateTime(...)` with that non-UTC `DateTime`, and then validate that the orchestration correctly completes in ~3 seconds. Feel free to tell me if that's not possible.

The problem is that the `DateTime` object does not store the timezone information, by design. It only has a datetime value and an indicator of whether this value is supposed to be treated as local or UTC (the `Kind` property). When `Kind` is `DateTimeKind.Local`, there is no way to know how to properly convert it to UTC - the timezone information is lost. The code converting to UTC needs to either get the timezone information from a different source, or just assume the current timezone (which is what `DateTime.ToUniversalTime` does). I believe this is the main reason why `DateTimeOffset` was introduced. The code at that StackOverflow link shows how to construct `DateTime` objects for an arbitrary timezone, _assuming that the code that is supposed to convert it has a way to know what this timezone was_. This is why they bundle both `DateTime` and `TimeZoneInfo` into a single struct, assuming this association can be somehow preserved.

But I understand the value of the test you are proposing. I see two ways to implement it:
- Let `CreateTimer` take a `DateTimeOffset` object (but then we'll not be testing the original `CreateTimer` behavior).
- Invent another way to pass the desired timezone to the production code, either an "official" one or for testing purposes only.

AnatoliB: > We might need to make the code change in AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync, which is where we commit new history events both to table storage and to queue storage.

Yes, conversion was needed in both places. It works now, thank you!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1106,6 +1106,11 @@ async Task AbandonAndReleaseSessionAsync(OrchestrationSession session)
             TaskMessage continuedAsNewMessage,
             OrchestrationState orchestrationState)
         {
+            foreach (var timerMessage in timerMessages)
+            {
+                Utils.ConvertTimeToUtc(timerMessage.Event);
+            }   
+
```
davidmrdavid, line 1113: nit - would like to document this change so we know why this is DTFx.AS-specific
```suggestion
        {
            // for backwards compatibility, we transform timestamps to UTC prior to persisting in AzStorage.
            // see: https://github.com/Azure/durabletask/pull/1138
            foreach (var timerMessage in timerMessages)
            {
                Utils.ConvertTimeToUtc(timerMessage.Event);
            }   

```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -846,6 +846,7 @@ public override Task StartAsync(CancellationToken cancellationToken = default)
                 bool isFinalEvent = i == newEvents.Count - 1;
 
                 HistoryEvent historyEvent = newEvents[i];
+                Utils.ConvertTimeToUtc(historyEvent);
```
davidmrdavid, line 849: same as above, just a small nit for our own future reference:
```suggestion
                // For backwards compatibility, we convert timer timestamps to UTC prior to persisting to Azure Storage
                // see: https://github.com/Azure/durabletask/pull/1138
                Utils.ConvertTimeToUtc(historyEvent);
```

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -272,5 +272,27 @@ public static object DeserializeFromJson(JsonSerializer serializer, string jsonS
             }
             return obj;
         }
+
+        public static void ConvertTimeToUtc(HistoryEvent historyEvent)
```
davidmrdavid, line 276: ultra nitpick - this is specifically converts timer events to Utc. Wondering if we should rename this to reflect that. Non-blocking thought

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -3304,11 +3333,11 @@ public override Task<DateTime> RunTask(OrchestrationContext context, string inpu
                 }
             }
 
-            internal class DelayedCurrentTimeInline : TaskOrchestration<DateTime, string>
+            internal class DelayedCurrentTimeInline : TaskOrchestration<DateTime, DateTime>
```
davidmrdavid, line 3336: so this class wasn't being used? Just surprised we changed the signature and no other tests needed to be refactored.

src/DurableTask.AzureStorage/Utils.cs
```diff
@@ -272,5 +272,27 @@ public static object DeserializeFromJson(JsonSerializer serializer, string jsonS
...
+
+        public static void ConvertTimerEventsToUTC(HistoryEvent historyEvent)
+        {
+            switch (historyEvent.EventType)
+            {
+                case EventType.TimerCreated:
+                    var timerCreatedEvent = (TimerCreatedEvent)historyEvent;
+                    if (timerCreatedEvent.FireAt.Kind != DateTimeKind.Utc)
+                    {
+                        timerCreatedEvent.FireAt = timerCreatedEvent.FireAt.ToUniversalTime();
```
davidmrdavid, line 284: @AnatoliB - `ToUniversalTime()` assumes that if the `Kind` property is unspecified, it is local. Just for safety - does the Azure Storage V1 SDK make the same assumption?
In other words, if we pass a `DateTime` of unspecified `Kind`, does it translate it to `UTC` by assuming it is using local time?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
davidmrdavid: > Let CreateTimer take a DateTimeOffset object (but then we'll not be testing the original CreateTimer behavior).
I'd advice against doing that in tihs PR, for the same reasons you list - we would not be testing the original `CreateTimer`, defeating the purpose of the backwards compatibility test.

So long as we can make an answer [this](https://github.com/Azure/durabletask/pull/1138/files#r1686842547) question, then I think the change we're doing here is be safe.

I did a quick search of the [OrchestrationContext](https://github.com/Azure/durabletask/blob/ac82c7f1cc37eec4e728f750ca013bef9231bf14/src/DurableTask.Core/OrchestrationContext.cs#L27) class as well and it seems this is the only orchestration API where we pass in a `DateTIme` object, so I think we're covered from _there_.

But the `TaskHubClient` does API a few more `DateTime` APIs that I now wonder whether they need updating with this same backwards compatibility. In theory, I think the answer is _yes_, but it does expand the scope of what we were originally envisioning

davidmrdavid: If we want to add backwards compatibility to all APIs that accept a `DateTime` parameter, I think the easiest way would be to add more cases to your `ConvertTimerEventsToUTC` method: effectively to catch all `HistoryEvent` types that have a `DateTime` property in there.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
AnatoliB: That's correct, it was not used, so I took it over for this test purposes.

src/DurableTask.AzureStorage/Utils.cs
AnatoliB: Yes, I discovered this by manual testing.

AnatoliB: Strictly speaking, this logic is in the .NET implementation (https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/TimeZoneInfo.cs,a3145a8ab7f5e115), and the Azure Storage V1 SDK probably just invokes this method.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
davidmrdavid: I don't think that should be needed, at least I won't ask for every single history-event to be tested. I'm ok with the testing done so far.

davidmrdavid: So I think taht so long as you add more HistoryEvent cases (look for the sub-classes of `HistoryEvent` that have `DateTime` props), I'd be ready to approve.

davidmrdavid: I suppose it also depends on how many more cases we have to add. I'm hoping it will just be 1 or 2 more.

cgillum: > The problem is that the DateTime object does not store the timezone information, by design. It only has a datetime value and an indicator of whether this value is supposed to be treated as local or UTC (the Kind property). When Kind is DateTimeKind.Local, there is no way to know how to properly convert it to UTC - the timezone information is lost. The code converting to UTC needs to either get the timezone information from a different source, or just assume the current timezone (which is what DateTime.ToUniversalTime does). I believe this is the main reason why DateTimeOffset was introduced.

Got it. So, it sounds like what I proposed isn't really possible. In that case, I'm fine with the test code as it is today. I agree with David that it doesn't make sense to introduce new product changes for this.

AnatoliB: > I suppose it also depends on how many more cases we have to add. I'm hoping it will just be 1 or 2 more.

Found just one: `ExecutionStartedEvent.ScheduledStartTime`. Hopefully, not hidden in nested data structures somewhere else.

davidmrdavid: Looking at the references of `ExecutionStartedEvent.ScheduledStartTime.get` will show us the right places.
In particular, this "scheduled start time" is used when creating orchestrators, so you'll have to modify durableClient code **as well,** which falls outside the `CompleteTaskOrchestrationWorkItemAsync`.

I think you want to modify `CreateTaskOrchestrationAsync` (the "creationMessage" is the message to convert). Other than that, the `orchestratorMessages` array in `CompleteTaskOrchestrationWorkItemAsync` should indeed contain any `ExecutionStartedEvent`s generated by orchestrators or entities.

In summary, you just need to modify `CreateTaskOrchestrationAsync` as well

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -1285,6 +1285,34 @@ public async Task TimerExpiration(bool enableExtendedSessions)
...
+
+                var actualDelay = DateTime.UtcNow - startTime.ToUniversalTime();
+                Assert.IsTrue(
+                    actualDelay >= delay && actualDelay < delay + TimeSpan.FromSeconds(10),
+                    $"Expected delay: {delay}, ActualDelay: {actualDelay}");
+
+                await host.StopAsync();
+            }
+        }
+
```
davidmrdavid, line 1315: nit - proposed unit test for the `ExecutionStarted` case
```suggestion
        [DataTestMethod]
        [DataRow(false)]
        [DataRow(true)]
        public async Task OrchestratorStartAtAcceptsAllDateTimeKinds(bool useUtc)
        {
            using (TestOrchestrationHost host = TestHelpers.GetTestOrchestrationHost(false))
            {
                await host.StartAsync();
                // by convention, DateTime objects are expected to be in UTC, but previous version of DTFx.AzureStorage
                // performed a implicit conversions to UTC when different timezones where used. This test ensures
                // that behavior is backwards compatible, despite not being recommended.

                // set up orchestrator start time
                var currentTime = DateTime.Now;
                var delay = TimeSpan.FromSeconds(5);
                var startAt = currentTime.Add(delay);

                if (useUtc)
                {
                    startAt.ToUniversalTime();
                }


                var client = await host.StartOrchestrationAsync(typeof(Orchestrations.JustReturnCurrentUtcTime), input: string.Empty, startAt: startAt);

                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30));
                Assert.AreEqual(OrchestrationStatus.Completed, status?.OrchestrationStatus);

                var orchestratorState = await client.GetStateAsync(client.InstanceId);
                var actualScheduledStartTime = status.ScheduledStartTime;

                // internal representation of DateTime is always UTC
                var expectedScheduledStartTime = startAt.ToUniversalTime();
                Assert.AreEqual(expectedScheduledStartTime, actualScheduledStartTime);
                await host.StopAsync();
            }
        }
```

davidmrdavid (commented):
Running the CI on this branch shortly, let's wait for that to go green before merging. Unsure why it did not run automatically

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
AnatoliB: I believe `ToUniversalTime` doesn't change the object in-place, it returns a new object, so line 1335 will have no effect.

davidmrdavid (approved):
LGTM :)

PR #1139, opened 2024-07-16
Add 1ES CI for DTFx.AS v2

davidmrdavid:
The primary goal of this PR is to add a public 1ES ADO testing CI for DTFx.AS v2. This pipeline's results (and a messy development history) can be found here: https://dev.azure.com/azfunc/public/_build?definitionId=881&_a=summary

While migrating our CI pipeline, I encountered previously stable tests becoming flaky, as well as the fact that some tests were previously not running (and were failing once enabled) plus other 1ES-specific idiosyncrasies. This PR aims to resolve them all. 

To help with reviews, I'll add comments throughout the diff to clarify why some part of the code had to be changed. I feel comfortable about most changes except for one: the addition of  `System.Runtime.CompilerServices.Unsafe  v6.0.0` as an explicit dependency in DTFx.Core.

Without this change, some DTFx.Core would fail with the following exception trace:

```
Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Could not find or load a specific file. (0x80131621)
File name: 'System.Runtime.CompilerServices.Unsafe, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
at System.Diagnostics.ActivityTraceId.SetToRandomBytes(Span1 outBytes)    at System.Diagnostics.ActivitySpanId.CreateRandom()    at DurableTask.Core.TaskOrchestrationDispatcher.ProcessScheduleTaskDecision(ScheduleTaskOrchestratorAction scheduleTaskOrchestratorAction, OrchestrationRuntimeState runtimeState, Boolean includeParameters, Activity parentTraceActivity) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 1016    at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 412    at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 598    at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 211    at DurableTask.Core.WorkItemDispatcher1.ProcessWorkItemAsync(WorkItemDispatcherContext context, Object workItemObj) in /_/src/DurableTask.Core/WorkItemDispatcher.cs:line 374
```

This exception occurs in the Distributed Tracing path, as you can see by the mention of `System.Diagnostics.ActivitySpanId` in the stack, which is a type we get from our `System.Diagnostics.DiagnosticSource` dependency.

I tried researching this, and found this [GH thread](https://github.com/dotnet/runtime/issues/67196#issuecomment-1081657399) with the following excerpt:

> Unsafe package isn't referenced anymore on netstandard2.0 or net462 which means that consumers of this package now get the transitive Unsafe dependency of System.Memory which is super old (4.5.3 vs 6.0.0).

Since we recently changed DTFx.Core to use only netstandard2.0 (as per https://github.com/Azure/durabletask/pull/1125/files), it's possible this is the reason why this error is only showing up now. Unfortunately, I'm unable to reproduce this locally, but this reproduces pretty consistently on the 1ES CI. 

**As a relevant aside:** Durable Functions v2 has been incompatible with Azure Functions v1 for a while now. We never got to the bottom of that investigation, but I have a rough recollection that the reason for it was an exception _like this_ being thrown in Functions V1. Perhaps that's worth revisiting.

In any case - I need some .NET expertise to help me validate what's the problem here and if the right solution is indeed to add the explicit dependency to `CompilerServices.Unsafe v6.x`.

Other than that, I'll annotate the PR below. Thanks

_See the CI passing for the latest commit here: https://dev.azure.com/azfunc/public/_build/results?buildId=172877&view=results_

### DIFF
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -53,7 +53,8 @@ extends:
     - stage: DTFxCoreValidate
       jobs:
       - job: Validate
-
+        strategy:
+          parallel: 13
         steps:
         # Build the code and the tests
         - template: /eng/templates/build-steps.yml@self
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -68,7 +68,7 @@ extends:
         - template: /eng/templates/test.yml@self
           parameters:
             # TODO: for some reason we're unable to run the net462 tests in the CI
-            testAssembly: '**\bin\Debug\netcoreapp3.1\DurableTask.Core.Tests.dll'
+            testAssembly: '**\bin\Debug\**\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
       jobs:
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -67,7 +67,7 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\**\DurableTask.Core.Tests.dll'
+            testAssembly: '**\bin\Debug\netcoreapp3.1\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
       jobs:
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -68,7 +68,7 @@ extends:
         - template: /eng/templates/test.yml@self
           parameters:
             # TODO: for some reason we're unable to run the net462 tests in the CI
-            testAssembly: '**\bin\Debug\**\DurableTask.Core.Tests.dll'
+            testAssembly: '**\bin\**\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
       jobs:
@@ -87,7 +87,7 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\Debug\**\DurableTask.AzureStorage.Tests.dll'
+            testAssembly: '**\bin\**\DurableTask.AzureStorage.Tests.dll'
     - stage: DTFxEmulatorValidate
       dependsOn: []
       jobs:
@@ -106,4 +106,4 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\Debug\**\DurableTask.Emulator.Tests.dll'
+            testAssembly: '**\bin\**\DurableTask.Emulator.Tests.dll'
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -31,7 +31,7 @@ resources:
     ref: refs/tags/release
 
 extends:
-  # The template we extend injects compliance-checks into the pipleine, such as SDL and CodeQL
+  # The template we extend injects compliance-checks into the pipeline, such as SDL and CodeQL
   template: v1/1ES.Unofficial.PipelineTemplate.yml@1es
   parameters:
     pool:
@@ -67,7 +67,6 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            # TODO: for some reason we're unable to run the net462 tests in the CI
             testAssembly: '**\bin\**\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -67,7 +67,8 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\**\netcoreapp3.1\DurableTask.Core.Tests.dll'
+            # TODO: for some reason we're unable to run the net462 tests in the CI
+            testAssembly: '**\bin\Debug\netcoreapp3.1\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
       jobs:
@@ -86,7 +87,7 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\**\netcoreapp3.1\DurableTask.AzureStorage.Tests.dll'
+            testAssembly: '**\bin\Debug\**\DurableTask.AzureStorage.Tests.dll'
     - stage: DTFxEmulatorValidate
       dependsOn: []
       jobs:
@@ -105,4 +106,4 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\**\netcoreapp3.1\DurableTask.Emulator.Tests.dll'
+            testAssembly: '**\bin\Debug\**\DurableTask.Emulator.Tests.dll'
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -53,8 +53,7 @@ extends:
     - stage: DTFxCoreValidate
       jobs:
       - job: Validate
-        strategy:
-          parallel: 13
+
         steps:
         # Build the code and the tests
         - template: /eng/templates/build-steps.yml@self
@@ -67,7 +66,7 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\Debug\**\DurableTask.Core.Tests.dll'
+            testAssembly: '**\bin\**\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
       jobs:
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -67,7 +67,7 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\Debug\netcoreapp3.1\DurableTask.Core.Tests.dll'
+            testAssembly: '**\bin\Debug\**\DurableTask.Core.Tests.dll'
     - stage: DTFxASValidate
       dependsOn: []
       jobs:
--- a/eng/templates/test.yml
+++ b/eng/templates/test.yml
@@ -20,16 +20,14 @@ steps:
   - task: VSTest@2
     displayName: 'Run tests'
     inputs:
-      testAssemblyVer2: |
-        ${{ parameters.testAssembly }}
-        !**\obj\**
-        testFiltercriteria: 'TestCategory!=DisabledInCI'
-        vsTestVersion: 17.0
-        distributionBatchType: basedOnExecutionTime
-        platform: 'any cpu'
-        configuration: 'Debug'
-        diagnosticsEnabled: True
-        collectDumpOn: always
-        rerunFailedTests: true
-        rerunFailedThreshold: 30
-        rerunMaxAttempts: 3
+      testAssemblyVer2: ${{ parameters.testAssembly }}
+      testFiltercriteria: 'TestCategory!=DisabledInCI'
+      vsTestVersion: 17.0
+      distributionBatchType: basedOnExecutionTime
+      platform: 'any cpu'
+      configuration: 'Debug'
+      diagnosticsEnabled: True
+      collectDumpOn: always
+      rerunFailedTests: true
+      rerunFailedThreshold: 30
+      rerunMaxAttempts: 3
--- a/eng/templates/test.yml
+++ b/eng/templates/test.yml
@@ -17,7 +17,7 @@ steps:
     displayName: 'Install and Run Azurite'
 
   # Run tests
-  - task: VSTest@3
+  - task: VSTest@2
     displayName: 'Run tests'
     inputs:
       testAssemblyVer2: ${{ parameters.testAssembly }}
--- a/eng/templates/test.yml
+++ b/eng/templates/test.yml
@@ -17,7 +17,7 @@ steps:
     displayName: 'Install and Run Azurite'
 
   # Run tests
-  - task: VSTest@2
+  - task: VSTest@3
     displayName: 'Run tests'
     inputs:
       testAssemblyVer2: ${{ parameters.testAssembly }}
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -40,7 +40,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.40.0" />
+    <PackageReference Include="Azure.Core" Version="1.41.0" />
     <PackageReference Include="Azure.Data.Tables" Version="12.8.3" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.20.0" />
     <PackageReference Include="Azure.Storage.Queues" Version="12.18.0" />
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -46,7 +46,7 @@
     <PackageReference Include="Azure.Storage.Queues" Version="12.18.0" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -13,7 +13,8 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>embedded</DebugType>
     <IncludeSymbols>false</IncludeSymbols>
-    <!--NuGet licenseUrl and PackageIconUrl/iconUrl deprecation. -->
+	<PackageReadmeFile>.\README.md</PackageReadmeFile>
+	<!--NuGet licenseUrl and PackageIconUrl/iconUrl deprecation. -->
     <NoWarn>NU5125;NU5048;CS7035</NoWarn> <!-- TODO: addition of CS7035 (version format doesn't follow convention) is a temporary workaround during 1ES migration -->
   </PropertyGroup>
 
@@ -51,6 +52,10 @@
 
   <ItemGroup>
     <ProjectReference Include="..\DurableTask.Core\DurableTask.Core.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include=".\..\..\README.md" Pack="true" PackagePath="\"/>
   </ItemGroup>
 
   <ItemGroup Condition="'$(Configuration)'=='Release'">
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -13,7 +13,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>embedded</DebugType>
     <IncludeSymbols>false</IncludeSymbols>
-	<PackageReadmeFile>.\README.md</PackageReadmeFile>
+    <PackageReadmeFile>.\README.md</PackageReadmeFile>
 	<!--NuGet licenseUrl and PackageIconUrl/iconUrl deprecation. -->
     <NoWarn>NU5125;NU5048;CS7035</NoWarn> <!-- TODO: addition of CS7035 (version format doesn't follow convention) is a temporary workaround during 1ES migration -->
   </PropertyGroup>
@@ -41,13 +41,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.41.0" />
+    <PackageReference Include="Azure.Core" Version="1.40.0" />
     <PackageReference Include="Azure.Data.Tables" Version="12.8.3" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.20.0" />
     <PackageReference Include="Azure.Storage.Queues" Version="12.18.0" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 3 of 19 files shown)


src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -40,7 +40,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.40.0" />
+    <PackageReference Include="Azure.Core" Version="1.41.0" />
```
davidmrdavid, line 43: 1ES reports Azure.Core 1.40.0 as vulnerable

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -513,7 +513,7 @@ public override async IAsyncEnumerable<OrchestrationState> GetStateAsync(IEnumer
                 yield break;
             }
 
-            IEnumerable<Task<OrchestrationState>> instanceQueries = instanceIds.Select(instance => this.GetStateAsync(instance, allExecutions: true, fetchInput: false, cancellationToken).SingleAsync().AsTask());
+            IEnumerable<Task<OrchestrationState>> instanceQueries = instanceIds.Select(instance => this.GetStateAsync(instance, allExecutions: true, fetchInput: false, cancellationToken).SingleOrDefaultAsync().AsTask());
```
davidmrdavid, line 516: There used to be a bug here (which the new timing differences in 1ES revealed). It's possible for `GetStateAsync` to return an empty collection (if there's no state in the instance table for a given ID yet). If that's the case, then `SingleAsync` will throw an error of the type `Sequence contains no elements`. 

By using `SingleOrDefaultAsync`, we make DTFx.Core tolerate this case and return `null` if the collection is empty. That allows us to more gracefully exit this method in line ~522 below.

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2252,6 +2252,7 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
         /// End-to-end test which validates a simple orchestrator function that calls an activity function
         /// and checks the OpenTelemetry trace information
         /// </summary>
+        [TestCategory("DisabledInCI")]
```
davidmrdavid, line 2255: these disabled tests have **not** been running for a long time, and they're failing in the main branch as well. They weren't running because they're conditionally compiled not to run in NET462, and our previous testing task in ADO was configured to run tests found in the `net462` dll.

Now, the test task does discover these tests, but I've disabled them as they aren't passing. @bachuv can you please add this as a To-Fix before Dist. Tracing GA?

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -5,28 +5,31 @@
...
   <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
     <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" version="3.2.1" />
     <PackageReference Include="WindowsAzure.Storage" version="8.7.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net462'">
     <PackageReference Include="Azure.Monitor.OpenTelemetry.Exporter" Version="1.0.0-beta.3" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.6" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
davidmrdavid, line 18: newtonsoft.json is inherited by the DTFx.Core project

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -5,28 +5,31 @@
...
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.5.0" />
     <PackageReference Include="WindowsAzure.Storage" version="9.3.3" />
   </ItemGroup>
 
   <!-- Common package dependencies -->
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.2.0" />
```
davidmrdavid, line 22: I added this `Logging.Console` dependency because I've configured the tests to emit traces to STDOUT, which makes debugging failed ADO tests much easier.  I decided to unify the dependencies to version 2.2.0 for simplicy sake and to avoid conditional compilation.

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
```diff
@@ -48,9 +51,6 @@
     <None Update="large.jpeg">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
-    <None Update="testhost.dll.config" Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
```
davidmrdavid, line 53: we don't have a netcoreapp2.1 TFM

test/DurableTask.AzureStorage.Tests/TestHelpers.cs
```diff
@@ -40,9 +41,9 @@ static class TestHelpers
                 TaskHubName = GetTestTaskHubName(),
 
                 // Setting up a logger factory to enable the new DurableTask.Core logs
-                // TODO: Add a logger provider so we can collect these logs in memory.
-                LoggerFactory = new LoggerFactory(),
+                LoggerFactory = new LoggerFactory().AddConsole(LogLevel.Trace),
             };
+#pragma warning restore CS0618 // Type or member is obsolete
```
davidmrdavid, line 46: as mentioned, I've added logging to STDOUT on our tests to facilitate debugging. The current logging library we use is pretty old, which is why we have to disable the warning here. I tried using a newer library, but then had issues with the NET462 TFM. For speed, I decided to use the old library instead, which works.

test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj
```diff
@@ -5,19 +5,13 @@
     <TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="EnterpriseLibrary.SemanticLogging.EventSourceAnalyzer" Version="2.0.1406.1" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
-    <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
-    <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
-  </ItemGroup>
-
```
davidmrdavid, line 14: there was no reason for separate dependencies anymore

test/DurableTask.Core.Tests/RetryInterceptorTests.cs
```diff
@@ -31,7 +31,7 @@ public async Task Invoke_WithFailingRetryCall_ShouldThrowCorrectException()
             await Assert.ThrowsExceptionAsync<IOException>(Invoke, "Interceptor should throw the original exception after exceeding max retry attempts.");
         }
 
-        [TestMethod]
+        [DataTestMethod]
```
davidmrdavid, line 34: this was throwing a warning. `TestMethod`s cannot have `DataRows`, for that you need `DataTestMethod` instead. The VSTesting task is forgiving enough here, but I changed this to remove the warning

test/DurableTask.Emulator.Tests/DurableTask.Emulator.Tests.csproj
```diff
@@ -7,8 +7,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.1" />
-    <PackageReference Include="MSTest.TestAdapter" Version="3.1.1" />
-    <PackageReference Include="MSTest.TestFramework" Version="3.1.1" />
+    <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
```
davidmrdavid, line 11: We have an inconsistent use of MSTest, some use version 1.4.0 and others use versoin 3.1.1. The use of different frameworks led to warnings. I've downgraded to the common denominator, as I encountered issues upgrading them all to 3.1.1 (can't remember which unfortunately).

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
```diff
@@ -409,6 +409,11 @@ public async override Task<string> RunTask(OrchestrationContext context, bool us
                 responderOrchestration = Task.FromResult("Herkimer is done");
             }
 
+            // before sending the event, wait a few seconds to ensure the sub-orchestrator exists
+            // otherwise, we risk a race condition where the event is dicarded because the instances table
+            // does not yet have the sub-orchestrator instance in it.
+            await context.CreateTimer<object>(context.CurrentUtcDateTime.AddSeconds(10), state: null);
+
```
davidmrdavid, line 416: this test suddenly became flaky. So added this to make it less flakey

eng/ci/public-build.yml
```diff
@@ -67,7 +67,8 @@ extends:
         # Run tests
         - template: /eng/templates/test.yml@self
           parameters:
-            testAssembly: '**\bin\**\netcoreapp3.1\DurableTask.Core.Tests.dll'
+            # TODO: for some reason we're unable to run the net462 tests in the CI
+            testAssembly: '**\bin\Debug\netcoreapp3.1\DurableTask.Core.Tests.dll'
```
davidmrdavid, line 71: In theory, I want to have the pattern `\**\bin\**\DurableTask.Core.Tests.dll` here. However, when I input that, 1ES complains that there's not tests to be found in the `net462` dll. This makes no sense to me since I can run those tests locally. If I may, I'd like to push for this to remain a "ToDo"

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
```diff
@@ -38,8 +40,14 @@ public class DispatcherMiddlewareTests
         [TestInitialize]
         public async Task Initialize()
         {
+            // configure logging so traces are emitted during tests.
+            // This facilitates debugging when tests fail.
+
+#pragma warning disable CS0618 // Type or member is obsolete
```
cgillum, line 46: Can we add comments here with reminders to update these dependencies?

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
cgillum: This is the right thing to do given the timeline we're working with, but I'm curious to better understand why this particular pattern isn't working reliably.

eng/ci/public-build.yml
cgillum: Is the `netcoreapp3.1` part of the path necessary? I see you didn't include it in line 90 and 109, further below.

cgillum (approved):
Overall LGTM. Left a few comments.

I don't know how to explain the need for explicitly adding System.Runtime.CompilerServices.Unsafe, unless perhaps it's related to differences in how binplacing works on these machines. When running locally, do you see this assembly anywhere in your build output on the file system?

test/DurableTask.Test.Orchestrations/SimpleOrchestrations.cs
davidmrdavid: I _think_ I know the root cause - if the ExternalEvent arrives before the ExecutionStarted event, then the `IsExecutableInstance` check will return `false`. We do that here:

https://github.com/Azure/durabletask/blob/d4b98d46158e470c6b769601cd76a86d9bb6d096/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L1038-L1060

So we return "no such instance" as seen above, and we end up **discarding** the message as it is for an instanceID that never existed. That code is here:

https://github.com/Azure/durabletask/blob/d4b98d46158e470c6b769601cd76a86d9bb6d096/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L782-L836

To be clear, the piece where we end up sending external events to the discard list is this particular section:

https://github.com/Azure/durabletask/blob/d4b98d46158e470c6b769601cd76a86d9bb6d096/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L801-L810

So it's an ordering issue. In theory, I think we could be more conservative and not discard events immediately, but that can creating poison messages. Please let me know if I'm missing something.

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
davidmrdavid: Yes, just did here -> https://github.com/Azure/durabletask/pull/1139/commits/bc9a5eab040d4950669c51dfbdd539f0ff568e18

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
bachuv: Yes, I'll add this task to the list.

eng/templates/test.yml
bachuv, line 1: Can you explain the changes that were made in this file?

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -39,6 +39,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.1" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
```
jviau, line 42: If we are going v3 with DTFx, I think we should try and just upgrade some packages here:

`Microsoft.Extensions.Logging.Abstractions` -> 6.0.0
`System.Diagnostics.DiagnosticSource` -> 6.0.0

See if that fixes your problem

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
jviau: 2.2.0 is quite old. Can we standardize to `6.0.0` as part of the v3 change?

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
```diff
@@ -36,23 +38,30 @@ public class DispatcherMiddlewareTests
...
                 .AddTaskActivities(typeof(SimplestGetUserTask), typeof(SimplestSendGreetingTask))
-                .StartAsync();
+                .StartAsync().RunSynchronously();
 
             this.client = new TaskHubClient(service);
         }
 
-        [TestCleanup]
-        public async Task TestCleanup()
+        [TestCleanup()]
```
jviau, line 61: nit: undo this change

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
```diff
@@ -36,23 +38,30 @@ public class DispatcherMiddlewareTests
...
+        public void CleanupTests()
         {
-            await this.worker!.StopAsync(true);
+            this.worker!.StopAsync(true).RunSynchronously();
```
jviau, line 64: Where is `RunSynchronously` from? Typically you sync wait for tasks with `.GetAwaiter().GetResult();`

jviau (commented):
I believe your assembly load issue is a combo of .netstandard and that we reference `System.Diagnostics.DiagnosticSource/5.0.0`, but then Azure Storage brings in a higher version. That package depends on the unsafe package, so I guess that is not transitively being upleveled?

Either way, I don't know if referencing Unsafe directly is the answer vs taking this chance to upgrade DiagnosticSource package.

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
davidmrdavid: You caught this PR as I was testing some weird stuff here. It's been reverted.

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
davidmrdavid: As mentioned - You caught this PR as I was testing some weird stuff here. It's been reverted.

src/DurableTask.Core/DurableTask.Core.csproj
davidmrdavid: Ah, maybe that's it. I'll try to diagnostics source change first, that could be it

davidmrdavid: No luck with just upgrading DiagnosticsSource -> https://dev.azure.com/azfunc/public/_build/results?buildId=173059&view=ms.vss-test-web.build-test-results-tab .

Let me try to rev the logging abstractions package as well next

eng/templates/test.yml
davidmrdavid: Yes. So previously I had:

```yml
  - task: VSTest@2
    displayName: 'Run tests'
    inputs:
      testAssemblyVer2: |
        ${{ parameters.testAssembly }}
        !**\obj\**
        testFiltercriteria: 'TestCategory!=DisabledInCI'
        vsTestVersion: 17.0
        distributionBatchType: basedOnExecutionTime
        platform: 'any cpu'
        configuration: 'Debug'
        diagnosticsEnabled: True
        collectDumpOn: always
        rerunFailedTests: true
        rerunFailedThreshold: 30
        rerunMaxAttempts: 3
```
As you can see, everything below `testAssemblyVer2` is at the same level indentation. That's an issue, since `testFiltercriteria` and the params below are not an input to `testAssemblyVer2`, instead they're properties that are supposed to be at the same level. So I've indented them back/out so that they're properly parsed by 1ES

src/DurableTask.Core/DurableTask.Core.csproj
davidmrdavid: Still no luck that failed too -> https://dev.azure.com/azfunc/public/_build/results?buildId=173069&view=ms.vss-test-web.build-test-results-tab

I'll keep the packages on the `6.x` version, and also include the `compilerServices.Unsafe` pkg

eng/ci/public-build.yml
davidmrdavid: I had `netcoreapp3.1` simply because there is a subset of tests in `net462` that fail to run on 1ES (again, not locally). It's the tests in `test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs`.

I've now changed the pattern back to `'**\bin\**\DurableTask.Core.Tests.dll'` and simply excluded the tests in `test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs` through and `#if !NET462` exclusion. It's hacky - but otherwise the tests fail with the errors seen here: https://dev.azure.com/azfunc/public/_build/results?buildId=173021&view=ms.vss-test-web.build-test-results-tab

test/DurableTask.AzureStorage.Tests/DurableTask.AzureStorage.Tests.csproj
davidmrdavid: I had some issues getting the code to build on 1ES with more recent versions. I'm open to upgrading, but since this is the test project, I'd like to push that to be a separate PR.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -23,6 +23,7 @@ namespace DurableTask.Core
     using DurableTask.Core.Logging;
     using DurableTask.Core.Middleware;
     using DurableTask.Core.Tracing;
+    using ActivityStatusCode = Tracing.ActivityStatusCode;
```
davidmrdavid, line 26: after upgrading DTFx.Core's `System.Diagnostics.DiagnosticSource` to 6.x, there's some ambiguous definitions here. So I needed to add this alias to disambiguate

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -85,7 +85,7 @@ public class TraceHelper
             DateTimeOffset startTime = startEvent.ParentTraceContext.ActivityStartTime ?? default;
 
             Activity? activity = ActivityTraceSource.StartActivity(
-                name: activityName,
+                activityName,
```
davidmrdavid, line 88: after upgrading DTFx.Core's `System.Diagnostics.DiagnosticSource` to 6.x, these named parameters create ambiguous method invocations here (there's multiple versions of this method with the same parameters in different orders). Using a positional argument here disambiguates the call.

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
```diff
@@ -10,6 +10,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
+#if !NET462 // for some reasons these tests are not discoverable on 1ES, leading to the test getting aborted. TODO: Needs investigation
```
davidmrdavid, line 13: FYI @cgillum, @jviau

test/DurableTask.Core.Tests/app.config
```diff
@@ -6,11 +6,5 @@
     <add key="TaskHubName" value="test" />
   </appSettings>
   <runtime>
-    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
-      <dependentAssembly>
-        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-7.0.0.0" newVersion="7.0.0.0" />
-      </dependentAssembly>
-    </assemblyBinding>
```
davidmrdavid, line 14: we don't use newtonsoft.json v7.0.0 anymore, so this binding redirect is unecessary

src/DurableTask.Core/DurableTask.Core.csproj
davidmrdavid: I increased the `System.Diagnostics.DiagnosticSource` pkg to `6.0.1` and that **indeed** fixed the issue :-) .
I did opt against increasing the logging abstractions package just because I want to avoid too many transitive dependency changes that may create issues unifying with the rest of bundles. If we have reason to think this is fine, I'm happy to add that in a follow up PR. For now though, I'd like to proceed as-is to unblock a few PRs.

davidmrdavid:
As mentioned [here](https://github.com/Azure/durabletask/pull/1139#discussion_r1682095306), upgrading `System.Diagnostics.DiagnosticSource` to `6.0.1` did fix the DLL loading issue.

I'm going to take a bit of optimistic liberty here to speed up some blocked PRs and merged on the basis of the previous approval. Goes without saying I'm happy to incorporate any follow-up feedback that may arise after this. I just would like to unblock some of the PRs that are waiting for this CI to be in place.

So merging now, but please reach out for feedback / follow up requests. Thanks!

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
cgillum: Do you mean the `.net462` flavor of the tests are not discoverable? The .NET Core flavor _is_ working?

cgillum:
The latest changes look good to me. ✅

I had one clarifying question about some tests that were disabled, but it's non-blocking.

PR #1140, opened 2024-07-17
Fix DurableTaskStorageException Constructor Null Handling

wsugarman:
The ctor for `DurableTaskStorageException` will throw an exception if the given `RequestFailedException` is `null` or the `ErrorCode` property is `null`. Both cases should be handled gracefully, and this change involves some refactoring to the class.

Included in this change is also an update to libraries including `Azure.Core` which had a vulnerability.

### DIFF
--- a/src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
+++ b/src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
@@ -35,12 +35,12 @@ namespace DurableTask.AzureStorage.Storage
         }
 
         public DurableTaskStorageException(RequestFailedException requestFailedException)
-            : base(requestFailedException.Message, requestFailedException)
+            : base("An error occurred while communicating with Azure Storage", requestFailedException)
         {
-            this.HttpStatusCode = requestFailedException.Status;
-            if (requestFailedException?.ErrorCode == BlobErrorCode.LeaseLost)
+            if (requestFailedException != null)
             {
-                LeaseLost = true;
+                this.HttpStatusCode = requestFailedException.Status;
+                this.LeaseLost = requestFailedException.ErrorCode != null && requestFailedException.ErrorCode == BlobErrorCode.LeaseLost;
             }
         }
 
--- a/src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
+++ b/src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
@@ -24,17 +24,17 @@ namespace DurableTask.AzureStorage.Storage
         {
         }
 
-        public DurableTaskStorageException(string message)
+        public DurableTaskStorageException(string? message)
             : base(message)
         {
         }
 
-        public DurableTaskStorageException(string message, Exception inner)
+        public DurableTaskStorageException(string? message, Exception? inner)
             : base(message, inner)
         {
         }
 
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
```diff
@@ -35,12 +35,12 @@ public DurableTaskStorageException(string message, Exception inner)
         }
 
         public DurableTaskStorageException(RequestFailedException requestFailedException)
```
cgillum, line 37: Should we make this nullable to clarify our expectations about this parameter?
```suggestion
        public DurableTaskStorageException(RequestFailedException? requestFailedException)
```

src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
```diff
@@ -35,12 +35,12 @@ public DurableTaskStorageException(string message, Exception inner)
...
-            : base(requestFailedException.Message, requestFailedException)
+            : base("An error occurred while communicating with Azure Storage", requestFailedException)
```
cgillum, line 38: Are we losing any information about using a hardcoded message string instead of `requestFailedException.Message` (when `requestFailedException` is non-null)?

wsugarman: It shouldn't be the case. The `Message` property will be changed, but its `ToString()` will reflect the expanded context, like other `Exception` types.

E.g.
```csharp
RequestFailedException inner = new((int)statusCode, "Inner error message!", errorCode, innerException: null);
DurableTaskStorageException before = new(inner.Message, inner);
DurableTaskStorageException after = new(inner);
```

The `before` value of `ToString()` is somewhat redundant:
```
DurableTask.AzureStorage.Storage.DurableTaskStorageException: Inner error message!
 ---> Azure.RequestFailedException: Inner error message!
```

Whereas the `after` value now looks like:
```
DurableTask.AzureStorage.Storage.DurableTaskStorageException: An error occurred while communicating with Azure Storage
 ---> Azure.RequestFailedException: Inner error message!
```

wsugarman: The same behavior should be seen in the standard library too. Consider this example:
```csharp
FormatException inner = new("Bad format");
InvalidOperationException before = new(f.Message, inner);
InvalidOperationException after = new(null, inner);
```
The respective `ToString()` values would be seen:
```
System.InvalidOperationException: Bad format
 ---> System.FormatException: Bad format
```
and
```
System.InvalidOperationException: Exception of type 'System.InvalidOperationException' was thrown.
 ---> System.FormatException: Bad format
```

cgillum: Thanks! I appreciate the detailed explanation, with examples. :)

cgillum (approved):

src/DurableTask.AzureStorage/Storage/DurableTaskStorageException.cs
wsugarman: I think that makes sense. I updated the other arguments to match `Exception` in .NET 8: https://learn.microsoft.com/en-us/dotnet/api/system.exception.-ctor?view=net-8.0

davidmrdavid (approved):

PR #1142, opened 2024-07-18
Ensure queue is listened to on a sudden PMv3 restart

davidmrdavid:
Replaces: https://github.com/Azure/durabletask/pull/1136/files, which targets `azure-storage-v12` which got _squashed_ when it got merged to main (on @cgillum's recommendation for a clean history on main). I figured that re-creating the PR was the easier path forward, as the other PR's history would need to be rebased and squashed as well.

Also made some minor edits to our pipeline config so the pipelines run on the DTFx.ASv1 "frozen" branch.

### DIFF
--- a/eng/ci/code-mirror.yml
+++ b/eng/ci/code-mirror.yml
@@ -4,6 +4,7 @@ trigger:
     # These are the branches we'll mirror to our internal ADO instance
     # Keep this set limited as appropriate (don't mirror individual user branches).
     - main
+    - durabletask-core-v2
 
 resources:
   repositories:
--- a/eng/ci/official-build.yml
+++ b/eng/ci/official-build.yml
@@ -3,6 +3,7 @@ trigger:
     branches:
         include:
             - main
+            - durabletask-core-v2
 
 # CI only, does not trigger on PRs.
 pr: none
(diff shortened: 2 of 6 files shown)


eng/ci/code-mirror.yml
```diff
@@ -4,6 +4,7 @@ trigger:
     # These are the branches we'll mirror to our internal ADO instance
     # Keep this set limited as appropriate (don't mirror individual user branches).
     - main
+    - durabletask-core-v2
```
bachuv, line 7: How come we didn't also add the durabletask-azurestorage-v1 branch?

bachuv (approved):
LGTM! Just left a question.

eng/ci/code-mirror.yml
davidmrdavid: @bachuv I can add it, but I was thinking that it might get confusing to have separate branches for the old DTFx.Core and DTFx.AS pkgs. Say we need to make a change to DTFx.AS v1 - would that mean we need to push changes to the `durabletaks-azurestorage-v1` branch **and** then cherry-pick to durabletask-core-v2`? I figured it might be best just to use the `durabletask-core-v2` branch for all of that, and delete `durabletak-azurestorage-v1`, just for reducing process. What do you think?

bachuv: I think that makes sense. We don't need to add it then. Thanks for the explanation!

PR #1147, opened 2024-07-18
Add source link reconstruction

davidmrdavid:
When building in 1ES, source-link information gets lost due to the code-mirror'ing. This creates issues during debugging, as Visual Studio would be unable to determine where to download the source code from. This PR aims to fix that by following conventions set by the Azure Functions Host.


Directory.Build.props
```diff
@@ -0,0 +1,12 @@
+<Project>
+  <!-- This is copied from: https://github.com/Azure/azure-functions-host/blob/dev/Directory.Build.props -->
+  <!-- To learn more about Directory.Build.props, see: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022
+  The basic idea is that it allows you to modify the MSBuild behavior of a `.sln` / solution by placing this at the project's root. -->
+
+  <!-- Define constants like `RepoRoot` and `EngRoot` (used for 1ES pipelines, amongst others) -->
+  <PropertyGroup>
+    <RepoRoot>$(MSBuildThisFileDirectory)</RepoRoot>
+    <EngRoot>$(RepoRoot)eng/</EngRoot>
+  </PropertyGroup>
```
davidmrdavid, line 10: on a second look, I don't really need this, right?

davidmrdavid: removed it

bachuv (approved):

jviau (approved):

PR #1149, opened 2024-07-23
Un-flake table TestWorkerRestartBeforeLeaseExpired test

davidmrdavid:
The test `TestWorkerRestartBeforeLeaseExpired` has been a bit flakey on the CI. I believe the following refactor should make it more deterministic.

### DIFF
--- a/test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
+++ b/test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
@@ -699,7 +699,7 @@ namespace DurableTask.AzureStorage.Tests
             var azureStorageClient = new AzureStorageClient(settings);
             Table partitionTable = azureStorageClient.GetTableReference(azureStorageClient.Settings.PartitionTableName);
 
-            // read the partions table
+            // read the partition table
             var results = partitionTable.ExecuteQueryAsync<TablePartitionLease>();
             var numResults = results.CountAsync().Result;
             Assert.AreEqual(numResults, 1); // there should only be 1 partition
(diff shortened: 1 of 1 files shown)


test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
```diff
@@ -665,68 +666,58 @@ public async Task EnsureOwnedQueueExclusive()
...
+
+            // We want to test that worker 0 starts listening to the control queue without claiming the lease.
+            // Therefore, we force the table to be in a state where worker 0 is still the current owner of the partition.
+            var partitionData = results.FirstAsync().Result;
+            partitionData.NextOwner = null;
+            partitionData.IsDraining = false;
+            partitionData.CurrentOwner = "0";
+            partitionData.ExpiresAt = DateTime.UtcNow.AddMinutes(5);
+            await partitionTable.InsertOrMergeEntityAsync(partitionData);
+
```
nytian, line 715: Can we add a Assert here verifying that the partition current owner is worker 0 after the updates?

test/DurableTask.AzureStorage.Tests/TestTablePartitionManager.cs
```diff
@@ -665,68 +666,58 @@ public async Task EnsureOwnedQueueExclusive()
...
+            // create and start new worker with the same settings, ensure it is actively listening to the queue
+            worker = new TaskHubWorker(service);
+            await worker.StartAsync();
+            await Task.Delay(TimeSpan.FromSeconds(15));
```
nytian, line 719: I think the wait time can be less as adding queue won't require that much time. We don't want to block the thread for a very long time.

nytian (commented):
Thanks for the PR! Left a few small suggestions.

nytian (approved):

PR #1152, opened 2024-07-24
Remove netcoreapp3.1 from tests

davidmrdavid:
Given our recent TFM changes, and that `netcoreapp3.1` is out of support and will _soon_ no longer be the used in Azure Functions' extension bundles, this PR changes our test project to use `.net6.0` as it's TFM over `netcoreapp3.1`. This removes some warnings from our solution, which should help with compliance and reduce "warning fatigue".

### DIFF
--- a/test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -3229,7 +3229,7 @@ namespace DurableTask.AzureStorage.Tests
             internal class AutoStartOrchestration : TaskOrchestration<string, string>
             {
                 private readonly TaskCompletionSource<string> tcs
-                    = new TaskCompletionSource<string>(TaskContinuationOptions.ExecuteSynchronously);
+                    = new TaskCompletionSource<string>();
 
                 // HACK: This is just a hack to communicate result of orchestration back to test
                 public static bool OkResult;
@@ -3273,7 +3273,7 @@ namespace DurableTask.AzureStorage.Tests
                 public class Responder : TaskOrchestration<string, string, RequestInformation, string>
                 {
                     private readonly TaskCompletionSource<string> tcs
-                        = new TaskCompletionSource<string>(TaskContinuationOptions.ExecuteSynchronously);
+                        = new TaskCompletionSource<string>();
 
                     public async override Task<string> RunTask(OrchestrationContext context, string input)
                     {
--- a/test/DurableTask.AzureStorage.Tests/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
+++ b/test/DurableTask.AzureStorage.Tests/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
@@ -311,10 +311,7 @@ namespace DurableTask.AzureStorage.Tests.Correlation
             if (initializeFromCurrent)
             {
                 opTelemetry.Id = activity.SpanId.ToHexString();
-                if (activity.ParentSpanId != null)
-                {
-                    opTelemetry.Context.Operation.ParentId = activity.ParentSpanId.ToHexString();
-                }
+                opTelemetry.Context.Operation.ParentId = activity.ParentSpanId.ToHexString();
             }
             else
             {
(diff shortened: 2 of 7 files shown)


test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -3229,7 +3229,7 @@ public async override Task<string> RunTask(OrchestrationContext context, string
             internal class AutoStartOrchestration : TaskOrchestration<string, string>
             {
                 private readonly TaskCompletionSource<string> tcs
-                    = new TaskCompletionSource<string>(TaskContinuationOptions.ExecuteSynchronously);
+                    = new TaskCompletionSource<string>();
```
davidmrdavid, line 3232: After changing the TFM to `net6.0`, this was failing to build with the following error: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2247.

In short, the constructor for `TaskCompletionSource` does not take a TaskContinuationOptions` enum. It either takes an `object`, or a `TaskCreationOptions` enum. When a `TaskContnuactionOptions` object is passed in to the constructor, that becomes the `AsyncState` of the Task, which is only relevant when using Tasks in conjunction with APM (the old "Asynchronous Programming Model" technology that Tasks eventually replaced). As a result, passing in this argument had not effect on the tests, so I've removed it.

test/DurableTask.AzureStorage.Tests/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -311,10 +311,7 @@ internal static void UpdateTelemetry(ITelemetry telemetry, Activity activity, bo
             if (initializeFromCurrent)
             {
                 opTelemetry.Id = activity.SpanId.ToHexString();
-                if (activity.ParentSpanId != null)
-                {
-                    opTelemetry.Context.Operation.ParentId = activity.ParentSpanId.ToHexString();
-                }
```
davidmrdavid, line 317: after changing to `.net6.0` this was failing to build. It appears that `activity.ParentSpanId` never returns null (instead, a "null" ParentSpanId is a one of all zeroes), so this if statement is always taken. So I removed the if statement in the first place.

test/DurableTask.Core.Tests/DispatcherMiddlewareTests.cs
```diff
@@ -255,7 +255,7 @@ public void EnsureOrchestrationExecutionContextSupportsDataContractSerialization
             StringReader stringReader = new StringReader(writtenString);
             XmlReader xmlReader = XmlReader.Create(stringReader);
 
-            OrchestrationExecutionContext deserializedContext = (OrchestrationExecutionContext)dataContractSerializer.ReadObject(xmlReader);
+            OrchestrationExecutionContext deserializedContext = (OrchestrationExecutionContext)dataContractSerializer.ReadObject(xmlReader)!;
```
davidmrdavid, line 258: there's technically a possible null-dereference here, which the build trace was complaining about. Since this is a test, I added the `!` operatior to suppress that warning.

test/DurableTask.Stress.Tests/DurableTask.Stress.Tests.csproj
```diff
@@ -12,16 +12,16 @@
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' != 'net462'">
     <None Remove="eventFlowConfig.json" />
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' != 'net462'">
```
jviau, line 18: could combine some item groups in this file:

``` XML
<ItemGroup Condition="'$(TargetFramework)' != 'net462'">
  <None Update="DurableTask.Stress.Tests.dll.config" CopyToOutputDirectory="PreserveNewest" />
  <None Remove="eventFlowConfig.json" />
  <Content Include="eventFlowConfig.json" CopyToOutputDirectory="PreserveNewest" />
</ItemGroup>
```

jviau (approved):

test/DurableTask.Stress.Tests/DurableTask.Stress.Tests.csproj
davidmrdavid: To be honest, part of why I didn't do that is my lack of familiarity with `eventFlowConfig.json` and my confusion around a `Remove="eventFlowConfig...` follow by `Include="EventFlowConfig...`. For now, I'll keep this as-is, I'd love to investigate this further.

PR #1154, opened 2024-08-07
Update DTFx.AzureStorage v2 Queue to use Base64 Encoding

nytian:
DTFx.AzureStorage v1.x and v2.x with different version Azure Storage SDK has different default encoding style.  This pull request updates the queue encoding mechanism in the DTFx.AzureStorage v2 to use Base 64 Encoding style. This change ensures that the queue messages created by DTFx.AzureStorage v2 are forward compatible with those created by DTFx.AzureStorage v1. 

Unit test has been added.


src/DurableTask.AzureStorage/StorageServiceClientProvider.cs
```diff
@@ -116,9 +116,11 @@ public static class StorageServiceClientProvider
                 throw new ArgumentNullException(nameof(connectionString));
             }
 
+            options = options ?? new QueueClientOptions();
+            options.MessageEncoding = QueueMessageEncoding.Base64;
```
davidmrdavid, line 120: @nytian: I usually like to follow the "rule" that, if we're doing the same work in 3 places or more, we should probably refactor that logic into a common helper function.

Otherwise, I see the following risk: we could end up adding a new method to create a queue client, and then forget to add this "base64" encoding trick. 

Can you look into a refactoring of these 3 methods where they call common logic to ensure the client options is always populated with base64 encoding?

davidmrdavid (commented):
Left some feedback. It would also be good to add a test here. Are we able to add an automated test that this is able to read messages generated by Azure Storage SDK v1?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2312,6 +2314,40 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
             }
         }
 
+        /// <summary>
+        /// Test queue message generated by DTFx.AzureStorage v2.x can be successfully read by DTFx.AzureStorage v1.x
+        /// </summary>
+        [TestMethod]
+        public async Task TestBackwardCompatibilityWithASv2()
+        {
+            //string connectionString = TestHelpers.GetTestStorageAccountConnectionString();
```
davidmrdavid, line 2323: let's remove commented out code?
```suggestion
            //string connectionString = TestHelpers.GetTestStorageAccountConnectionString();
```

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2312,6 +2314,40 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
             }
         }
 
+        /// <summary>
+        /// Test queue message generated by DTFx.AzureStorage v2.x can be successfully read by DTFx.AzureStorage v1.x
+        /// </summary>
+        [TestMethod]
+        public async Task TestBackwardCompatibilityWithASv2()
```
davidmrdavid, line 2321: Here you're testing that backwards compatibility is possible with the previous version, so I believe it should be `v1`
```suggestion
        public async Task TestBackwardCompatibilityWithASv1()
```

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2312,6 +2314,40 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
             }
         }
 
+        /// <summary>
```
davidmrdavid, line 2317: can you please also add a test that this can still read messages sent by DTFX.AS V2.0.0 (which doesn't specify encoding)? Or do we already have that test?

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2312,6 +2314,40 @@ public async Task ScheduledStart_Activity_GetStatus_Returns_ScheduledStart(bool
...
+
+            // Add a orchestration message to the control queue
+            await client.CreateOrchestrationInstanceAsync(typeof(Orchestrations.LongRunningOrchestrator), input: "0");
+
+            var message = await queue.GetMessageAsync();
+            string content = message.AsString;
+
+            // Check the message can be successfully read without any error
+            Assert.IsNotNull(content);
+            Assert.IsInstanceOfType<string>(content);
```
davidmrdavid, line 2348: can you please also check the contents of the message? Want to make sure it's de-serialized to the right string, and not a random string (which could happen under the wrong encoding)

davidmrdavid (commented):
Almost there. Left some feedback on the test (thank you for adding it)

nytian:
Close as this changes are no longer needed.

PR #1167, opened 2024-09-16
Update Durable Functions support status

AnatoliB:
AzureStorage is not the only backend available for Durable Functions, Netherite and SqlServer are also available


davidmrdavid (approved):
Good catch!

cgillum (approved):

PR #1168, opened 2024-09-23
Modifies WorkItem renew logic to start running after item is fetched.

### LINKED ISSUE #1150: Consider starting the renewal of messages in DTFx.Core as soon as they are fetched
In DTFx.Core, the method `RenewTaskOrchestrationWorkItemLockAsync` is used to ensure a given worker maintains exclusivity over a given partition message. For example, in the Azure Storage backend, this messages renews the "message visibility timeout" so that the message does not get dequeued again, or at least until the visibility timeout expires.

This renewal flow is invoked when the message is being **processed**, which has a very specific meaning: we have not exceeded the "maxConcurrentOrchestrations" / "maxConcurrentActivities" limit, and therefore have enough capacity to process more messages.

This means that a message may be received by a given worker, but not become processable for a long time if the active orchestrators/activities match their "max concurrent" settings and are long-running. In that time, since we're not actively extending the message's visibilityTimeout, it is possible for the message to become visible again (possibly being dequeued by the same worker that already has that message!), therefore changing it's `popReceipt`, which in turn prevents us from successfully processing the copy of the message with the old popReceipt. This can lead to a cascade of errors.

I believe framework-level fix to this is to start renewing messages as soon as they're fetched/received, not just when they're being processed. This may require some refactoring in DTFx.Core's `WorkItemDispatcher` class, so it needs to be done with care.

sig5:
Fixes #1150 
The summary of changes is as follows:
1. Adds a `WorkItemBase` to extract common properties for Work Items.
2. Pulls the logic to renew work item lock in `WorkItemDispatcher`.

Open Question:
Do we want to add a renewal policy for Tracking store items too? If not, why?

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -68,7 +68,8 @@ namespace DurableTask.Core
                 "TaskOrchestrationDispatcher",
                 item => item == null ? string.Empty : item.InstanceId,
                 this.OnFetchWorkItemAsync,
-                this.OnProcessWorkItemSessionAsync)
+                this.OnProcessWorkItemSessionAsync,
+                this.RenewWorkItem)
             {
                 GetDelayInSecondsAfterOnFetchException = orchestrationService.GetDelayInSecondsAfterOnFetchException,
                 GetDelayInSecondsAfterOnProcessException = orchestrationService.GetDelayInSecondsAfterOnProcessException,
@@ -155,7 +156,7 @@ namespace DurableTask.Core
             {
                 // Keep track of orchestrator generation changes, maybe update target position
                 string executionId = message.OrchestrationInstance.ExecutionId;
-                if(previousExecutionId != executionId)
+                if (previousExecutionId != executionId)
                 {
                     // We want to re-position the ExecutionStarted event after the "right-most"
                     // event with a non-null executionID that came before it.
@@ -217,7 +218,7 @@ namespace DurableTask.Core
 
                 CorrelationTraceClient.Propagate(
                     () =>
-                    {                
+                    {
                         // Check if it is extended session.
                         // TODO: Remove this code - it looks incorrect and dangerous
                         isExtendedSession = this.concurrentSessionLock.Acquire();
@@ -305,7 +306,7 @@ namespace DurableTask.Core
             var isCompleted = false;
             var continuedAsNew = false;
             var isInterrupted = false;
-            
+
             // correlation
             CorrelationTraceClient.Propagate(() => CorrelationTraceContext.Current = workItem.TraceContext);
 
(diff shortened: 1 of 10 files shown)


sakarsinghal:
cc: @davidmrdavid for Review

src/DurableTask.Core/TaskOrchestrationWorkItem.cs
```diff
@@ -32,11 +32,6 @@ public class TaskOrchestrationWorkItem
         /// </summary>
         public OrchestrationRuntimeState OrchestrationRuntimeState;
 
-        /// <summary>
-        /// The datetime this orchestration work item is locked until
-        /// </summary>
-        public DateTime LockedUntilUtc;
```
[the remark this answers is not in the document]
davidmrdavid: nevermind, I see this was added to `WorkItemBase`

src/DurableTask.Core/TrackingWorkItem.cs
```diff
@@ -19,18 +19,13 @@ namespace DurableTask.Core
...
     {
         /// <summary>
         /// The instance id of this tracking work item
         /// </summary>
         public string InstanceId;
 
-        /// <summary>
-        /// The datetime this work item is locked until
-        /// </summary>
-        public DateTime LockedUntilUtc;
```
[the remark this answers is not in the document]
davidmrdavid: nevermind, I see this was added to `WorkItemBase`

src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs
```diff
@@ -221,7 +221,8 @@ public ServiceBusOrchestrationService(
                     "TrackingDispatcher",
                     item => item == null ? string.Empty : item.InstanceId,
                     FetchTrackingWorkItemAsync,
-                    ProcessTrackingWorkItemAsync)
+                    ProcessTrackingWorkItemAsync,
+                    (T,token) => Task.CompletedTask)
```
davidmrdavid, line 225: can you please explain why this is the implementation for ServiceBus? Not sure I follow why we can just return Task.CompletedTask

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -635,6 +613,7 @@ static OrchestrationExecutionContext GetOrchestrationExecutionContext(Orchestrat
         static TimeSpan MinRenewalInterval = TimeSpan.FromSeconds(5); // prevents excessive retries if clocks are off
         static TimeSpan MaxRenewalInterval = TimeSpan.FromSeconds(30);
```
davidmrdavid, line 615: please remove this newline

src/DurableTask.Core/WorkItemDispatcher.cs
```diff
@@ -92,20 +94,23 @@ public class WorkItemDispatcher<T> : IDisposable
...
+        /// <param name="processWorkItem">Action to process the work item.</param>
+        /// <param name="renewWorkItem">Action to renew the work item to maintain locking.</param>
         public WorkItemDispatcher(
-            string name, 
+            string name,
             Func<T, string> workItemIdentifier,
             Func<TimeSpan, CancellationToken, Task<T>> fetchWorkItem,
-            Func<T, Task> processWorkItem)
+            Func<T, Task> processWorkItem,
+            Func<T, CancellationToken, Task> renewWorkItem)
```
davidmrdavid, line 106: since this is a public class, we should avoid adding a new manadatory parameter, that will break existing implementations. Please make this optional, or create another constructor

davidmrdavid (commented):
Sorry for the delay. Left some notes, overall the work makes sense based on what I had originally suggested [here](https://github.com/Azure/durabletask/issues/1150#issuecomment-2369394041), but now that I'm reviewing the code and looking at the logic more closely, I don't think this really gets to the root problem. Sorry, I think I misunderstood the original problem.

Let me explain. In `WorkItemDispatcher.DispatchAsync`, we have we in line [227](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/WorkItemDispatcher.cs#L227) the logic for ensuring we don't exceed our concurrency limits:

>   if (!await this.concurrencyLock.WaitAsync(TimeSpan.FromSeconds(5)))

In line [262](https://github.com/Azure/durabletask/blob/2795981f114d2f315adf55f07d01cfb7ce3018b0/src/DurableTask.Core/WorkItemDispatcher.cs#L262) we have the `FetchWorkItem` invocation. At that point ,we have already guaranteed that we're not going over the concurrency limit, so it doesn't make a big difference whether we start renewing there, or a bit later in `ProcessWorkItemAsync`. In both cases, we have already satisfied the check in line 227 (the `.WaitAsync`). So it's "too late".

I think the issue is the `FetchWorkItem` async is simply pulling of messages already read from the queues, and it is those messages that may have already expired (because they weren't getting renewe until after we call `FetchWorkItem`). So it feels to me that the solution really is to guarantee the renewing is ocurring as soon as messages are **obtained** from storage, in `GetMessageAsync` here: https://github.com/Azure/durabletask/blob/2795981f114d2f315adf55f07d01cfb7ce3018b0/src/DurableTask.AzureStorage/Storage/Queue.cs#L88-L99

Please let me know if this doesn't make sense, I'm happy to clarify. In any case, I need to think about the design a bit more, if it ends up being too complicated, I may not want to tackle this edge case just yet. I'm happy to hear your ideas if the problem makes sense though, please let me know

PR #1172, opened 2024-10-13
Update DurableTask.ApplicationInsights and DurableTask.AzureStorage Version

nytian:
As titled. 
 DurableTask.ApplicationInsights v0.1.6 -> v2.0.0 as it uses DurableTask.Core v3.0.0. 
DurableTask.AzureStorage v2.0.0- v2.0.1 for new changes in https://github.com/Azure/durabletask/pull/1159

### DIFF
--- a/src/DurableTask.ApplicationInsights/DurableTask.ApplicationInsights.csproj
+++ b/src/DurableTask.ApplicationInsights/DurableTask.ApplicationInsights.csproj
@@ -11,8 +11,8 @@
     <!-- Version Info -->
     <PropertyGroup>
         <MajorVersion>0</MajorVersion>
-        <MinorVersion>1</MinorVersion>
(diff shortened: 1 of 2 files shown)


cgillum (approved):
The code change looks good. However, I see that the "Build & Sign" pipeline failed with what looks like an ESRP authentication issue. However, this doesn't seem related to the code change, so I don't consider it a blocker for this PR.

nytian:
> The code change looks good. However, I see that the "Build & Sign" pipeline failed with what looks like an ESRP authentication issue. However, this doesn't seem related to the code change, so I don't consider it a blocker for this PR.

Thanks for the review, Chris! The pipeline failed because I attempted to create the package for the bug bash using our old DevOps pipelines. Therefore, we don't need to worry about that failure. :)

PR #1175, opened 2024-11-11
Handle poisoned work items by marking them as failed

ItalyPaleAle:
When orchestrations and activity tasks fail with un-retriable errors (by any worker), they should not be abandoned; instead, they should be marked as completed and with state failed.

This is implemented by catching exceptions of type `WorkItemPoisonedException` which can be raised downstream, for example by a persistence store provider. When that happens, instead of abandoning the work item, the dispatcher marks it as completed.

### DIFF
--- a/src/DurableTask.Core/Logging/StructuredEventSource.cs
+++ b/src/DurableTask.Core/Logging/StructuredEventSource.cs
@@ -712,7 +712,7 @@ namespace DurableTask.Core.Logging
                 // TODO: Use WriteEventCore for better performance
                 this.WriteEvent(
                     EventIds.EntityLockReleased,
-                    EntityId, 
+                    EntityId,
                     InstanceId,
                     Id,
                     AppName,
@@ -808,6 +808,32 @@ namespace DurableTask.Core.Logging
                 // TODO: Use WriteEventCore for better performance
                 this.WriteEvent(
                     EventIds.TaskActivityAborted,
+                    InstanceId,
+                    ExecutionId,
+                    Name,
+                    TaskEventId,
+                    Details,
+                    AppName,
+                    ExtensionVersion);
+            }
+        }
+
+        [Event(EventIds.TaskActivityPoisoned, Level = EventLevel.Warning, Version = 1)]
+        internal void TaskActivityPoisoned(
+            string InstanceId,
+            string ExecutionId,
+            string Name,
+            int TaskEventId,
+            string Details,
+            string AppName,
+            string ExtensionVersion
+        )
+        {
+            if (this.IsEnabled(EventLevel.Warning))
+            {
+                // TODO: Use WriteEventCore for better performance
+                this.WriteEvent(
+                    EventIds.TaskActivityPoisoned,
                     InstanceId,
                     ExecutionId,
                     Name,
--- a/src/DurableTask.Core/Logging/StructuredEventSource.cs
+++ b/src/DurableTask.Core/Logging/StructuredEventSource.cs
@@ -599,6 +599,26 @@ namespace DurableTask.Core.Logging
             }
         }
 
+        [Event(EventIds.OrchestrationPoisoned, Level = EventLevel.Warning, Version = 1)]
+        internal void OrchestrationPoisoned(
+            string InstanceId,
+            string ExecutionId,
+            string Details,
+            string AppName,
+            string ExtensionVersion)
+        {
+            if (this.IsEnabled(EventLevel.Warning))
+            {
+                this.WriteEvent(
+                    EventIds.OrchestrationPoisoned,
+                    InstanceId,
+                    ExecutionId,
+                    Details,
+                    AppName,
+                    ExtensionVersion);
+            }
+        }
+
         [Event(EventIds.DiscardingMessage, Level = EventLevel.Warning, Version = 1)]
         internal void DiscardingMessage(
             string InstanceId,
--- a/src/DurableTask.Core/OrchestrationRuntimeState.cs
+++ b/src/DurableTask.Core/OrchestrationRuntimeState.cs
@@ -98,6 +98,19 @@ namespace DurableTask.Core
         }
 
         /// <summary>
+        /// Returns a deep copy of the object.
+        /// </summary>
+        /// <returns>Cloned object</returns>
+        public OrchestrationRuntimeState Clone()
+        {
+            return new OrchestrationRuntimeState(this.Events)
+            {
+                Size = this.Size,
+                Status = this.Status,
+            };
+        }
+
+        /// <summary>
         /// Gets the execution started event
         /// </summary>
         public ExecutionStartedEvent? ExecutionStartedEvent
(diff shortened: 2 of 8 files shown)


src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -599,6 +599,26 @@ internal void OrchestrationAborted(
...
+            string AppName,
+            string ExtensionVersion)
+        {
+            if (this.IsEnabled(EventLevel.Warning))
+            {
+                this.WriteEvent(
+                    EventIds.OrchestrationPoisoned,
+                    InstanceId,
+                    ExecutionId,
+                    Details,
```
cgillum, line 616: Do we need to check for null here? Passing `null` to `WriteEvent` for any parameter will cause the entire log event to be lost. If I'm reading the code correctly, this comes from the `WorkItemPoisonedException.Message` property, which doesn't have any guards against null values.

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -818,6 +838,32 @@ internal void TaskActivityAborted(
...
+            if (this.IsEnabled(EventLevel.Warning))
+            {
+                // TODO: Use WriteEventCore for better performance
+                this.WriteEvent(
+                    EventIds.TaskActivityPoisoned,
+                    InstanceId,
+                    ExecutionId,
+                    Name,
+                    TaskEventId,
+                    Details,
```
cgillum, line 861: Same question about null checks.

src/DurableTask.Core/OrchestrationRuntimeState.cs
```diff
@@ -97,6 +97,19 @@ public OrchestrationRuntimeState(IList<HistoryEvent>? events)
             }
         }
 
+        /// <summary>
+        /// Returns a deep copy of the object.
+        /// </summary>
+        /// <returns>Cloned object</returns>
+        public OrchestrationRuntimeState Clone()
+        {
+            return new OrchestrationRuntimeState(this.Events)
```
cgillum, line 106: I don't think this is technically a deep clone operation. The state of the two objects may potentially be different. For example, if the object being cloned as a set of history events in both the `NewEvents` and `PastEvents` lists, then the cloned object will instead have all the history event objects only in the `PastEvents` list.

I suggest either we carefully document that this is not an exact clone or give this method a different name.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -268,6 +268,22 @@ await this.dispatchPipeline.RunAsync(dispatchContext, async _ =>
...
+                this.logHelper.TaskActivityPoisoned(orchestrationInstance, scheduledEvent!, poisonedException.Message);
+                TraceHelper.TraceInstance(TraceEventType.Warning, "TaskActivityDispatcher-ExecutionPoisoned", orchestrationInstance, "{0}: {1}", scheduledEvent?.Name ?? "", poisonedException.Message);
+                await this.orchestrationService.CompleteTaskActivityWorkItemAsync(
+                    workItem, new TaskMessage()
+                    {
+                        Event = new TaskFailedEvent(
+                            -1,
+                            // Guaranteed to be not null because of the "when" clause in the catch block
+                            scheduledEvent!.EventId,
+                            poisonedException.Message, string.Empty),
```
cgillum, line 283: We need to also populate the `failureDetails` parameter of the `TaskFailedEvent` constructor since there will be some consumers that use that instead of the (legacy) `reason` + `details` values.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -316,11 +316,10 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
 
             workItem.OrchestrationRuntimeState.LogHelper = this.logHelper;
             OrchestrationRuntimeState runtimeState = workItem.OrchestrationRuntimeState;
+            OrchestrationRuntimeState originalOrchestrationRuntimeState = runtimeState.Clone();
```
cgillum, line 319: I see that you've repurposed this variable to be used only for the new poison message handling logic, which I think is fine given that the previous behavior used `originalOrchestrationRuntimeState` as a reference to the live `runtimeState` object (I don't recall why - the previous implementation seems confusing to me, but I haven't gone back to look at the change history). This is fine, but just confirming that this was your intention, and that you intended to do it in a way that doesn't change existing behavior (per your comment further down)?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -610,15 +610,44 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
+                failedRuntimeState.AddEvent(new OrchestratorCompletedEvent(-1));
+
+                await this.orchestrationService.CompleteTaskOrchestrationWorkItemAsync(
+                    workItem,
+                    failedRuntimeState,
+                    outboundMessages: Array.Empty<TaskMessage>(),
+                    orchestratorMessages: Array.Empty<TaskMessage>(),
+                    timerMessages: Array.Empty<TaskMessage>(),
+                    continuedAsNewMessage: null,
+                    instanceState);
```
cgillum, line 644: Do we need to update `instanceState` to also reflect the fact that this instance is now in a failed state?

cgillum (changes requested):
Thanks for this PR @ItalyPaleAle, I like these changes as a first step in helping us deal with poison messages more generally. Some of the cases we normally deal with are on the work item fetch side (failing to deserialize received work items, for example) but my hope is that we can at least borrow some of the ideas here in separate PRs to also address those cases.

I added some comments about things I think we should fix or discuss before merging. They should be pretty minor.

FYI @davidmrdavid since this is a topic you've also spent a lot of time thinking about. Note that the motive in this case is for a different class of poison messages than what you were thinking about, but there might be some common pieces we can consider reusing.

FYI @sebastianburckhardt as well.

src/DurableTask.Core/OrchestrationRuntimeState.cs
ItalyPaleAle: Do you reckon the issue is with the description/comment only, or should the behavior be changed too?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
ItalyPaleAle: In the existing code, `originalOrchestrationRuntimeState` is pointing to the same memory as `runtimeState`, so by the time it is used (when `workItem.RestoreOriginalRuntimeStateDuringCompletion` is true), it is pointing to the updated state.

To me that smells a bit like it could _potentially_ be a bug in the existing implementation. However, I didn't want to alter the behavior in this PR as I'm not fully understand what the purpose (or intention) were, or whether fixing the bug (if it is) would break any existing user.

However, I can change the behavior if you think it should be.

cgillum: I think we're on the same page. Since it doesn't affect the existing behavior (buggy or not), I have no objection to this.

src/DurableTask.Core/OrchestrationRuntimeState.cs
cgillum: I don't have a strong opinion. Here's what I think the code would need to be to do a proper deep clone.

```csharp
var cloned = new OrchestrationRuntimeState(this.PastEvents)
{
    CompressedSize = this.CompressedSize,
    Size = this.Size,
    Status = this.Status,
};

foreach (HistoryEvent e in this.NewEvents)
{
    cloned.AddEvent(e);
}

return cloned;
```

PR #1176, opened 2024-11-11
Increase Azure Storage SDK version and DTFx.AzureStorage ver

nytian:
as titled. 

Azure.Core updated from 1.41.0 to 1.43.0
Azure.Data.Tables updated from 12.8.3 to 12.9.1
Azure.Storage.Blobs updated from 12.20.0 to 12.22.1
Azure.Storage.Queues updated from 12.18.0 to 12.20.0

### DIFF
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -40,10 +40,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.44.1" />
+    <PackageReference Include="Azure.Core" Version="1.43.0" />
     <PackageReference Include="Azure.Data.Tables" Version="12.9.1" />
-    <PackageReference Include="Azure.Storage.Blobs" Version="12.22.2" />
-    <PackageReference Include="Azure.Storage.Queues" Version="12.20.1" />
+    <PackageReference Include="Azure.Storage.Blobs" Version="12.22.1" />
+    <PackageReference Include="Azure.Storage.Queues" Version="12.20.0" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
   </ItemGroup>
--- a/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
+++ b/src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
@@ -22,7 +22,7 @@
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>2</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <FileVersion>$(VersionPrefix).0</FileVersion>
     <!-- FileVersionRevision is expected to be set by the CI. This is useful for distinguishing between multiple builds of the same version. -->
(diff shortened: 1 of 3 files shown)


test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="MSTest.TestFramework" Version="3.5.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" />
+	<PackageReference Include="System.Text.Json" Version="6.0.10" />
```
nytian, line 15: Add this to pass the test..

bachuv: Can you fix the indentation here? Also, what test wasn't passing without this package reference?

src/DurableTask.AzureStorage/DurableTask.AzureStorage.csproj
```diff
@@ -40,10 +40,10 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.41.0" />
-    <PackageReference Include="Azure.Data.Tables" Version="12.8.3" />
-    <PackageReference Include="Azure.Storage.Blobs" Version="12.20.0" />
-    <PackageReference Include="Azure.Storage.Queues" Version="12.18.0" />
+    <PackageReference Include="Azure.Core" Version="1.44.1" />
+    <PackageReference Include="Azure.Data.Tables" Version="12.9.1" />
+    <PackageReference Include="Azure.Storage.Blobs" Version="12.22.2" />
+    <PackageReference Include="Azure.Storage.Queues" Version="12.20.1" />
```
bachuv, line 46: Are you just upgrading the packages to be on the latest version or is there another reason?

nytian: To be the latest. 

Actually I am doing this mainly because I want to upgrade the Azure.Identity version to the latest in WebJobs.Extensions.DurableTask, where we uses v1.11.4, while the latest version is [v1.13.1 ](https://www.nuget.org/packages/Azure.Identity). Although Azure.Identity v1.11.4 hasn't been deprecated, but I just worry as it would be soon. :(

But again, I am also a little worried that if this dependency update qualifies a DTFx.AzureStorage batch version update. If you think we shouldn't do this either, I will just skip this and  release the WebJobs.Extensions.DurableTask v3 with the current DTFx.AzureStorage v2.0.1 that is already out, not this new coming one.

test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="MSTest.TestFramework" Version="3.5.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" />
+    <PackageReference Include="System.Text.Json" Version="6.0.10" />
```
nytian, line 15: @bachuv , our comments were gone as I updated here.  To answer your question, i added this line to pass the tests, as tests FailureDetailsOnUnhandled and CatchInvalidOperationException failed, with error message:

_Unable to create instance of class DurableTask.Core.Tests.ExceptionHandlingIntegrationTests. Error: System.IO.FileLoadException: Could not load file or assembly 'System.Text.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040).Unable to create instance of class DurableTask.Core.Tests.ExceptionHandlingIntegrationTests. Error: System.IO.FileLoadException: Could not load file or assembly 'System.Text.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)._

Failed CI: https://azfunc.visualstudio.com/public/_build/results?buildId=189162&view=ms.vss-test-web.build-test-results-tab&runId=1454649&resultId=100039&paneView=debug

I am not sure exactly about why this shows up as I didn't update DTFx.Core, but this should be some dependency versions load conflicts somewhere. The latest Azure Storage SDK uses System.Text.Json >6.0.10, so I added this.

cgillum:
> Azure.Core updated from 1.41.0 to 1.43.0

Just FYI, we're targeting Azure.Core v1.43.0 even though there are newer versions because the newer versions contain a breaking dependency change that causes problems with the current Azure Functions host versions. v1.43.0 is the latest version that works. Once the Functions team has updated the host to support newer versions of Azure.Core, we should update our dependency as well to match.

cgillum (approved):

test/DurableTask.Core.Tests/DurableTask.Core.Tests.csproj
bachuv: Thanks for the explanation!

bachuv (approved):

PR #1183, opened 2025-01-26
CreateTimer validation and more context for activity implementations

cgillum:
### Changes

- **Add CreateTimer input validation**: the design of the `CreateTimer` API makes it easy to accidentally pass a `CancellationToken` as the `state` parameter, which is never appropriate and is also theoretically dangerous because of how cancellation token deserialization can corrupt memory. This PR adds validation ensuring that the API is not being used incorrectly.

- **Add more context for activities**: General enhancement to allow activities to know their name, version, and task ID. This can be useful for debugging and logging purposes. It can also be used for simplifying the implementation of some versioning scenarios.

- Incremented DurableTask.Core package version to 3.1.0.


nytian (dismissed):

src/DurableTask.Core/TaskContext.cs
```diff
@@ -23,15 +23,41 @@ public class TaskContext
...
         /// <param name="orchestrationInstance"></param>
         public TaskContext(OrchestrationInstance orchestrationInstance)
+            : this(orchestrationInstance, string.Empty, null, -1)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new TaskContext with the supplied OrchestrationInstance and taskId
+        /// </summary>
+        internal TaskContext(OrchestrationInstance orchestrationInstance, string name, string? version, int taskId)
```
jviau, line 33: Consider making this `public`. Or give some way for this to be constructed in unit tests with all the properties set.

cgillum: Made it public

jviau (dismissed):

nytian (approved):

PR #1189, opened 2025-03-06
Don't fail orchestrations on missing blobs from previous executions

AnatoliB:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/3022.

Sometimes a duplicate message for an execution that is already finished is received late, when this execution is already completed. For example, this can happen when a message is in fact processed, but the worker failed to remove it from the queue for any reason, so the message automatically reappears in the queue later, potentially when the execution is already finished. In most cases, this is not a problem, and this message will eventually be discarded without any negative consequences. However, here is a sequence of events that leads to a stuck orchestration:

1. A `TaskScheduled` message is delivered to a worker and the worker that starts executing an activity.
2. After successfully executing an activity, the worker fails to remove the `TaskScheduled` message from the queue, but the orchestration continues.
3. The activity returns a large amount of data (>~45K). This data is stored in a blob, and the history table entry contains a reference to this blob.
4. Eventually, the `TaskScheduled` message reappears in the queue and is eventually picked up by a worker. The worker starts executing the activity again (which is fine because Durable Tasks guarantee _at-least-once_ execution, but not _exactly-once_).
5. In the meantime, the orchestrator function invokes `ContinueAsNew`. As a result, all the blobs for the previous execution are deleted (but the history table entries remain).
6. Eventually, the worker finishes the second activity execution and produces a `TaskCompleted` message, which is eventually picked up.
7. This `TaskCompleted` message has the previous execution id, so the worker tries to load the history for that execution id. However, all the blobs are missing. The worker repeatedly tries to retrieve the blobs, hitting the same error, and making no progress on the orchestration anymore. The orchestration is stuck indefinitely for no good reason.

This PR addresses the problem by making the history retrieval logic rely on the execution ID recorded in the sentinel row and skip history entries for different execution IDs, so there will be no attempt to retrieve missing blobs or interpret these history entries in any other way.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -169,7 +169,7 @@ namespace DurableTask.AzureStorage.Tracking
                                                expectedExecutionId == sentinel?.GetString("ExecutionId")))
             {
                 // The most recent generation will always be in the first history event.
-                executionId = results.Entities[0].GetString("ExecutionId");
+                executionId = sentinel?.GetString("ExecutionId") ?? results.Entities[0].GetString("ExecutionId");
 
                 // Convert the table entities into history events.
                 var events = new List<HistoryEvent>(results.Entities.Count);
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -158,9 +158,12 @@ namespace DurableTask.AzureStorage.Tracking
 
             IList<HistoryEvent> historyEvents;
             string executionId;
-            TableEntity sentinel = null;
+
+            // The sentinel row should always be the last row
+            TableEntity sentinel = results.Entities.LastOrDefault(e => e.RowKey == SentinelRowKey);
+
             TrackingStoreContext trackingStoreContext = new TrackingStoreContext();
-            if (results.Entities.Count > 0)
+            if (results.Entities.Count > 0 && sentinel?.GetString("ExecutionId") == expectedExecutionId)
             {
                 // The most recent generation will always be in the first history event.
                 executionId = results.Entities[0].GetString("ExecutionId");
@@ -176,11 +179,9 @@ namespace DurableTask.AzureStorage.Tracking
                         break;
                     }
 
-                    // The sentinel row does not contain any history events, so save it for later
-                    // and continue
-                    if (entity.RowKey == SentinelRowKey)
+                    // The sentinel row does not contain any history events, so ignore and continue
+                    if (entity == sentinel)
                     {
-                        sentinel = entity;
                         continue;
                     }
 
@@ -215,7 +216,6 @@ namespace DurableTask.AzureStorage.Tracking
             // A sentinel won't exist only if no instance of this ID has ever existed or the instance history
             // was purged.The IsCheckpointCompleteProperty was newly added _after_ v1.6.4.
             DateTime checkpointCompletionTime = DateTime.MinValue;
-            sentinel = sentinel ?? results.Entities.LastOrDefault(e => e.RowKey == SentinelRowKey);
             ETag? eTagValue = sentinel?.ETag;
             if (sentinel != null &&
                 sentinel.TryGetValue(CheckpointCompletedTimestampProperty, out object timestampObj) &&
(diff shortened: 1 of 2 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -184,7 +185,20 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
...
+                    bool success = await this.TryDecompressLargeEntityPropertiesAsync(
+                        entity,
+                        trackingStoreContext.Blobs,
+                        instanceId,
+                        executionId,
+                        cancellationToken);
+                    if (!success)
+                    {
+                        // Some properties were not retrieved because we are apparently trying to load
+                        // outdated history. No reason to raise an exception here, as this does not
```
cgillum, line 197: I'm a bit confused about how/why we'd find outdated history at this point. Shouldn't we have filtered out all the outdated history events via the check on line 173 (where we check the `ExecutionId` value)?

cgillum (commented):
Thank you for this PR. If I understand the description correctly, the root cause of the problem is that we're attempting to load old, partially deleted history. If I understand the fix correctly, we're trying to catch a very specific exception (blob not found) for a very narrow use case in order to detect this condition.

I wonder if we can make a broader fix and still keep things relatively simple. Note that whenever we load history, we're already going to load the sentinel row, so there shouldn't be a need to fetch it explicitly like we are in this PR. What if we instead loaded the history as normal but validated the history first before attempting to download or deserialize any blobs. By "validating", I mean (at a minimum) confirming that each entity has an execution ID that matches the sentinel row. If validation, then we don't even attempt to load the blob at all and discard the message. This would be more efficient and resolve a broader range of potential issues.

Thoughts on this alternate approach?

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
AnatoliB: Exactly, this is the source of our troubles. Line 173 doesn't perform proper filtering because `executionId` contains an old execution ID, which happens because we passed an old `expectedExecutionId` to `GetHistoryEntitiesResponseInfoAsync`, so `results.Entities[0]` also belongs to the outdated history, and line 173 filters out everything unrelated to the _old_ execution. Eventually, if we don't encounter missing blobs, we hit this line: https://github.com/Azure/durabletask/blob/3bc255cb4df1565e5ac760ebd1dc040f58f02e5e/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L1072

AnatoliB:
> Thank you for this PR. If I understand the description correctly, the root cause of the problem is that we're attempting to load old, partially deleted history. If I understand the fix correctly, we're trying to catch a very specific exception (blob not found) for a very narrow use case in order to detect this condition.

@cgillum Yes, you understand the description and the approach correctly. I've been intentionally trying to make it a very targeted fix because I'm not super-familiar with this code, so I may not realize all the consequences.

> I wonder if we can make a broader fix and still keep things relatively simple. Note that whenever we load history, we're already going to load the sentinel row, so there shouldn't be a need to fetch it explicitly like we are in this PR. What if we instead loaded the history as normal but validated the history first before attempting to download or deserialize any blobs. By "validating", I mean (at a minimum) confirming that each entity has an execution ID that matches the sentinel row. If validation, then we don't even attempt to load the blob at all and discard the message. This would be more efficient and resolve a broader range of potential issues.

Thank you for pointing out that we already have the sentinel row, so no need to fetch it separately _or_ even change the query 
in `GetHistoryEntitiesResponseInfoAsync`, I missed this somehow! I wanted to avoid hurting performance just for the purposes of mitigating a relatively rare corner case. But, since we already have the sentinel anyway, this may be a good idea, I'll try it.

AnatoliB:
@cgillum I've updated the fix to rely on the execution ID in the sentinel row and skip loading history if it doesn't match the expected ID.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -155,11 +155,18 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
...
-            if (results.Entities.Count > 0)
+
+            // If expectedExecutionId is provided but it does not match the sentinel executionId,
+            // it may belong to a previous generation. In that case, treat it as an unknown executionId
+            // and skip loading history.
+            if (results.Entities.Count > 0 && (expectedExecutionId == null ||
+                                               expectedExecutionId == sentinel?.GetString("ExecutionId")))
             {
                 // The most recent generation will always be in the first history event.
                 executionId = results.Entities[0].GetString("ExecutionId");
```
cgillum, line 172: Would it make sense to change this line as well to something like the following?

```suggestion
                executionId = sentinel?.GetString("ExecutionId") ?? results.Entities[0].GetString("ExecutionId");
```

I'm just wondering if we want to find a way to ensure we never attempt to load irrelevant table entities. The check above ensures that we don't load history for messages that have an old execution ID, but I wonder if there could be other situations where we're unintentionally loading irrelevant table entities. I think this likely won't happen because we won't have a sentinel that doesn't match the first entity we get back (see line 179) but it's something we could potentially consider as defense-in-depth.

cgillum (approved):
I like the simplicity of the new approach! Does the PR description need to be updated?

In any case, I'm good with this change. I left a comment for an optional change, but I'm fine if you think it's not worth risking further changes.

AnatoliB:
Updated the PR description

PR #1190, opened 2025-03-08
Distributed Tracing for Entities (In-Process)

sophiatev:
This PR adds distributed tracing for entities, specifically for

- Calling entities (from an orchestration)
- Signaling entities (from an orchestration, a client, or another entity)
- An entity starting an orchestration

In order to correctly connect parent and child spans in the case of an entity starting an orchestration, we need to pass the trace information to the `TaskOrchestrationDispatcher.ProcessCreateSubOrchestrationInstanceDecision` method, which is ultimately what the `DurableEntityContext.StartNewOrchestration` call hits. To do this, we attach the trace information to the orchestration tags via the new `OrchestrationTags.TraceParent` and `OrchestrationTags.TraceState` constants. 

We also remove some existing tracing in favor of the entities tracing now occurring in the WebJobs repo. To do this we add information to `TaskOrchestrationDispatcher` and `TaskHubClient` to indicate whether entities are enabled (or in the latter case, whether the event being raised corresponds to an entity). `TraceHelper.StartTraceActivityForEventRaisedFromWorker` and  `TraceHelper.StartActivityForNewEventRaisedFromClient` then use this information as well as the instance ID format to determine whether the event corresponds to an entity, and if so no `Activity` is created. 

For more details and an example of the new distributed tracing logic, please see this [PR](https://github.com/Azure/azure-functions-durable-extension/pull/3066) in the WebJobs repo.


src/DurableTask.Core/Command/CreateSubOrchestrationAction.cs
```diff
@@ -50,5 +51,25 @@ public class CreateSubOrchestrationAction : OrchestratorAction
...
+
+        /// <summary>
+        /// Parent trace flags of the sub-orchestration.
+        /// </summary>
+        public ActivityTraceFlags ParentTraceFlags { get; set; }
+
+        /// <summary>
+        /// Parent trace state of the sub-orchestration.
+        /// </summary>
+        public string? ParentTraceState { get; set; }
```
cgillum, line 73: Instead of adding so many properties (or method parameters in other updated files) can we just introduce a new class that encapsulates all these properties? Or even better, can we just reuse the existing `DistributedTraceContext` class?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -357,6 +357,14 @@ internal static void EmitTraceActivityForSubOrchestrationFailed(
             OrchestrationInstance? instance,
             string? targetInstanceId)
         {
+
+            // If either the target or source of this event is an entity, we don't want to start a trace activity for it because entity tracing
+            // is handled separately in the WebJobs extension
```
cgillum, line 362: Ideally DTFx shouldn't have any knowledge of WebJobs or dependencies on its behavior. Rather, the WebJobs stuff should be consider a wrapper around DTFx. This comment makes me wonder if there's something wrong with our abstractions/layering. If we think the code here is still the right thing to do, then I suggest we describe the reasoning differently here. Alternatively, we could pass in some parameter to indicate whether or not to start the trace activity.

sophiatev: Unfortunately I don't see a way around it in this case since we basically need to tell the DTFx trace activity to "turn itself off" in favor of the WebJobs one. An alternative I was considering was to instead move the conditional logic out to [this point](https://github.com/Azure/durabletask/blob/3bc255cb4df1565e5ac760ebd1dc040f58f02e5e/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L1160) where we actually call this method, but the if check would remain the same so I don't think it actually improves anything... maybe I'm missing something but it seems unavoidable to me

bachuv: I think we can update the comment to include an explanation about how we don't want to include spans about external events when it's about an entity. We can also pass in a parameter or even do a check to see if it's an entity before calling this method.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1083,7 +1089,7 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
             CreateSubOrchestrationAction createSubOrchestrationAction,
             OrchestrationRuntimeState runtimeState,
             bool includeParameters,
-            Activity? parentTraceActivity)
+            DistributedTraceContext? parentTraceContext)
```
bachuv, line 1092: Can you explain this update and why we're replacing `Activity` with `DistributedTraceContext`?

src/DurableTask.Core/Tracing/TraceHelper.cs
cgillum: Got it. In that case, let's keep it as-is but update the comment per Varshi's suggestion just so we provide a bit more context.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
sophiatev: `DistributedTraceContext` is how we pass the parent trace information from WebJobs to DurableTask.Core when an entity is starting an orchestration. Ultimately that information is used by `ProcessCreateSubOrchestrationInstanceDecision` to set the parent trace context of the event for creating the suborchestration. So in the case that we need to use this propagated information I needed to change this method to accept a `DistributedTraceContext`, and then create a `DistributedTraceContext` using the current activity as the parent [here](https://github.com/Azure/durabletask/blob/d3f25f3a8d4d1e9cae7348494c4d259e6f3890eb/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L431) in the case that we lack propagated information.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -425,12 +425,18 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
-                                        this.ProcessCreateSubOrchestrationInstanceDecision(
-                                            createSubOrchestrationAction,
-                                            runtimeState,
-                                            this.IncludeParameters,
-                                            traceActivity));
+                                    this.ProcessCreateSubOrchestrationInstanceDecision(
+                                        createSubOrchestrationAction,
+                                        runtimeState,
+                                        this.IncludeParameters,
+                                        createSubOrchestrationAction.ParentTraceContext ?? traceContext));
```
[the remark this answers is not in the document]
sophiatev: Hmm doesn't it match the previous indentation of the other parameters in the method call? Maybe I'm not understanding what you're referring to

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -425,12 +425,22 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                                     break;
                                 case OrchestratorActionType.CreateSubOrchestration:
                                     var createSubOrchestrationAction = (CreateSubOrchestrationAction)decision;
+                                    DistributedTraceContext? traceContext = null;
+                                    if (traceActivity != null)
+                                    {
+                                        traceContext = new DistributedTraceContext(
+                                            traceActivity.TraceId.ToString(),
```
jviau, line 432: `traceparent` is the id, not the traceid.

```suggestion
                                            traceActivity.Id.ToString(),
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1096,7 +1106,7 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
                 historyEvent.Input = createSubOrchestrationAction.Input;
             }
 
-            ActivitySpanId clientSpanId = ActivitySpanId.CreateRandom();
+            ActivitySpanId clientSpanId = parentTraceContext?.ParentSpanId != null ? ActivitySpanId.CreateFromString(parentTraceContext.ParentSpanId.AsSpan()) : ActivitySpanId.CreateRandom();
```
jviau, line 1109: This doesn't look right to me. Creating a sub orchestration should always be a new span ID. I think I understand the problem you are trying to solve though. During entity processing, there is a disconnect between when a user schedules an orchestration and when that actually occurs.

I'd like to avoid needing this change. See my proposal for using `tags` to override trace parent and state.

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -357,6 +357,13 @@ internal static void EmitTraceActivityForSubOrchestrationFailed(
             OrchestrationInstance? instance,
             string? targetInstanceId)
         {
+
+            // We don't want to emit tracing for external events when they are related to entities
+            if ((targetInstanceId != null && targetInstanceId.StartsWith("@")) || (instance != null && instance.InstanceId.StartsWith("@")))
```
jviau, line 362: I am not a fan of this approach of assuming any id that starts with `@` will have their events emitted elsewhere. There may be standalone customers not using entities with use `@` as a prefix. We shouldn't break their experience here.

jviau (changes requested):
I am concerned the current approach is leaking too many implementation details from DF WebJobs package into this layer. I'd like to avoid that, we cannot compromise this feature for standalone customers as part of supporting DF.

I also want to avoid any breaking changes.

src/DurableTask.Core/Tracing/TraceHelper.cs
sophiatev: I tried implementing an alternative approach where I added a flag to  the `RequestMessage` that indicates whether or not we should create a trace. I then attempt to parse the `EventRaisedEvent.Input` into a `RequestMessage` here and check for this flag. The problem turned out to be that that we do not catch returns from a `CallEntity` request. These returns do not have type `RequestMessage` (they just look like `{result: 24}` for example) and as far as I can tell the only way to catch them is to check the format of the instance ID. It was my understanding that starting with "@" was reserved only for entities, and there seem to be examples of this assumption in the code already like [here](https://github.com/Azure/durabletask/blob/894674800424c4dc10a4b865949a018878b6cc72/src/DurableTask.Core/Common/Entities.cs#L30). But if you would still rather we not use this check, do you have any suggestions for alternative approaches?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
sophiatev: I think in our case we don't want to create a new span ID. Here is an example of a trace where we do not create a new span ID:

image

As you can see the start of the orchestration (the `orchestration:SumOfCountsOrchestration` event) appears under the event indicating that an entity created an orchestration (`entity:create_orchestration`), but if we create a new span ID these are instead disjoint

Screenshot 2025-03-18 002521

src/DurableTask.Core/Tracing/TraceHelper.cs
jviau: > it was my understanding that starting with "@" was reserved only for entities

Starting with `@` is only restricted to entities when entities are enabled, which is not by default. The history is that entities was only added in the past years, and using `@` for orchestrations has been supported long before then.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -425,12 +425,31 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                                     break;
                                 case OrchestratorActionType.CreateSubOrchestration:
                                     var createSubOrchestrationAction = (CreateSubOrchestrationAction)decision;
+                                    ActivityContext? parentTraceContext = traceActivity?.Context;
+                                    var createNewSpanId = true;
+                                    if (createSubOrchestrationAction.Tags != null 
+                                        && createSubOrchestrationAction.Tags.ContainsKey(OrchestrationTags.TraceParent)
```
jviau, line 431: Consider using `TryGetValue()` to perform the check and get the value in one call.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1083,7 +1102,8 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
             CreateSubOrchestrationAction createSubOrchestrationAction,
             OrchestrationRuntimeState runtimeState,
             bool includeParameters,
-            Activity? parentTraceActivity)
+            ActivityContext? parentTraceContext,
+            bool createNewSpanId)
```
jviau, line 1106: Would it be possible to refactor this so that we pass in `ActivitySpanId? spanId`? Having the `createNewSpanId` co-dependent on `parentTraceContext` is not ideal to me, as if you pass in `null, false` respectively, it will fault.

src/DurableTask.Core/Tracing/DistributedTraceContext.cs
```diff
@@ -81,4 +81,4 @@ public string? TraceState
         [DataMember]
         public DateTimeOffset? ActivityStartTime { get; set; }
     }
-}
+}
```
jviau, line 84: Small nit: when you see this red icon it means you do not have a new line at the end of the file. The "proper" way to end a file is an empty newline. I have it in quotes because I have no idea why it is "proper", just that analyzers and github will complain if you don't.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1122,9 +1142,13 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
                 Version = createSubOrchestrationAction.Version
             };
 
-            if (parentTraceActivity != null)
+            if (parentTraceContext is ActivityContext parentContext)
             {
-                ActivityContext activityContext = new ActivityContext(parentTraceActivity.TraceId, clientSpanId, parentTraceActivity.ActivityTraceFlags, parentTraceActivity.TraceStateString);
+                ActivityContext activityContext = new ActivityContext(
```
jviau, line 1147: I am a bit confused by this logic (I think it predates your PR here): the span ID of both the parent trace context _and_ `historyEvent.ClientSpanId` are the same. What is the reason for having both if they are identical?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1083,7 +1102,8 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
             CreateSubOrchestrationAction createSubOrchestrationAction,
             OrchestrationRuntimeState runtimeState,
             bool includeParameters,
-            Activity? parentTraceActivity)
```
jviau, line 1086: When the traceparent orchestration tags are present, do we still want to create this activity? Genuinely asking - not sure what the proper behavior should be.

src/DurableTask.Core/Tracing/TraceHelper.cs
cgillum: For the purposes of unblocking this PR, I suggest we go ahead with Sophia's current approach. I agree that these assumptions already exist in many places, and we're not breaking any existing behavior with this particular change. We can certainly document it in our distributed tracing documentation, and I suggest we add a comment here too explaining the possibility of mis-labeling an orchestration as an entity.

@sophiatev rather than checking the string manually like this, let's leverage the API you linked to. Something like this:
```suggestion
            if (Entities.IsEntityInstance(targetInstanceId) || Entities.IsEntityInstance(instance.InstanceId))
```

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Could it be that the parent span ID we're using was initialized already to be used in this way?

I agree that we want the first image.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
sophiatev: Maybe I'm misunderstanding the question, but this isn't the created activity but rather the parent trace context to the `ExecutionStartedEvent` made [here](https://github.com/Azure/durabletask/blob/894674800424c4dc10a4b865949a018878b6cc72/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L1106). In the case that tags are present we make sure that the parent trace context is extracted from those tags. Otherwise we just use `Activity.Current?.Context` as the parent trace context.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
sophiatev: Hm, I'm not sure, I guess I'm not clear how the `historyEvent` and `startedEvent` are used. But it doesn't seem obviously wrong to me that they both would have the same parent span ID? Since the span ID of the parent trace context is what becomes the parent span ID of the `startedEvent`.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -425,12 +425,29 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
+                                    ActivityContext? parentTraceContext = traceActivity?.Context;
+                                    var spanId = ActivitySpanId.CreateRandom();
+                                    if (createSubOrchestrationAction.Tags != null 
+                                        && createSubOrchestrationAction.Tags.TryGetValue(OrchestrationTags.TraceParent, out string traceParent) 
+                                        && createSubOrchestrationAction.Tags.TryGetValue(OrchestrationTags.TraceState, out string traceState))
+                                    {
+                                        try
+                                        {
+                                            parentTraceContext = ActivityContext.Parse(traceParent, traceState);
+                                            spanId = parentTraceContext.Value.SpanId;
```
jviau, line 437: The approach you have here is that the tag provided span will replace the orchestration _client_ span. Want to make sure that is the intention. If so, we will need to update another area.

Currently the `historyEvent.clientSpanId` will be used when the sub-orchestration completes to emit the client span (which represents dispatching the orchestration). With a tag provided parent context this span is now external to us and out of our control. We need to skip emitting that client span.

_ALTERNATIVELY_ -- the other option is to have the tag-provided span be the _parent_ of the client span we will emit. In which case you will need to refactor this to generate an `ActivityContext` with `parentTraceContext` as the parent, and a `ActivitySpanId.CreateRandom()` span ID.

I am not sure which approach you want - 1/ the external trace context replaces the client span, or 2/ the external trace context is the parent of the client span.

jviau (changes requested):

src/DurableTask.Core/Entities/EventFormat/ResponseMessage.cs
```diff
@@ -32,6 +34,9 @@ internal class ResponseMessage : EntityMessage
         [IgnoreDataMember]
         public bool IsErrorResult => this.ErrorMessage != null || this.FailureDetails != null;
 
+        [DataMember(Name = "requestInfo", EmitDefaultValue = false)]
+        public RequestInformation? RequestInfo { get; set; }
+
```
bachuv, line 39: Why do we need to add the request information to the `ResponseMessage"?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1122,7 +1122,32 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
...
+                        EntityId.FromString(runtimeState.OrchestrationInstance!.InstanceId).Name,
+                        createSubOrchestrationAction.InstanceId!,
+                        parentTraceContext,
+                        requestTime);
+                    if (createOrchestrationActivity != null)
+                    {
+                        startedEvent.SetParentTraceContext(createOrchestrationActivity);
+                    }
+                }
+            }
```
bachuv, line 1149: Can we add this code to a helper method to improve code readability?

src/DurableTask.Core/Entities/EventFormat/ResponseMessage.cs
```diff
@@ -47,5 +52,22 @@ public override string GetShortDescription()
                 return $"[OperationSuccessful ({Result?.Length ?? 0} chars)]";
             }
         }
+        internal class RequestInformation
+        {
+            [DataMember]
+            public string? Operation { get; set; }
+
+            [DataMember]
+            public DateTime? ScheduledTime { get; set; }
```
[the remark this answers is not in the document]
sophiatev: I'm only using `DateTime `here to mimic the fact that `RequestMessage `also does for its [ScheduledTime](https://github.com/Azure/durabletask/blob/1514129bac2cbdc67b664fca61f919ff1c225ee3/src/DurableTask.Core/Entities/EventFormat/RequestMessage.cs#L66) field (and this is meant to replicate that information)

src/DurableTask.Core/Tracing/Schema.cs
```diff
@@ -33,5 +33,16 @@ internal static class Status
...
         }
+
+        internal static class Entity
+        {
+            internal const string Type = Task.Type;
+            internal const string EntityId = "durabletask.entity.entity_id";
+            internal const string TargetInstanceId = "durabletask.entity.target_instance_id";
+            internal const string TargetEntityId = "durabletask.entity.target_entity_id";
+            internal const string EntityOperation = "durabletask.entity.entity_operation";
+            internal const string ErrorMessage = "durabletask.entity.error_message";
```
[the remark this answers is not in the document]
sophiatev: Ah sorry, this bled over from the isolated app case. But either way, I think in this case it makes sense to have a separate attribute - this is used specifically in the case that an entity failed to process a request, in which case we use the [OperationResult.FailureDetails](https://github.com/Azure/durabletask/blob/1514129bac2cbdc67b664fca61f919ff1c225ee3/src/DurableTask.Core/Entities/OperationFormat/OperationResult.cs#L48) (or `ErrorMessage`) field to populate the trace created for the entity invocation with information about why it failed.

src/DurableTask.Core/TaskHubClient.cs
```diff
@@ -745,7 +745,8 @@ void CreateAndTrackDependencyTelemetry(TraceContextBase? requestTraceContext)
         /// <param name="orchestrationInstance">Instance in which to raise the event</param>
         /// <param name="eventName">Name of the event</param>
         /// <param name="eventData">Data for the event</param>
-        public async Task RaiseEventAsync(OrchestrationInstance orchestrationInstance, string eventName, object eventData)
+        /// <param name="entityEvent">Whether or not this event corresponds to an entity (false by default)</param>
+        public async Task RaiseEventAsync(OrchestrationInstance orchestrationInstance, string eventName, object eventData, bool entityEvent = false)
```
jviau, line 749: Add a parameter (even optional) is a breaking change. This will need to be split out to a different method.

jviau (changes requested):

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -1122,8 +1124,22 @@ TaskMessage ProcessCreateSubOrchestrationInstanceDecision(
                 Version = createSubOrchestrationAction.Version
             };
 
-            if (parentTraceActivity != null)
+            // If this is an orchestration triggered by an entity, we will attempt to use the trace context provided in the CreateSubOrchestrationAction.Tags as the parent trace context rather than the current Activity.
```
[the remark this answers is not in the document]
sophiatev: I'm struggling to find a way to do it without referencing WebJobs from this repo, but maybe you have a suggestion?

The reason we do this is that we create the `Activity` for the entity starting the orchestration on the WebJobs side, but when we reach this point in the code, empirically I discovered that `Activity.Current` does not reflect the one created in WebJobs (maybe due to a thread-switch?) This is why I have to manually pass the `ActivityContext` from WebJobs to DurableTask.Core in this case via the orchestration tags. 

If it helps, I can add something like "since `Activity.Current` is not the `Activity` created for the orchestration start" but I feel like that might cause more questions than answers.

bachuv (approved):

PR #1194, opened 2025-03-25
Partial Purge Support

YunchuWang:
add a flag in purgeresult class to indicate whether purge is fully completed or partially. this is part of changes required for making sdk support partial purge instances 

related changes:
https://github.com/microsoft/durabletask-protobuf/pull/39
https://github.com/microsoft/durabletask-dotnet/pull/400


src/DurableTask.Core/PurgeResult.cs
```diff
@@ -22,14 +22,21 @@ public class PurgeResult
         /// Initializes a new instance of the <see cref="PurgeResult" /> class.
         /// </summary>
         /// <param name="deletedInstanceCount">The number of instances deleted.</param>
-        public PurgeResult(int deletedInstanceCount)
+        /// <param name="isComplete">A value indicating whether the purge operation is complete.</param>
+        public PurgeResult(int deletedInstanceCount, bool? isComplete = null)
```
cgillum, line 26: Adding a parameter to a method is a binary-breaking change. To make this non-breaking, you'll need to remove the parameter and make the new property below settable.

cgillum: Actually, a better solution for this project will be to create a second constructor overload that has both properties (non-null).

YunchuWang: make sense, updated

src/DurableTask.Core/PurgeResult.cs
```diff
@@ -27,9 +27,28 @@ public PurgeResult(int deletedInstanceCount)
...
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PurgeResult" /> class.
+        /// </summary>
+        /// <param name="deletedInstanceCount">The number of instances deleted.</param>
+        /// <param name="isComplete">A value indicating whether the purge operation is complete. 
+        /// If true, the purge operation is complete. All instances were purged.
+        /// If false, not all instances were purged. Please purge again.
+        /// If null, default to legacy purge behavior. All instances were purged.</param>
```
cgillum, line 37:
```suggestion
        /// If null, whether or not all instances were purged is undefined.</param>
```

cgillum (approved):

PR #1197, opened 2025-03-27
Fix renew/delete race in DT.AzureStorage

### LINKED ISSUE #1196: Duplicate activity execution due to pop-receipt mismatch
In DurableTask.AzureStorage v2.0.1, there is a race condition where we attempt to renew a message at the same time we delete it. If the renew happens slightly before the delete, it can result in the delete operation failing, and the activity getting run a second time after the visibility timeout expires.

This might be a regression introduced in v2.0.0 when we moved to the new Azure Storage SDK. We had this exact problem in earlier versions of the SDK. However, the retry fix we had for this appears to not work correctly in the newer version.

cgillum:
Fixes #1196 

This issue was discovered as part of investigating ICM#608312432. More details in the issue description. The fix is to retry the delete operation if it fails due to a pop-receipt mismatch. This PR also makes small logging improvements that should help make these kinds of problems easier to diagnose.

I created a console app to reproduce the problem locally and confirm the error details for this kind of scenario.

### DIFF
--- a/src/DurableTask.AzureStorage/Storage/Queue.cs
+++ b/src/DurableTask.AzureStorage/Storage/Queue.cs
@@ -83,6 +83,8 @@ namespace DurableTask.AzureStorage.Storage
                     queueMessage.PopReceipt,
                     cancellationToken)
                 .DecorateFailure();
+
+            this.stats.MessagesUpdated.Increment();
         }
 
         public async Task<QueueMessage?> GetMessageAsync(TimeSpan visibilityTimeout, CancellationToken cancellationToken = default)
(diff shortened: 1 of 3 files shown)


src/DurableTask.AzureStorage/Storage/Queue.cs
```diff
@@ -83,6 +83,8 @@ await this.queueClient
                     queueMessage.PopReceipt,
                     cancellationToken)
                 .DecorateFailure();
+
+            this.stats.MessagesUpdated.Increment();
```
bachuv, line 87: Is this change part of the logging improvements that were made in this PR?

nytian (approved):

src/DurableTask.AzureStorage/Storage/Queue.cs
cgillum: It's part of it, yes. I want to make sure we're tracking message deletes as part of the counts that we report in our logs.

PR #1198, opened 2025-03-28
Distributed Tracing for Entities (Isolated)

sophiatev:
This PR adds support for distributed tracing for entities in the .NET isolated framework. This repo is where the trace `Activities` are actually created for signaling and calling entities and for entities starting orchestrations from an isolated app. 

- `RequestMessage`, `OperationRequest`, `OperationResult`, `SendSignalOperationAction`, and `StartNewOrchestrationOperationAction` were altered to include extra information from the durabletask-dotnet repo where the requests to entities are actually generated, and where the requests are also executed. This extra information includes the time of the requests, the end time of the execution and any error messages, and parent trace contexts. 
    - It's worth noting that the `CreateTrace` field added to `RequestMessage` is only used in the isolated case to indicate that we want to make an entity-specific trace for this request (it is set to true by the appropriate signal/call methods in durabletask-dotnet). In the in-process case, all of the traces are created in the WebJobs repo, so this field is not populated (and will be false by default). 
- `TraceHelper`, `Schema`, and `TraceActivityConstants` were updated with the instantiation of the entity-specific trace `Activities`  
-  `ClientyEntityHelpers` and `OrchestrationEntityContext` methods that generate `EntityMessageEvents` which are used by the durabletask-dotnet repo were updated to attach this above-mentioned additional information to the message events. 
- `TaskEntityDispatcher`, which is where all entity requests end up (orchestrations calling/signaling entities, clients signaling entities, entities signaling other entities, and entities starting orchestrations), and where the entities are actually invoked to fulfill the requests, was updated to instantiate the corresponding traces. One exception is that clients signaling entities via gRPC (i.e., when the `DurableEntityClient` is a `GrpcDurableEntityClient`) is handled in the WebJobs repo, where the call ultimately reaches the `LocalGrpcListener`. The PR for this repo is linked below. 
    - `TaskEntityDispatcher.StartTraceActivityForSignalingEntity` is used to create the `Activity` in the case of a client signaling an entity (via `ShimDurableEntityClient` in the dotnet repo, since the gRPC client call is handled by WebJobs) or in the case of an orchestration signaling an entity. In the former case, `ShimDurableEntityClient` has access to the correct parent trace context via `Activity.Current.Context` so it attaches this context to the request message itself.  `StartTraceActivityForSignalingEntity` then parses and uses this context as the parent to the `Activity` for signaling the entity. For an orchestration signaling an entity, the dotnet repo does not have access to the orchestration trace context and neither does `TaskEntityDispatcher`. In this case, the way the parent trace context is attached is via `TaskOrchestrationDispatcher.ProcessSendEvent`, where `Activity.Current.Context` holds the orchestration context. This method only has access to the associated `EventRaisedEvent`, so this is what it attaches the parent trace context to and is what is eventually parsed and used by `StartTraceActivityForSignalingEntity`. Finally, in the case of an orchestration calling an entity, the `Activity` is only created at the very end once the call has completed. The code at that point only has access to the `RequestMessage`, so `StartTraceActivityForSignalingEntity` attaches the parent trace context from the `EventRaisedEvent` to the `RequestMessage` such that it can eventually be used when making the `Activity` for the call. 

The various other PRs related to this effort are
- [durabletask-dotnet](https://github.com/microsoft/durabletask-dotnet/pull/404)
- [protobuf](https://github.com/microsoft/durabletask-protobuf/pull/40)
- [WebJobs](https://github.com/Azure/azure-functions-durable-extension/pull/3076)

It is worth noting that the `Activities` for signaling an entity in the isolated case will have longer durations than in the in-process case. In the in-process case, the `Activity` for a signal to an entity is created upon the request and almost immediately disposed. In the isolated case, we cannot immediately dispose the `Activity` upon the request since this would require creating the `Activity` in the dotnet repo where the request is generated. Instead, we create the `Activity` once the signal request reaches DurableTask.Core and is actually processed by the `TaskEntityDispatcher`, and pass the request time as the start time of the `Activity`. Its end time will therefore be much more offset from its start time (the request time) relative to the in-process case. This is not an issue for calls to entities since these are only ended once the operation completes (in the isolated case, once we send the result back to the orchestration instance, and in the in-process case once the entity invocation completes). 

This is also true in the case of a client creating an orchestration using the `ShimDurableTaskClient` - we only create the `Activity` for the orchestration once the request reaches DurableTask.Core and is processed by the `TaskOrchestrationDispatcher`. Therefore the duration of the create orchestration `Activity` will be much longer than in all other cases where the `Activity` is started upon the request and almost immediately ended afterwards. 

An example trace generated by this simple orchestration
image
looks as follows
image

Each signal request has type ActivityKind.Producer and each call request has type ActivityKind.Client (an entity starting an orchestration is also of type ActivityKind.Producer). When an entity actually processes the request, for a signal the span has type ActivityKind.Producer and for a call the span has type ActivityKind.Server. Note that the call to `add_to_other_entity_step_1` starts a cascade of entities signaling other entities until eventually the last call is simply an `add` to the third entity. 

If instead of starting the orchestration via an HTTP request we signal an entity to start the orchestration, the trace would look like this
image


src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -446,6 +473,124 @@ internal static void EmitTraceActivityForTimer(
...
+            }
+
+            newActivity.SetTag(Schema.Task.Type, TraceActivityConstants.Entity);
+            newActivity.SetTag(Schema.Task.Operation, signalEntity ? TraceActivityConstants.SignalEntity : TraceActivityConstants.CallEntity);
+            newActivity.SetTag(Schema.Task.InstanceId, entityId);
+
+            return newActivity;
+        }
+
+        internal static void EndActivitiesForProcessingEntityInvocation(List<Activity> traceActivities, List<OperationResult> results, FailureDetails? batchFailureDetails)
```
bachuv, line 550: Can you add a comment for this method?

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -446,6 +473,124 @@ internal static void EmitTraceActivityForTimer(
...
+                }
+            }
+            // This can happen if some of the operations failed and have no corresponding OperationResult
+            // There is no way to map the successful operation results to the corresponding operation requests or trace activities, so we will just "fail" the trace activities in this case and dispose them
+            else
+            {
+                string errorMessage = "Unable to generate a trace activity for the entity invocation even though it may have succeeded.";
+                if (batchFailureDetails is FailureDetails failureDetails)
+                {
+                    errorMessage += $" If it failed, it may be due to {failureDetails.ErrorMessage}";
```
bachuv, line 581: Do we only have an error message for the batch of operations and not for each operation?

sophiatev: Ideally no - ideally the `OperationResult.ErrorMessage` or `OperationResult.FailureDetails` field is populated in which case we can just add a tag to the `Activity` explaining why the entity invocation failed (done a few lines above [here](https://github.com/Azure/durabletask/blob/27858607ca282df9b9fd28f1b412374112d87dbb/src/DurableTask.Core/Tracing/TraceHelper.cs#L560)). The logic here is a sort of edge case in the situation that the batch wasn't fully processed, as in not all of the requests came back with an `OperationResult`. In that situation, since there's no way to know which requests failed as far as I can tell, we try to just use the batch-level failure details as a potential explanation for the failure and end all of the Activities.

We could presumably add extra information to the `OperationResult` object (like a request ID field to correlate it to the `OperationRequest` object) so that we know which requests were processed successfully and which failed. But since it doesn't seem possible that dotnet could ever fail to return the correct number of `OperationResult`s I figured it's not worth it (see the logic [here](https://github.com/microsoft/durabletask-dotnet/blob/3a59439fc577f920c8000d7503b111b6caac3f5c/src/Worker/Core/Shims/TaskEntityShim.cs#L78). In the case of an exception it catches it and adds the failure information to the  `OperationResult.FailureDetails` field)

bachuv: Makes sense. Thanks for the explanation!

src/DurableTask.Core/Tracing/TraceHelper.cs
```diff
@@ -39,9 +44,23 @@ public class TraceHelper
...
+            {
+                return null;
+            }
+
+            DateTimeOffset? startTime = null;
+            if (startEvent.Tags != null && startEvent.Tags.ContainsKey(OrchestrationTags.RequestTime) &&
+                DateTimeOffset.TryParse(startEvent.Tags[OrchestrationTags.RequestTime], CultureInfo.InvariantCulture, DateTimeStyles.AssumeUniversal, out DateTimeOffset requestTime))
+            {
+                startTime = requestTime;
+            }
```
bachuv, line 57: Can you add a comment for this code?

bachuv (approved):

bachuv (approved):

bachuv (approved):

PR #1199, opened 2025-04-06
Fix Concurrent Modification Error in IsControlQueueProcessingMessages

nytian:
Fixes an `InvalidOperationException` caused by concurrent modifications to `activeOrchestrationSessions` while `IsControlQueueProcessingMessages` is enumerating its values. This issue could surface as a `PartitionManagerError`.

This method is called when there are drain tasks, where concurrent access is expected. To ensure thread safety, we now add a lock before enumeration.

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
@@ -116,7 +116,8 @@ namespace DurableTask.AzureStorage
 
         public bool IsControlQueueProcessingMessages(string partitionId)
         {
-            return this.activeOrchestrationSessions.Values.Where(session => string.Equals(session.ControlQueue.Name, partitionId)).Any();
+            var sessionsSnapshot = this.activeOrchestrationSessions.Values.ToList();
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -116,7 +116,8 @@ public bool IsControlQueueReceivingMessages(string partitionId)
 
         public bool IsControlQueueProcessingMessages(string partitionId)
         {
-            return this.activeOrchestrationSessions.Values.Where(session => string.Equals(session.ControlQueue.Name, partitionId)).Any();
+            var sessionsSnapshot = this.activeOrchestrationSessions.Values.ToList();
```
cgillum, line 119: Looking at other uses of `this.activeOrchestrationSessions`, it seems we typically access it via a lock.

```csharp
lock (this.messageAndSessionLock)
{
    // ...
}
```

Could that be the proper fix for this case instead of copying the values to a new list?

nytian: Oh yes I noticed that. Let me update it—I just thought this might be the easier way. Thanks for the review!

cgillum (approved):

PR #1200, opened 2025-04-10
Add timeout wehn Read&Write Parititon Table 

nytian:
- Add a 2-second timeout when calling `ReadAndWriteTableAsync`  at TablePartitionManager.

This helps guard against cases where the read or write hangs silently, as Azure Storage client has its own retry policies. If the timeout is hit, a `PartitionManagerWarning` is logged and the operation is retried.
- Bump DTFx.AzureStorage and DTFx.ApplicationInsights minor to align with the  DTFx.Core version increment.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
@@ -171,6 +171,14 @@ namespace DurableTask.AzureStorage
         public bool AllowReplayingTerminalInstances { get; set; } = false;
 
         /// <summary>
+        /// Specifies the timeout (in seconds) for read and write operations on the partition table in partition manager V3 (table partition manager).
+        /// This helps detect and recover from potential silent hangs caused by Azure Storage client's internal retries. 
+        /// If the operation exceeds the timeout, a PartitionManagerWarning is logged and the operation is retried.
+        /// The default time is 2 seconds.
+        /// </summary>
+        public TimeSpan PartitionTableOperationTimeout { get; set; } = TimeSpan.FromSeconds(2);
+
+        /// <summary>
         /// If UseAppLease is true, gets or sets the AppLeaseOptions used for acquiring the lease to start the application.
         /// </summary>
         public AppLeaseOptions AppLeaseOptions { get; set; } = AppLeaseOptions.DefaultOptions;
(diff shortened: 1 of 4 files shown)


src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -118,7 +118,10 @@ async Task PartitionManagerLoop(CancellationToken gracefulShutdownToken, Cancell
 
                 try
                 {
-                    ReadTableReponse response = await this.tableLeaseManager.ReadAndWriteTableAsync(isShuttingDown, forcefulShutdownToken);
+                    using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(2));
```
nytian, line 121: Reading and writing the partition table should typically complete within milliseconds. And since we also need time to process the table info, I think a 2-second timeout is a reasonable bound. But please let me know if you have any thoughts about this.

cgillum: I suggest making this configurable. That way if we find out that this gets triggered too often in production and causes unexpected issues, we can give customers instructions for how to mitigate.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -170,6 +170,14 @@ public class AzureStorageOrchestrationServiceSettings
         /// </remarks>
         public bool AllowReplayingTerminalInstances { get; set; } = false;
 
+        /// <summary>
+        /// Specifies the timeout (in seconds) for read and write operations on the partition table in partition manager V3 (table partition manager).
+        /// This helps detect and recover from potential silent hangs caused by Azure Storage client's internal retries. 
+        /// If the operation exceeds the timeout, a PartitionManagerWarning is logged and the operation is retried.
+        /// The default time is 2 seconds.
+        /// </summary>
+        public TimeSpan PartitionTableOperationTimeout { get; set; } = TimeSpan.FromSeconds(2);
```
nytian, line 179: To make this configurable for DF customers, I also need to add this at[ webjobs extensions here ](https://github.com/Azure/azure-functions-durable-extension/blob/a5bfceeed793d204645f6268bd7514dc9f3555df/src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs). I will update that once this PR is merged

cgillum (approved):
LGTM!

PR #1201, opened 2025-04-16
Add versioning to DTFx orchestration dispatch

halspang:
This commit adds versioning similar to the feature created in durabletask-dotnet. This portion of versioning allows for an orchestration to be failed or abandonned based on the configuration provided.

The actual configuration will be provided in a follow-up PR in the web extensions as it should be handled through the host.json file.

### DIFF
--- a/src/DurableTask.Core/Settings/VersioningSettings.cs
+++ b/src/DurableTask.Core/Settings/VersioningSettings.cs
@@ -1,4 +1,17 @@
-﻿using System;
+﻿//  ----------------------------------------------------------------------------------
+//  Copyright Microsoft Corporation
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+
+using System;
 
 namespace DurableTask.Core.Settings
 {
--- a/src/DurableTask.Core/Settings/VersioningSettings.cs
+++ b/src/DurableTask.Core/Settings/VersioningSettings.cs
@@ -1,0 +1,106 @@
+﻿using System;
+
+namespace DurableTask.Core.Settings
+{
+    /// <summary>
+    /// Collection of settings that define the overall versioning behavior.
+    /// </summary>
+    public class VersioningSettings
+    {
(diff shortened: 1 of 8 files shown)


src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -13,6 +13,12 @@
 
 namespace DurableTask.Core
 {
+    using DurableTask.Core.Entities;
+    using DurableTask.Core.Exceptions;
+    using DurableTask.Core.Logging;
+    using DurableTask.Core.Middleware;
+    using DurableTask.Core.Settings;
+    using Microsoft.Extensions.Logging;
```
cgillum, line 21: Can these `using` reorderings across the different files be reverted? I'm guessing you have some auto-formatting which does this?

src/DurableTask.Core/TaskHubWorker.cs
```diff
@@ -76,13 +78,15 @@ public TaskHubWorker(IOrchestrationService orchestrationService)
         /// </summary>
         /// <param name="orchestrationService">Reference the orchestration service implementation</param>
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to use for logging</param>
-        public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null)
+        /// <param name="versioningSettings">The <see cref="VersioningSettings"/> that define how orchestration versions are handled</param>
+        public TaskHubWorker(IOrchestrationService orchestrationService, ILoggerFactory loggerFactory = null, VersioningSettings versioningSettings = null)
```
cgillum, line 82: Adding a new parameter, even nullable, is technically a binary breaking change. The right thing to do in this case is to create a new constructor overload (or if you don't want to do that, just make this a property that people can set separately).

cgillum (commented):
Looks great! Just a couple small things.

src/DurableTask.Core/TaskHubWorker.cs
halspang: Yes, it's the default formatter for VS. I'll try and make it play nice.

src/DurableTask.Core/TaskHubWorker.cs
halspang: I always forget that's a breaking change, thanks for the callout. I've duplicated the constructors and made `VersioningSettings` non-null in the new constructors as there was some ambiguity in signatures when using multiple default/optional values.

src/DurableTask.Core/Settings/VersioningSettings.cs
```diff
@@ -0,0 +1,106 @@
+﻿using System;
```
cgillum, line 1: Sorry for missing this in the first review, but both new files need to have the copyright header added.

halspang: Oh, good call, I thought I had this setup to automatically add the header but I guess not :)

cgillum (approved):

PR #1204, opened 2025-04-22
Fix blob URL handling for large messages with special characters in instanceId

nytian:
When the instanceId contains special characters like '|',' :', or spaces, the generated blob URL becomes invalid because these characters are not allowed unescaped in a URI. This results in the orchestrator returning a blob URL instead of the expected large message content. This PR fixes it by escaping invalid characters when constructing the blob URL.

Solves https://github.com/orgs/Azure/projects/745/views/1?pane=issue&itemId=96361588&issue=Azure%7Cazure-functions-durable-extension%7C3010


cgillum:
@nytian looks like the `LargeOverallTextMessagePayloads_BlobUrl` test is failing.

PR #1205, opened 2025-04-27
Added support for .net 8 - DTF Service Fabric

jabhalla:
# Added Support for .NET 8 - DTF Service Fabric Provider

## Summary
This update introduces support for .NET 8 within the DTF Service Fabric Provider, alongside existing support for .NET Framework. Key changes and improvements are outlined below.

## Changes
1. **Service Structure**
   - Services are now created per target framework.
   - Each framework has a separate folder containing:
     - `Controller`
     - `TaskHubProxyListener`
     - `Startup`

2. **Framework-specific Integrations**
   - **.NET Framework** continues to use **OWIN**.
   - **.NET Core** (including .NET 8) integrates with **Kestrel**.

3. **TaskHubProxyListenerBase**
   - Introduced a new `TaskHubProxyListenerBase` class to abstract and centralize common functionality across frameworks.

4. **Endpoint Adjustments**
   - In .NET Core, the same path cannot be used for two endpoints.
   - Created separate endpoints for:
     - `orchestration`
     - `orchestrationAll`
   - Updated `RemoteOrchestrationServiceClient` to align with these changes.


jabhalla:
@microsoft-github-policy-service agree company="Microsoft"

jabhalla:
@cgillum / @jviau - Can you please add @shankarsama as reviewer? I would appreciate a review from one of you gus as well. Also, can you pelase approve the ci pipeline?

Thanks

Directory.Packages.props
jviau, line 1: There is a bunch of package changes here that will impact all projects in this repo. Have you verified what the impact to other packages are? Keep in mind we **cannot** have any major version revs of any transitive dependency. This is considered a breaking change to customers and we would need to also major version rev.

src/DurableTask.AzureServiceFabric/Service/.net8/FabricOrchestrationServiceController.cs
jviau, line 1: Recommend removing the `.` from this folder name. `.` has some implicit meaning of being a hidden folder, so it isn't good to keep source code in them. Same with the `.net462` folder.

src/DurableTask.AzureServiceFabric/DurableTask.AzureServiceFabric.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),DurableTask.sln))\tools\DurableTask.props" />
+
   <PropertyGroup>
-    <TargetFrameworks>net462;net472</TargetFrameworks>
+    <TargetFrameworks>net8.0;net462;net472</TargetFrameworks>
```
jviau, line 6: These changes look fairly extensive, especially adding a new TFM. Have you evaluated what version change needs to happen? if there are any breaking changes (even revving transitive dependencies across major versions), then this package will need a major version rev as well.

jviau (changes requested):

jviau:
These changes are very large and impactful. I think it will need to go through a breaking change review and potential major version rev.

src/DurableTask.AzureServiceFabric/Service/.net8/FabricOrchestrationServiceController.cs
jabhalla: Renamed.

Directory.Packages.props
jabhalla: None of the other project's target .net8. All projects continue to be on the same version. Updated the Packages.props to a more understandable structure.

src/DurableTask.AzureServiceFabric/DurableTask.AzureServiceFabric.csproj
```diff
@@ -14,29 +15,42 @@
...
     <PackageReference Include="Newtonsoft.Json" />
+	<Compile Remove="Service\net462\**" />
+	<Compile Remove="Service\net8\**" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <PackageReference Include="Newtonsoft.Json" />
+	<PackageReference Include="Microsoft.AspNetCore.Mvc" />
+	<PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" />
+	<Compile Remove="Service\net462\**" />
```
cgillum, line 43: There are a mix of tabs and spaces, causing indentation inconsistencies. Please standardize on spaces.

Directory.Packages.props
```diff
@@ -15,28 +15,26 @@
...
     <PackageVersion Include="Microsoft.Data.Services.Client" Version="5.8.5" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="2.1.1" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.1" />
-    <PackageVersion Include="Microsoft.ServiceFabric" Version="6.4.617" />
-    <PackageVersion Include="Microsoft.ServiceFabric.Data" Version="3.3.617" />
-    <PackageVersion Include="Microsoft.ServiceFabric.Services" Version="3.3.617" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="2.1.1" Condition="'$(TargetFramework)' != 'net8.0'" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.1" Condition="'$(TargetFramework)' != 'net8.0'" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="6.0.0" Condition="'$(TargetFramework)' != 'net8.0'" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" Condition="'$(TargetFramework)' != 'net8.0'" />
```
cgillum, line 27: I'm worried about all these `!= 'net8.0'` checks, and how we will maintain these if/when we need to move to `net10` in the future. Is there another way we can refactor this to make it easier to make future changes?

jviau: These property methods from MSBuild might do what you are asking: https://learn.microsoft.com/en-us/visualstudio/msbuild/property-functions?view=vs-2022#msbuild-targetframework-and-targetplatform-functions

jabhalla: @jviau - Can you please elaborate? I din't follow. Do you mean using IsTargetFrameworkCompatible?

`Condition="!$([MSBuild]::IsTargetFrameworkCompatible('net8.0', '$(TargetFramework)'))"`

jviau: Yes `IsTargetFrameworkCompatible` _might_ be what you want. But I can't remember the exact behavior/syntax off the top of my head. But you should be able to use it to effectively do a `>= net8.0`

PR #1208, opened 2025-05-01
Fix DrainAsync Reason at Logging and Prevent Queue Listening During Partition Release

nytian:
This PR fixes several issues on partition manager v3:

- The drain reason was always logged as "Shutdown," even when caused by lease loss. This is fixed by passing a CloseReason reason argument to CheckDrainTask.

- When draining completes and the lease is released, the worker incorrectly resumes listening to the queue. This is fixed by adding a !ReleasedPartition check to the isRenewingToDrainQueue logic.

### DIFF
--- a/src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
+++ b/src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
@@ -412,24 +412,6 @@ namespace DurableTask.AzureStorage.Partitioning
                         try
                         {
                             await this.partitionTable.ReplaceEntityAsync(partition, etag, forcefulShutdownToken);
-
-                            // Ensure worker is listening to the control queue iff either:
-                            // 1) worker just claimed the lease,
-                            // 2) worker was already the owner in the partitions table and is not actively draining the queue.
-                            //    Note that during draining, we renew the lease but do not want to listen to new messages.
-                            //    Otherwise, we'll never finish draining our in-memory messages.
-                            // When drain completes, and the worker may decide to release the lease. In that moment,
-                            // IsDrainingPartition can still be true but renewedLease is false — without checking
-                            // !releasedLease, the worker could incorrectly resume listening just before releasing the lease.
-                            bool isRenewingToDrainQueue = renewedLease && response.IsDrainingPartition && !releasedLease;
-                            if (claimedLease || !isRenewingToDrainQueue)
-                            {
-                                // Notify the orchestration session manager that we acquired a lease for one of the partitions.
-                                // This will cause it to start reading control queue messages for that partition.
-                                await this.service.OnTableLeaseAcquiredAsync(partition);
-                            }
-
-                            this.LogHelper(partition, claimedLease, stoleLease, renewedLease, drainedLease, releasedLease, previousOwner);
                         }
                         catch (DurableTaskStorageException ex) when (ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
                         {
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -412,6 +412,24 @@ public async Task<ReadTableReponse> ReadAndWriteTableAsync(bool isShuttingDown,
...
+                            if (claimedLease || !isRenewingToDrainQueue)
+                            {
+                                // Notify the orchestration session manager that we acquired a lease for one of the partitions.
+                                // This will cause it to start reading control queue messages for that partition.
+                                await this.service.OnTableLeaseAcquiredAsync(partition);
+                            }
+
+                            this.LogHelper(partition, claimedLease, stoleLease, renewedLease, drainedLease, releasedLease, previousOwner);
                         }
                         catch (DurableTaskStorageException ex) when (ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
```
cgillum, line 434: Does increasing the scope of the try/catch change any behavior? It seems the only place that would throw `DurableTaskStorageException` is the call to `ReplaceEntityAsync` on line 414, so it doesn't look like this change does anything. Or is there some other code path I'm not seeing, like something in `OnTableLeaseAcquiredAsync` which can throw?

The PR description mentioned logging, but again I'm not sure I understand how the current logging behavior is changed. Either way, if an exception is thrown, we rethrow and exit from this method.

cgillum (approved):

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
nytian: I moved the logging inside the try block because we should only log when `ReplaceEntityAsync` succeeds. Previously, the logging was outside the try block and relied solely on local variables like claimedLease, stoleLease, renewedLease, drainedLease, and releasedLease. So even if the table update failed, we would still log actions that didn’t actually happen.

nytian: My bad --  please forget my earlier message, I didn't see we already throw the exception. I will change it back

src/DurableTask.AzureStorage/Partitioning/TablePartitionManager.cs
```diff
@@ -426,8 +426,13 @@ public async Task<ReadTableReponse> ReadAndWriteTableAsync(bool isShuttingDown,
...
                         // 1) worker just claimed the lease,
-                        // 2) worker was already the owner in the partitions table and is not actively draining the queue. Note that during draining, we renew the lease but do not want to listen to new messages. Otherwise, we'll never finish draining our in-memory messages.
-                        bool isRenewingToDrainQueue = renewedLease & response.IsDrainingPartition;
+                        // 2) worker was already the owner in the partitions table and is not actively draining the queue.
+                        //    Note that during draining, we renew the lease but do not want to listen to new messages.
+                        //    Otherwise, we'll never finish draining our in-memory messages.
+                        // When drain completes, and the worker may decide to release the lease. In that moment,
+                        // IsDrainingPartition can still be true but renewedLease is false — without checking
+                        // !releasedLease, the worker could incorrectly resume listening just before releasing the lease.
```
bachuv, line 434: suggestion:
```suggestion
                        // Note that during draining, we renew the lease but do not want to listen to new messages.
                        // Otherwise, we'll never finish draining our in-memory messages.
                        // When draining completes, the worker may decide to release the lease. In that moment,
                        // IsDrainingPartition can still be true but renewedLease can be false — without checking
                        // !releasedLease, the worker could incorrectly resume listening just before releasing the lease.
```

bachuv (approved):

cgillum (approved):

PR #1211, opened 2025-05-07
add tag to TaskScheduledEvent

YunchuWang:
This pull request introduces a new property, `Tags`, to the `TaskScheduledEvent` class in `src/DurableTask.Core/History/TaskScheduledEvent.cs`. This property is a dictionary designed to store key-value pairs of strings, providing a flexible way to include additional metadata.

Related prs:
https://github.com/microsoft/durabletask-protobuf/pull/43/files
https://github.com/microsoft/durabletask-dotnet/pull/426
https://msazure.visualstudio.com/One/_git/AAPT-DTMB/pullrequest/12388270

### Key Change:
* Added a new `Tags` property to the `TaskScheduledEvent` class. This property is an `IDictionary<string, string>` with a default initialization to an empty dictionary, allowing for the storage of string-based key-value metadata.


src/DurableTask.Core/History/TaskScheduledEvent.cs
```diff
@@ -84,5 +85,11 @@ public TaskScheduledEvent(int eventId)
         /// </summary>
         [DataMember]
         public DistributedTraceContext? ParentTraceContext { get; set; }
+        
+        /// <summary>
+        /// Gets or sets a dictionary of tags of string, string
+        /// </summary>
+        [DataMember]
+        public IDictionary<string, string> Tags { get; set; } = new Dictionary<string, string>();
```
cgillum, line 93: Make this nullable by default with no dictionary, so that we're not allocating this additional memory for each of these objects when we don't need it. Then it will also match the `Tags` property for `ExecutionStartedEvent`.
```suggestion
        public IDictionary<string, string>? Tags { get; set; }
```

cgillum (changes requested):
It's a bit odd that we create this history event but don't expose any way for DTFx users to set it. I'm okay with this if we want to expose it via other libraries, but let's wait to merge it until we have a PR (in this repo or another repo) that actually uses the property in some end-to-end way.

YunchuWang:
hi @cgillum , i made more changes, can you take a look when have chance? thanks

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -206,6 +206,22 @@ public virtual Task<T> ScheduleWithRetry<T>(string name, string version, RetryOp
...
+        /// </summary>
+        /// <typeparam name="T">Return Type of the TaskActivity.Execute method</typeparam>
+        /// <param name="name">Name of the orchestration as specified by the ObjectCreator</param>
+        /// <param name="version">Name of the orchestration as specified by the ObjectCreator</param>
+        /// <param name="retryOptions">Retry policy</param>
+        /// <param name="tags">Dictionary of key/value tags associated with this instance</param>
+        /// <param name="parameters">Parameters for the TaskActivity.Execute method</param>
+        /// <returns>Task that represents the execution of the specified TaskActivity</returns>
+        public virtual Task<T> ScheduleWithRetry<T>(string name, string version, RetryOptions retryOptions,
+            IDictionary<string, string> tags, params object[] parameters)
```
cgillum, line 221: I'm wondering if this could be a breaking change? It seems that if someone called `ScheduleWithRetry` previously and specified an `IDictionary<string, string>` for `parameters`, their updated code would now call this overload instead of the previous overload.

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -286,6 +302,19 @@ public virtual Task<TResult> ScheduleTask<TResult>(Type activityType, params obj
...
 
+        /// <summary>
+        ///     Schedule a TaskActivity by type, version, and tags.
+        /// </summary>
+        /// <typeparam name="TResult">Return Type of the TaskActivity.Execute method</typeparam>
+        /// <param name="name">Name of the orchestration as specified by the ObjectCreator</param>
+        /// <param name="version">Name of the orchestration as specified by the ObjectCreator</param>
+        /// <param name="parameters">Parameters for the TaskActivity.Execute method</param>
+        /// <param name="tags">Dictionary of key/value tags associated with this instance</param>
+        public virtual Task<TResult> ScheduleTask<TResult>(string name, string version, IDictionary<string, string> tags, params object[] parameters)
```
cgillum, line 313: Same concern about whether this is a breaking change for the overload below this one.

cgillum (commented):
I'm worried that these overloads might be accidental breaking changes. Maybe we should revert those changes and only update the history events, which is what we did in the first version of your PR.

src/DurableTask.Core/OrchestrationContext.cs
YunchuWang: good find, its ambiguous for compiler. i updated to use builder design pattern its backward compatible and future proof (hopefully)

src/DurableTask.Core/DurableTask.Core.csproj
```diff
@@ -18,7 +18,7 @@
   <PropertyGroup>
     <MajorVersion>3</MajorVersion>
     <MinorVersion>2</MinorVersion>
-    <PatchVersion>0</PatchVersion>
```
cgillum, line 21: If we're making public API changes, then we need to increment the minor version.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -86,10 +86,25 @@ public override async Task<TResult> ScheduleTask<TResult>(string name, string ve
             return result;
         }
 
+        public override async Task<TResult> ScheduleTask<TResult>(string name, string version,
```
cgillum, line 89: For a new API like this, I think we will need to add test coverage to ensure that they actually work.

test/DurableTask.AzureStorage.Tests/ScheduleTaskTests.cs
```diff
@@ -40,7 +41,7 @@ public async Task TaskReturnsVoid_OrchestratorFails()
                 TestInstance<int> instance = await host.StartInlineOrchestration(
                     input: 123,
                     orchestrationName: "TestOrchestration",
-                    implementation: (ctx, input) => ctx.ScheduleTask<int>("Activity", "", input),
+                    implementation: (ctx, input) => ctx.ScheduleTask<int>("Activity", "", new Dictionary<string, string>()),
```
cgillum, line 44: What is this change for?

YunchuWang: bummer not cleaned up, updated

src/DurableTask.Core/OrchestrationContext.cs
YunchuWang: updated

src/DurableTask.Core/TaskOrchestrationContext.cs
YunchuWang: tests added

cgillum (commented):
I haven't not read the new tests closely yet, but here is some feedback for now.

YunchuWang:
@cgillum hi Chris, updated all tests are passing, can you take another look?

cgillum (approved):

PR #1219, opened 2025-05-21
Fix parameter ordering bug in TaskHubClient.cs

ajchalla:
Fixed a bug in `CreateOrchestrationInstanceAsync` where the `startAt` parameter was incorrectly passed as the 8th argument to `InternalCreateOrchestrationInstanceWithRaisedEventAsync`, which expects `startAt` as the 9th parameter.

Updated the method call to correctly pass `startAt` as the 9th argument, preserving the correct order of optional parameters

### DIFF
--- a/src/DurableTask.Core/TaskHubClient.cs
+++ b/src/DurableTask.Core/TaskHubClient.cs
@@ -182,6 +182,7 @@ namespace DurableTask.Core
                 null,
                 null,
                 null,
+                null,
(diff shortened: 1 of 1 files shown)


cgillum (approved):
Thanks for this fix! Indeed, it looks like this parameter was not correctly assigned (one of the reasons I wish we were using named parameters for these overloads).

PR #1220, opened 2025-05-22
Fix history events buildup from version reject

halspang:
When an orchestration was being rejected due to a versioning mismatch an issue would cause the history events of the orchestration to build continuously. This was caused by the orchestration adding an OrchestrationStarted event before the versioning took place and was saved even on a rejected version.

This change causes the versioning code to exit the process method entirely instead of just prematurely end the process loop. This avoids the commit of the history.


AnatoliB:
Do you think we have sufficient test coverage for this? Wondering if it would be possible (and practical) to add a test that _fails_ without this change.

halspang:
@AnatoliB - Updated a test that looks at this scenario. I confirmed manually that it passes with this change and fails without it.

cgillum (approved):

PR #1230, opened 2025-06-26
Support Configurable Queue Client Encoding Strategy and Cross-Encoding Handling

nytian:
This PR adds support for configurable encoding strategies in Azure Storage Queue clients, enabling compatibility between UTF8 and Base64 encodings.

- Settings Update: Added num QueueClientEncodingStrategy to AzureStorageOrchestrationServiceSettings for user configuration.
- Client Configuration: Applied encoding strategy in AzureStorageClient, with a MessageDecodingFailed handler to manage cross-encoding scenarios, especially for base64 received utf8 message.
- Improved DeserializeQueueMessageAsync to support both encoding types with fallback logic.

### DIFF
--- a/Test/DurableTask.AzureStorage.Tests/QueueClientEncodingStrategyIntegrationTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/QueueClientEncodingStrategyIntegrationTests.cs
@@ -21,7 +21,7 @@ namespace DurableTask.AzureStorage.Tests
     using System.Threading.Tasks;
 
     [TestClass]
-    public class QueueClientEncodingStrategyIntegrationTests
+    public class QueueClientMessageEncodingIntegrationTests
     {
         private const string TestConnectionString = "UseDevelopmentStorage=true";
 
@@ -37,7 +37,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.Base64,
+                QueueClientMessageEncoding = QueueClientMessageEncoding.Base64,
             };
 
             var base64Service = new AzureStorageOrchestrationService(base64Settings);
@@ -55,7 +55,7 @@ namespace DurableTask.AzureStorage.Tests
                 {
                     TaskHubName = testName,
                     StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                    QueueClientEncodingStrategy = QueueClientEncodingStrategy.UTF8,
+                    QueueClientMessageEncoding = QueueClientMessageEncoding.UTF8,
                 };
 
                 var utf8Service = new AzureStorageOrchestrationService(utf8Settings);
@@ -99,7 +99,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.UTF8,
+                QueueClientMessageEncoding = QueueClientMessageEncoding.UTF8,
             };
 
             var utf8Service = new AzureStorageOrchestrationService(utf8Settings);
@@ -117,7 +117,7 @@ namespace DurableTask.AzureStorage.Tests
                 {
                     TaskHubName = testName,
                     StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                    QueueClientEncodingStrategy = QueueClientEncodingStrategy.Base64,
+                    QueueClientMessageEncoding = QueueClientMessageEncoding.Base64,
                 };
 
                 var base64Service = new AzureStorageOrchestrationService(base64Settings);
@@ -160,7 +160,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.Base64,
+                QueueClientMessageEncoding = QueueClientMessageEncoding.Base64,
             };
 
             var base64Service = new AzureStorageOrchestrationService(base64Settings);
@@ -206,7 +206,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.UTF8,
+                QueueClientMessageEncoding = QueueClientMessageEncoding.UTF8,
             };
 
             var utf8Service = new AzureStorageOrchestrationService(utf8Settings);
@@ -253,7 +253,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.Base64,
+                QueueClientMessageEncoding = QueueClientMessageEncoding.Base64,
             };
 
             var base64Service = new AzureStorageOrchestrationService(base64Settings);
--- a/Test/DurableTask.AzureStorage.Tests/QueueClientEncodingStrategyIntegrationTests.cs
+++ b/Test/DurableTask.AzureStorage.Tests/QueueClientEncodingStrategyIntegrationTests.cs
@@ -55,7 +55,7 @@ namespace DurableTask.AzureStorage.Tests
                 {
                     TaskHubName = testName,
                     StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                    QueueClientEncodingStrategy = QueueClientEncodingStrategy.None,
+                    QueueClientEncodingStrategy = QueueClientEncodingStrategy.UTF8,
                 };
 
                 var utf8Service = new AzureStorageOrchestrationService(utf8Settings);
@@ -99,7 +99,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.None,
+                QueueClientEncodingStrategy = QueueClientEncodingStrategy.UTF8,
             };
 
             var utf8Service = new AzureStorageOrchestrationService(utf8Settings);
@@ -206,7 +206,7 @@ namespace DurableTask.AzureStorage.Tests
             {
                 TaskHubName = testName,
                 StorageAccountClientProvider = new StorageAccountClientProvider(TestConnectionString),
-                QueueClientEncodingStrategy = QueueClientEncodingStrategy.None,
+                QueueClientEncodingStrategy = QueueClientEncodingStrategy.UTF8,
             };
 
             var utf8Service = new AzureStorageOrchestrationService(utf8Settings);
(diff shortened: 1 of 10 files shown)


Test/DurableTask.AzureStorage.Tests/QueueClientEncodingStrategyIntegrationTests.cs
```diff
@@ -0,0 +1,269 @@
...
+    public class QueueClientEncodingStrategyIntegrationTests
+    {
+        private const string TestConnectionString = "UseDevelopmentStorage=true";
+
+        [TestMethod]
+        // Verifies that messages sent with Base64 encoding can be processed by a worker with UTF8 encoding
+        public async Task CrossEncodingCompatibility_Base64ToUtf8()
+        {
+            string testName = "Base64ToUtf8";
+            string input = "世界! 🌍 Test with émojis and spéciål chàracters: ñáéíóúü";
```
cgillum, line 33: Would any of these tests have failed prior to this fix?

src/DurableTask.AzureStorage/QueueClientEncodingStrategy.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Specifies the encoding strategy used for Azure Storage Queue messages.
+    /// This enum maps to the Azure Storage SDK's <see cref="QueueMessageEncoding"/> values.
+    /// </summary>
+    public enum QueueClientEncodingStrategy
+    {
+        /// <summary>
+        /// Use UTF8 encoding for queue messages. Maps to <see cref="QueueMessageEncoding.None"/>.
+        /// </summary>
+        None = 0,
```
cgillum: Let's call this `UTF8` instead of `None`

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -288,5 +288,11 @@ internal LogHelper Logger
         /// Consumers that require separate dispatch (such as the new out-of-proc v2 SDKs) must set this to true.
         /// </summary>
         public bool UseSeparateQueueForEntityWorkItems { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the encoding strategy used for Azure Storage Queue messages.
+        /// The default is <see cref="QueueClientEncodingStrategy.None"/>.
+        /// </summary>
+        public QueueClientEncodingStrategy QueueClientEncodingStrategy { get; set; } = QueueClientEncodingStrategy.None;
```
cgillum: nit: rather than calling this a "strategy", I recommend we name it like `QueueClientMessageEncoding`.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -156,16 +156,71 @@ public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMe
...
+            MessageData envelope = null;
+
+            if (this.azureStorageClient.Settings.QueueClientEncodingStrategy == QueueClientEncodingStrategy.Base64)
             {
-                envelope = this.DeserializeMessageData(Encoding.UTF8.GetString(body));
+                try
+                {
+                    // For a queue client using the Base64 encoding strategy, error handling is configured via the options.
+                    // Therefore, we should always receive messages encoded in Base64 here.
+                    string messageText = queueMessage.Body.ToString();
```
cgillum: I'm wondering what the performance characteristics of this code are. What does the `ToString()` method do? I'm wondering if we can more efficiently deserialize directly from `queueMessage.Body` in a way that avoids expensive allocations for large messages. I'm not familiar with the Azure Storage SDK so I apologize if this is a dumb question.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -156,16 +156,71 @@ public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMe
...
+                        string base64String = Encoding.UTF8.GetString(body);
+                        byte[] decodedBytes = Convert.FromBase64String(base64String);
+                        string decodedMessage = Encoding.UTF8.GetString(decodedBytes);
+                        envelope = this.DeserializeMessageData(decodedMessage);
+                        
+                        // Log successful cross-encoding processing
+                        this.azureStorageClient.Settings.Logger.GeneralWarning(
+                            this.azureStorageClient.QueueAccountName,
+                            this.azureStorageClient.Settings.TaskHubName,
+                            $"Successfully processed Base64 message with UTF8 client. MessageId: {queueMessage.MessageId}");
```
cgillum: I recommend we move this warning up to the top of the catch block and say something like "Failed to deserialize queue message using UTF-8 encoding. First few characters are {...}. Will assume that it is a base64 encoded message that was created by an instance of the app with a different message encoding configuration.".

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -64,6 +65,63 @@ void ConfigureClientPolicies<TClientOptions>(TClientOptions options) where TClie
...
+                    QueueClientEncodingStrategy.Base64 => QueueMessageEncoding.Base64,
+                    _ => throw new ArgumentException($"Unsupported encoding strategy: {this.Settings.QueueClientEncodingStrategy}")
+                };
+
+                // Base64-encoded clients will fail to decode messages sent in UTF-8 format.
+                // This handler catches decoding failures and update the message with its the original content,
+                // so that the client can successfully process it on the next attempt.
+                if (this.Settings.QueueClientEncodingStrategy == QueueClientEncodingStrategy.Base64)
+                {
+                    options.MessageDecodingFailed += async (QueueMessageDecodingFailedEventArgs args) =>
```
cgillum, line 84: It seems this handler will try to handle the decoding error automatically. But it seems we also have exception handling logic in other places. Which exception handling logic will actually get triggered?

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
```diff
@@ -64,6 +65,63 @@ void ConfigureClientPolicies<TClientOptions>(TClientOptions options) where TClie
...
+                    {
```
cgillum, line 85: We should log a message whenever we enter into this decoding-failed handler.

src/DurableTask.AzureStorage/MessageManager.cs
nytian: tbh I am not sure if there it's any better ways and it seems not. :( https://learn.microsoft.com/en-us/azure/storage/queues/storage-tutorial-queues#dequeue-messages The official tutorial is using queueMessage.body.tostring(). but I will further confirm this part.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -156,16 +156,72 @@ public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMe
         {
             // TODO: Deserialize with Stream?
             byte[] body = queueMessage.Body.ToArray();
```
cgillum: I'm looking closely at all the references to `body` and `queueMessage.Body` in this method, and it seems like we're allocating memory to read the full message multiple times. I worry that this can cause significant memory pressure on the app and CPU pressure on the .NET garbage collector, especially for larger messages.

Looking more closely, it seems we *always* need to get the UTF-8 string contents of the message as a first step, whether we decode as base64 or not. Can we change the code to do `string bodyAsString = queueMessage.Body.ToString()` here, and then reuse `bodyAsString` in all the logic below? For example, replacing subsequent calls to `queueMessage.Body.ToString` (line 167), `Encoding.UTF8.GetString(body)` (line 186, 199, and 203)?

src/DurableTask.AzureStorage/MessageManager.cs
cgillum: OK, no problem. See my newer comment where I suggest some refactoring to reduce the total number of allocations involved in this code path.

src/DurableTask.AzureStorage/Storage/AzureStorageClient.cs
nytian: This one is only for the case when a base64 queue client receives a message sent by a utf8 queue client.  I added more comments to make it more clear.

Test/DurableTask.AzureStorage.Tests/QueueClientEncodingStrategyIntegrationTests.cs
nytian: We originally supported UTF-8 clients receiving either UTF-8 or Base64 messages. However, if a Base64 client receives a UTF-8 message, it should fail unless with the error handling implemented in AzureStorageClient.cs.

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -154,74 +154,61 @@ internal static bool TryGetLargeMessageReference(string messagePayload, out Uri
...
-            byte[] body = queueMessage.Body.ToArray();
+            // No matter what the queue client encoding is (Base64 or UTF8),
+            // we should always successfully get the string here.
+            string bodyAsString = queueMessage.Body.ToString();
             MessageData envelope = null;
 
-            if (this.azureStorageClient.Settings.QueueClientMessageEncoding == QueueClientMessageEncoding.Base64)
+            try
+            {
+                envelope = this.DeserializeMessageData(bodyAsString);
```
cgillum: It looks like we're always assuming UTF-8 encoding and relying on exception handling to tell us if it's base64. This is a bit of an anti-pattern since exceptions should only be used in exceptional cases, especially since they can result in a performance impact. Instead, I suggest we check to see if the first character is a `{`, in which case we deserialize it directly. If it's not `{`, then we assume base64.

Or am I misunderstanding the use case for this code?

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -161,38 +161,30 @@ public async Task<MessageData> DeserializeQueueMessageAsync(QueueMessage queueMe
...
+                    envelope = this.DeserializeMessageData(bodyAsString);
+                }
+                else
+                {
+                    // The message we got is not a valid json message (doesn't start with '{').
+                    // This could happen in the case where our queue client is UTF8 encoding 
+                    // while receiving a message sent by a Base64 queue client.
+                    // So we try to re-decode it as Base64.
+
                     this.azureStorageClient.Settings.Logger.GeneralWarning(
```
cgillum: Does it still make sense to emit a warning here, or should we remove this? If someone configures their default encoding to base64, we will always go down this code path, right?

nytian: No, if the queue client is base64 we should not go to this else block. Because queueMessage.body.AsString() should return a valid json message for base64 queue client. We will only get to this else block when we have a utf8 queue client but receive a base64 queue message.

cgillum: Oh, so the `else` block is a very special case that only happens during a migration from UTF-8 to base64? I didn't realize this. Maybe we should make that clearer in the comments. Either way, I think we need to change the logging. Since everything here is working as designed, it doesn't make sense to issue a warning. Also, there's no deserialization failure, so we should instead log an informational message that we received a non-JSON queue message and explains why this may have happened.

nytian: Yes this should only happen for that case, and I will update it as informational. 

We have four scenarios:

- A UTF-8 client receiving a UTF-8 message.  ( no issue, if block)
- A UTF-8 client receiving a Base64  message.  ( else block)
In this case,  queue client can decode message successfully but it will decode it with utf8 encoding. Thus queueMessage.Body.Asstring() will return base64 characters instead of a valid string. So we go to else block here to re-decode the message. 

- A Base64 client receiving a Base64 message. ( no issue, if block)
- A Base64 client receiving a UTF8 message. 
Base64 client cannot decode UTF-8 encoded message correctly and thus Base64 client will never receive any UTF-8 encoded message. Unless we manually update the message content to be Base64. In this case, handling is done in the failure event logic, which is what I added in AzureStorageClient.cs. 

The if section handles the following scenarios:
- A UTF-8 client receiving a UTF-8 message.
- A Base64 client receiving a Base64 message.

nytian: Hmmm I didn't find any appropriate informational level logging [here](https://github.com/Azure/durabletask/blob/main/src/DurableTask.AzureStorage/Logging/LogHelper.cs) we can use for this case ( utf8 client receives base64 message). Is it ok to remove the logging? Or what do you think about logging it as `ReceivedOutOfOrderMessage`?

We used to have a warning logging for this case is because we only enable utf8 encoding for queue client before for DTFx.AzureStorage v2, so receiving a base64 message will only happen when upgrading from DTFx.AzureStorage v1. But since right now we make the encoding configurable, this will happen more frequently, and maybe it's ok to remove the logging?

cgillum: I'm okay with removing the logging.

cgillum (approved):

PR #1232, opened 2025-07-14
Extended Sessions for Isolated (Orchestrations)

sophiatev:
This PR introduces enabling extended sessions for orchestrations in the .NET isolated framework. The main changes involve adding  a check for a result being injected from the middleware in the case of an extended session, as well as additional information to the `dispatchContext` so that the middleware can correctly execute the extended session. I also made the `IsCompleted` method of the `TaskOrchestrationExecutor` public so that it can be used by the dotnet SDK to remove an extended session from its cache upon completion of the orchestration. The majority of the changes for this feature are in the dotnet SDK, which is responsible for actually executing the orchestrations for .NET isolated.

Other PRs:
- [DotNet SDK](https://github.com/microsoft/durabletask-dotnet/pull/449)
- [WebJobs](https://github.com/Azure/azure-functions-durable-extension/pull/3154)
- [Protobuf](https://github.com/microsoft/durabletask-protobuf/pull/47)

### Open question
I found this existing TODO and agree that this section of code looks incorrect and dangerous. It basically checks if it is able to acquire the extended session lock (which means that `maxConcurrentSessions` has not yet been reached at that moment in time), and if it is able to, it sets `isExtendedSession` to true and immediately releases the lock. I think the more correct logic is that we are in an extended session whenever `OnProcessWorkItemAsync` is called from within the loop of the `OnProcessWorkItemSessionAsync` call. That is why I moved the `workItem.IsExtendedSession = true` statement there. If an extended session ends, then we break out of the loop, and will reload the `workItem` from storage upon the next orchestration execution, in which case `workItem.IsExtendedSession` will be false by default. 

I removed this section of code in this PR, but the question is - does the `CorrelationTraceClient.Propagate` call need to be replaced with something else now? What is its point? To log something upon the start of every extended session? 

```c#
                CorrelationTraceClient.Propagate(
                    () =>
                    {
                        // Check if it is extended session.
                        // TODO: Remove this code - it looks incorrect and dangerous
                        isExtendedSession = this.concurrentSessionLock.Acquire();
                        this.concurrentSessionLock.Release();
                        workItem.IsExtendedSession = isExtendedSession;
                    });
```

### Manual testing done thus far
- If extended sessions are not enabled, everything still proceeds as expected (the orchestration history is replayed upon every execution)
- If an extended session expires, a history is sent along with the orchestration request to the worker. The worker replays the orchestration and then saves the state in a new extended session (state is only saved if the orchestration is not completed)
- Upon completion of an orchestration, the extended session state is evicted from the cache
- If the worker ends the extended session before the host does (this should be very rare but can happen if e.g. there is a network delay when sending the orchestration request to the worker, or something along those lines), then it will throw a `SessionAbortedException`. The work item will then be retried again.
- Other OOProc scenarios still work as expected (they do not have extended sessions enabled so the orchestration is replayed upon every request. If they attempt to enable extended sessions an error occurs upon startup indicating that the extended session feature does not work for these other languages).

### Performance testing
Two scenarios were run in in-process with extended sessions enabled/disabled, and in isolated with extended sessions enabled/disabled. Multiple trials were run, and the time it took to complete the orchestration was recorded as well as the number of times the history was loaded in each of these settings (with `extendedSessionIdleTimeoutInSeconds` set to 30 seconds). The two scenarios tested were

- Fan out to a large number of tasks (30,000) and then fan-in. For both isolated and in-process, without extended sessions this took about 35 minutes and with extended sessions around 25 minutes. In the isolated model there were consistently two history loads with ES enabled, with the second corresponding to a `SessionAbortedException` being thrown (so the worker ends the extended session before the host, and as expected, throws the exception which leads to a retry of the work item, this time with a history attached). In in-process, there was just one history load. Without ES, there were about 100 history loads in each. 
- Run a large number of sequential tasks (1000) with large payloads (10,000 character-length strings). For both isolated and in-process, without extended sessions this took about 30 minutes and with extended sessions enabled around 5 minutes. Without ES there were around 100 history loads, and with ES just 1. Interestingly, in isolated this scenario actually ran somewhat faster both with/without ES as opposed to in-process.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1215,11 +1215,16 @@ namespace DurableTask.AzureStorage
                     this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, session.TraceActivityId, CancellationToken.None);
                 }
             }
-            catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.PreconditionFailed)
+            // Handle the case where the 'ETag' has changed, which implies another worker has taken over this work item while
+            // we were trying to process it. We detect this in 2 cases:
+            // Common case: the resulting code is 'PreconditionFailed', which means our ETag no longer matches the one stored.
+            // Edge case: the resulting code is 'Conflict'. This can occur if this was the first orchestration work item.
+            // The 'Conflict' represents that we attempted to insert a new orchestration history when one already exists.
+            catch (DurableTaskStorageException dtse) when (dtse.HttpStatusCode == (int)HttpStatusCode.Conflict || dtse.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
             {
                 // Precondition failure is expected to be handled internally and logged as a warning.
                 // The orchestration dispatcher will handle this exception by abandoning the work item
-                throw new SessionAbortedException("Aborting execution due to failed precondition.", rfe);
+                throw new SessionAbortedException("Aborting execution due to conflicting completion of the work item by another worker.", dtse);
             }
             catch (Exception e)
             {
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -1203,7 +1203,11 @@ namespace DurableTask.AzureStorage.Tracking
             }
             catch (DurableTaskStorageException ex)
             {
-                if (ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
+                // Handle the case where the the history has already been updated by another caller.
+                // Common case: the resulting code is 'PreconditionFailed', which means "eTagValue" no longer matches the one stored, and TableTransactionActionType is "Update".
+                // Edge case: the resulting code is 'Conflict'. This is the case when eTagValue is null, and the TableTransactionActionType is "Add",
+                // in which case the exception indicates that the table entity we are trying to "add" already exists.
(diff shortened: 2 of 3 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -242,6 +242,8 @@ async Task OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem)
                         // If the provider provided work items, execute them.
                         if (workItem.NewMessages?.Count > 0)
                         {
+                            // As long as we are in this execution loop, the extended session is maintained.
+                            workItem.IsExtendedSession = true;
```
cgillum: This assignment is a little confusing. Right above this, on line 236, we set a local variable named `IsExtendedSession` to `false`, yet we set `workItem.IsExtendedSession = true` here without changing the `isExtendedSession` local variable. Same inconsistency below on line 259 where we assign `isExtendedSession` but don't do anything with `workItem.IsExtendedSession`. These fields have the same name and semantic meaning but different values. I haven't yet reasoned about correctness, but is there any way we can clean this up to make the variable assignments more consistent (or remove a variable if it's redundant)?

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -786,12 +792,23 @@ async Task ResumeOrchestrationAsync(TaskOrchestrationWorkItem workItem)
             dispatchContext.SetProperty(cursor.TaskOrchestration);
             dispatchContext.SetProperty(cursor.RuntimeState);
             dispatchContext.SetProperty(workItem);
+            dispatchContext.SetProperty(IsExtendedSession, true);
+            dispatchContext.SetProperty(IncludePastEvents, false);
```
cgillum: This is okay, but it would be a bit easier to maintain if we just created a `WorkItemMetadata` class which included these two properties in a type-safe way and set the `WorkItemMetadata` as the one property. Then we could also add more stuff to it later if needed without making more changes here.

cgillum (approved):

PR #1235, opened 2025-08-20
Adding an Exception Case to CompleteTaskOrchestrationWorkItemAsync

sophiatev:
`AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync` is not correctly catching the exception thrown in the case of an etag mismatch when calling `ITrackingStore.UpdateStateAsync` method. The `InstanceStoreBackedTrackingStore` implementation throws a `RequestFailedException`, which is correctly caught, while the `AzureTableTrackingStore` implementation wraps this exception in a `DurableTaskStorageException`, so this was not being correctly caught. This PR adds this exception check to the method as well.

Note: It is difficult to add testing for this - the best I can think to do is add a unit test to confirm that `AzureTableTrackingStore.UpdateStateAsync` throws the expected `DurableTaskStorageException` with `StatusCode` equal to `HttpStatusCode.PreconditionFailed`, but this method expects a parameter of [type](https://github.com/Azure/durabletask/blob/931cbe0f3d8036b925339c5d3440dd376fd4a5c4/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L821)`TrackingStoreContext` which is a private class and can't be instantiated from within the tests. So there seems to be no easy way to test this (this parameter is passed around as a generic `object` in all other classes, and cast to its type from within this method).

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1215,11 +1215,17 @@ namespace DurableTask.AzureStorage
                     this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, session.TraceActivityId, CancellationToken.None);
                 }
             }
+            // If the tracking store is of type InstanceStoreBackedTrackingStore, then it will throw a RequestFailedException
             catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.PreconditionFailed)
             {
                 // Precondition failure is expected to be handled internally and logged as a warning.
                 // The orchestration dispatcher will handle this exception by abandoning the work item
                 throw new SessionAbortedException("Aborting execution due to failed precondition.", rfe);
+            }
+            // If the tracking store is of type AzureStorageTrackingStore, then it will throw a DurableTaskStorageException which decorates the RequestFailedException
+            catch (DurableTaskStorageException dtse) when (dtse.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
+            {
+                throw new SessionAbortedException("Aborting execution due to failed precondition.", dtse);
             }
             catch (Exception e)
             {


davidmrdavid:
non-blocking tip (I'm just an OSS contributor now) - this seems like an involved failure case. I recommend adding in the PR description a short example of how this error case is reached, in the form of a timeline.

For example:
At
*  t=1, such and such happens.
* t=2, then this happens
* t=3, this happens, and now we're in an inconsistent state because Y
* t=4, the inconsistency is observed, and the exception is thrown.

It might make it easier to determine that the PR is doing what's expected. Though others caught up with the context might not need this level of detail.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1215,11 +1215,12 @@ await this.CommitOutboundQueueMessages(
                     this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, session.TraceActivityId, CancellationToken.None);
                 }
             }
-            catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.PreconditionFailed)
+            // If this is the first work item for the orchestration, then session.Etag will be null and the status code will be "Conflict" because a new history will attempt to be inserted when one already exists.
+            // If this is not the first work item, then session.Etag will be non-null but will not match the etag stored when attempting to update the history due to a conflicting completion of the work item.
```
davidmrdavid: This second sentence is phrased a bit strange:
>             // If this is not the first work item, then session.Etag will be non-null but will not match the etag stored when attempting to update the history due to a conflicting completion of the work item.
>             // In that case, the status code will be "PreconditionFailed".


It suggests to me, if this is not the first work item, then we _necessarily_ end up with a "PreconditionFailed" state. I think you mean to replace the "_will_" verb with a "may be" instead? Not sure.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1215,11 +1215,12 @@ await this.CommitOutboundQueueMessages(
                     this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, session.TraceActivityId, CancellationToken.None);
                 }
             }
-            catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.PreconditionFailed)
+            // If this is the first work item for the orchestration, then session.Etag will be null and the status code will be "Conflict" because a new history will attempt to be inserted when one already exists.
```
davidmrdavid: This is interesting - how was this working before? Are you saing that, one any first work item, we were throwing a "conflict" error? I must be missing something.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sophiatev: Good point, this has been fixed

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sophiatev: A confusion of my badly-written comment - I was trying to say that if this the first work item for the orchestration, and it was _already completed_ by another worker, then the conflict error is thrown. Hopefully the new version of this comment is clearer

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1214,7 +1215,7 @@ static string GetBlobName(TableEntity entity, string property)
                         numberOfTotalEvents,
                         historyEventNamesBuffer.ToString(0, historyEventNamesBuffer.Length - 1), // remove trailing comma
                         stopwatch.ElapsedMilliseconds,
-                        eTagValue?.ToString());
+                        eTagValue is null ? string.Empty : eTagValue.ToString());
```
davidmrdavid, line 1221: ultra nit - is this change necessary? I think it's the same as the original: `eTagValue?.ToString());`

It probably doesn't matter much, but figured I'd check.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1203,7 +1203,8 @@ static string GetBlobName(TableEntity entity, string property)
             }
             catch (DurableTaskStorageException ex)
             {
-                if (ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
+                // The status code is "Conflict" for a null eTagValue and TableTransactionActionType.Add, and "PreconditionFailed" for a non-null eTagValue and TableTransactionActionType.UpdateMerge
+                if (ex.HttpStatusCode == (int)HttpStatusCode.Conflict || ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
```
davidmrdavid: ultra nit -

I think it's useful to structure htis in terms of what's the 'general case' versus the edge case.
In general, what you're detecting here is that the ETag has changed (PreconditionFailed).
And the edge case is that if the ETag changed _from `null`_ to a value, then the status is 'Conflict'.

So I recommend organizing this in the order of 'most expected' to 'edge case', to make it easier to parse.
Like this:

```suggestion
                // Handle case where the 'ETag' has changed. This implies another worker has taken over this
                // work item while we were trying to process it. We detect this in 2 cases:
                // Common case: the resulting code is 'PreconditionFailed', it means our ETag is no longer current.
                // Edge case: the resulting code is 'Conflict'. This can occur if this was the first orchestrator work item, 
                // as no ETag exists then. The 'Conflict' represents that we attempted to create a history that already
                // existed .
                if (ex.HttpStatusCode == (int)HttpStatusCode.Conflict || ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
```

davidmrdavid: @sophiatev - in the 'conflict' case - should we not also be checking if ETag is 'null'? Otherwise, we're assuming it is null. Not sure, you're the expert :)

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1215,11 +1215,15 @@ await this.CommitOutboundQueueMessages(
                     this.orchestrationSessionManager.AddMessageToPendingOrchestration(session.ControlQueue, messages, session.TraceActivityId, CancellationToken.None);
                 }
             }
-            catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.PreconditionFailed)
+            // If this is the first work item for the orchestration and it has already been completed, then session.Etag will be null and a new history will attempt to be inserted when one already exists.
```
davidmrdavid, line 1218: nits - feel free to disregard.
```suggestion
            // If this is the first work item for the orchestration but it was already completed, by another worker then our session.Etag will be null. This means we'll try to create a new history despite one already existing.

```

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1215,11 +1215,15 @@ await this.CommitOutboundQueueMessages(
...
+            // In this case, the status code will be "Conflict".
+            // If this is not the first work item and it has already been completed, then session.Etag will be non-null but will not match the etag stored when attempting to update the history.
+            // In this case, the status code will be "PreconditionFailed".
+            catch (DurableTaskStorageException dtse) when (dtse.HttpStatusCode == (int)HttpStatusCode.Conflict || dtse.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
             {
                 // Precondition failure is expected to be handled internally and logged as a warning.
                 // The orchestration dispatcher will handle this exception by abandoning the work item
-                throw new SessionAbortedException("Aborting execution due to failed precondition.", rfe);
+                throw new SessionAbortedException("Aborting execution due to conflicting completion of the work item.", dtse);
```
davidmrdavid: nit - if this is possibly customer facing, it's useful to add a little hint here that the error is benign. Consider:

```suggestion
                throw new SessionAbortedException("Aborting execution because the work item was already completed. This can occur when another worker took over processing, and is benign when it occurs rarely", dtse);

```

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sophiatev: Ah yeah you're right, I just tracked this through to the ultimate log message and it actually checks for the null case: `ETag ?? string.Empty`, so I probably don't need this. I initially added it just because I noticed that the logging methods expect a `string` as opposed to a `string?`, which makes me tempted to keep it just in case anything ever changes underneath, just to be extra safe.

davidmrdavid: That reasoning seems perfectly valid. Resolving!

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sophiatev: Good idea :) I think based on these [lines](https://github.com/Azure/durabletask/blob/931cbe0f3d8036b925339c5d3440dd376fd4a5c4/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L1195) above that ETag has to be null in this case. I _think_ the conflict error is only thrown with the "add" action

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sophiatev: I _think_ `SessionAbortedExceptions` do not surface, though someone else from the team correct me if I'm wrong, please.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
davidmrdavid: Agreed. In theory, those lines above may be refactored in the future and so the assumptions here can be accidentally invalided, but let's not overthink it. LGTM. Resolving

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -119,7 +128,7 @@ async Task<AzureStorageOrchestrationService> EnsureTaskHubAsync(
             // Control queues
             Assert.IsNotNull(service.AllControlQueues, "Control queue collection was not initialized.");
             ControlQueue[] controlQueues = service.AllControlQueues.ToArray();
-            Assert.AreEqual(4, controlQueues.Length, "Expected to see the default four control queues created.");
+            Assert.AreEqual(partitionCount, controlQueues.Length, "Expected to see the default four control queues created.");
```
davidmrdavid: given that we've generalized the assertion to check for a variable number of partitions, I think the error string ("Expected to see the default four ...") should also be generalized.

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -483,6 +492,138 @@ await TestHelpers.WaitFor(
...
+            // Have worker 2 dequeue the same work item and start "working" on it
+            var workItem2 = await service2.LockNextTaskOrchestrationWorkItemAsync(
+                TimeSpan.FromMinutes(5),
+                CancellationToken.None);
+            workItem2.OrchestrationRuntimeState = runtimeState;
+
+            // Worker 2 completes the work item
+            await service2.CompleteTaskOrchestrationWorkItemAsync(workItem2, runtimeState, new List<TaskMessage>(), new List<TaskMessage>(), new List<TaskMessage>(), null, null);
+            // Now worker 1 will attempt to complete the same work item. Since this is the first attempt to complete a work item and add a history for the orchestration (by worker 1),
+            // there is no etag stored for the OrchestrationSession, and so the exception that will be thrown is a conflict since worker 2 already created a history for the orchestration in its completion of the work item.
```
davidmrdavid: nit on phrasing:

``` suggestion
            // there is no etag stored for the OrchestrationSession, and so the a "conflict" exception will be thrown as worker 2 already created a history for the orchestration.
```

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -483,6 +492,138 @@ await TestHelpers.WaitFor(
...
+            // Worker 1 also acquires the work item and starts "working" on it
+            workItem1 = await service1.LockNextTaskOrchestrationWorkItemAsync(
+                TimeSpan.FromMinutes(5),
+                CancellationToken.None);
+            workItem1.OrchestrationRuntimeState = runtimeState;
+
+            // Worker 1 completes the work item
+            await service1.CompleteTaskOrchestrationWorkItemAsync(workItem1, runtimeState, new List<TaskMessage>(), new List<TaskMessage>(), new List<TaskMessage>(), null, null);
+            // Now worker 2 attempts to complete the same work item. Since this is not the first work item for the orchestration, now an etag exists for the OrchestrationSession, and the exception
+            // that is thrown will be a precondition failed (since the etag doesn't match after worker 1 already completed the work item and updated the orchestration history).
```
davidmrdavid: nit on phrasing

```suggestion
            // that is thrown will be "precondition failed" as the Etag is stale after worker 1 completed the work item.
```

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -483,6 +492,138 @@ await TestHelpers.WaitFor(
             Assert.IsTrue(queueMessages.All(msg => msg.DequeueCount == 1));
         }
 
+        [TestMethod]
+        public async Task MultipleWorkersAttemptingToCompleteSameWorkItem()
```
davidmrdavid, line 506: ultra nit - there are 2 cases being tested in this method, one for each method being handled 'precondition failed' and 'conflict'.

As a general tip - I recommend writing a small blurb at the beginning of the test when _multiple things_ are being tested. It helps the reader predict where the test is going as they're reading it. You could just describe what are the 2 high level behaviors you're trying to assert.

davidmrdavid (approved):
Other than my comment changes, and the "partitionCount" log that should be generalized, the behavior here looks right to me.

Great job with the test, I personally don't think I ever had to write a test with such low level control over the orchestration processing state machine, so it's cool to see that in action.

As always - I'm just an OSS contributor, **so please wait for someone actively in the team to approve** before moving forward. Just wanted to signal that behavior-wise this _seems_ correct, and that test seems complete.

**Update:** I cleared my approval just to avoid accidentally confusing the team into thinking the PR is already approved by a team member.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: LGTM, resolving.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: ty, resolving.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
davidmrdavid: I won't insist on this one, it's my personal style to augment these nonetheless but it really doesn't matter. Feel free to dismiss.

PR #1237, opened 2025-09-02
Include Exception Types at FailureDetails

nytian:
This PR introduces support for including custom exception properties in `FailureDetails`.

Customers can register the interface `IExceptionPropertiesProvider` at TaskhubWorker that specifies which exception types and properties should be included.

`IExceptionPropertiesProvider` is a property on `TaskHubWorker`. We pass the provider to each dispatcher and then to the `TaskContext`. As a result, when orchestration or activity tasks are executed, use the provided implementation at TaskContext to generate `FailureDetails` to include the specified exception properties.


src/DurableTask.Core/FailureDetails.cs
```diff
@@ -38,14 +38,16 @@ public class FailureDetails : IEquatable<FailureDetails>
         /// <param name="stackTrace">The exception stack trace.</param>
         /// <param name="innerFailure">The inner cause of the failure.</param>
         /// <param name="isNonRetriable">Whether the failure is non-retriable.</param>
+        /// <param name="properties">Additional properties associated with the failure.</param>
         [JsonConstructor]
-        public FailureDetails(string errorType, string errorMessage, string? stackTrace, FailureDetails? innerFailure, bool isNonRetriable)
+        public FailureDetails(string errorType, string errorMessage, string? stackTrace, FailureDetails? innerFailure, bool isNonRetriable, IDictionary<string, object>? properties = null)
```
cgillum: Since this is a public API, I don't think we can change the parameter list. Rather, we'll need to create a new constructor overload.

src/DurableTask.Core/ExceptionPropertiesProviderRegistry.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// This registry is intended for use by the durabletask-dotnet layer to set a provider
+    /// that will be used when creating FailureDetails from exceptions.
+    /// </summary>
+    public static class ExceptionPropertiesProviderRegistry
+    {
+        /// <summary>
+        /// Gets or sets the global exception properties provider.
+        /// This will be set by the durabletask-dotnet layer.
+        /// </summary>
+        public static IExceptionPropertiesProvider? Provider { get; set; }
```
cgillum: public static settable properties like this are an anti-pattern. This should instead be a property of the `TaskHubWorker`.

src/DurableTask.Core/ExceptionPropertiesProviderRegistry.cs
```diff
@@ -0,0 +1,29 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//  ----------------------------------------------------------------------------------
+#nullable enable
+namespace DurableTask.Core
+{
+    /// <summary>
+    /// Registry for the global exception properties provider.
+    /// This registry is intended for use by the durabletask-dotnet layer to set a provider
```
cgillum: It's generally better to **not** describe this as a feature of another library, but rather as a standalone feature that can be used in the current library.

If you want to add a note about any systems that depend on this, then do it as a regular code comment and not as a documentation comment.

nytian: deleted this file

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -54,7 +54,7 @@ public FailureDetails(string errorType, string errorMessage, string? stackTrace,
         /// <param name="e">The exception used to generate the failure details.</param>
         /// <param name="innerFailure">The inner cause of the failure.</param>
         public FailureDetails(Exception e, FailureDetails innerFailure)
-            : this(e.GetType().FullName, GetErrorMessage(e), e.StackTrace, innerFailure, false)
+            : this(e, innerFailure, null)
```
nytian: I added this ctor because I think I might need to use it for dotnet sdk..

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -63,8 +63,31 @@ public FailureDetails(Exception e, FailureDetails innerFailure)
...
+            : this(e, (IExceptionPropertiesProvider?)null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FailureDetails"/> class from an exception object.
+        /// </summary>
+        /// <param name="e">The exception used to generate the failure details.</param>
+        /// <param name="exceptionPropertiesProvider">The provider to extract custom properties from the exception.</param>
+        public FailureDetails(Exception e, IExceptionPropertiesProvider? exceptionPropertiesProvider)
```
cgillum: The approach of creating a new constructor that takes an `IExceptionPropertiesProvider` doesn't seem correct to me. I think we should instead add a new constructor that takes in a `IDictionary<string, object>?` object, and then make that the new `[JsonConstructor]` constructor so that these properties can be correctly deserialized from JSON. The existing `[JsonConstructor]` parameter should then call into this new one, passing in `null`, and its `[JsonConstructor]` attribute should be removed.

Also, I don't think we should expose `IExceptionPropertiesProvider` in this class at all. Just have callers pass in `Properties` dictionary. How they get the properties can be an implementation detail of the caller. This makes the types more loosely coupled, which is a best practice.

src/DurableTask.Core/FailureDetails.cs
```diff
@@ -204,7 +243,32 @@ static string GetErrorMessage(Exception e)
...
-            return e == null ? null : new FailureDetails(e);
+            return FromException(e, null);
+        }
+
+        static FailureDetails? FromException(Exception? e, IExceptionPropertiesProvider? provider)
+        {
+            return e == null ? null : new FailureDetails(e, provider);
+        }
+
+        static IDictionary<string, object>? GetExceptionProperties(Exception exception, IExceptionPropertiesProvider? provider)
```
cgillum: I suggest making this into an internal static method that's defined on the `IExceptionPropertiesProvider` interface so that it can be called by the callers of `FailureDetails`.

src/DurableTask.Core/ReflectionBasedTaskActivity.cs
```diff
@@ -140,7 +140,7 @@ public override async Task<string> RunAsync(TaskContext context, string input)
                 }
                 else
                 {
-                    failureDetails = new FailureDetails(exception);
+                    failureDetails = new FailureDetails(exception, context.ExceptionPropertiesProvider);
```
cgillum: Similar feedback that I mentioned in FailureDetails.cs: Just use the provider here and pass in the properties as the constructor parameter.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -36,19 +36,41 @@ public sealed class TaskActivityDispatcher
...
+        /// <param name="logHelper">The log helper</param>
+        /// <param name="errorPropagationMode">The error propagation mode</param>
+        /// <param name="exceptionPropertiesProvider">The exception properties provider for extracting custom properties from exceptions</param>
+        internal TaskActivityDispatcher(
+            IOrchestrationService orchestrationService,
+            INameVersionObjectManager<TaskActivity> objectManager,
+            DispatchMiddlewarePipeline dispatchPipeline,
+            LogHelper logHelper,
+            ErrorPropagationMode errorPropagationMode,
+            IExceptionPropertiesProvider? exceptionPropertiesProvider)
```
cgillum, line 56: These constructors are internal, so you don't need to add a new constructor. Just update the existing constructor.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -42,19 +42,41 @@ public class TaskEntityDispatcher
...
+        /// <param name="logHelper">The log helper</param>
+        /// <param name="errorPropagationMode">The error propagation mode</param>
+        /// <param name="exceptionPropertiesProvider">The exception properties provider for extracting custom properties from exceptions</param>
+        internal TaskEntityDispatcher(
+            IOrchestrationService orchestrationService,
+            INameVersionObjectManager<TaskEntity> entityObjectManager,
+            DispatchMiddlewarePipeline entityDispatchPipeline,
+            LogHelper logHelper,
+            ErrorPropagationMode errorPropagationMode,
+            IExceptionPropertiesProvider exceptionPropertiesProvider)
```
cgillum, line 62: Same feedback - no need to create a new constructor.

src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -52,6 +53,16 @@ public TaskOrchestrationContext(
...
+            : this(orchestrationInstance, taskScheduler, entityParameters, errorPropagationMode, null)
+        {
+        }
+
+        public TaskOrchestrationContext(
+            OrchestrationInstance orchestrationInstance,
+            TaskScheduler taskScheduler,
+            TaskOrchestrationEntityParameters entityParameters,
+            ErrorPropagationMode errorPropagationMode,
+            IExceptionPropertiesProvider exceptionPropertiesProvider)
```
cgillum: Same feedback - this type is internal, so even if the constructors are public, it can't be called by user code so making changes to the existing constructor is non-breaking.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -57,6 +58,28 @@ internal TaskOrchestrationDispatcher(
...
+        /// <param name="versioningSettings">The versioning settings</param>
+        /// <param name="exceptionPropertiesProvider">The exception properties provider for extracting custom properties from exceptions</param>
+        internal TaskOrchestrationDispatcher(
+            IOrchestrationService orchestrationService,
+            INameVersionObjectManager<TaskOrchestration> objectManager,
+            DispatchMiddlewarePipeline dispatchPipeline,
+            LogHelper logHelper,
+            ErrorPropagationMode errorPropagationMode,
+            VersioningSettings versioningSettings,
+            IExceptionPropertiesProvider? exceptionPropertiesProvider)
```
cgillum, line 71: Same feedback - just update the existing constructor.

src/DurableTask.Core/FailureDetails.cs
nytian: updated as a extension method

cgillum (approved):

cgillum (approved):

PR #1245, opened 2025-09-22
Add Safe FileVersion for DurableTask.ServiceBus

nytian:
This pull request makes a change to the way the `FileVersion` is set in the `DurableTask.ServiceBus.csproj` project . The update ensures that the `FileVersionRevision` value always stays within the valid range (0-65535) by applying a modulo operation. This prevents potential build versioning issues if the revision number exceeds the allowed limit.

### DIFF
--- a/src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
+++ b/src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
@@ -16,9 +16,8 @@
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <FileVersion>$(VersionPrefix).0</FileVersion>
     <!-- FileVersionRevision is expected to be set by the CI. This is useful for distinguishing between multiple builds of the same version. -->
-    <!-- Ensure FileVersionRevision fits within valid range (0-65535) by using modulo operation -->
-    <FileVersionRevision Condition="'$(FileVersionRevision)' != ''">$([MSBuild]::Modulo($(FileVersionRevision), 65536))</FileVersionRevision>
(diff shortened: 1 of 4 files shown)


src/DurableTask.ServiceBus/DurableTask.ServiceBus.csproj
```diff
@@ -15,7 +16,9 @@
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <FileVersion>$(VersionPrefix).0</FileVersion>
     <!-- FileVersionRevision is expected to be set by the CI. This is useful for distinguishing between multiple builds of the same version. -->
-    <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevision)</FileVersion>
+    <!-- Ensure FileVersionRevision fits within valid range (0-65535) by using modulo operation -->
+    <FileVersionRevisionSafe Condition="'$(FileVersionRevision)' != ''">$([MSBuild]::Modulo($(FileVersionRevision), 65536))</FileVersionRevisionSafe>
+    <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevisionSafe)</FileVersion>
```
cgillum: A couple things:
1. Rather than defining a new property, can we just reuse `FileVersionRevision`?
```xml
<FileVersionRevision Condition="'$(FileVersionRevision)' != ''">$([MSBuild]::Modulo($(FileVersionRevision), 65536))</FileVersionRevision>
```
2. Do we need a similar fix for all our other nuget packages in this repo, which also use the same mechanism for getting a file version revision?

cgillum (approved):

PR #1253, opened 2025-10-21
Fixing Azure Storage History/Instance Table Inconsistency

### LINKED ISSUE #1252: Completed orchestrations sometimes get permanently stuck in the Running status
### Description
This issue is specific to the Azure Storage backend.

If an orchestration completes, but the process hosting it recycles after the History table is updated but before the Instances table is updated, the orchestration will get permanently stuck in the "Running" state.

This issue can be detected by checking for a "Discarding work item" warning in the `DurableTask-AzureStorage` logs with a reason of "Instance is Completed" (or Terminated, or Failed), even though the orchestration reports as being in the "Running" state.

This problem happens rarely, but we have seen recent cases of it, including in a recent support case (ICM 676943847). It's also been observed in Azure/azure-functions-durable-extension#2364 and Azure/azure-functions-durable-extension#3079. There currently isn't a great workaround beyond manually manipulating the data in Azure Storage.

The impact of this issue is that app workflows that depend on status changes for orchestrations will also get stuck waiting for the orchestration to transition to the Completed (or Terminated/Failed) state.

### Expected behavior
Orchestrations should be able to run to completion in the face of any transient failures and have that completion be reflected in status queries.

### Actual behavior
This is one case where the orchestration gets stuck in an inconsistent state.

### Known workarounds
Manually update the status of the orchestration in the `<TaskHubName>Instances` table to the correct status ("Completed", "Failed", or "Terminated"). This will cause status queries to return the proper status.

sophiatev:
If a process ends right after the history table is updated but before the instance table is updated in a call to `AzureStorageOrchestrationService.CompleteTaskOrchestrationWorkItemAsync` for a terminal orchestration, then storage is left in an inconsistent state. The instance table shows the orchestration as running while the history table shows the orchestration completed with an `ExecutionCompletedEvent` at the end. For non-terminal orchestrations, this is not a problem because the call to complete the next work item will reconcile these two tables. However, in the current implementation, the call to `AzureStorageOrchestrationService.LockNextTaskOrchestrationWorkItemAsync` will recognize that the orchestration is in a terminal state based on its history and simply [discard](https://github.com/Azure/durabletask/blob/29321fbbaef3998fcaefd38e576b08b526d4d6f2/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L1077) the work item, meaning the instance table is never updated. To fix this, we add a check in the case that a work item is received for an orchestration in a terminal state to confirm that the instance table has been correctly updated. If not, it is updated, and any orphaned blobs deleted. 
This PR also adds an integration test for this scenario.

Fixes #1252

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -796,13 +796,18 @@ namespace DurableTask.AzureStorage.Tracking
         }
 
         /// <inheritdoc />
-        public override async Task UpdateStatusForTerminationAsync(string instanceId, string output, CancellationToken cancellationToken = default)
+        public override async Task UpdateStatusForTerminationAsync(
+            string instanceId,
+            string output,
+            DateTime lastUpdatedTime,
+            CancellationToken cancellationToken = default)
         {
             string sanitizedInstanceId = KeySanitation.EscapePartitionKey(instanceId);
             TableEntity entity = new TableEntity(sanitizedInstanceId, "")
             {
                 ["RuntimeStatus"] = OrchestrationStatus.Terminated.ToString("G"),
-                ["LastUpdatedTime"] = DateTime.UtcNow,
+                ["LastUpdatedTime"] = lastUpdatedTime,
+                ["CompletedTime"] = DateTime.UtcNow,
                 [OutputProperty] = output
             };
 
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -1075,7 +1075,11 @@ namespace DurableTask.AzureStorage
                     TaskMessage executionTerminatedEventMessage = newMessages.LastOrDefault(msg => msg.Event is ExecutionTerminatedEvent);
                     if (executionTerminatedEventMessage is not null)
                     {
-                        await this.trackingStore.UpdateStatusForTerminationAsync(instanceId, ((ExecutionTerminatedEvent)executionTerminatedEventMessage.Event).Input);
+                        var executionTerminatedEvent = (ExecutionTerminatedEvent)executionTerminatedEventMessage.Event;
+                        await this.trackingStore.UpdateStatusForTerminationAsync(
+                            instanceId,
+                            executionTerminatedEvent.Input,
+                            executionTerminatedEvent.Timestamp);
                         return $"Instance is {OrchestrationStatus.Terminated}";
                     }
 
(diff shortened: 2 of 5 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1043,6 +1043,49 @@ public override Task StartAsync(CancellationToken cancellationToken = default)
...
+            string sanitizedInstanceId = KeySanitation.EscapePartitionKey(instanceId);
+            var instanceEntity = new TableEntity(sanitizedInstanceId, string.Empty)
+            {
+                // TODO: Translating null to "null" is a temporary workaround. We should prioritize 
+                // https://github.com/Azure/durabletask/issues/477 so that this is no longer necessary.
+                ["CustomStatus"] = runtimeState.Status ?? "null",
+                ["ExecutionId"] = executionId,
+                ["LastUpdatedTime"] = runtimeState.Events.Last().Timestamp,
+                ["RuntimeStatus"] = runtimeState.OrchestrationStatus.ToString(),
+                ["CompletedTime"] = runtimeState.Events.Last().Timestamp // do we want to do this as a rough proxy or DateTime.UtcNow?
```
cgillum: How about `runtimeState.CompletedTime`?

One slightly quirky thing about this method is that there's an implicit assumption that the orchestration is complete, but technically somebody could call this for a non-completed orchestration. It might be worth either a) adding some comments indicating that this should only be called for completed orchestrations (maybe even having an explicit check for this) or b) making the implementation defensive so that it works regardless of the orchestration state. In that case, we'd have to check to confirm that the orchestration is complete before attempting to set the "CompletedTime" field.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1043,6 +1043,49 @@ public override Task StartAsync(CancellationToken cancellationToken = default)
...
+            };
+
+            Stopwatch orchestrationInstanceUpdateStopwatch = Stopwatch.StartNew();
+            await this.InstancesTable.InsertOrMergeEntityAsync(instanceEntity);
```
cgillum, line 1090: Should we wait to do this last? Otherwise if this succeeds but there is a crash right after, we won't retry the deletion of the orphaned blobs, which currently happens after the Instances table is updated.

cgillum (commented):
I like the test you added! A couple comments:

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
sophiatev: Good point, I went with option a) (comments + method name change + method does nothing if the orchestration is not in a terminal state)

cgillum (approved):

PR #1260, opened 2025-10-28
Adding a warning for a continue-as-new call on a failed orchestration

sophiatev:
This PR adds a warning when the `TaskOrchestrationContext` detects that continue-as-new has been called on a failed orchestration (in which case the continue-as-new is ignored and the orchestration is completed in a "failed" state). It does this by introducing a new generic `OrchestrationCompletedWithWarning` log and a new `Tags` field to the `OrchestrationCompleteOrchestratorAction` which can be leveraged generically in the future for other such cases.

Solves #1261

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationContext.cs
+++ b/src/DurableTask.Core/TaskOrchestrationContext.cs
@@ -709,15 +709,6 @@ namespace DurableTask.Core
             }
             else
             {
-                if (this.continueAsNew != null && orchestrationStatus == OrchestrationStatus.Failed)
-                {
-                    TraceHelper.TraceSession(
-                        TraceEventType.Warning,
-                        "TaskOrchestrationContext-ContinueAsNewForFailedOrchestration",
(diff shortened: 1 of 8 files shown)


src/DurableTask.Core/TaskOrchestrationContext.cs
```diff
@@ -709,6 +709,15 @@ public void CompleteOrchestration(string result, string details, OrchestrationSt
             }
             else
             {
+                if (this.continueAsNew != null && orchestrationStatus == OrchestrationStatus.Failed)
+                {
+                    TraceHelper.TraceSession(
```
cgillum: This is the old, legacy mechanism for tracing. The updated/correct way to add new logs is following the pattern in LogEvents.cs.

sophiatev: Thanks, this has been updated to the new model. Though one thing I'm not quite sure about is if changes need to be made elsewhere since I introduced a new `EventId` now

PR #1265, opened 2025-11-17
Extended sessions for entities in .NET isolated

sophiatev:
This PR introduces the extended sessions feature for entities in .NET isolated. Most of the framework was already in place, just a few tweaks needed to be made to the `TaskEntityDispatcher` to ensure that metadata related to the extended session is retained by the host process while the session is active (and also sent to the worker process via the `WorkItemMetadata` object). The other changes are simply to make the `WorkItemMetadata` fields more generic since this object is now being used for both orchestration and entity execution.

Note: I had to change the `TaskEntityDispatcher.OnProcessWorkItemAsync` method from `protected` to `private` since it now accepts a `SchedulerState` parameter, which is "less visible than the accessibility of the method". Alternatively, I could have changed the visibility of the `SchedulerState` class. Is this change acceptable?

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -1326,6 +1326,12 @@ namespace DurableTask.Core
             out List<TaskMessage> subOrchestrationRewindMessages,
             out OrchestrationRuntimeState newRuntimeState)
         {
+
+            /* WARNING!!!:
+             * If any changes are made to how this method modifies the orchestration's history, then corresponding changes *must* 
+             * be made in the backend implementations that rely on this method for executing a rewind.
+             */
+
             HashSet<int> failedTaskIds = new();
             subOrchestrationRewindMessages = new();
 


src/DurableTask.Core/WorkItemMetadata.cs
```diff
@@ -11,9 +11,10 @@ public class WorkItemMetadata
...
 
         /// <summary>
-        /// Gets or sets whether or not to include past events in the orchestration history when executing the work item via middleware.
-        /// This assumes that the middleware is able to handle extended sessions and does not require history for replays.
+        /// Gets or sets whether or not to include instance state when executing the work item via middleware.
+        /// This assumes that the middleware is able to handle extended sessions and does not require the instance
+        /// state to fulfill the request.
         /// </summary>
-        public bool IncludePastEvents { get; set; }
+        public bool IncludeState { get; set; }
```
cgillum: This is a breaking API change. Have we already shipped a version of DT.Core that uses the old property?

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -131,37 +131,47 @@ async Task OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem)
...
+
+                            // The entity has been deleted, so we end the extended session.
+                            if (this.EntityIsDeleted(schedulerState))
                             {
                                 break;
                             }
 
+                            // We do this to avoid keeping the entity state in memory between executions within the extended session.
+                            // The middleware will keep the entity state cached, so it does not need it to fulfill requests within the extended session.
+                            schedulerState.EntityState = null;
```
cgillum, line 167: If the entity state is still kept in memory by the middleware, do we actually gain anything by setting this property to `null`?

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -208,7 +218,9 @@ internal class WorkItemEffects
         /// Method to process a new work item
         /// </summary>
         /// <param name="workItem">The work item to process</param>
-        protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem)
+        /// <param name="schedulerState">If extended sessions are enabled, the scheduler state that is being cached across executions.
+        /// If they are not enabled, or if this is the first execution from within an extended session, this parameter is null.</param>
+        private async Task<SchedulerState> OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem, SchedulerState schedulerState)
```
cgillum, line 224: Ideally, we'd have nullable reference types enabled on this file and mark the new property as `SchedulerState?` for safety. I'm not sure how big of a change that would be, though.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -245,19 +257,20 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                 }
                 else
                 {
+                    bool firstExecution = schedulerState == null;
```
cgillum: According to the comments, `schedulerState` is null if it's the first execution *or if extended sessions are not enabled*, so this variable name seems to be misleading.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -474,11 +488,13 @@ void DetermineWork(OrchestrationRuntimeState runtimeState, out SchedulerState sc
...
 
-
-                        if (runtimeState.Input != null)
+                        // Only attempt to deserialize the scheduler state if we don't already have it in memory (which can be true 
+                        // for extended sessions)
+                        if (runtimeState.Input != null && deserializeState)
                         {
                             try
                             {
+                                schedulerState = new SchedulerState();
```
cgillum: Is this needed? It looks like you already initialized the scheduler state on line 480. Or are you intentionally overwriting the value here? If so, should we instead leave it null above and only initialize `schedulerState` when required?

src/DurableTask.Core/TaskEntityDispatcher.cs
sophiatev: Yeah, in the sense that the entity state isn't kept in memory redundantly in multiple places (for example for isolated, we don't want it kept in-memory in both the worker process _and_ the host process)

src/DurableTask.Core/TaskEntityDispatcher.cs
sophiatev: Yeah, I thought about it, but when I enabled it it looked like quite a big change so I decided to just leave it alone.

src/DurableTask.Core/TaskEntityDispatcher.cs
sophiatev: I see what you mean, but this variable is only used in the `ExecuteViaMiddlewareAsync` method, where it translates to the boolean `includeEntityState`. This bool only has meaning if extended sessions are enabled, so in that context, we only want to include entity state if this is the first execution within an extended session. The bool is ignored entirely by the middleware if extended sessions aren't enabled. So all that to say, I think the variable name ultimately makes sense.

src/DurableTask.Core/TaskEntityDispatcher.cs
sophiatev: Oh, you're right! I should get rid of the initialization on line 480. It's redundant and unnecessary.

src/DurableTask.Core/WorkItemMetadata.cs
sophiatev: Discussed offline and decided this is ultimately okay since this class should only be used by apps running in isolated, in which case there is no explicit reference to the DT.Core package anyway.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -460,10 +473,11 @@ string SerializeSchedulerStateForNextExecution(SchedulerState schedulerState)
 
         #region Preprocess to determine work
 
-        void DetermineWork(OrchestrationRuntimeState runtimeState, out SchedulerState schedulerState, out Work batch)
+        SchedulerState DetermineWork(OrchestrationRuntimeState runtimeState, SchedulerState schedulerState, out Work batch)
```
cgillum: nit: instead of returning `SchedulerState`, I suggest making it a `ref` argument (I see it was previously an `out` argument). This makes the intent to mutate an existing object more clear. Otherwise, the method signature gives the impression that you're returning an entirely new object which, based on my reading of the code, you are not.

src/DurableTask.Core/TaskEntityDispatcher.cs
cgillum: But your comment mentions that you have this state also cached in middleware. What middleware are you referring to? We have middleware that runs here in the host and also middleware that runs in the language worker, so this is a bit ambiguous to me.

Assuming you're talking about the middleware in the external worker process, is the intent to allow this state to be garbage collector a bit sooner since we only need the state in the first loop iteration but not subsequent iterations?

sophiatev: Oh sorry, I was being intentionally vague to avoid leaking implementation details in this repo. I mean specifically the language worker middleware, yes. Am I allowed to add that specificity here? 

And yes, the intent is to allow GC to run on subsequent loop iterations seeing as we only need the state in the first iteration. That being said, it will be "regenerated" every time we commit the state to storage [here](https://github.com/Azure/durabletask/blob/f09ccd60e3594b30d1cde7e40c0d0516f278e74a/src/DurableTask.Core/TaskEntityDispatcher.cs#L344), based on the [result](https://github.com/Azure/durabletask/blob/f09ccd60e3594b30d1cde7e40c0d0516f278e74a/src/DurableTask.Core/TaskEntityDispatcher.cs#L303) of the entity batch operation. But the intent is to allow it to be garbage collected after this point while we wait for the next batch of messages in the extended session.

cgillum: I think what you can say instead is that we only send the entity state just once and expect the handler (which may be in another process) to do their own caching as appropriate. We then set the local reference here to null so that the garbage collector can reclaim it as soon as it detects it's not being used rather than always waiting for the session loop to complete (which could theoretically take some time).

I think we have to mention that the entity might be running externally, otherwise setting this to null doesn't really make sense (which could lead someone to misunderstand the intent).

cgillum (approved):

PR #1272, opened 2025-12-04
Adding Missing Fields to Instance Entity and Fix Blob References

sophiatev:
Unfortunately I missed some things in this earlier [PR](https://github.com/Azure/durabletask/pull/1271). In particular, many fields of the instance entity were not being set in the case that no instance entity existed in the first place (and, in fact, I wasn't setting the output at all in any situation). 

I also noticed I had other bugs. I actually should not be deleting the blob references in the call to synchronize the instance table with the history table. I misunderstood - the blobs should only be deleted in the case of a continue-as-new call when the history is overridden, which is never the case for an orchestration in a terminal state. I also updated the "input" and "output" fields of the instance entity to reference the blobs when appropriate. I also noticed that for the case of terminating a pending orchestration, the orchestration output was not set correctly if it was too large and needed to be stored in blob storage.

I also noticed some other places where fields were missing or incorrectly set (for example, for the case of a suborchestration, the tags, generation, and task hub name fields were not populated for the instance entity. 

I made sure to add tests for synchronizing the instance and history tables for a failed and terminated orchestration (not just a completed one), and tested large inputs and outputs for all three of these cases. I also added tests for terminating a pending orchestration with a large output (termination reason).

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -1095,6 +1095,23 @@ namespace DurableTask.AzureStorage.Tracking
                 instanceEntity["ScheduledStartTime"] = executionStartedEvent.ScheduledStartTime;
             }
 
+            static TableEntity GetSingleEntityFromHistoryTableResults(IReadOnlyList<TableEntity> entities, string dataType)
+            {
+                try
+                {
+                    TableEntity singleEntity = entities.SingleOrDefault();
+
+                    return singleEntity ?? throw new DurableTaskStorageException($"The history table query to determine the blob storage URL " +
+                        $"for the large orchestration {dataType} returned no rows. Unable to extract the URL from these results.");
+                }
+                catch (InvalidOperationException)
+                {
+                    throw new DurableTaskStorageException($"The history table query to determine the blob storage URL for the large orchestration " +
+                        $"{dataType} returned more than one row, when exactly one row is expected. " +
+                        $"Unable to extract the URL from these results.");
+                }
+            }
+
             // Set the output.
             // In the case that the output is too large and is stored in blob storage, extract the blob name from the ExecutionCompleted history entity.
             if (this.ExceedsMaxTablePropertySize(runtimeState.Output))
@@ -1102,7 +1119,7 @@ namespace DurableTask.AzureStorage.Tracking
                 string filter = $"{nameof(ITableEntity.PartitionKey)} eq '{KeySanitation.EscapePartitionKey(instanceId)}'" +
                     $" and {nameof(OrchestrationInstance.ExecutionId)} eq '{executionId}'" +
                     $" and {nameof(HistoryEvent.EventType)} eq '{nameof(EventType.ExecutionCompleted)}'";
-                TableEntity executionCompletedEntity = (await this.QueryHistoryAsync(filter, instanceId, cancellationToken)).Single();
+                TableEntity executionCompletedEntity = GetSingleEntityFromHistoryTableResults(await this.QueryHistoryAsync(filter, instanceId, cancellationToken), "output");
                 this.SetInstancesTablePropertyFromHistoryProperty(
                     executionCompletedEntity,
                     instanceEntity,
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -798,21 +798,22 @@ public override async Task UpdateStatusForRewindAsync(string instanceId, Cancell
...
             string sanitizedInstanceId = KeySanitation.EscapePartitionKey(instanceId);
-            TableEntity entity = new TableEntity(sanitizedInstanceId, "")
+            TableEntity instanceEntity = new TableEntity(sanitizedInstanceId, "")
             {
                 ["RuntimeStatus"] = OrchestrationStatus.Terminated.ToString("G"),
-                ["LastUpdatedTime"] = lastUpdatedTime,
+                ["LastUpdatedTime"] = executionTerminatedEvent.Timestamp,
                 ["CompletedTime"] = DateTime.UtcNow,
-                [OutputProperty] = output
+                [OutputProperty] = executionTerminatedEvent.Input,
```
cgillum, line 811: Might be worth adding a comment here that this line (mapping the event input to the output column) is intentional and not a typo.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1063,28 +1066,51 @@ public override async Task UpdateInstanceStatusAndDeleteOrphanedBlobsForComplete
...
-
             string sanitizedInstanceId = KeySanitation.EscapePartitionKey(instanceId);
+            ExecutionStartedEvent executionStartedEvent = runtimeState.ExecutionStartedEvent;
+
+            // We need to set all of the fields of the instance entity in the case that it was never created for the orchestration.
+            // This can be the case for a suborchestration that completed in one execution, for example.
             var instanceEntity = new TableEntity(sanitizedInstanceId, string.Empty)
             {
                 // TODO: Translating null to "null" is a temporary workaround. We should prioritize 
                 // https://github.com/Azure/durabletask/issues/477 so that this is no longer necessary.
```
cgillum: This comment now needs to be moved three lines down

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1063,28 +1066,51 @@ public override async Task UpdateInstanceStatusAndDeleteOrphanedBlobsForComplete
...
+            if (runtimeState.ExecutionStartedEvent.ScheduledStartTime.HasValue)
+            {
+                instanceEntity["ScheduledStartTime"] = executionStartedEvent.ScheduledStartTime;
+            }
+
+            // Set the input and output
+            // In the case that the input or output are too large and are stored in blob storage, we must extract the blob names from the history entities.
+            TableQueryResults<TableEntity> results = await this
+                .GetHistoryEntitiesResponseInfoAsync(instanceId, executionId, null, cancellationToken)
+                .GetResultsAsync(cancellationToken: cancellationToken);
```
cgillum: This could be really expensive for long orchestrations. Are we sure this is the right thing to do? I didn't check closely, but do we only go down this code path if we don't have an item already in the Instances table? It seems a little crazy to me that we'd potentially fetch the full history and re-serialize the input for each orchestration when it's only needed in a >0.001% of cases.

Let me know if I'm mixed up about the scenario.

sophiatev: This is only necessary in a few niche scenarios, you're right. Let me fix this (those scenarios being: the output is in blob storage, in which case we need to extract the execution completed event from the history table to get the blob URL. Or the input is too large, and an instance entity was not already stored, in which case we need to extract the URL from the execution started event in the history table).

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1095,22 +1095,14 @@ public override async Task UpdateInstanceStatusForCompletedOrchestrationAsync(
...
-                    .GetResultsAsync(cancellationToken: cancellationToken);
-            }
-
-            if (outputTooLarge)
-            {
-                TableEntity executionCompletedEntity = results.Entities.LastOrDefault(e => (string)e["EventType"] == EventType.ExecutionCompleted.ToString());
+                string filter = $"{nameof(ITableEntity.PartitionKey)} eq '{KeySanitation.EscapePartitionKey(instanceId)}'" +
+                    $" and {nameof(OrchestrationInstance.ExecutionId)} eq '{executionId}'" +
+                    $" and {nameof(HistoryEvent.EventType)} eq '{nameof(EventType.ExecutionCompleted)}'";
+                TableEntity executionCompletedEntity = (await this.QueryHistoryAsync(filter, instanceId, cancellationToken)).Single();
```
cgillum: The `.Single()` method will throw a very generic exception if the number of rows returned != 1.  Consider using `.FirstOrDefault()` and then throwing a friendlier exception if we get zero results back.

Same thing for both queries.

PR #1280, opened 2025-12-09
Adding ETag Usage to the Instance Table

sophiatev:
This PR introduces the ability to use etags when attempting to update the instance table in Azure Storage upon completion of a work item. This behavior will be "off" by default (elaborated on below). This is to help detect the following scenario.

1. Worker A attempts to complete a work item for an orchestration: it commits all outgoing messages, successfully updates the history table, then stalls.
2. Worker B picks up the next work item, which ends the orchestrations, and successfully completes it (updates the history and instance tables to reflect the completed status of the orchestration and deletes all control queue messages for the orchestration).
3. Worker A then resumes and updates the instance table to have status "Running".

Since the orchestration was completed in step 2 and all control queue messages for it deleted, there is no way to detect this scenario (i.e., no future messages will "retrigger" this orchestration to run). The only way to prevent this from happening, as far as I can tell, is to introduce etag usage for the instance table. Then, when worker A attempts to update the instance table in step 3, it will fail due to an etag mismatch.

This new behavior would require doing a read on the instance table to get the latest instance table etag for every new orchestration work item (assuming extended sessions are not enabled). After running some performance tests to validate the impact of this new I/O, I discovered that:

1. When running 1000 fanout orchestrations in parallel, each of which does a `Task.WhenAll` on 10 activity calls, the existing code without the instance table etag usage took around 14.5 minutes to complete across 3 trials whereas this new code took around 17.5 minutes.
2. When running 1000 sequential orchestrations in parallel, each of which sequentially awaits the result of 10 activity calls, the existing code took around 22 minutes to complete across 3 trials and the new code took around 25.

Given the negative performance impact of enabling this new etag usage, this PR hides it behind a feature flag in the `AzureStorageOrchestrationServiceSettings` which is off by default.

### DIFF
--- a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
+++ b/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
@@ -284,9 +284,6 @@ namespace DurableTask.AzureStorage
                 {
                     // Happy path: The message matches the table status. Alternatively, if the table doesn't have an ExecutionId field (older clients, pre-v1.8.5),
                     // then we have no way of knowing if it's a duplicate. Either way, allow it to run.
-
-                    // There's actually no extra cost from setting and using the instance etag in this case, I just don't for consistency since it will not be set and
-                    // used for future work items. Is this a good reason?
                     if (this.settings.UseInstanceTableEtag)
                     {
                         message.MessageMetadata = remoteInstance.ETag;


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1424,6 +1412,63 @@ bool ExceedsMaxTablePropertySize(string data)
             return false;
         }
 
+        async Task<ETag?> TryUpdateInstanceTableAsync(TableEntity instanceEntity, ETag? eTag, string instanceId, string executionId, OrchestrationStatus runtimeStatus, int episodeNumber)
```
cgillum: nit: It's .NET convention that all `TryXXX` methods return `bool`. Since you're instead returning `ETag`, I suggest removing the `Try` prefix from the name.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1424,6 +1412,63 @@ bool ExceedsMaxTablePropertySize(string data)
...
+                    // indicates that the table entity we are trying to "add" already exists.
+                    if (ex.HttpStatusCode == (int)HttpStatusCode.Conflict || ex.HttpStatusCode == (int)HttpStatusCode.PreconditionFailed)
+                    {
+                        this.settings.Logger.SplitBrainDetected(
+                            this.storageAccountName,
+                            this.taskHubName,
+                            instanceId,
+                            executionId,
+                            newEventCount: 0,
+                            totalEventCount: 1, // these fields don't really make sense for the instance table case. do we want to introduce a new log? or are we okay with this since "InstanceEntity"
```
cgillum: A new event probably makes the most sense since this isn't necessarily a split-brain problem, but at the same time I don't know if it's worth the effort to create a new event just for this case. As you mentioned, "InstanceEntity" gives enough clarity as far as how to interpret the failure.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -294,5 +294,22 @@ internal LogHelper Logger
...
+        /// 1. Worker A completes an orchestration work item, sends outgoing messages, updates the history table, then stalls.
+        /// 2. Worker B picks up the next and final orchestration work item for the same instance and completes it, updating the history 
+        /// table and instance table with the new terminal status.
+        /// 3. Worker A resumes and overrides the terminal status in the instance table with an incorrect non-terminal status.
+        /// This will leave the instance status of the orchestration permanently as "Running" even though it has actually completed.
+        /// To prevent such scenarios, enabling this setting will ensure that instance table updates also use etags. This would prevent
+        /// worker A's update in step 3 from completing. Enabling this setting will also introduce a performance overhead since the instance
+        /// table must now be read before processing every orchestration work item to obtain the latest etag.
+        /// </remarks>
+        public bool UseInstanceTableEtag { get; set; } = false;
```
cgillum, line 313: One important thing we should include in this documentation is the tradeoff because it's not obvious why this would be set to false by default: which is because there is a performance penalty associated with turning on the feature.

src/DurableTask.AzureStorage/OrchestrationETags.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Azure;
```
cgillum: All new files need to include the copyright header.

src/DurableTask.AzureStorage/OrchestrationSessionManager.cs
```diff
@@ -519,9 +530,23 @@ async Task ScheduleOrchestrationStatePrefetch(
...
                     batch.OrchestrationState = new OrchestrationRuntimeState(history.Events);
-                    batch.ETag = history.ETag;
+                    batch.ETags.HistoryETag = history.ETag;
                     batch.LastCheckpointTime = history.LastCheckpointTime;
                     batch.TrackingStoreContext = history.TrackingStoreContext;
+
+                    // Try to get the instance ETag from the tracking store if it wasn't already provided
+                    if (this.settings.UseInstanceTableEtag && batch.ETags.InstanceETag == null)
+                    {
+                        // Do we want to introduce a new method to just get the ETag without fetching the full instance status?
```
cgillum: Agreed that it's probably not worth creating a new method for this.

src/DurableTask.AzureStorage/MessageData.cs
```diff
@@ -111,6 +111,8 @@ internal void Update(UpdateReceipt receipt)
         {
             this.OriginalQueueMessage = this.OriginalQueueMessage.Update(receipt);
         }
+
+        internal object MessageMetadata { get; set; }
```
cgillum, line 115: What is this new field? Is it being used?

test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -33,6 +26,13 @@ namespace DurableTask.AzureStorage.Tests
     using DurableTask.Core.Exceptions;
     using DurableTask.Core.History;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Net;
+    using System.Threading;
+    using System.Threading.Tasks;
```
cgillum: nit: the expected sort order for all these files is to have `System` using statements first in the list. Not sure how this change was made, but there's a setting in Visual Studio to enforce this if you're using Visual Studio for development.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
sophiatev: That's actually the last sentence of the comment:

"Enabling this setting will also introduce a performance overhead since the instance table must now be read before processing every orchestration work item to obtain the latest etag."

Is it too buried though, should I make it more upfront?

src/DurableTask.AzureStorage/MessageData.cs
sophiatev: Yeah, I use it in the case that we already make a call to get the instance status upon verifying `ExecutionStartedEvent` messages here: https://github.com/Azure/durabletask/pull/1280/changes#diff-e0d4790c90d444980109aa613526aaa58cb662244025589fff10ca7e3984eb05R292

This is to prevent making a redundant call to get the instance etag later.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
cgillum: I guess I was reading too quickly. OK, glad to know it's there. I'll leave it up to you if you want to call this out more explicitly. If you have other changes to make, maybe make this part into its own paragraph.

PR #1293, opened 2025-12-23
Add Tags to SubOrchestrationInstanceCreatedEvent

sebastianburckhardt:
Previously the history event for creating suborchestration did not record the `Tags` dictionary. This creates challenges for the rewind implementation in the DTS backend, since it uses this history event to reconstruct the ExecutionStartedEvent in cases where a suborchestration was purged prior to a rewind.

### DIFF
--- a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
+++ b/src/DurableTask.Core/TaskOrchestrationDispatcher.cs
@@ -1221,7 +1221,7 @@ namespace DurableTask.Core
             bool includeParameters,
             Activity? parentTraceActivity)
         {
-            IDictionary<string,string> mergedTags = OrchestrationTags.MergeTags(createSubOrchestrationAction.Tags, runtimeState.Tags);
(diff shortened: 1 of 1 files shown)


src/DurableTask.Core/History/SubOrchestrationInstanceCreatedEvent.cs
```diff
@@ -47,6 +48,7 @@ internal SubOrchestrationInstanceCreatedEvent(SubOrchestrationInstanceCreatedEve
             InstanceId = other.InstanceId;
             Input = other.Input;
             ClientSpanId = other.ClientSpanId;
+            Tags = other.Tags;
```
sophiatev, line 51: Should we do a deep copy here? Not sure how much it matters

sebastianburckhardt: Yeah I was also thinking about it. I don't think we have any code that mutates the tags dictionary so it is effectively readonly AFAIK.

PR #1298, opened 2026-01-15
Add SetUseSeparateQueueForEntityWorkItems

andystaples:
Allows configuring this setting after initializing AzureStorageOrchestrationService to allow post-config by extension based on worker indexing results

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -2151,14 +2151,15 @@ namespace DurableTask.AzureStorage
         }
 
         /// <summary>
-        /// Sets the value for settings.UseSeparateQueueForEntityWorkItems, allowing modification after initialization.
+        /// Whether to use separate work item queues for entities and orchestrators.
+        /// Wraps the value of AzureStorageOrchestrationServiceSettings.UseSeparateQueueForEntityWorkItems.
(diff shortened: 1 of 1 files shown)


src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2150,6 +2150,15 @@ public Task<string> DownloadBlobAsync(string blobUri)
             return cachedQueue;
         }
 
+        /// <summary>
+        /// Sets the value for settings.UseSeparateQueueForEntityWorkItems, allowing modification after initialization.
+        /// </summary>
+        /// <param name="newValue"></param>
+        public void SetUseSeparateQueueForEntityWorkItems(bool newValue)
```
cgillum: Should we just make this a get/set property?

```suggestion
        public bool UseSeparateQueueForEntityWorkItems
        {
            get => this.settings.UseSeparateQueueForEntityWorkItems;
            set => this.UseSeparateQueueForEntityWorkItems = value;
        }
```

cgillum (approved):

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
andystaples: Good idea - added

sophiatev (approved):

PR #1303, opened 2026-02-23
Fix race condition with continue-as-new

### LINKED ISSUE #1302: Orchestration is stuck in the Running state
### Description
My workflow is designed to wait for a specific state from an external service. Since this can take days or even weeks, I use the **eternal orchestration** pattern. If the awaited status is not reached, the orchestration calls `ContinueAsNew` to start the next iteration.

Out of thousands of executions, the workflow occasionally becomes stuck. This issue started occurring after upgrading the **Durable Task extension for .NET isolated functions** to version **1.13.0**. Upgrading further to **1.14.0** did not resolve the issue — it has happened again.

Using **Durable Functions Monitor**, I identified a stuck orchestration with ID:

```
efd34532203e593297b36e38bd89590f
```

Its last update timestamp is:

```
2026-02-16T09:41:54 UTC
```

which is more than 10 hours ago.

From the logs, I can see that the orchestration was running normally for approximately 50 minutes before it stopped making progress.

```
2026-02-16 08:57:47.9259236 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' started.
2026-02-16 08:58:51.9662074 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' completed.
2026-02-16 08:58:51.9693582 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' started.
2026-02-16 08:59:51.7197587 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' completed.
2026-02-16 08:59:51.7239817 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' started.
2026-02-16 09:00:51.6458984 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' completed.
2026-02-16 09:00:51.6488428 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' started.
2026-02-16 09:01:51.5722086 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' completed.
...
2026-02-16 09:41:54.6424303 | 'wait-for-task' orchestration with ID 'efd34532203e593297b36e38bd89590f' started.
```

I suspect there may be an issue with the control queue message processing, as I can see errors in the Log Stream
```
2026-02-16T20:45:04   [Warning]   Error response [ff130abf-c869-44d5-9a42-d6a7d41f42cf] 404 The specified queue does not exist. (00.0s)
Server:Windows-Azure-Queue/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id:b5953f2c-e003-000d-2b85-9f01c4000000
x-ms-client-request-id:ff130abf-c869-44d5-9a42-d6a7d41f42cf
x-ms-version:2025-07-05
x-ms-error-code:QueueNotFound
Date:Mon, 16 Feb 2026 20:45:03 GMT
Content-Length:217
Content-Type:application/xml
```

I also attempted to identify any stuck messages in the control queue by searching for messages with a high DequeueCount, but I was unable to find any such messages. There are no obvious poison messages or repeatedly dequeued items that would explain the behavior.

### Expected behavior
Workflow should not stop progressing

### App Details

- **Durable Functions extension version 1.14.0**:
- **Azure Functions runtime version 2.0**:
- **Programming language used**: C# .Net 8

### If deployed to Azure
- **Timeframe issue observed**: 2026-02-16T09:41:54 UTC
- **Function App name**: odmt-eu-red-files-workers
- **Function name(s)**: wait-for-task
- **Azure region**: North Europe
- **Orchestration instance ID(s)**: efd34532203e593297b36e38bd89590f
- **Azure storage account name**: odmteuredfidata

sophiatev:
Currently we have a subtle race condition that can occur if an orchestration attempts to continue-as-new due to clock skew across different instances. The flow is as follows

1. An orchestration continues-as-new with a new execution ID, and the `TaskOrchestrationDispatcher` calls `CompleteTaskOrchestrationWorkItemAsync`.
2. In the completion call, outbound messages are committed. Say one of these is a `TaskScheduled` event to start a new Activity.
3. The Activity completes on instance B and sends a `TaskCompleted` event with timestamp x back to the orchestration, all before the `CompleteTaskOrchestrationWorkItemAsync` has updated the orchestration's state in storage to reflect the new execution ID.
4. A call to `LockNextTaskOrchestrationWorkItemAsync` is made by instance A which retrieves the `TaskCompleted` event. The `TaskCompleted` event is addressed to the new execution ID, but since the orchestration's state has not yet been updated in storage, there is no record or history for that execution ID. This [call](https://github.com/Azure/durabletask/blob/7c4e018197c8b10c4985465ed4de5716ba81df0b/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs#L140) to determine out of order messages should detect that this is potentially an "out of order" `TaskCompleted` message, since the instance does "not yet exist". However, `IsOutOfOrderMessage` decides that the message is okay, because this [condition](https://github.com/Azure/durabletask/blob/1b04239f01b8376e084d9b957bf15b546700dd64/src/DurableTask.AzureStorage/Messaging/OrchestrationSession.cs#L163) evaluates to true. This is due to the clock skew across instance A and B. Even though the `TaskCompleted` event was actually generated _after_ the old history checkpoint for the previous execution ID, its timestamp x is less than `this.LastCheckpointTime` due to clock skew. The timestamp x on the `TaskCompleted` event is set by role instance B, whereas `this.LastCheckpointTime` is set by role instance A.
5. Later on, in the `LockNextTaskOrchestrationWorkItemAsync` method, when we attempt to retrieve information about this orchestration instance with the new execution ID in storage, we find none, and fail at this [point](https://github.com/Azure/durabletask/blob/7c4e018197c8b10c4985465ed4de5716ba81df0b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L1090). We delete the `TaskCompleted` event, which leaves the orchestration permanently stuck in a running state.

The core of the issue is that the time check in `IsOutOfOrderMessage` is potentially comparing timestamps across different role instances. As such this can never be a reliable check.

This PR removes this check entirely, and also add tests for the remaining checks (non-existent-instance, no matching task scheduled, no matching event sent, etc.) 

Resolves #1302

### DIFF
--- a/test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
+++ b/test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
@@ -3481,8 +3481,8 @@ namespace DurableTask.AzureStorage.Tests
             {
                 var orchestrationInstance = new OrchestrationInstance
                 {
-                    InstanceId = "instance_id",
-                    ExecutionId = "execution_id",
+                    InstanceId = Guid.NewGuid().ToString(),
+                    ExecutionId = Guid.NewGuid().ToString(),
                 };
 
                 ExecutionStartedEvent startedEvent = new(-1, string.Empty)
@@ -3574,8 +3574,8 @@ namespace DurableTask.AzureStorage.Tests
             {
                 var orchestrationInstance = new OrchestrationInstance
                 {
-                    InstanceId = "instance_id",
-                    ExecutionId = "execution_id",
+                    InstanceId = Guid.NewGuid().ToString(),
+                    ExecutionId = Guid.NewGuid().ToString(),
                 };
 
                 ExecutionStartedEvent startedEvent = new(-1, string.Empty)


test/DurableTask.AzureStorage.Tests/AzureStorageScaleTests.cs
```diff
@@ -643,281 +643,6 @@ await TestHelpers.WaitFor(
             }
         }
 
-        /// <summary>
```
sophiatev, line 646: I incorrectly added these tests to this class in earlier PRs. They always should have been in the `AzureStorageScenarioTests` class so I took this opportunity to move them there

PR #1315, opened 2026-03-13
Clarify README support status for DTFx

### LINKED ISSUE #1314: A little heads up would have been nice
So this library is now unsupported, and we're told to use the Durable Task SDKs, but according to this...

https://learn.microsoft.com/en-us/azure/azure-functions/durable/choose-orchestration-framework

"It's the recommended backend for Durable Functions and the only supported backend for the Durable Task SDKs."

Not clear how much more the Durable Task Scheduler is going to be, but are there any plans to implement additional backends for the Durable Task SDKs?

torosent:
Rephrases the README disclaimer to clarify that DTFx is a community-maintained open-source project that continues to be actively used in production (including by Microsoft engineering teams), rather than implying it is abandoned.

### Callout box
- Replaced "legacy library that is not officially supported" with "community-maintained open-source project"
- Clarified that "no official support" means no Microsoft support tickets -- not that the project is being abandoned
- Explicitly states the project will continue to be maintained on a best-effort basis
- Softened migration language from "looking to migrate" to "if you need official Microsoft support"

### Intro paragraph
- Removed "predates them and doesn't include the latest features" framing
- Now states DTFx "continues to be maintained and used in production"
- Clarifies newer Durable Task SDKs offer additional features + official support

### Documentation reference
- Removed "Legacy" label and "no longer maintained" from the docs callout

## Context

Based on the feedback in #1314 and the clarification from @cgillum:
- DTFx was never officially "supported" in the Microsoft support-ticket sense
- Nothing is actually changing -- DTFx is NOT going away
- It will continue to be kept up-to-date and compliant
- The newer Durable Task SDKs + Durable Task Scheduler will offer full Microsoft support

Closes #1314

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 >
 > These alternatives offer full Microsoft support, including the ability to open support tickets. For more details, see [Choosing an orchestration framework](https://learn.microsoft.com/azure/azure-functions/durable/choose-orchestration-framework#community-and-experimental-durable-task-sdks).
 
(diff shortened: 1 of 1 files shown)


cgillum (approved):
LGTM!

PR #1331, opened 2026-03-24
Add Poison Message Handling to the Dispatchers

sophiatev:
This PR adds poison message handling to the activity, entity, and orchestration dispatchers. The general policy followed is that we want to make sure when poison message handling is enabled:

1. exceptions are not thrown from the dispatchers which would cause repeated aborting of the work item in the case of an irrecoverable error
2. we stop dispatching a message after its dispatch count exceeds the user-configured maximum
3. whenever possible, we surface this information to the customer via a failed orchestration or entity operation

Depending on the type of "irrecoverable" error, the backends might have to add special edge-case handling for the poison message. The SDK's responsibility is simply to mark the message as poisoned and prevent its processing.

Note that we have intentionally chosen _not_ to include poison message handling for unlock requests. This is because failing to unlock an entity could leave an entire task hub in a bad state, so we retain the current behavior.

### DIFF
--- a/src/DurableTask.Core/History/HistoryEvent.cs
+++ b/src/DurableTask.Core/History/HistoryEvent.cs
@@ -90,6 +90,12 @@ namespace DurableTask.Core.History
         public virtual EventType EventType { get; private set; }
 
         /// <summary>
+        /// Gets or sets the number of times this event has been dispatched.
+        /// </summary>
+        [DataMember]
+        public int DispatchCount { get; set; }
+
+        /// <summary>
         /// Implementation for <see cref="IExtensibleDataObject.ExtensionData"/>.
         /// </summary>
         public ExtensionDataObject? ExtensionData { get; set; }
--- a/src/DurableTask.Core/History/HistoryEvent.cs
+++ b/src/DurableTask.Core/History/HistoryEvent.cs
@@ -96,6 +96,14 @@ namespace DurableTask.Core.History
         public int DispatchCount { get; set; }
 
         /// <summary>
+        /// Gets or sets whether or not this event has been marked as "poisoned".
+        /// This can occur if the event's dispatch count exceeds a certain threshold,
+        /// or if some other error occurs during dispatch.
+        /// </summary>
+        [DataMember]
+        public bool IsPoisoned { get; set; }
+
+        /// <summary>
         /// Implementation for <see cref="IExtensibleDataObject.ExtensionData"/>.
         /// </summary>
         public ExtensionDataObject? ExtensionData { get; set; }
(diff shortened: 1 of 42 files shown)


src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -432,6 +434,34 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
...
+                                    runtimeState.OrchestrationInstance!,
+                                    poisonEvent,
+                                    $"Orchestration has received an event with dispatch count {poisonEvent.DispatchCount} which exceeds the maximum dispatch" +
+                                    $"count of {this.maxDispatchCount} and will be failed.");
+                            }
+
+                            var failureAction = new OrchestrationCompleteOrchestratorAction
+                            {
+                                Id = runtimeState.PastEvents.Count,
+                                FailureDetails = new FailureDetails(
```
sophiatev: Given that we perform this logic before any other orchestration processing, we will potentially override the failure of an already-failing-orchestration with this failure. Are we okay with this? The problem of course is that if we attempt to process the poison messages that correspond to the already-failing-orchestration, we could run into issues or exceptions (they are "poisoned" for a reason).

src/DurableTask.Core/History/HistoryEvent.cs
```diff
@@ -89,6 +89,20 @@ protected HistoryEvent(int eventId)
         [DataMember]
         public virtual EventType EventType { get; private set; }
 
+        /// <summary>
+        /// Gets or sets the number of times this event has been dispatched.
+        /// </summary>
+        [DataMember]
```
cgillum: I suggest we have these new properties be omitted by default if not populated, especially `IsPoisoned`. This will reduce the chance that there's some compatibility issue with older versions of the SDK that try to deserialize the history.

```suggestion
        [DataMember(EmitDefaultValue = false)]
```

cgillum (commented):
Sharing some initial feedback - I unfortunately haven't reviewed everything yet.

cgillum (commented):
Finished reviewing all the changes.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -626,6 +699,14 @@ void DetermineWork(OrchestrationRuntimeState runtimeState, ref SchedulerState sc
                     }
 
                     var request = schedulerState.Dequeue();
+                    if (request.DispatchCount > this.maxDispatchCount)
+                    {
+                        this.logHelper.PoisonMessageDetected(
+                            runtimeState.OrchestrationInstance,
+                            request,
+                            $"Entity request has dispatch count {request.DispatchCount} which exceeds the maximum dispatch count " +
+                            $"of {this.maxDispatchCount} and will be failed.");
```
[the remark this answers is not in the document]
sophiatev: We do want to process it in the sense that we want to make sure to send a calling orchestration the failure details if we can

PR #1333, opened 2026-03-30
Adding new logic for extended sessions to enable the DTS use-case

sophiatev:
This PR introduces new logic in the extended sessions flow that is necessary for the DTS use-case. In particular, it introduces a new `EndSessionAsync` call which will allow DTS to release the work item once the extended session ends.

### DIFF
--- a/src/DurableTask.Core/IOrchestrationSession.cs
+++ b/src/DurableTask.Core/IOrchestrationSession.cs
@@ -11,6 +11,7 @@
 //  limitations under the License.
 //  ----------------------------------------------------------------------------------
 
+#nullable enable
 namespace DurableTask.Core
 {
     using System.Collections.Generic;
@@ -30,5 +31,11 @@ namespace DurableTask.Core
         /// and the dispatcher will shut down the session.
         /// </remarks>
         Task<IList<TaskMessage>> FetchNewOrchestrationMessagesAsync(TaskOrchestrationWorkItem workItem);
+
+        /// <summary>
+        /// Ends the session.
+        /// </summary>
+        /// <returns>A task that completes when the session has been ended.</returns>
+        Task EndSessionAsync();
     }
 }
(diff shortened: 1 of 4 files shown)


src/DurableTask.Core/IOrchestrationSession.cs
sebastianburckhardt, line 23: changing a public interface is a breaking change and, as such, requires a major version change.

This could break a lot of code all over the place.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -552,10 +553,6 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                                     orchestratorMessages.AddRange(subOrchestrationRewindMessages);
                                     workItem.OrchestrationRuntimeState = newRuntimeState;
                                     runtimeState = newRuntimeState;
-                                    // Setting this to true here will end an extended session if it is in progress.
-                                    // We don't want to save the state across executions, since we essentially manually modify
-                                    // the orchestration history here and so that stored by the extended session is incorrect.
-                                    isRewinding = true;
```
sebastianburckhardt, line 558: why is this assignment no longer needed?

sophiatev: This was leftover from a previous PR where I added rewind support. I realized that I had moved where this was set to [here](https://github.com/Azure/durabletask/blob/1107cd15d91baede0b60b7c3fbd00000a007af1a/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L349) but forgot to remove this redundant setting later on.

src/DurableTask.Core/IOrchestrationSession.cs
sophiatev: Technically yes, but only Azure Storage implements this interface and it is updated in this same PR so in practice it won't be one

sophiatev: That being said, I will discuss the nature of the "breaking change" and whether we do want this to be a major version update with the rest of the team. The version change will appear in a separate PR either way

cc @torosent

sebastianburckhardt (approved):

PR #1366, opened 2026-06-10
Add poison message handling to Azure Storage

sophiatev:
This PR introduces poison message handling to the Azure Storage. This is a very primitive implementation which simply stores a message in poison storage if its dequeue count exceeds the (user-configurable) maximum, and deletes it from the queue.


src/DurableTask.Core/History/ExecutionRewoundEvent.cs
```diff
@@ -39,6 +39,12 @@ public ExecutionRewoundEvent(int eventId, string? reason)
             this.Reason = reason;
         }
 
+        // Private ctor for JSON deserialization (required by some storage providers and out-of-proc executors)
```
sophiatev, line 42: Unrelated to this PR but I bug I found when testing (JSON was not able to deserialize this event because it lacked a 0-arg constructor and the other constructors all had multiple parameters)

src/DurableTask.Core/Tracing/TraceHelper.cs
sophiatev, line 1: Also unrelated to this PR, but I realized while working on it that this code I wrote a while back had some incorrect assumptions so I took the opportunity to fix it

src/DurableTask.Core/Entities/EventFormat/RequestMessage.cs
```diff
@@ -122,6 +122,17 @@ internal class RequestMessage : EntityMessage
         /// </summary>
         public string? ClientSpanId { get; set; }
 
+        /// <summary>
+        /// If the request message is poisoned, the reason it is poisoned.
+        /// Otherwise, null.
+        /// </summary>
+        public string? PoisonReason { get; set; }
```
cgillum: What's the use case for `PoisonReason` in entity request messages?

src/DurableTask.Core/Logging/LogEvents.cs
```diff
@@ -1976,5 +1976,59 @@ void IEventSourceEvent.WriteEventSource() =>
...
+            public int DispatchCount { get; }
+
+            [StructuredLogField]
+            public string Details { get; }
+
+            public override EventId EventId => new EventId(
+                EventIds.PoisonMessageDetected,
+                nameof(EventIds.PoisonMessageDetected));
+
+            public override LogLevel Level => LogLevel.Error;
```
cgillum: This should be `Warning` since it's something we've handled gracefully.

src/DurableTask.Core/Logging/StructuredEventSource.cs
```diff
@@ -656,6 +656,33 @@ internal void DiscardingMessage(
             }
         }
 
+        [Event(EventIds.PoisonMessageDetected, Level = EventLevel.Error, Version = 1)]
```
cgillum: Same comment about making this `Warning` instead of Error.

src/DurableTask.Core/IPoisonMessageHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public interface IPoisonMessageHandler
+    {
+        /// <summary>
+        /// Determines whether the given <see cref="HistoryEvent"/> is a poison message.
+        /// </summary>
+        /// <param name="historyEvent">The history event being dispatched.</param>
+        /// <param name="reason">Why the message is considered poisoned.</param>
+        /// <returns><c>true</c> if the message should be treated as poisoned; otherwise <c>false</c>.</returns>
+        public bool IsPoisonMessage(HistoryEvent historyEvent, out string? reason);
```
cgillum: Why do we need the backend to decide whether something is a poison message? Shouldn't the dispatcher be making this decision?

src/DurableTask.Core/IPoisonMessageHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Invoked to handle a poison message in the case that a message cannot necessarily
+        /// be "failed" by the dispatchers, so  the <see cref="IPoisonMessageHandler"/> must
+        /// decide what to do.
+        /// </summary>
+        /// <param name="orchestrationInstance">The orchestration instance the event was sent to, or null
+        /// if this information is not available.</param>
+        /// <param name="historyEvent">The "poisoned" history event.</param>
+        /// <param name="reason">The reason the event is "poisoned".</param>
+        /// <returns>True if the poison message was successfully handled, otherwise false.</returns>
```
cgillum: It would be useful to explain in the summary or in `<remarks>` what is expected to happen if this method returns `false`.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -441,7 +482,7 @@ protected async Task<bool> OnProcessWorkItemAsync(TaskOrchestrationWorkItem work
                             "Executing user orchestration: {0}",
                             JsonDataConverter.Default.Serialize(runtimeState.GetOrchestrationRuntimeStateDump(), true));
 
-                        if (!versioningFailed)
+                        if (!versioningFailed && !poisonEvents.Any())
```
cgillum: nit: here we do `!poisonEvents.Any()` but earlier on line 452 the code does `poisonEvents.Count > 0`. I suggest we use the latter check to avoid any confusion about lazy enumerator evaluation or allocations associated with calls to `.Any()`, and to keep the code consistent.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
```diff
@@ -873,14 +914,31 @@ await this.dispatchPipeline.RunAsync(dispatchContext, _ =>
...
+        /// <param name="reason">In the case that the work item should be dropped (this method return false), provides the reason why.</param>
         /// <returns>True if workItem should be processed further. False otherwise.</returns>
-        internal static bool ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem, string dispatcher, ErrorPropagationMode errorPropagationMode, LogHelper logHelper)
+        internal static bool ReconcileMessagesWithState(
+            TaskOrchestrationWorkItem workItem,
+            string dispatcher,
+            ErrorPropagationMode errorPropagationMode,
+            LogHelper logHelper,
+            bool isPoisonMessageHandlingEnabled,
+            out string? reason)
```
cgillum: We should give this a more descriptive name, like `errorMessage` to make it more obvious that it's associated with some failure. We should also mark this with `[NotNullWhen(false)]` so that the compiler understands that this variable is always non-null when the method returns `false`.

cgillum (commented):
Some initial comments. I haven't gone through the dispatcher code yet (those are bigger diffs).

src/DurableTask.Core/Entities/EventFormat/RequestMessage.cs
sophiatev: It's used when generating the failure response for a poisoned entity operation, i.e. [here](https://github.com/Azure/durabletask/blob/85ae25016d8f9f087c388b2fefaf5734d8db11f7/src/DurableTask.Core/TaskEntityDispatcher.cs#L1171). 

Since processing the history event that corresponds to the entity request is [decoupled](https://github.com/Azure/durabletask/blob/85ae25016d8f9f087c388b2fefaf5734d8db11f7/src/DurableTask.Core/TaskEntityDispatcher.cs#L606) from sending the response, we need to store the poison reason in the request message so we can use it later to populate the failure details.

src/DurableTask.Core/IPoisonMessageHandler.cs
sophiatev: We don't, and it definitely can. I thought it would be the responsibility of the "poison message handler" since it's sort of in the name but this was just a somewhat arbitrary choice on my end. I can return responsibility to the dispatchers

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
sophiatev: Will definitely change the field name but unfortunately netstandard2.0 doesn't support the `NotNullWhen` attribute

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -117,9 +119,16 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
             {
                 if (orchestrationInstance == null || string.IsNullOrWhiteSpace(orchestrationInstance.InstanceId))
                 {
-                    this.logHelper.TaskActivityDispatcherError(
-                        workItem,
-                        $"The activity worker received a message that does not have any OrchestrationInstance information.");
+                    string message = "The activity worker received a message that does not have any OrchestrationInstance information.";
+                    this.logHelper.TaskActivityDispatcherError(workItem, message);
```
cgillum: I would move this error log down below, right before the point where we throw the exception. If poison message handling is handling the error, then we don't need to also log this message.

src/DurableTask.Core/TaskActivityDispatcher.cs
```diff
@@ -128,151 +137,189 @@ async Task OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
 
                 if (taskMessage.Event.EventType != EventType.TaskScheduled)
                 {
-                    this.logHelper.TaskActivityDispatcherError(
-                        workItem, 
-                        $"The activity worker received an event of type '{taskMessage.Event.EventType}' but only '{EventType.TaskScheduled}' is supported.");
+                    string message = $"The activity worker received an event of type '{taskMessage.Event.EventType}' but only " +
+                        $"'{EventType.TaskScheduled}' is supported.";
+                    this.logHelper.TaskActivityDispatcherError(workItem, message);
```
cgillum: Same comment about using error logging only when we throw (move it down).

src/DurableTask.Core/IPoisonMessageHandler.cs
cgillum: One thing that's tricky about reviewing this interface is that we don't yet have any implementations of it. I'm thinking we should probably implement this for Azure Storage before committing to these new public APIs.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -1,4 +1,4 @@
-﻿//  ----------------------------------------------------------------------------------
+//  ----------------------------------------------------------------------------------
```
cgillum: It might be good to have @sebastianburckhardt review the changes to this file as I'm less familiar with the details of entity dispatching.

src/DurableTask.Core/TaskOrchestrationDispatcher.cs
cgillum: Darn, I forgot about that `netstandard2.0` limitation. I wonder when we should look into dropping `.netstandard2.0` support...

cgillum (commented):
Finished reviewing. Just a few more comments (and some responses).

src/DurableTask.Core/IPoisonMessageHandler.cs
sophiatev: Actually one thing I remembered is that doing it this way meant I didn't have to pass a `maxDispatchCount` when instantiating the dispatchers, which meant we didn't need yet another overload for the `TaskHubWorker` to accept this parameter when it creates them. But if we want to keep this decision on the dispatcher side and avoid another parameter, we can maybe expose this via `IPoisonMessageHandler.MaxDispatchCount`, if that sounds reasonable?

cgillum: Yeah, that seems reasonable. It's at least consistent with some other properties we have on `IOrchestrationService`.

src/DurableTask.Core/TaskEntityDispatcher.cs
```diff
@@ -454,7 +481,22 @@ void ProcessLockRequest(WorkItemEffects effects, SchedulerState schedulerState,
             {
                 // send lock acquisition completed response back to originating orchestration instance
                 var target = new OrchestrationInstance() { InstanceId = request.ParentInstanceId, ExecutionId = request.ParentExecutionId };
-                this.SendLockResponseMessage(effects, target, request.Id);
+
+                // In the case of a poison message, it will be the locking instance's responsibility to unlock any other entities for whom the
```
sebastianburckhardt: I think this "locking instance's responsibility" has not yet been implemented?

For example in durabletask-dotnet src/Worker/Core/Shims/TaskOrchestrationEntityContext.cs is where I would expect release messages to be sent

sebastianburckhardt (commented):
I think the design is o.k. - it is not ideal in the sense that it is not simple, and not completely reliable. 

Meaning that it adds a lot of special paths where (a) some of those paths catch things that I don't think can happen, and (b) not all poison messages can be caught at this layer (e.g. if something goes wrong before execution even reaches the dispatcher). 

But we discussed this, and there is not a clear path to a better implementation (because a lot of complexity was introduced into the codebase over time and cannot be easily removed now). As is it should still make the code more resilient as it will catch some poison messages and produce detailed traces in those situations. Which is probably good enough.

src/DurableTask.Core/TaskEntityDispatcher.cs
sophiatev: Addressed in the .NET SDK PR: https://github.com/microsoft/durabletask-dotnet/pull/743/changes#diff-4ce2acdc2e29f811f25e432163b9b27d3f9684041d4d141534280dc74c693514R85

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2230,6 +2273,408 @@ private static OrchestrationQueryResult ConvertFrom(DurableStatusQueryResult sta
...
+            // There must have been a deserialization error with the ExecutionStartedEvent, or the work item is somehow
+            // otherwise invalid (i.e. missing an ExecutionStartedEvent).
+            // We make a best effort attempt to send a failure response for any entity calls included in the work item,
+            // but do not complete the work item since that would cause deletion of the other entity messages included in it and we
+            // have no real way to mark an entity as "failed".
+            // These messages will presumably continue to be retried since there is no completion to delete them, so all subsequent
+            // attempts will land in poison storage as well.
+            if (isEntity)
+            {
+                foreach (TaskMessage message in workItem.NewMessages)
```
sophiatev: Do we want to try and complete the work item in this case too? Otherwise the messages will keep getting redispatched. The reason I chose not to is because the entity messages would be permanently deleted but with no way to mark the entity as "failed" (same happens in the orchestration case a few lines down, but at least there we can mark the orchestration as "failed")

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -796,7 +808,11 @@ async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAsync(boo
                     {
                         InstanceId = session.Instance.InstanceId,
                         LockedUntilUtc = session.CurrentMessageBatch.Min(msg => msg.OriginalQueueMessage.NextVisibleOn.Value.UtcDateTime),
-                        NewMessages = session.CurrentMessageBatch.Select(m => m.TaskMessage).ToList(),
+                        NewMessages = session.CurrentMessageBatch.Select(m =>
+                        {
+                            m.TaskMessage.Event.DispatchCount = (int)m.OriginalQueueMessage.DequeueCount;
```
cgillum: Shouldn't we be updating this in the message dequeue path - i.e. TaskQueue.cs or ControlQueue.cs?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2230,6 +2273,403 @@ private static OrchestrationQueryResult ConvertFrom(DurableStatusQueryResult sta
...
+            OrchestrationInstance? entityInstance,
+            HistoryEvent historyEvent,
+            PoisonMessageReason reason,
+            string details)
+        {
+            if (!this.settings.IsPoisonMessageStorageEnabled || 
+                !await this.StorePoisonMessagesAsync(
+                entityInstance,
+                new[] { new TaskMessage { OrchestrationInstance = entityInstance, Event = historyEvent } },
+                this.instancePoisonMessageContainer))
```
cgillum: nit: these parameters should be indented.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2230,6 +2273,403 @@ private static OrchestrationQueryResult ConvertFrom(DurableStatusQueryResult sta
...
+                        : null;
+
+                    await this.TrySendEntityMessageAsync(new OrchestrationInstance { InstanceId = parentInstanceId }, unlockEvent);
+                }
+
+            }
+            else
+            {
+                // The only type of poison message we could ever receive via this API should be an EventRaisedEvent
+                this.settings.Logger.GeneralError(
```
cgillum: nit: this should be a warning and not an error.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2230,6 +2273,403 @@ private static OrchestrationQueryResult ConvertFrom(DurableStatusQueryResult sta
...
+                        eventId: -1,
+                        result: null,
+                        orchestrationStatus: OrchestrationStatus.Failed,
+                        failureDetails: failureDetails)
+                    {
+                        Timestamp = DateTime.UtcNow,
+                    });
+
+                    // Commit the updated history and instance table metadata (RuntimeStatus = Failed).
+                    await this.trackingStore.UpdateStateAsync(
```
cgillum: One thing that really concerns me about this change is that we're introducing write paths outside of the main work-item completion path. This adds a lot more complexity to the overall design of the Azure Storage backend because you can no longer make certain assumptions about when and how we write data, what race conditions are possible, and what the side effects of failure (e.g., process crashes) are. There are certain invariants we've tried to maintain over the years as we evolve this code, and it seems this design breaks some of those invariants.

cgillum (commented):
Adding a few quick comments for now.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
sophiatev: We can, but this felt the more natural place for me since we are setting it at the point where it will actually be used (this value is only read by the dispatchers, and so it made sense to me to set it when we are generating the work item to be sent to the dispatchers). And this way all the changes are localized in `AzureStorageOrchestrationService`

That being said I don't feel strongly about this so I can move it if that's better

src/DurableTask.Core/IPoisonMessageHandler.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// <summary>
+    /// Categorizes why a message or work item was considered "poisoned" so that an
+    /// <see cref="IPoisonMessageHandler"/> can decide how to handle it without parsing human-readable strings.
+    /// </summary>
+    public enum PoisonMessageReason
+    {
+        /// <summary>
+        /// A message or event could not be deserialized
+        /// </summary>
+        DeserializationError,
```
sophiatev: I added all of these for future-proofing (and consistency, to cover all our poison message cases) in case another backend needs them but really the Azure Storage implementation only needs a handful of them for its own poison-message-handling. 

I suppose if we ever need all of them, we could add them at that point and then make a new release that the backend could reference. Or we can keep them to be consistent

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -186,6 +188,16 @@ public AzureStorageOrchestrationService(AzureStorageOrchestrationServiceSettings
                 this.settings.TaskHubName.ToLowerInvariant() + "-applease",
                 this.settings.TaskHubName.ToLowerInvariant() + "-appleaseinfo",
                 this.settings.AppLeaseOptions);
+
+            if (this.settings.IsPoisonMessageStorageEnabled)
+            {
+                string prefix = string.IsNullOrEmpty(this.settings.PoisonMessageStorageContainerNamePrefix)
+                    ? "durable-task-poison"
+                    : this.settings.PoisonMessageStorageContainerNamePrefix;
```
cgillum: Generally speaking all containers we create should be prefixed with the task hub name. Otherwise, we'll leak data across apps if they share the same storage account.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -1278,7 +1294,15 @@ await this.CommitOutboundQueueMessages(
                 throw;
             }
 
-            // Finally, delete the messages which triggered this orchestration execution. This is the final commit.
+            // Finally, store any poison messages and delete the messages which triggered this orchestration execution. This is the final commit.
+            if (this.settings.IsPoisonMessageStorageEnabled)
+            {
+                await this.StorePoisonMessagesAsync(
```
cgillum: I'm a bit surprised to see this inside of CompleteTaskOrchestrationWorkItemAsync and not AbandonTaskOrchestrationWorkItemAsync. This is the code path we generally go down when a work item completes successfully, but a message is a poison message, I wouldn't expect us to get here...unless we can somehow guarantee that we gracefully handle every possible poison message situation (which doesn't seem possible to guarantee)?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -2230,6 +2273,396 @@ private static OrchestrationQueryResult ConvertFrom(DurableStatusQueryResult sta
...
+            OrchestrationInstance? orchestrationInstance = workItem.OrchestrationRuntimeState?.OrchestrationInstance;
+
+            if (!await this.StorePoisonMessagesAsync(orchestrationInstance, workItem.NewMessages, this.instancePoisonMessageContainer))
+            {
+                return false;
+            }
+
+            // There must have been a deserialization error with the ExecutionStartedEvent, or the work item is somehow
+            // otherwise invalid (i.e. missing an ExecutionStartedEvent).
+            // We make a best effort attempt to send a failure response for any entity calls included in the work item
```
cgillum: How do we know that this is a request/response operation and not a one-way operation?

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
cgillum: It would make sense to me if this setting were decoupled from `DequeueCount`. But the fact that it's always assigned to `DequeueCount`, I'm not sure why we wouldn't just assign it in one place shortly after receiving the message. It seems like that would be more maintainable. I don't feel strongly about it but it does seem more intuitive to me.

src/DurableTask.AzureStorage/Messaging/ControlQueue.cs
```diff
@@ -125,12 +128,22 @@ await batch.ParallelForEachAsync(async delegate (QueueMessage queueMessage)
                                     0 /* TaskEventId */,
                                     e.ToString());
 
+                                if (await this.TryMoveMessageToPoisonStorageAsync(queueMessage, linkedCts.Token))
```
sophiatev: Note that this will only store the message in poison storage if the dequeue count has been exceeded even for this deserialization error case. I'm not sure if this makes sense... this logic makes me think it might? https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.AzureStorage/MessageManager.cs#L173

Something like multiple Function apps talking to the same task hub (which is illegal anyway but customers do sometimes)

cgillum: nit: based on the name of this method, a reader might assume that we will always attempt to move the message to poison storage and return false only if we _failed_ to do so. I'm wondering if we could name this something that makes it sound more conditional, like `CheckForAndHandlePoisonMessage(...)`

cgillum: I think it's a good, conservative place to start. It's possible that the deserialization error can be fixed by upgrading the extension version or changing a setting, and the customer may prefer to do that before the message is permanently removed to avoid unnecessary disruption.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -389,6 +392,171 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase? s
...
+        protected abstract string PoisonMessageContainerName { get; }
+
+        protected async Task<bool> TryMoveMessageToPoisonStorageAsync(MessageData messageData, CancellationToken cancellationToken = default)
+        {
+            if (!this.settings.IsPoisonMessageStorageEnabled || messageData.OriginalQueueMessage.DequeueCount <= this.settings.MaxDequeueCount)
+            {
+                return false;
+            }
+
+            string serializedMessageData = JsonConvert.SerializeObject(
```
cgillum: Should we instead use our existing message encoding/decoding helper function(s) here instead of using Newtonsoft.Json directly? We need to be really careful and intentional about how we do message serialization because there are a ton of things that could go wrong, and this is a security-sensitive area in general.

cgillum: I also wonder if it would be better to just save the raw message content because right now it looks like we save two different types of payloads depending on where we detected the failure: either raw queue message content or deserialized (and re-serialized) message data payloads. Standardizing on one would make it far easier for us to introduce a mechanism to re-add these messages down the road. It also means we don't have to serialize anything.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -389,6 +392,171 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase? s
...
+        {
+            try
+            {
+                BlobContainer container = this.poisonMessageContainer ??=
+                    this.azureStorageClient.GetBlobContainerReference(this.PoisonMessageContainerName);
+
+                // Replace any invalid characters with a dash
+                string sanitizedInstanceId = SanitizeString(orchestrationInstance?.InstanceId, '-');
+                string sanitizedExecutionId = SanitizeString(orchestrationInstance?.ExecutionId, '-');
+                string blobNamePrefix = $"{sanitizedInstanceId}~{sanitizedExecutionId}";
```
cgillum: The `~` seems a bit risky to me since it's a special character with special meaning in certain path contexts (like on Unix-based systems). Maybe do `_` instead?

cgillum: Some messages, like external event or terminate messages, won't have an execution ID. Should we just use the instance ID in that case?

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -389,6 +392,171 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase? s
...
+                // Blob name length limit is 1024 characters and we attach an extra character (~) and GUID (32) at the end
+                // From https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata?#blob-names
+                const int MaxPrefixLength = 1024 - 32 - 1;
+                if (blobNamePrefix.Length > MaxPrefixLength)
+                {
+                    blobNamePrefix = blobNamePrefix.Substring(0, MaxPrefixLength);
+                }
+                string blobName = $"{blobNamePrefix}~{Guid.NewGuid():N}";
```
cgillum: Since these are queue messages, I suggest we use the existing queue message ID (which is also a GUID) as the identifier rather than a random GUID. That way users can easily correlate these blobs to actual messages that show up in their logs.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -389,6 +392,171 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase? s
...
+
+                await container.CreateIfNotExistsAsync(cancellationToken);
+
+                Blob blob = container.GetBlobReference(blobName);
+                await blob.UploadTextAsync(serializedPoisonMessage, cancellationToken: cancellationToken);
+
+                this.settings.Logger.GeneralWarning(
+                    this.azureStorageClient.BlobAccountName,
+                    this.settings.TaskHubName,
+                    $"Stored a poison message for instance InstanceId='{orchestrationInstance?.InstanceId ?? string.Empty}' " +
```
cgillum: Is there a log event we can use which takes the instance ID as a parameter? That would be better, because then we can much more easily find these warnings when filtering based on orchestration instance ID.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -389,6 +392,171 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase? s
...
+
+                await this.storageQueue.DeleteMessageAsync(queueMessage, cancellationToken: cancellationToken);
+            }
+            catch (Exception e)
+            {
+                this.settings.Logger.GeneralError(
+                    this.azureStorageClient.BlobAccountName,
+                    this.settings.TaskHubName,
+                    $"Error when attempting to store poison message for instance " +
+                        $"InstanceId={orchestrationInstance?.InstanceId ?? string.Empty} " +
```
cgillum: Same question about whether there's a log event we can use that takes InstanceId as a first-class parameter for improved filtering.

src/DurableTask.AzureStorage/Messaging/WorkItemQueue.cs
```diff
@@ -31,6 +31,9 @@ public WorkItemQueue(
 
         protected override TimeSpan MessageVisibilityTimeout => this.settings.WorkItemQueueVisibilityTimeout;
 
+        protected override string PoisonMessageContainerName =>
+            $"{this.settings.TaskHubName.ToLowerInvariant()}-{this.settings.PoisonMessageStorageContainerNamePrefix}-activity-messages";
```
cgillum: I suggest a common method for generating the container names in TaskHubQueue.cs which takes a suffix as a parameter.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -311,5 +311,80 @@ internal LogHelper Logger
...
+        {
+            get => this.poisonMessageStorageContainerNamePrefix;
+            set
+            {
+                ValidatePoisonMessageStorageContainerNamePrefix(value);
+                this.poisonMessageStorageContainerNamePrefix = value;
+            }
+        }
+
+        string poisonMessageStorageContainerNamePrefix = "durable-task-poison";
```
cgillum: Instance fields should be defined at the top of the class, before all properties and constructors.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
```diff
@@ -311,5 +311,80 @@ internal LogHelper Logger
...
+        /// <summary>
+        /// Gets or sets a flag whether poison message storage is enabled.
+        /// </summary>
+        /// <remarks>
+        /// <para>If enabled, orchestration, entity, and activity messages that have been dequeued for processing
+        /// more than <see cref="MaxDequeueCount"/> times will be moved to poison message storage in Azure Blob
+        /// Storage and deleted from their source queue.
+        /// </para>
+        /// </remarks>
+        public bool IsPoisonMessageStorageEnabled { get; set; } = false;
```
cgillum: I wonder if we should remove this property and just rely on `MaxDequeueCount`. The default value could be something really large, like 10K, which would represent roughly a month of retries and pretty much guarantees that customer doesn't care about this message. Having this enabled by default would allow many customers with existing poison messages to automatically self-heal.

cgillum (commented):
I haven't looked at the test cases yet, but here's some feedback based on the core implementation.

src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs
sophiatev: We can remove it but I can also imagine that there might be customers which really do want to turn this off and not have to think about it. Maybe we should keep it and just leave it true by default? But I don't feel strongly about this at all, we can just remove it entirely if you prefer

cgillum: I'm fine with true by default as a compromise.

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -23,6 +23,7 @@ namespace DurableTask.AzureStorage.Messaging
     using DurableTask.AzureStorage.Storage;
     using DurableTask.Core;
     using DurableTask.Core.History;
+    using Newtonsoft.Json;
```
cgillum: Can this `using` be removed? I get nervous every time I see `Newtonsoft`. :)

cgillum (dismissed):

cgillum (dismissed):

src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs
```diff
@@ -389,6 +381,171 @@ public virtual async Task DeleteMessageAsync(MessageData message, SessionBase? s
...
+                    this.azureStorageClient.BlobAccountName,
+                    this.settings.TaskHubName,
+                    eventType: string.Empty,
+                    taskEventId: 0,
+                    queueMessage.MessageId,
+                    orchestrationInstance?.InstanceId ?? string.Empty,
+                    orchestrationInstance?.ExecutionId ?? string.Empty,
+                    this.storageQueue.Name,
+                    queueMessage.DequeueCount,
+                    blobName);
```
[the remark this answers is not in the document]
cgillum: @sophiatev this looks like a good thing to fix. It will make it easier to know what kind of message was removed as poisoned.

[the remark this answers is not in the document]
sophiatev: Yep, I agree, I actually did fix it :) I believe this is outdated

PR #1367, opened 2026-06-12
Add per-attempt retry metadata tags in OrchestrationContext.ScheduleWithRetry

bachuv:
## Summary

`OrchestrationContext.ScheduleWithRetry<T>` now stamps two reserved tags on
every per-attempt schedule:

| Tag | Value |
|---|---|
| `dt.retry.attempt` | attempt counter (starts at 1), decimal, `InvariantCulture` |
| `dt.retry.maxAttempts` | `retryOptions.MaxNumberOfAttempts`, same format |

Tags flow through the existing
`ScheduleTaskOptions.Tags` → `ScheduleTaskOrchestratorAction.Tags` →
persisted `TaskScheduledEvent.Tags` pipeline. No dispatcher change.

This is the foundational change for the native activity retry visibility
feature.

## Design

Counter is **scoped to a single `ScheduleWithRetry` call** (local variable
captured by the nested `RetryCall` function):

- Per-call-site counter — no shared state, no cross-call leakage.
- Replay-deterministic — `RetryInterceptor.Invoke` re-walks history the same
  way on replay, producing identical tag values.
- No infinite recursion — the recursive `ScheduleTask<T>` call passes
  `options.RetryOptions == null`, taking the non-retry path.
- `dt.*` tag prefix reserved for DTFx-injected metadata (no existing code or
  docs use it).

## Files

- `src/DurableTask.Core/OrchestrationContext.cs` — `ScheduleWithRetry`
  closure builds `ScheduleTaskOptions` with the two tags.
- `src/DurableTask.Core/RetryTags.cs` (new) — internal constants.
- `test/DurableTask.Core.Tests/ScheduleWithRetryTagsTests.cs` (new, 360 lines)
  — covers monotonic counter starting at 1, per-call-site independence, replay
  determinism, `Handle()` short-circuit, `InvariantCulture` formatting,
  multi-digit attempts, `maxAttempts=1`.

## Compatibility

- No API breaks.
- Backends that ignore `TaskScheduledEvent.Tags` (Azure Storage, MSSQL,
  Netherite) silently drop them at persistence.

### DIFF
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -210,7 +210,7 @@ namespace DurableTask.Core
         ///     under the reserved <c>dt.retry.*</c> tag namespace (<c>dt.retry.attempt</c>, <c>dt.retry.maxAttempts</c>).
         ///     These tags flow through <c>ScheduleTaskOrchestratorAction.Tags</c> into the persisted
         ///     <c>TaskScheduledEvent.Tags</c>, where consumers (Functions extension, DTS Dashboard) can read them.
-        ///     Backends that do not roundtrip <c>TaskScheduledEvent.Tags</c> (currently Azure Storage, MSSQL, Netherite)
+        ///     Backends that do not roundtrip <c>TaskScheduledEvent.Tags</c>
         ///     simply drop the tags at persistence; consumers fall back to a "metadata unavailable" path with no error.
         ///     The attempt counter is closure-local, so each call site builds its own counter; deterministic replay
         ///     reproduces the same per-attempt values because <see cref="RetryInterceptor{T}.Invoke"/> invokes the
--- a/src/DurableTask.Core/OrchestrationContext.cs
+++ b/src/DurableTask.Core/OrchestrationContext.cs
@@ -15,6 +15,7 @@ namespace DurableTask.Core
 {
     using System;
     using System.Collections.Generic;
+    using System.Globalization;
     using System.Threading;
     using System.Threading.Tasks;
     using Castle.DynamicProxy;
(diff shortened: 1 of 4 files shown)


src/DurableTask.Core/RetryTags.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <see cref="History.TaskScheduledEvent.Tags"/>) when an activity is scheduled under a retry policy.
+    /// </summary>
+    /// <remarks>
+    /// These key names are part of the cross-stack wire contract consumed by the Functions extension,
+    /// the JS SDK helpers, and the DTS Dashboard. They are frozen at v1 and must not be renamed.
+    /// The full design and test-vector fixture live in the <c>df-retry-information</c> investigation doc.
+    /// The <c>dt.</c> prefix is reserved for DTFx-injected metadata; customers must not use it on
+    /// caller-supplied tags.
+    /// </remarks>
+    internal static class RetryTags
```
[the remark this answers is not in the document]
nytian: This comment looks correct to me/ But as long as after testing internal works then I vote for internal

nytian (dismissed):

src/DurableTask.Core/RetryTags.cs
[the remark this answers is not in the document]
bachuv: Keeping it as internal.

test/DurableTask.Core.Tests/ScheduleWithRetryTagsTests.cs
```diff
@@ -0,0 +1,438 @@
+// ---------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// ---------------------------------------------------------------
```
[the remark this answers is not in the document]
bachuv: Fixed.

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -204,10 +205,31 @@ public virtual Task<T> ScheduleWithRetry<T>(Type taskActivityType, RetryOptions
         /// <param name="retryOptions">Retry policy</param>
         /// <param name="parameters">Parameters for the TaskActivity.Execute method</param>
         /// <returns>Task that represents the execution of the specified TaskActivity</returns>
+        /// <remarks>
+        ///     Per-attempt retry metadata is injected into the scheduled task's <see cref="ScheduleTaskOptions.Tags"/>
+        ///     under the reserved <c>dt.retry.*</c> tag namespace (<c>dt.retry.attempt</c>, <c>dt.retry.maxAttempts</c>).
+        ///     These tags flow through <c>ScheduleTaskOrchestratorAction.Tags</c> into the persisted
+        ///     <c>TaskScheduledEvent.Tags</c>, where consumers (Functions extension, DTS Dashboard) can read them.
```
sophiatev, line 212: Can we remove this part of the comment?

```
        ///     where consumers (Functions extension, DTS Dashboard) can read them.
        ///     Backends that do not roundtrip <c>TaskScheduledEvent.Tags</c>
        ///     simply drop the tags at persistence; consumers fall back to a "metadata unavailable" path with no error.
```

I feel like it's leaking implementation details/we probably shouldn't be referencing Functions and DTS from here

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -204,10 +205,31 @@ public virtual Task<T> ScheduleWithRetry<T>(Type taskActivityType, RetryOptions
...
+        ///     Backends that do not roundtrip <c>TaskScheduledEvent.Tags</c>
+        ///     simply drop the tags at persistence; consumers fall back to a "metadata unavailable" path with no error.
+        ///     The attempt counter is closure-local, so each call site builds its own counter; deterministic replay
```
sophiatev, line 215: Just double-checking, I'm assuming we confirmed this is actually true in testing (like a "live test", I see that we did test it in the unit tests)?

"deterministic replay reproduces the same per-attempt values because <see cref="RetryInterceptor{T}.Invoke"/> invokes the closure the same number of times in the same order on replay."

src/DurableTask.Core/OrchestrationContext.cs
```diff
@@ -204,10 +205,31 @@ public virtual Task<T> ScheduleWithRetry<T>(Type taskActivityType, RetryOptions
...
             params object[] parameters)
         {
-            Task<T> RetryCall() => ScheduleTask<T>(name, version, parameters);
+            int attempt = 0;
+            Task<T> RetryCall()
+            {
+                attempt++;
+                ScheduleTaskOptions options = ScheduleTaskOptions.CreateBuilder()
+                    .AddTag(RetryTags.Attempt, attempt.ToString(CultureInfo.InvariantCulture))
+                    .AddTag(RetryTags.MaxAttempts, retryOptions.MaxNumberOfAttempts.ToString(CultureInfo.InvariantCulture))
```
sophiatev, line 228: What happens if `MaxNumberOfAttempts` is not set here?

test/DurableTask.Core.Tests/ScheduleWithRetryTagsTests.cs
```diff
@@ -0,0 +1,438 @@
...
+            await Assert.ThrowsExceptionAsync<DurableTask.Core.Exceptions.TaskFailedException>(() => resultTask);
+
+            // Assert: exactly 1 schedule, tagged attempt=1, maxAttempts=1.
+            Assert.AreEqual(1, this.context.ScheduledTasks.Count);
+            Assert.AreEqual("1", this.context.ScheduledTasks[0].Options.Tags[RetryTags.Attempt]);
+            Assert.AreEqual("1", this.context.ScheduledTasks[0].Options.Tags[RetryTags.MaxAttempts]);
+        }
+
+        // ---------------------------------------------------------------------
+        // Core test #8: MaxNumberOfAttempts == 0 — closure is never invoked.
```
sophiatev, line 304: So if I understand this test correctly, we should never invoke the Activity in this case, right? I haven't looked at the other PRs yet so forgive me if we're already doing this but I wonder if we should add a check preventing a 0 value here since it seems non-sensical in that case...

sophiatev:
One thing I just realized - will this work with rewind? I think it should be fine since we just delete the `TaskScheduled` events from the history that failed in that case, and the already-completed events should presumably be fine under replay... but maybe worth a check

src/DurableTask.Core/OrchestrationContext.cs
bachuv: MaxNumberOfAttempts is a required, non-nullable constructor argument (there's no parameterless RetryOptions ctor), so it's always set.

PR #1369, opened 2026-06-17
Prevent OData injection by using parameterized filters and escaping quote characters

AnatoliB:
### Summary

Hardens Azure Table Storage OData filter construction against injection. User-influenced values — orchestration **instance IDs**, **instance-ID prefixes**, **task hub names**, and **execution IDs** — were interpolated into OData `$filter` strings inside single quotes but were not escaped. Since `KeySanitation.EscapePartitionKey` does **not** escape single quotes, a value containing `'` could alter or break the generated query.

### The problem

OData string literals are single-quoted, so a literal `'` must be doubled (`'` → `''`). Several filters were built via raw interpolation, e.g.:

```csharp
string filter = $"PartitionKey eq '{KeySanitation.EscapePartitionKey(instanceId)}'";
```

`EscapePartitionKey` only escapes storage-key characters (`^ / \ # ?` and control chars), not `'`. A crafted instance ID like `a' or PartitionKey ne '` therefore changes the query's meaning.

### Changes

**`DurableTask.AzureStorage`** (uses `Azure.Data.Tables`)
- New internal helper **`AzureTableQueryFilter`** with atomic, self-escaping predicate builders — `PartitionKeyEquals`, `ColumnEquals`, `PartitionKeyGreaterOrEqual`, `PartitionKeyLessThan` — built on `TableClient.CreateQueryFilter(...)`, which parameterizes and escapes values. `PartitionKeyEquals` also applies `KeySanitation.EscapePartitionKey`.
- Refactored all filter-building sites to use it: `AzureTableTrackingStore` (`GetHistoryEntitiesResponseInfoAsync`, `RewindHistoryAsync`, `PurgeInstanceHistoryAsync`, `UpdateInstanceStatusForCompletedOrchestrationAsync`) and `OrchestrationInstanceStatusQueryCondition` (`TaskHubNames`, `InstanceIdPrefix`, `InstanceId`).

**`DurableTask.ServiceBus`** (builds filters from `string.Format` templates)
- Added a small `EscapeODataValue(value) => value?.Replace("'", "''")` helper, applied to user-influenced values in `AzureTableClient` query generation (instance ID, execution ID, name, version, status, partition/row-key ranges).

The two approaches reflect the different table SDKs: AzureStorage can use the modern parameterized `CreateQueryFilter`; ServiceBus builds filters from format-string templates, so values are escaped before formatting.

### A note on `executionId`

`executionId` is escaped even though it's framework-generated at runtime, because the public `IOrchestrationServiceClient.GetOrchestrationHistoryAsync(instanceId, executionId)` API forwards a **caller-supplied** execution ID into the history query. Escaping all filter values uniformly also avoids fragile per-value "is this user-controlled?" judgments.

### Testing

- **`AzureTableQueryFilterTests`** (new) — data-driven unit tests per helper: normal values, single-quote escaping, empty value, partition-key sanitization.
- **`OrchestrationInstanceStatusQueryConditionTest`** (new cases) — single-quote escaping for task hub name, instance-ID prefix, instance ID.
- **`AzureTableClientTest`** (new cases) — single-quote escaping in ServiceBus queries.
- **`PurgeInstanceHistory_InstanceIdWithSingleQuote_Succeeds`** (new) — end-to-end purge with a single-quote instance ID; verified to fail without the fix.
- Existing `TestWorkerFailingDuringCompleteWorkItemCallLargeInputOutput` continues to exercise the `UpdateInstanceStatus` large-payload paths.

Filter output is **byte-identical** to prior behavior for inputs without special characters (confirmed by the unchanged composition tests).

### Compatibility

No behavior change for normal inputs — only quote handling differs. Compile-time constants (sentinel row key, event-type names) flow through the same helpers for consistency; escaping them is a harmless no-op.


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -233,10 +233,14 @@ public override async Task<OrchestrationHistory> GetHistoryEventsAsync(string in
...
-            string filter = $"{nameof(ITableEntity.PartitionKey)} eq '{KeySanitation.EscapePartitionKey(instanceId)}'";
+            string escapedInstanceId = KeySanitation.EscapePartitionKey(instanceId);
+            string filter = TableClient.CreateQueryFilter($"PartitionKey eq {escapedInstanceId}");
             if (!string.IsNullOrEmpty(expectedExecutionId))
             {
-                filter += $" and ({nameof(ITableEntity.RowKey)} eq '{SentinelRowKey}' or {nameof(OrchestrationInstance.ExecutionId)} eq '{expectedExecutionId}')";
+                // Use parameterized filter to prevent OData injection via crafted execution IDs
+                string sentinelCondition = TableClient.CreateQueryFilter($"RowKey eq {SentinelRowKey}");
+                string executionIdCondition = TableClient.CreateQueryFilter($"ExecutionId eq {expectedExecutionId}");
+                filter += $" and ({sentinelCondition} or {executionIdCondition})";
```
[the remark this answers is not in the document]
AnatoliB: Addressed

src/DurableTask.AzureStorage/Tracking/AzureTableQueryFilter.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        /// <summary>
+        /// Builds a <c>{columnName} eq '...'</c> condition with <paramref name="value"/> OData-escaped.
+        /// The <paramref name="columnName"/> must be a trusted constant (never user input), since it is
+        /// emitted as literal filter text rather than an escaped value.
+        /// </summary>
+        public static string ColumnEquals(string columnName, string value)
+        {
+            return TableClient.CreateQueryFilter(FormattableStringFactory.Create(columnName + " eq {0}", value));
```
cgillum: What's the purpose of using `FormattableStringFactory.Create(...)` here?

src/DurableTask.AzureStorage/Tracking/AzureTableQueryFilter.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    static class AzureTableQueryFilter
+    {
+        /// <summary>
+        /// Builds a <c>PartitionKey eq '...'</c> condition for the given orchestration instance ID.
+        /// Applies <see cref="KeySanitation.EscapePartitionKey"/> for storage-key sanitization and then
+        /// OData quote-escaping. Pass the <b>raw</b> instance ID; do not pre-sanitize it, because
+        /// <see cref="KeySanitation.EscapePartitionKey"/> is not idempotent.
+        /// </summary>
+        public static string PartitionKeyEquals(string instanceId)
```
cgillum: I'm worried that some of these APIs make assumptions about whether the input has already been sanitized without baking it into the method signature (you have to actually read the summary docs). I suggest we make this more explicit. A couple suggestions:
- Rename the input parameter, adding a "sanitized" or "unsanitized" prefix (you're doing that with the latter two methods but not the former two)
- Add a Boolean flag indicating whether the input has been sanitized and act accordingly to accept the value as-is or sanitize it.

src/DurableTask.AzureStorage/Tracking/AzureTableQueryFilter.cs
AnatoliB: We can't just pass `$"{columnName} eq {value}"` to `TableClient.CreateQueryFilter`, because it escapes/quotes every interpolated value, including `columnName`. We'd get `'myColumn' eq 'myValue'` instead of the intended `myColumn eq 'myValue'`. That's why the column name has to be part of the literal format text, which is what `FormattableStringFactory.Create` lets us do. I realize this is a bit obscure though, so added a comment.

src/DurableTask.AzureStorage/Tracking/AzureTableQueryFilter.cs
AnatoliB: Good point. Renamed the input parameters.

cgillum (approved):

PR #1380, opened 2026-07-27
Avoid serializing inline Azure Storage queue messages twice

### LINKED ISSUE #1378: Avoid serializing inline Azure Storage queue messages twice
## What is the issue

`MessageManager.SerializeMessageDataAsync` serializes every `MessageData` instance into `rawContent`, uses that string to calculate its UTF-8 size and select inline versus blob storage, and then serializes the same object a second time in the common `InlineJson` branch:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.AzureStorage/MessageManager.cs#L103-L123

The only mutation between those two serializations is `TotalMessageSizeBytes`. That property is internal and is not a `[DataMember]`, while `MessageData` is a `[DataContract]` whose wire properties are explicitly marked:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.AzureStorage/MessageData.cs#L24-L27

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.AzureStorage/MessageData.cs#L94-L108

Each serialization also creates a fresh `StringBuilder`, `StringWriter`, and result string:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.AzureStorage/Utils.cs#L168-L184

This method is called for every outbound Azure Storage task-hub queue message:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs#L91-L110

The blob-offload branch legitimately serializes a small wrapper after uploading `rawContent`; only the inline branch repeats the original serialization.

## Performance impact

Inline messages are the normal path for payloads below the 45 KiB threshold. Every such orchestration, activity, timer, sub-orchestration, and external-event message currently incurs two complete Newtonsoft JSON traversals and two sets of temporary buffers/strings instead of one.

The duplicate CPU and allocation cost scales linearly with message throughput and payload size. At high task-hub throughput this increases serialization CPU, memory bandwidth, Gen-0 pressure, and queue-send latency without changing the resulting wire payload.

## Proposed backward-compatible solution

Return the already-produced `rawContent` in the `InlineJson` branch:

```csharp
if (messageFormat != MessageFormatFlags.InlineJson)
{
    // Existing blob upload and wrapper serialization remain unchanged.
}

return rawContent;
```

This preserves the exact JSON that was already used for the byte-count/format decision. It changes no public API, queue schema, serializer settings, or blob behavior.

## Validation

- Add regression coverage for both `UseDataContractSerialization` modes and representative custom type-binder settings, asserting that the optimized inline result exactly matches the current second serialization.
- Cover payloads immediately below and above the inline threshold to ensure the blob-wrapper path is unchanged.
- Benchmark representative small, medium, and near-threshold messages with allocation diagnostics; the inline path should perform one `MessageData` serialization instead of two.

berndverst:
## Summary
- reuse the `MessageData` JSON already produced for the UTF-8 size check when a queue message stays inline
- verify byte-for-byte inline payload equivalence with a deterministic custom type binder in both `UseDataContractSerialization` modes
- cover the exact 45 KiB inline boundary and the above-threshold blob payload and queue wrapper
- document custom type-binder expectations and the compatibility caveat for stateful serialization hooks

## Backward compatibility
The initial serialization already determines both the queue payload size and the inline/blob decision. `TotalMessageSizeBytes` is internal and is not a data member. For stable messages with deterministic, side-effect-free serialization, returning that existing JSON on the inline path is byte-for-byte equivalent to the previous second serialization. The size threshold, serializer configuration, public API signatures, queue schema, blob upload content, and blob-reference wrapper serialization are unchanged.

Serialization invocation counts are implementation details, not an API contract. Inline messages now undergo one full serialization pass instead of two. Custom `ICustomTypeBinder.BindToName` implementations, getters, or serialization callbacks on transport types that change their output or rely on side effects across passes can observe a behavior change and should be reviewed when upgrading. The existing blob path already serializes the full message only once, followed by a separate serialization of the blob-reference wrapper. These expectations are documented in the Azure Storage provider guide and custom type-binder API documentation.

## Testing
- `dotnet test Test\DurableTask.AzureStorage.Tests\DurableTask.AzureStorage.Tests.csproj --framework net8.0 --filter "FullyQualifiedName~DurableTask.AzureStorage.Tests.MessageManagerTests"`
- `dotnet test Test\DurableTask.AzureStorage.Tests\DurableTask.AzureStorage.Tests.csproj --framework net48 --filter "FullyQualifiedName~DurableTask.AzureStorage.Tests.MessageManagerTests"`
- Rechecked the inline-payload regression cases on both `net8.0` and `net48` after the documentation update to verify the test helper's binder assignment; both serializer modes passed on both targets.

Fixes #1378


nytian (dismissed):

src/DurableTask.AzureStorage/MessageManager.cs
```diff
@@ -119,7 +119,7 @@ public async Task<string> SerializeMessageDataAsync(MessageData messageData, Can
                 return Utils.SerializeToJson(serializer, wrapperMessageData);
             }
 
-            return Utils.SerializeToJson(serializer, messageData);
+            return rawContent;
```
nytian, line 122: was quite surprised about this. When will we hit this line?

berndverst (commented):
.

src/DurableTask.AzureStorage/MessageManager.cs
berndverst: This is the normal inline-message path: we hit this line whenever the full serialized `MessageData` is **<= 45 KiB (46,080 UTF-8 bytes)**, before Base64 encoding. `TaskHubQueue.AddMessageAsync` calls this method before enqueueing the message. Only payloads above that threshold enter the preceding `if` block, which uploads the full message to blob storage and returns the blob-reference wrapper early.

For inline messages, we already have the complete JSON in `rawContent` from the serialization used to measure the payload size. The only intervening mutation is setting `TotalMessageSizeBytes`, which is internal and not a serialized data member. The old line therefore serialized the same payload a second time; returning `rawContent` avoids that redundant work.

test/DurableTask.AzureStorage.Tests/MessageManagerTests.cs
```diff
@@ -94,17 +195,108 @@ private string GetMessage(string dictionaryType)
...
+        }
+
+        private static string SerializeUsingMessageSettings(
+            AzureStorageOrchestrationServiceSettings settings,
+            MessageData message)
+        {
+            var serializerSettings = new JsonSerializerSettings
+            {
+                TypeNameHandling = TypeNameHandling.Objects,
+                SerializationBinder = new TypeNameSerializationBinder(settings.CustomMessageTypeBinder),
```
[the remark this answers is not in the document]
berndverst: The conditional here is evaluated when the provider library is compiled, not when the test project is compiled. `DurableTask.AzureStorage.csproj` targets `netstandard2.0`, and both test targets consume it via `ProjectReference`; they do not compile `MessageManager.cs` themselves. The referenced `TypeNameSerializationBinder` therefore implements `ISerializationBinder` for both `net8.0` and `net48`, so `SerializationBinder` is the correct assignment.

I rebuilt both test targets and ran `SerializeMessageDataAsync_InlinePayloadMatchesPreviousWirePayload`: both serializer-mode cases passed on each target. No binder change is needed; using the test project's `NETSTANDARD2_0` conditional would select the wrong property.

src/DurableTask.AzureStorage/MessageManager.cs
berndverst: One qualification to my explanation above: identical output assumes stable messages and deterministic, side-effect-free serialization. A stateful custom type binder, getter, or serialization callback could observe the removed second pass. The full-message blob path already runs only one pass, so relying on two passes would also be payload-size-dependent.

I've kept the optimization and documented this caveat, including that binder invocation counts are not an API contract, in f09bb565. The PR's compatibility section is also updated to avoid claiming unconditional byte-for-byte equivalence.

nytian (approved):

PR #1381, opened 2026-07-27
Avoid eager trace payload construction when diagnostics are disabled

### LINKED ISSUE #1377: Avoid eager trace payload construction when diagnostics are disabled
## What is the issue

`TraceHelper` does not actually defer message construction until after the corresponding `EventSource` level is known to be enabled:

- The `Func<string>` overloads call `generateMessage()` while evaluating the arguments to `DefaultEventSource.Log.TraceEvent(...)`.
- The format overloads call `FormatString(...)` at the same point.
- `DefaultEventSource` checks `IsTraceEnabled` / `IsInfoEnabled` only after those arguments have already been evaluated.
- Expensive expressions passed to the format overload are evaluated even earlier, before entering `TraceHelper` at all.

This means disabled diagnostics still execute payload-building work on central dispatch paths. For example, orchestration dispatch serializes a runtime-state dump before every execution, eagerly joins every decision after execution, and uses the nominally lazy overload to serialize another dump on completion:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L437-L442

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L471-L477

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L1047-L1051

The helper overloads evaluate their factories and formatting before the eventual level check:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/Tracing/TraceHelper.cs#L713-L777

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/Tracing/DefaultEventSource.cs#L99-L118

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/Tracing/DefaultEventSource.cs#L145-L152

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.Core/Tracing/DefaultEventSource.cs#L197-L203

The same pattern eagerly enumerates and formats Service Bus message batches on receive and send paths:

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs#L549-L556

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs#L1363-L1367

https://github.com/Azure/durabletask/blob/5217032961abf45846f462732a5e2813316e3747/src/DurableTask.ServiceBus/ServiceBusOrchestrationService.cs#L1576-L1592

## Performance impact

When no diagnostics listener is attached, or when the requested level is disabled, every affected work item still pays for payload construction that is immediately discarded:

- runtime-state dump objects, lists, JSON writers, and result strings;
- LINQ enumerators and per-action/per-message strings;
- `string.Join`, interpolation, `params` arrays, and final formatted strings.

This work scales directly with orchestration dispatch rate and Service Bus batch size. It consumes CPU, increases Gen-0 allocation/collection frequency, and can increase tail latency despite producing no telemetry.

## Proposed backward-compatible solution

1. Add a single internal level check in `TraceHelper` (or an internal `DefaultEventSource.IsEnabled(TraceEventType)` helper) and return before `generateMessage()` or `FormatString(...)` is invoked.
2. Apply the guard consistently to the factory, format, and exception-trace helper paths.
3. Convert expensive pre-evaluated call sites such as the runtime-state serialization and Service Bus `string.Join` expressions to the factory overload, with all expensive work inside the lambda. A central helper guard alone cannot defer expressions that were evaluated before the helper call.
4. Preserve the existing event IDs, levels, payload text, and schemas whenever the level is enabled.

This changes no public API or wire format and leaves enabled diagnostics unchanged.

## Validation

- Add tests with a message factory that increments a counter: it must not run with the level disabled and must run exactly once with an appropriate `EventListener` enabled.
- Assert that enabled event payloads remain byte-for-byte equivalent.
- Benchmark orchestration dispatch with diagnostics disabled and track allocated bytes/op and Gen-0 collections before and after.

berndverst:
## Summary
- guard `TraceHelper` factory, format, and exception payload construction with the matching `EventSource` level check
- defer orchestration runtime-state serialization, decision joins, and Service Bus batch formatting until diagnostics are enabled
- add regression coverage proving factories and formatting are skipped when disabled, execute exactly once when enabled, and preserve event IDs, schemas, event types, messages, and exception return values

## Compatibility and performance
This does not change public APIs or EventSource event IDs/schemas. Enabled diagnostic payloads remain unchanged. When the requested level is disabled, high-throughput dispatch paths now avoid runtime-state serialization, LINQ enumeration, `string.Join`, interpolation, and final formatting allocations.

## Validation
- `dotnet test Test\DurableTask.Core.Tests\DurableTask.Core.Tests.csproj -f net8.0 --filter "FullyQualifiedName~TraceHelperTests|FullyQualifiedName~CommonTests.ShouldValidateEventSource"`
- `dotnet test Test\DurableTask.Core.Tests\DurableTask.Core.Tests.csproj -f net48 --filter "FullyQualifiedName~TraceHelperTests|FullyQualifiedName~CommonTests.ShouldValidateEventSource"`
- `dotnet build src\DurableTask.ServiceBus\DurableTask.ServiceBus.csproj`

Fixes #1377

### DIFF
--- a/Test/DurableTask.Core.Tests/TraceHelperTests.cs
+++ b/Test/DurableTask.Core.Tests/TraceHelperTests.cs
@@ -14,8 +14,10 @@
 #nullable enable
 namespace DurableTask.Core.Tests
 {
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Diagnostics.Tracing;
     using DurableTask.Core.Entities.OperationFormat;
     using DurableTask.Core.Tracing;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -23,6 +25,7 @@ namespace DurableTask.Core.Tests
     using TraceActivityStatusCode = DurableTask.Core.Tracing.ActivityStatusCode;
 
     [TestClass]
+    [DoNotParallelize]
     public class TraceHelperTests
     {
         [TestMethod]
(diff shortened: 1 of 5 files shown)


cgillum (approved):
This is a nice improvement. One caveat, however, is that I think Windows App Service-hosted apps capture all ETW trace levels by default, which means we won't observe improvements for those apps. However, I expect this will help with anything hosted on Linux or outside of App Service.

PR #1386, opened 2026-08-14
Enforce safe JSON deserialization in ServiceFabric

nytian:
Ensure the Service Fabric JSON formatter always uses the allowed-types binder. This prevents untrusted requests from deserializing arbitrary .NET types through type metadata. Tests are added.


bachuv (approved):
Approving, but please add detailed notes about this behavior change in the release notes. Thanks!

PR #1387, opened 2026-08-17
Persist parent orchestration metadata in Azure Storage

### LINKED ISSUE #618: GetOrchestrationStateAsync does not populate `ParentInstance` (StorageAccount provider)
Getting the state of a sub orchestration using `GetOrchestrationStateAsync`, the resulting OrchestrationState.ParentInstance is `null`.

However, if I request the history using `GetOrchestrationHistoryAsync`, the parent instance is populated correctly:
```json
{
    "EventType": 12,
    "EventId": -1,
    "IsPlayed": false,
    "Timestamp": "2021-10-01T02:46:41.3972531Z"
  },
  {
    "OrchestrationInstance": {
      "InstanceId": "30aebd816f1f4755b0710229bda170ed:5",
      "ExecutionId": "9fde139382dc4d8280d14916692b8870"
    },
    "EventType": 0,
    "ParentInstance": {
      "Name": "DeploymentOrchestration",
      "OrchestrationInstance": {
        "InstanceId": "cd8af473-fcb4-4e25-99c4-5cc75c64aeb1",
        "ExecutionId": "30aebd816f1f4755b0710229bda170ed"
      },
      "TaskScheduleId": 5,
      "Version": "1"
   }
   
```

My persistent store is StorageAccount.

YunchuWang:
Fixes https://github.com/Azure/durabletask/issues/618

Enables https://github.com/Azure/azure-functions-durable-extension/issues/3515 to expose `parentInstanceId` for the Azure Storage provider, matching MSSQL semantics.

## Before / after

Before, Azure Storage retained parent metadata only in history, so existing get/query status paths returned `OrchestrationState.ParentInstance == null` for sub-orchestrations. After this change, the Instances row stores only the parent instance ID and the existing row-to-state conversion reconstructs `ParentInstance.OrchestrationInstance.InstanceId`.

## Implementation

- Adds one Instances-table string property: `ParentInstanceId`.
- Sources it from `ExecutionStartedEvent.ParentInstance.OrchestrationInstance.InstanceId`.
- Writes it in new-execution, execution-started update, and fast-completion/recreated-row paths.
- Preserves `ParentInstance` in `InstanceStoreBackedTrackingStore`.
- Uses an empty value for no parent because update/recreation paths use Azure Table merge/upsert-merge semantics; omitting a property could preserve a stale parent on a reused row. Reads treat empty and legacy missing properties as null.
- Adds no history reads, status-query round trips, N+1 behavior, or unrelated API/protobuf changes.

## Coverage

Behavioral coverage:

- Exact child parent ID via direct get and list/query.
- Top-level direct/query state remains null.
- Legacy missing-property rows remain null.
- Terminal-history repair recreates a missing Instances row with the parent ID, and overwrites a stale one.
- Child `ContinueAsNew` retains parent metadata.
- Alternate `InstanceStoreBackedTrackingStore` preserves the original `ParentInstance`.
- `ParentInstanceId` stays in the OData projection when `FetchInput`/`FetchOutput` are false, so status queries that exclude payloads still return the parent.

Write-path coverage against real Azure Table update semantics. Each production write site is covered by a test that asserts the **raw stored property**, not just the converted state, so a value merely retained by a merge cannot produce a false positive:

| Production site | Update mode | Covering tests |
| --- | --- | --- |
| `SetNewExecutionAsync` | Insert / Replace | `SetNewExecution_PersistsNonNullParentInstanceId`, `SetNewExecution_ClearsStaleParentInstanceId` |
| `UpdateStateAsync` ExecutionStarted → `UpdateInstanceTableAsync` | InsertOrMerge (`UseInstanceTableEtag=false`) and Merge with ETag (`true`) | `CheckpointWrite_PersistsParentInstanceId_ForBothEtagModes`, `CheckpointWrite_ClearsStaleParentInstanceId_ForBothEtagModes` |
| `UpdateInstanceStatusForCompletedOrchestrationAsync` | InsertOrMerge | `CompletedOrchestrationRepair_PersistsParentInstanceId`, `CompletedOrchestrationRepair_OverwritesStaleParentInstanceId`, `CompletedOrchestrationRepair_ClearsStaleParentInstanceId_WithInsertOrMerge` |
| `ConvertFromAsync` read mapping | n/a | `QueryStatus_WithContinuationToken_NoInputToken` |
| `InstanceStoreBackedTrackingStore` creation | n/a | `InstanceStoreBackedTrackingStore_PersistsParentOnCreation` |

Both `UseInstanceTableEtag` modes are genuinely exercised, not merely toggled: the tests seed a row and pass its real ETag so the `true` case reaches `MergeEntityAsync` rather than the insert branch. This was verified with a branch probe that drops the property in only one branch of `UpdateInstanceTableAsync` — probing the Merge branch fails exactly the two `(True)` rows, and probing the InsertOrMerge branch fails exactly the two `(False)` rows.

Mutation testing confirms each test actually proves its production behavior. Removing or weakening a site fails the listed tests:

| Mutation | Tests that fail |
| --- | --- |
| Parent assignment helper reverted to a conditional write (skip when empty) | 4: both `..._ClearsStale...` checkpoint rows, the InsertOrMerge repair clearing test, and `SetNewExecution_ClearsStaleParentInstanceId` |
| Remove the terminal-history repair write | 3 `CompletedOrchestrationRepair_*` tests |
| Remove the checkpoint ExecutionStarted write | all 4 `CheckpointWrite_*` rows |
| Remove the initial `SetNewExecutionAsync` write | both `SetNewExecution_*` tests |
| Drop `ParentInstanceId` from the OData projection | `OrchestrationInstanceQuery_ProjectionRetainsParentInstanceId` |
| Remove the read-side mapping | `QueryStatus_WithContinuationToken_NoInputToken` |
| Remove the alternate-store assignment | `InstanceStoreBackedTrackingStore_PersistsParentOnCreation` |

`ParentMetadataIsReturnedForFastCompletingChild` was renamed to `ParentMetadataIsReturnedByDirectGetAndQuery`, because it exercises the normal checkpoint write and the read paths rather than the terminal-history repair path; repair is covered directly by the `CompletedOrchestrationRepair_*` tests.

## Validation

Local:

- `dotnet build test\DurableTask.AzureStorage.Tests\DurableTask.AzureStorage.Tests.csproj` (clean build, `bin`/`obj` removed first) — passed, 0 warnings, 0 errors.
- `git diff --check` — clean.

Hosted Azure Storage (approved non-production subscription `edc48857-dd0b-4085-a2a9-5e7df12bd2fd`, verified by re-reading the subscription ID before provisioning):

- Resource group `dtfx-pmverify-5fab2bb3c3`, storage account `dtfxpv5fab2bb3c3` (`StorageV2`, `Standard_LRS`, `eastus`), tagged for validation and ephemeral cleanup.
- Harness: process-scoped `DurableTaskTestStorageConnectionString`.
- Command: `dotnet test test\DurableTask.AzureStorage.Tests\DurableTask.AzureStorage.Tests.csproj -f net8.0 --no-build --filter "FullyQualifiedName~ParentInstanceIdTrackingStoreTests|FullyQualifiedName~AzureTableTrackingStoreTest|FullyQualifiedName~OrchestrationInstanceStatusQueryConditionTest|FullyQualifiedName~ParentMetadata"`
- Result: **30/30 passed**, 0 failed, 0 skipped. Includes both `ParentMetadata*` provider scenarios, all 9 `ParentInstanceIdTrackingStoreTests` (both `CheckpointWrite` DataRow values), the tracking-store conversion tests, and the query projection tests.
- Cleanup proof: `az group exists --subscription edc48857-dd0b-4085-a2a9-5e7df12bd2fd --name dtfx-pmverify-5fab2bb3c3` → `false`. Earlier validation groups `dtfx-pmtests-e1997043f4` and `dtfx-pmfix-81e0870dc4` are also deleted and verified `false`; no resources remain.

Test-only follow-up commits on top of the implementation commit add the mutation-effective write-path tests and the corrected ETag merge coverage. Production behavior is unchanged by those commits.

### DIFF
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -956,6 +956,12 @@ namespace DurableTask.AzureStorage.Tracking
                 ["TaskHubName"] = this.settings.TaskHubName,
             };
 
+            // The parent is written on every checkpoint, not just the one carrying ExecutionStarted, so that
+            // it is always merged together with the ExecutionId above. Because the Instances row is keyed
+            // only by instance ID, a reused instance ID would otherwise end up advertising the current
+            // execution alongside a parent left behind by the previous orchestration.
+            SetParentInstanceId(instanceEntity, newRuntimeState.ParentInstance);
+
             // check if we are replacing a previous execution with blobs; those will be deleted from the store after the update. This could occur in a ContinueAsNew scenario
             List<string> blobsToDelete = null;
             if (oldRuntimeState != newRuntimeState && context.Blobs.Count > 0)
@@ -1002,7 +1008,6 @@ namespace DurableTask.AzureStorage.Tracking
                         instanceEntity["RuntimeStatus"] = OrchestrationStatus.Running.ToString();
                         instanceEntity["Tags"] = TagsSerializer.Serialize(executionStartedEvent.Tags);
                         instanceEntity["Generation"] = executionStartedEvent.Generation;
-                        SetParentInstanceId(instanceEntity, executionStartedEvent.ParentInstance);
                         if (executionStartedEvent.ScheduledStartTime.HasValue)
                         {
                             instanceEntity["ScheduledStartTime"] = executionStartedEvent.ScheduledStartTime;
--- a/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
+++ b/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
@@ -38,6 +38,7 @@ namespace DurableTask.AzureStorage.Tracking
     class AzureTableTrackingStore : TrackingStoreBase
     {
         const string NameProperty = "Name";
+        const string ParentInstanceIdProperty = "ParentInstanceId";
         const string InputProperty = "Input";
         const string ResultProperty = "Result";
         const string OutputProperty = "Output";
@@ -461,6 +462,16 @@ namespace DurableTask.AzureStorage.Tracking
                 InstanceId = instanceId,
                 ExecutionId = orchestrationInstanceStatus.ExecutionId,
             };
+            if (!string.IsNullOrEmpty(orchestrationInstanceStatus.ParentInstanceId))
+            {
+                orchestrationState.ParentInstance = new ParentInstance
+                {
+                    OrchestrationInstance = new OrchestrationInstance
+                    {
+                        InstanceId = orchestrationInstanceStatus.ParentInstanceId,
+                    },
+                };
+            }
 
             orchestrationState.Name = orchestrationInstanceStatus.Name;
             orchestrationState.Version = orchestrationInstanceStatus.Version;
@@ -799,6 +810,7 @@ namespace DurableTask.AzureStorage.Tracking
                 ["Generation"] = executionStartedEvent.Generation,
                 ["Tags"] = TagsSerializer.Serialize(executionStartedEvent.Tags),
             };
+            SetParentInstanceId(entity, executionStartedEvent.ParentInstance);
 
             // It is possible that the queue message was small enough to be written directly to a queue message,
             // not a blob, but is too large to be written to a table property.
@@ -990,6 +1002,7 @@ namespace DurableTask.AzureStorage.Tracking
                         instanceEntity["RuntimeStatus"] = OrchestrationStatus.Running.ToString();
                         instanceEntity["Tags"] = TagsSerializer.Serialize(executionStartedEvent.Tags);
                         instanceEntity["Generation"] = executionStartedEvent.Generation;
+                        SetParentInstanceId(instanceEntity, executionStartedEvent.ParentInstance);
                         if (executionStartedEvent.ScheduledStartTime.HasValue)
                         {
                             instanceEntity["ScheduledStartTime"] = executionStartedEvent.ScheduledStartTime;
@@ -1151,6 +1164,7 @@ namespace DurableTask.AzureStorage.Tracking
                 ["Tags"] = TagsSerializer.Serialize(executionStartedEvent.Tags),
                 ["TaskHubName"] = this.settings.TaskHubName,
             };
+            SetParentInstanceId(instanceEntity, executionStartedEvent.ParentInstance);
             if (runtimeState.ExecutionStartedEvent.ScheduledStartTime.HasValue)
             {
                 instanceEntity["ScheduledStartTime"] = executionStartedEvent.ScheduledStartTime;
@@ -1246,6 +1260,16 @@ namespace DurableTask.AzureStorage.Tracking
             }
 
             return estimatedByteCount;
+        }
+
+        // The value is always assigned, including when there is no parent. Several of the Instances
+        // table writes use merge semantics, so omitting the property would let a top-level or newly
+        // recreated orchestration inherit a stale parent ID from a previous row with the same
+        // instance ID. An empty string is used rather than null because merge semantics for null
+        // properties are ambiguous; reads treat empty and missing identically.
+        static void SetParentInstanceId(TableEntity entity, ParentInstance parentInstance)
+        {
+            entity[ParentInstanceIdProperty] = parentInstance?.OrchestrationInstance?.InstanceId ?? string.Empty;
         }
 
         Type GetTypeForTableEntity(TableEntity tableEntity)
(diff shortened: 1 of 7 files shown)


src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -1248,6 +1262,15 @@ static int GetEstimatedByteCount(TableEntity entity)
...
         }
 
+        static void SetParentInstanceId(TableEntity entity, ParentInstance parentInstance)
+        {
+            string parentInstanceId = parentInstance?.OrchestrationInstance?.InstanceId;
+            if (parentInstanceId != null)
+            {
+                entity[ParentInstanceIdProperty] = parentInstanceId;
+            }
+        }
```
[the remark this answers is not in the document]
YunchuWang: This assignment is intentional. Several Instances-table paths use Azure Table merge semantics (`UpsertEntityAsync(..., TableUpdateMode.Merge)` / `UpdateEntityAsync(..., TableUpdateMode.Merge)`). If `ParentInstanceId` is omitted for a top-level or recreated orchestration, an existing row can retain a stale parent ID. Writing `string.Empty` explicitly clears that value; the read path treats both empty and legacy-missing values as no parent.

This is covered by `CompletedOrchestrationRepair_ClearsStaleParentInstanceId_WithInsertOrMerge`, both rows of `CheckpointWrite_ClearsStaleParentInstanceId_ForBothEtagModes`, and `SetNewExecution_ClearsStaleParentInstanceId`. Reverting the helper to a conditional write makes those cases fail. The current inline comment and PR description now document this behavior, so no code change is needed here.

src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs
```diff
@@ -461,6 +462,16 @@ async Task<OrchestrationState> ConvertFromAsync(OrchestrationInstanceStatus orch
                 InstanceId = instanceId,
                 ExecutionId = orchestrationInstanceStatus.ExecutionId,
             };
+            if (!string.IsNullOrEmpty(orchestrationInstanceStatus.ParentInstanceId))
```
bachuv, line 465: Could we bind `ParentInstanceId` to the current `ExecutionId` to prevent stale parent metadata when a sub-orchestration instance ID is reused?

YunchuWang: Good catch — this was a real gap, and the fix is now in 5501727c951297b286d76a787f16b18afcadff63.

**The problem.** `UpdateStateAsync` always merges `["ExecutionId"] = executionId` into `instanceEntity`, but `SetParentInstanceId` was only called from the `EventType.ExecutionStarted` branch of the `NewEvents` loop. A checkpoint whose `NewEvents` does not include `ExecutionStarted` — i.e. any mid-orchestration checkpoint — therefore merged a current execution ID while leaving the previously stored `ParentInstanceId` untouched. Since the Instances row is keyed only by instance ID, a reused instance ID could end up advertising the current execution alongside the parent of the previous orchestration.

**The fix.** Rather than binding the parent to the execution with an extra column, the parent is now written unconditionally right after `instanceEntity` is constructed, using `newRuntimeState.ParentInstance`, and the call was removed from the `ExecutionStarted` case. `OrchestrationRuntimeState.ParentInstance` resolves from the runtime state's `ExecutionStartedEvent` regardless of whether that event is in `PastEvents` or `NewEvents`, so the value is always the parent of the execution being checkpointed. `ExecutionId` and `ParentInstanceId` are now updated — or cleared — in the same merge, which makes them consistent by construction. No new column was needed, and the existing empty-string clearing semantics are unchanged.

**Test.** `ParentInstanceIdTrackingStoreTests.CheckpointWithoutExecutionStarted_WritesParentForCurrentExecution` seeds a row with `ExecutionId=execution-0` / `ParentInstanceId=stale-parent`, builds an `OrchestrationRuntimeState` whose `ExecutionStartedEvent` is in `PastEvents` and whose only new event is an `OrchestratorStartedEvent`, then calls `UpdateStateAsync`. It asserts the row's `ExecutionId` and `ParentInstanceId` both correspond to the new execution. It is a 4-row `DataTestMethod` covering `UseInstanceTableEtag=false` (InsertOrMerge) and `true` (real ETag `MergeEntityAsync`), crossed with having a parent and having none, so the clearing case for a reused no-parent execution is covered too.

Before the production change all 4 rows failed with `Expected:<parent-...>. Actual:<stale-parent>`; after it, 4/4 pass. Re-removing the new write line fails 8 tests, so the site is mutation-covered. The full focused suite (`ParentInstanceIdTrackingStoreTests` + `AzureTableTrackingStoreTest` + `OrchestrationInstanceStatusQueryConditionTest`) is 32/32 green.

bachuv (approved):
Approving, but please merge after the CI passes. Thanks!

PR #1388, opened 2026-08-19
Remove  durabletask-core-v2 branch from public build triggers

bachuv:
The durabletask-core-v2 branch has been inactive for ~2 years and no longer needs scheduled or CI builds. This removes it from the CI trigger: and nightly schedules: in the public build pipeline so it no longer runs.

### DIFF
--- a/eng/ci/public-build.yml
+++ b/eng/ci/public-build.yml
@@ -6,7 +6,6 @@ trigger:
   branches:
     include:
     - main
-    - durabletask-core-v2
     - vabachu/v320-release
 
 # Run nightly to catch new CVEs and to report SDL often.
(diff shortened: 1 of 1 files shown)


eng/ci/public-build.yml
```diff
@@ -6,7 +6,6 @@ trigger:
   branches:
     include:
     - main
-    - durabletask-core-v2
     - vabachu/v320-release
```
andystaples: Should we drop this carryover too?

andystaples (dismissed):

eng/ci/public-build.yml
bachuv: Removed vabachu/v320-release

andystaples: It would be nice to branch-protect and include release/* and hotfix/* but probably outside the scope

andystaples (approved):

eng/ci/public-build.yml
bachuv: Yes, we can address that in a separate PR if we want to follow that rule.

PR #1399, opened 2026-09-09
Fix deleted external events during a continue-as-new

sophiatev:
We currently have a bug which deletes external events for a continue-as-new. Consider the following sequence:

1. An orchestration continues-as-new, and schedules a task. The task completes with a `TaskCompleted` event sent to the new execution ID of the orchestration, _before the new history of the orchestration has been committed._ In combination, an `EventRaised` is sent to the orchestration (with no target execution ID).
2. These messages both end up in the same message batch in the `OrchestrationSessionManager`. Since the `TaskCompleted` message has an execution ID, this becomes the execution ID of the message batch.
3. When performing the orchestration history fetch to process the message batch, no history is found because the history for the new execution ID has not been committed.
4. When attempting to generated the next work item, `IsOutOfOrder` correctly deems the `TaskCompleted` event as out of order since it has no matching `TaskScheduled` in the empty history. This message is correctly abandoned and retried. The `EventRaised` should also be retried, but the problem occurs at this [line](https://github.com/Azure/durabletask/blob/b385165ac10ecebbf183fdfdb07db33307756792/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L833) - since the `EventRaised` does not match the execution ID of the orchestration session, it is first correctly added to the abandon list. However, because there end up being no messages to delete, our current logic decides that all of the messages must have been external events with no target execution ID, and it must be the case that the orchestration has completed or been purged. It neglects to take into account the situation where an orchestration continues-as-new, which is why no history was returned for the session's execution ID. This PR simply updates the logic to take this scenario into account by only allowing a full delete of messages if the session's execution ID is null. This means that there cannot be a race condition involving different execution IDs. The most recently stored execution ID would have been prefetched, and so it is indeed safe to say that either there is no history or the orchestration has completed.

### DIFF
--- a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
+++ b/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
@@ -802,7 +802,7 @@ namespace DurableTask.AzureStorage
                         TraceContext = currentRequestTraceContext,
                     };
 
-                    string warningMessage = await this.IsExecutableInstanceAsync(
+                    (string warningMessage, bool isKnownTerminalInstance) = await this.IsExecutableInstanceAsync(
                         session.RuntimeState,
                         orchestrationWorkItem.NewMessages,
                         settings.AllowReplayingTerminalInstances,
@@ -829,12 +829,15 @@ namespace DurableTask.AzureStorage
 
                         // If no messages have a matching execution ID, then delete all of them. This means all the
                         // messages are external (external events, termination, etc.) and were sent to an instance that
-                        // doesn't exist or is no longer in a running state.
-                        // This is only safe to do if the session's execution ID is itself null, otherwise the history
-                        // could be missing due to trying to fetch a particular execution ID that no longer exists for
-                        // example due to a continue-as-new. In that case we do not want to delete the external messages
-                        // but rather retry them on the current execution
-                        if (messagesToDiscard.Count == 0 && session.Instance.ExecutionId == null)
+                        // doesn't exist or is no longer in a running state:
+                        // - If the session's ExecutionId is null, that means no history exists for this instance at all
+                        // (since the session adopts whatever ExecutionId it finds in history, if it was not already set)
+                        // - If the ExecutionId is non-null, then either there exists state for a terminal instance, or
+                        // some of the messages in the batch had an execution ID with no matching history. In the first case,
+                        // we want to delete the message, in the second case we want to abandon it. To distinguish between the two,
+                        // IsExecutableInstanceAsync explicitly reports whether the instance is known to be terminal.
+                        if (messagesToDiscard.Count == 0 &&
+                            (session.Instance.ExecutionId == null || isKnownTerminalInstance))
                         {
                             messagesToDiscard.AddRange(messagesToAbandon);
                             messagesToAbandon.Clear();
@@ -1070,7 +1073,7 @@ namespace DurableTask.AzureStorage
                 data.Episode.GetValueOrDefault(-1));
         }
 
-        async Task<string> IsExecutableInstanceAsync(
+        async Task<(string WarningMessage, bool IsKnownTerminalInstance)> IsExecutableInstanceAsync(
             OrchestrationRuntimeState runtimeState,
             IList<TaskMessage> newMessages,
             bool allowReplayingTerminalInstances,
@@ -1082,7 +1085,7 @@ namespace DurableTask.AzureStorage
 
                 if (DurableTask.Core.Common.Entities.AutoStart(instanceId, newMessages))
                 {
-                    return null;
+                    return (null, false);
                 }
                 else
                 {
@@ -1093,7 +1096,7 @@ namespace DurableTask.AzureStorage
                         await this.trackingStore.UpdateStatusForTerminationAsync(
                             instanceId,
                             executionTerminatedEvent);
-                        return $"Instance is {OrchestrationStatus.Terminated}";
+                        return ($"Instance is {OrchestrationStatus.Terminated}", true);
                     }
 
                     // A non-zero event count usually happens when an instance's history is overwritten by a
@@ -1103,7 +1106,9 @@ namespace DurableTask.AzureStorage
                     // the old history and we receive a message from the old instance (this happens frequently
                     // with canceled durable timer messages) we'll end up loading just the history that hasn't
                     // been fully overwritten. We know it's invalid because it's missing the ExecutionStartedEvent.
-                    return runtimeState.Events.Count == 0 ? "No such instance" : "Invalid history (may have been overwritten by a newer instance)";
+                    return (
+                        runtimeState.Events.Count == 0 ? "No such instance" : "Invalid history (may have been overwritten by a newer instance)",
+                        false);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -3950,6 +3950,171 @@ await service.SendTaskOrchestrationMessageAsync(
...
+                ControlQueueVisibilityTimeout = TimeSpan.FromSeconds(5),
+                UseAppLease = false,
+            };
+
+            try
+            {
+                service = new AzureStorageOrchestrationService(settings);
+                await service.CreateAsync();
+
+                OrchestrationHistory emptyHistory = await service.TrackingStore.GetHistoryEventsAsync(
```
nytian, line 4391: Can  we add some comments here for what each step is doing? I am trying to understand how we simulate the situation and why it works. Thanks

sophiatev: Yeah very good point, I added comments in the newest commit

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -830,7 +830,11 @@ async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAsync(boo
                         // If no messages have a matching execution ID, then delete all of them. This means all the
                         // messages are external (external events, termination, etc.) and were sent to an instance that
                         // doesn't exist or is no longer in a running state.
-                        if (messagesToDiscard.Count == 0)
+                        // This is only safe to do if the session's execution ID is itself null, otherwise the history
+                        // could be missing due to trying to fetch a particular execution ID that no longer exists for
+                        // example due to a continue-as-new. In that case we do not want to delete the external messages
+                        // but rather retry them on the current execution
+                        if (messagesToDiscard.Count == 0 && session.Instance.ExecutionId == null)
```
bachuv: Did you check this change across the different orchestration states and message types to make sure messages that should be discarded aren’t retried? Do we have tests for those cases?

sophiatev: Thank you so much for asking this question, it actually unearthed a bug I had in my implementation where indeed something that should have been discarded (external event sent to a completed orchestration) was retried. This has been fixed in the latest commit, and tests have been added to confirm that external events sent to non-existent orchestrations and to terminal orchestrations are deleted as they should be.

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -829,8 +829,15 @@ async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAsync(boo
...
-                        if (messagesToDiscard.Count == 0)
+                        // doesn't exist or is no longer in a running state:
+                        // - If the session's ExecutionId is null, that means no history exists for this instance at all
+                        // (since the session adopts whatever ExecutionId it finds in history, if it was not already set)
+                        // - If the ExecutionId is non-null, then either there exists state for a terminal instance, or
+                        // some of the messages in the batch had an execution ID with no matching history. In the first case,
+                        // we want to delete the message, in the second case we want to abandon it. To distinguish between the two,
+                        // IsExecutableInstanceAsync explicitly reports whether the instance is known to be terminal.
+                        if (messagesToDiscard.Count == 0 &&
+                            (session.Instance.ExecutionId == null || isKnownTerminalInstance))
```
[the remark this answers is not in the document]
sophiatev: I don't think we should accommodate this scenario - there are two scenarios where IsExecutable would return false, one is a terminal instances, another is an instance with invalid history (no ExecutionStarted) or empty history. This check means that for an invalid history we would continue retrying the message, which I don't think we should.

nytian (dismissed):

test/DurableTask.AzureStorage.Tests/AzureStorageScenarioTests.cs
```diff
@@ -2947,6 +2954,13 @@ public async Task TestWorkerFailingDuringCompleteWorkItemCallCompletedOrchestrat
                 Assert.AreEqual(OrchestrationStatus.Completed, status.OrchestrationStatus);
                 Assert.AreEqual(input, JToken.Parse(status.Output).ToString());
                 Assert.AreEqual(input, JToken.Parse(status.Input).ToString());
+                if (!terminate)
+                {
+                    Assert.AreEqual(
+                        0,
+                        await GetControlQueueMessageCountAsync(host.service),
+                        "The external event sent to the completed orchestration should be deleted.");
```
bachuv: Could we also assert that the queue is empty for the failed and terminated cases and add a terminal-orchestration test with a mixed batch of execution-specific and external-event messages?

sophiatev: Done!

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -827,10 +827,11 @@ async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAsync(boo
...
 
-                        // If no messages have a matching execution ID, then delete all of them. This means all the
-                        // messages are external (external events, termination, etc.) and were sent to an instance that
-                        // doesn't exist or is no longer in a running state.
-                        if (messagesToDiscard.Count == 0)
+                        // If the instance is known to be terminal, delete all messages.
+                        // We don't necessarily want to do this for an instance with an empty history, because it
+                        // could be that there is no history for the specific execution attached to session.Instance.ExecutionId.
+                        // In this case, we would want to retry any external events that do not target a specific execution ID.
+                        if (isKnownTerminalInstance)
```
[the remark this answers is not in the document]
sophiatev: But this logic could never impact a continue-as-new since the instance is non-terminal by definition in that case

src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs
```diff
@@ -827,10 +827,16 @@ async Task<TaskOrchestrationWorkItem> LockNextTaskOrchestrationWorkItemAsync(boo
                             }
                         }
 
-                        // If no messages have a matching execution ID, then delete all of them. This means all the
-                        // messages are external (external events, termination, etc.) and were sent to an instance that
-                        // doesn't exist or is no longer in a running state.
-                        if (messagesToDiscard.Count == 0)
+                        bool matchingExecutionMessageWasAbandoned = outOfOrderMessages?.Any(
+                            msg => msg.TaskMessage.OrchestrationInstance.ExecutionId == session.Instance.ExecutionId) == true;
```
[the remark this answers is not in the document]
sophiatev: With one stale `TaskCompleted(E1)` and `EventRaised(null)`:

1. `TaskCompleted` is marked out of order, abandoned with a visibility delay, and removed from the session.

2. `matchingExecutionMessageWasAbandoned` is true, so `EventRaised` is deferred.

3. On session release, only the deferred `EventRaised(null)` is immediately re-added to the pending manager. The abandoned `TaskCompleted` is not included.

4. The new null-ID batch fetches the latest terminal history.

5. On this second pass, there are no out-of-order messages, so `matchingExecutionMessageWasAbandoned` is false and `messagesToDiscard.Count` is zero.

6. The event is promoted to `messagesToDiscard` and deleted.

Therefore, it normally costs one extra dequeue but does not loop indefinitely. The stale `TaskCompleted` could theoretically become visible and rejoin before the event-only batch finishes prefetching, causing another deferral, but its abandonment uses exponential backoff. That is not the claimed tight release loop.

bachuv (approved):

